PR #3, opened 2023-12-18
chore(contributing): improve onboarding

steebchen:
- shell scripts logging and fail on error
- use npx --yes instead of manual global install
- run go libs directly instead of installing and running binaries
- add missing brew dependencies


abelanger5 (approved):
Looks good 👍

PR #6, opened 2023-12-19
chore(contributing): adapt deps, hardcode generator versions

steebchen:
- remove sqlc brew dependency and use go install instead
- use hardcoded oapi-codegen version


abelanger5 (approved):
looks good!

PR #14, opened 2023-12-19
feat: add initial docs site

abelanger5:
Adds initial docs site with Nextra.


frontend/docs/pnpm-lock.yaml
```diff
@@ -0,0 +1,2172 @@
+lockfileVersion: 5.3
```
steebchen, line 1: We should decide for either pnpm or npm, I'm fine with either 😅 (I'm usually using pnpm these days for various reasons but mostly speed and symlinked node_modules, the only other thing is that using pure npm might be easier for onboarding).

abelanger5: Removed package lock in favor of `pnpm` in #35 , will use that moving forward 👍

PR #16, opened 2023-12-19
chore: add docker builds

abelanger5:


build/package/migrate.Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+# -------------------
+FROM golang:1.21-alpine as base
+WORKDIR /hatchet
+
+RUN apk update && apk add --no-cache gcc musl-dev git
+
+COPY go.mod go.sum ./
+COPY /prisma ./prisma
+
+RUN go install github.com/steebchen/prisma-client-go@v0.31.2
```
steebchen, line 11: @abelanger5 We should just do `go mod download`, otherwise this version will get out of sync with the one in `go.mod`.

PR #20, opened 2023-12-21
feat(events): add event search

steebchen:


frontend/app/src/pages/main/events/index.tsx
```diff
@@ -170,10 +172,6 @@ function EventsTable() {
   //   }
   // }, [listEventsQuery.data?.pagination]);
 
-  if (listEventsQuery.isLoading) {
-    return <Loading />;
-  }
```
steebchen, line 178: removes global loading and instead moves it to the table

internal/repository/prisma/event.go
```diff
@@ -50,6 +50,11 @@ func (r *eventRepository) ListEvents(tenantId string, opts *repository.ListEvent
 		TenantId: *pgTenantId,
 	}
 
+	if opts.Search != nil {
+		queryParams.Search.String = *opts.Search
```
abelanger5:
```suggestion
		queryParams.Search = sqlchelpers.TextFromStr(opts.Search)
```

^ can use this method which is meant to make this a little cleaner (setting `.Valid` isn't ideal) - just noticed that the `TextFromStr` method panics if the scan fails, which I'm not sure is ideal, but we also don't want to case on an error each time we do a simple string conversion, so perhaps it's alright?

internal/repository/prisma/event.go
```diff
@@ -50,6 +50,11 @@ func (r *eventRepository) ListEvents(tenantId string, opts *repository.ListEvent
 		TenantId: *pgTenantId,
 	}
 
+	if opts.Search != nil {
+		queryParams.Search = sqlchelpers.TextFromStr(opts.Search)
+		queryParams.Search.Valid = true
```
steebchen:
```suggestion
```

steebchen: @abelanger5 this can be removed right :D

abelanger5: Yep!

frontend/app/src/components/molecules/data-table/data-table-toolbar.tsx
```diff
@@ -22,26 +23,28 @@ interface DataTableToolbarProps<TData> {
...
     <div className="flex items-center justify-between">
       <div className="flex flex-1 items-center space-x-2">
-        {/* <Input
-          placeholder="Filter tasks..."
-          value={(table.getColumn("title")?.getFilterValue() as string) ?? ""}
-          onChange={(event) =>
-            table.getColumn("title")?.setFilterValue(event.target.value)
-          }
+        <Input
+          placeholder="Search for tasks..."
```
abelanger5:
```suggestion
          placeholder="Search for events..."
```

abelanger5 (changes requested):
Looks great - just one small language change 👍

abelanger5 (approved):

abelanger5 (approved):

PR #18, opened 2023-12-20
feat: add `hatchet-admin` command and improve dockerfiles

abelanger5:

PR #21, opened 2023-12-21
feat(seed): set a default tenant ID

steebchen:
This way the developer doesn't have to set the tenant ID always, it's just the same and already set.

### DIFF
--- a/api/v1/server/handlers/tenants/create.go
+++ b/api/v1/server/handlers/tenants/create.go
@@ -36,8 +36,8 @@ func (t *TenantService) TenantCreate(ctx echo.Context, request gen.TenantCreateR
 	}
 
 	createOpts := &repository.CreateTenantOpts{
-		Slug: string(request.Body.Slug),
-		Name: string(request.Body.Name),
+		Slug: request.Body.Slug,
+		Name: request.Body.Name,
 	}
 
 	// write the user to the db
--- a/cmd/hatchet-admin/cli/seed.go
+++ b/cmd/hatchet-admin/cli/seed.go
@@ -131,55 +131,52 @@ func seedDev(repo repository.Repository, tenantId string) error {
 		"message": "Above message is: {{ .steps.echo1.message }}",
 	})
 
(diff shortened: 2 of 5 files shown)


internal/config/client/client.go
```diff
@@ -26,6 +26,7 @@ type ClientConfig struct {
 }
 
 func BindAllEnv(v *viper.Viper) {
+	v.SetDefault("tenantId", "f790792c-cd0f-4304-91f3-3cb91feb019c")
```
steebchen: I wanted to re-use `DefaultTenantID` here but wasn't sure where if it's worth to do

abelanger5: So I'm using the `defaults` package elsewhere for this behavior, so you can set this declaratively in the ConfigFile struct: 

```go 
type ClientConfigFile struct {
	TenantId string `mapstructure:"tenantId" json:"tenantId,omitempty" default:"f790792c-cd0f-4304-91f3-3cb91feb019c"`

	TLS ClientTLSConfigFile `mapstructure:"tls" json:"tls,omitempty"`
}
```

abelanger5 (changes requested):
Check out the recent changes on `main` that were pushed in #18 - the `seed` command was moved and the database config now includes a seed config: [`SeedConfigFile`](https://github.com/hatchet-dev/hatchet/blob/14ecac8b81d4dfaaff6d9c0283ec99e6e4fb8e84/internal/config/database/config.go#L19). You can add the default ID here which should be more straightforward.

internal/config/client/client.go
abelanger5: I do think setting the client with a default tenant ID probably isn't great practice, since this will lead to a lot of confusing errors. I was thinking about changing the client implementation so that the client can use the tenant slug for connections instead of the UUID, which should address a lot of the issues here.

steebchen: > probably isn't great practice, since this will lead to a lot of confusing errors

Like what?

abelanger5 (approved):

PR #30, opened 2023-12-28
feat(go-sdk): add ability to create workflows from the Go SDK more easily, quickstart improvements

abelanger5:


pkg/worker/workflow_test.go
```diff
@@ -0,0 +1,48 @@
+package worker
+
+// import (
```
abelanger5, line 3: Refactor tests to test method args/output decoding

internal/config/loader/loaderutils/files.go
```diff
@@ -0,0 +1,33 @@
+package loaderutils
```
abelanger5, line 1: Create `loaderutils` so that server packages wouldn't leak into client packages, like they were when client was importing `loader` directly.

frontend/docs/theme.config.tsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
```
steebchen: might be better off to drop commit containing the docs page and leave it in #14?

internal/repository/prisma/worker.go
```diff
@@ -173,28 +209,22 @@ func (w *workerRepository) CreateNewWorker(tenantId string, opts *repository.Cre
...
+				db.Worker.ID.Equals(workerId),
+			).Update(
+				db.Worker.Actions.Link(
+					db.Action.TenantIDID(
+						db.Action.TenantID.Equals(tenantId),
+						db.Action.ID.Equals(action),
+					),
 				),
-			))
+			).Tx())
```
steebchen, line 224: ~~@abelanger5 can you please elaborate on this? I don't quite understand~~ (resolved via DM)

PR #37, opened 2023-12-30
ci(workflows): add build for dockerfiles workflow

steebchen:
It's a little verbose but I think it can't hurt to have these additional checks to make sure the dockerfiles always work, so we can potentially catch errors in these early, and not when there is a release about to happen.


.github/workflows/build.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Build frontend
+        run: docker build -f ./build/package/frontend.dockerfile .
+
+  servers:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v4
+      - name: Build servers
+        run: docker build -f ./build/package/servers.dockerfile . --build-arg SERVER_TARGET=api
```
abelanger5, line 18: Shouldn't we build for the other targets as well? They don't have quite the same dependencies

steebchen: I was wondering about that, but I didn't see any places where the contents are different, but indeed different things are build via the scripts. Fixed.

abelanger5 (approved):

PR #78, opened 2024-01-03
feat: update go-sdk workflow syntax to make it less verbose

abelanger5:


examples/go-sdk-workflow/main.go
```diff
@@ -57,25 +57,17 @@ func main() {
...
-				Function: StepTwo,
-				ID:       "step-two",
+	err = w.On(
+		worker.Events("user:create", "user:update"),
+		&worker.WorkflowJob{
+			Name:        "test-job",
+			Description: "This is a test job.",
+			Steps: []worker.WorkflowStep{
+				worker.Fn(StepOne).SetName("step-one"),
+				worker.Fn(StepTwo).SetName("step-two"),
```
steebchen, line 67: @abelanger5 interesting concept, I'm just wondering if this makes it harder to get started (this is one of the big things in the Go client as it needs a little time to discover the syntax, vs just using structs and living off autocomplete)

PR #82, opened 2024-01-04
feat: add middleware to worker and services

### LINKED ISSUE #80: feat(go-sdk): middleware support at the `worker` and `service` level
Support middleware that can wrap each action with access to the function's `context.Context`.

It should be possible to register this globally with `worker.Use`, and per-service with `service.Use`. 

For example:

```go
worker.Use(func (ctx context.Context, next func(ctx context.Context) error) error {
  // can set values in context, handle and throw errors, etc
  next(ctx)
  return nil
})
```

Panic middleware (#71) should be present by default, but overwritten as needed.

abelanger5:
Fixes #80


steebchen (approved):
Looks solid! 💯

examples/go-sdk-workflow/main.go
```diff
@@ -57,7 +62,38 @@ func main() {
...
+	testSvc := w.NewService("test")
+
+	testSvc.Use(func(ctx context.Context, next func(context.Context) error) error {
+		err := next(context.WithValue(ctx, "svckey", "svcvalue"))
+
+		if err != nil {
+			return err
+		}
+
+		return nil
```
steebchen: Could be shortened to `return next(...)`, but maybe it's better to keep it verbose as it is an example

frontend/docs/pages/go-sdk/creating-a-workflow.mdx
```diff
@@ -149,4 +149,39 @@ w.On(
...
+```go
+w.Use(func(ctx context.Context, next func(context.Context) error) error {
+	err := next(context.WithValue(ctx, "testkey", "testvalue"))
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+})
```
steebchen: Uses different indentation. I will fix this eventually with automation using editorconfig and editorconfig lint CI

PR #92, opened 2024-01-10
fix: rabbitmq concurrent processing

abelanger5:
There were several issues with the previous implementation of the task queue - messages were sent and processed synchronously from RabbitMQ and the sent `msg` was using implicit memory aliasing which was causing some duplicate messages/dropped messages

### DIFF
--- a/internal/repository/prisma/dbsqlc/job_runs.sql
+++ b/internal/repository/prisma/dbsqlc/job_runs.sql
@@ -39,8 +39,8 @@ END, "finishedAt" = CASE
 END, "startedAt" = CASE 
     -- Started at is final, cannot be changed
     WHEN "startedAt" IS NOT NULL THEN "startedAt"
-    -- If a step is running, then the job has started
-    WHEN s.runningRuns > 0 THEN NOW()
+    -- If steps are running (or have finished), then set the started at time
+    WHEN s.runningRuns > 0 OR s.succeededRuns > 0 OR s.failedRuns > 0 AND s.cancelledRuns > 0 THEN NOW()
     ELSE "startedAt"
 END
 FROM stepRuns s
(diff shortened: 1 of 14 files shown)


internal/repository/prisma/dbsqlc/job_runs.sql
```diff
@@ -39,8 +39,8 @@ END, "finishedAt" = CASE
 END, "startedAt" = CASE 
     -- Started at is final, cannot be changed
     WHEN "startedAt" IS NOT NULL THEN "startedAt"
-    -- If a step is running, then the job has started
-    WHEN s.runningRuns > 0 THEN NOW()
+    -- If steps are running (or have finished), then set the started at time
+    WHEN s.runningRuns > 0 OR s.succeededRuns > 0 OR s.failedRuns > 0 AND s.cancelledRuns > 0 THEN NOW()
```
abelanger5, line 43: Small hotfix for `startedAt` not always being populated - I believe this is because `PENDING_ASSIGNMENT`, `RUNNING` and `COMPLETED` status updates may be processed out-of-order, but this will require a deeper investigation.

internal/services/dispatcher/dispatcher.go
```diff
@@ -168,11 +168,13 @@ func (d *DispatcherImpl) Start(ctx context.Context) error {
 
 			return err
 		case task := <-taskChan:
-			err = d.handleTask(ctx, task)
+			go func(task *taskqueue.Task) {
```
abelanger5, line 171: Task queue consumers should process tasks in a separate goroutine so they don't block.

internal/taskqueue/rabbitmq/rabbitmq.go
```diff
@@ -233,7 +230,10 @@ func (t *TaskQueueImpl) publish() {
 // }
 
 func (t *TaskQueueImpl) subscribe(ctx context.Context, subId, queue string, sessions chan chan session, messages chan *taskqueue.Task, tasks chan<- *taskqueue.Task) {
+	sessionCount := 0
+
 	for session := range sessions {
+		sessionCount++
```
abelanger5, line 236: It's good to track session count for debugging, because high session values may indicate some issue with the rabbit connection.

PR #97, opened 2024-01-10
feat(test): introduce basic e2e tests

steebchen:


examples/simple/simple_test.go
```diff
@@ -0,0 +1,27 @@
...
+func TestSimpleRun(t *testing.T) {
+	testutils.Prepare(t)
+
+	ch := make(chan interface{}, 1)
+
+	go func() {
+		time.Sleep(5 * time.Second)
+		ch <- struct{}{}
+		log.Printf("sent interrupt")
+	}()
```
steebchen: this just makes it work for now (so that test exits)

examples/simple/simple_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+	if err := run(ch); err != nil {
+		t.Fatalf("run() error = %v", err)
+	}
+
+	// TODO test here that the workflow ran by checking the database
```
steebchen: This is obviously missing and important, but I'm not sure about the best way to do it right now

examples/action-old/main.go
```diff
@@ -0,0 +1,120 @@
+package main
```
steebchen, line 1: we might also just delete this package now @abelanger5

abelanger5: I'm updating the `RegisterAction` usage today, so let's just keep it and I'll fix this case.

abelanger5 (approved):

PR #109, opened 2024-01-17
feat: python client

abelanger5:
Adds initial python SDK. Docs still to come.


python-client/examples/dag/worker.py
```diff
@@ -0,0 +1,40 @@
+from hatchet import workflow, step, Worker, Context
+
+@workflow(on_events=["user:create"])
```
grutt: nit style choice: most of the client libs I've used in Python will instantiate (or import) an object with config and then call decorators on that object.

FastAPI/Flask/Celery all use a similar pattern. I've had some instances where I've needed multiple Celery instances in one app because of bad decisions.

```py
hatchet = Hatchet(config)

@hatchet.workflow(...)
class MyWorkflow:

...

    @hatchet.step()
    def step()
        ...
```

python-client/hatchet/client.py
```diff
@@ -0,0 +1,82 @@
...
+from .clients.admin import AdminClientImpl, new_admin
+from .clients.events import EventClientImpl, new_event
+from .clients.dispatcher import DispatcherClientImpl, new_dispatcher
+
+from .loader import ConfigLoader, ClientConfig
+import grpc
+from dotenv import load_dotenv
+
+# TODO: don't load dotenv
+load_dotenv()
```
grutt: not sure if you missed removing this

python-client/hatchet/client.py
```diff
@@ -0,0 +1,82 @@
...
+
+    if config.tls_config is None:
+        raise ValueError("TLS config is required")
+    
+    if config.host_port is None:
+        raise ValueError("Host and port are required")
+        
+    root = open(config.tls_config.ca_file, "rb").read()
+    private_key = open(config.tls_config.key_file, "rb").read()
+    certificate_chain = open(config.tls_config.cert_file, "rb").read()
```
grutt: question: I noticed this in the initial go repo as well, is there a reason you're using this auth scheme instead of something simpler for the user?

python-client/hatchet/clients/admin.py
```diff
@@ -0,0 +1,80 @@
...
+            if e.code() == grpc.StatusCode.NOT_FOUND:
+                should_put = True
+            else:
+                raise ValueError(f"Could not get workflow: {e}")
+
+        # Determine if we should put the workflow
+        should_put = False
+        if auto_version:
+            if workflow.version == "":
+                workflow.version = "v0.1.0"
```
grutt: should should_put be True here?

python-client/hatchet/workflow.py
```diff
@@ -0,0 +1,110 @@
...
+   
+   return inner
+
+def step(name : str='', parents : List[str] = []):
+    def inner(func):
+        func._step_name = name or func.__name__
+        func._step_parents = parents
+
+        # print length of parents
+        print("len(parents)", len(parents), len(func._step_parents))
```
grutt: rm debug code?

python-client/hatchet/workflow.py
```diff
@@ -0,0 +1,110 @@
...
+from typing import Callable, List, Tuple, Any
+
+# TODO: don't load dotenv
+load_dotenv()
+
+def workflow(name : str='', on_events : list=[], on_crons : list=[]):
+   def inner(cls):
+        cls.on_events = on_events
+        cls.on_crons = on_crons
+        cls.name = name or str(cls.__name__)
```
grutt: i don't remember if an empty string evaluates to false in python

grutt: weird it does.

python-client/hatchet/workflow.py
```diff
@@ -0,0 +1,110 @@
...
+                readable_id=func_name,
+                action="default:" + func_name,
+                timeout="60s",
+                inputs='{}',
+                parents=[x for x in func._step_parents]  # Assuming this is how you get the parents
+            ) 
+            for func_name, func in attrs.items() if hasattr(func, '_step_name')
+        ]
+
+        print("createStepOpts", createStepOpts)
```
grutt: rm debug code

python-client/hatchet/workflow.py
```diff
@@ -0,0 +1,110 @@
...
+
+#     return func
+    
+
+# Initialize the counter for ordering steps
+step.order_counter = 0
+
+stepsType = List[Tuple[str, Callable[..., Any]]]
+
+class WorkflowMeta(type):
```
grutt: what is the point of this?

python-client/hatchet/client.py
grutt: oh looks like its required for grpc?

grutt (commented):
Overall makes sense and clear interface. Added a small suggestion and commented on things you might have overlooked while moving fast.

python-client/hatchet/workflow.py
abelanger5: I wasn't sure of a better/more pythonic way of tracking the attributes of the workflow class, like the steps, without using a metaclass, along with creating the workflow over the API when the child class is instantiated, which is why I modified `__new__`. As far as I understand, one use-case for metaclasses is adding classes/attributes of the class to separate data structures and potentially modifying the behavior of the child class. This could probably be done with a decorator function that modifies the class directly, I'm not sure which approach is better though.

python-client/hatchet/client.py
abelanger5: yeah part of it is that it's easier on the server side - if the server trusts the client certificates we don't need to add extra auth, which is nice for standalone tenant instances. the other thing is that it gets us message encryption. we'll probably want API tokens at some point for querying workflows/hatchet objects, but I think for worker <> engine communication it works well.

grutt (approved):
looks good enough for an initial merge

PR #110, opened 2024-01-18
feat: typescript client

grutt:
Depends on #109 #129 
References #107 #123

**Problem**
Hatchet users want to use hatchet within their typescript projects

**Solve**
A simple, typescript package published on NPM

**Tasks**
- [x] initialize typescript project
- [x] generate types from protobuf
- [x] client class
- [x] worker class
- [x] workflow class
- [x] dispatcher class
- [x] configuration 
- [x] clients: admin
- [x] clients: dispatcher
- [x] clients: dispatcher/action-listener
- [x] clients: events 
- [x] examples: dag
- [x] examples: simple
- [x] docs
- [x] tests
- [x] logging
- [x] github actions (tests and publish)
- [x] auto versioning 
- [x] clean up
- [x] swap to token

**Questions/Thoughts/Unknowns/Risks**
- This library, as written, will not work with serverless frameworks
- Should we automate Protobuf generation on CI/CD to prevent breaking?
- Permission needed to publish to NPM, this should likely be handled by CI/CD

**No Gos**
- Next route invocation wrapper (separate SDK/PR)

**Success**
- An example project can register workflows, invoke steps from events, and dispatch events.


typescript-client/hatchet/clients/dispatcher/action-listener.ts
```diff
@@ -0,0 +1,92 @@
...
+    this.client = client.client;
+    this.listener = listener;
+    this.workerId = workerId;
+  }
+
+  // actions
+  // parse_action_payload
+
+  actions = () =>
+    (async function* gen(client: ActionListener) {
```
grutt: I am not sure this is the most optimal way of writing this. I havent worked with async generators in js before and am surprised how similar it is to python. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols

@abelanger5 @steebchen did you have any thoughts on this?

grutt:
@abelanger5  @steebchen  this is chonky PR is ready for review! 

I still need to fix the auto versioning per #123 and cleanup a few wonky things, mostly in `worker.ts`. 

You can run with `pnpm install`, start the service with `pnpm worker:simple` or `pnpm worker:dag` and dispatch an event with `pnpm example:event`

Tests can be run with `pnpm test`

frontend/docs/pages/typescript-sdk/creating-a-worker.mdx
```diff
@@ -0,0 +1,35 @@
+# Creating a Worker
+
+Workers can be created via the `hatchet.run()` method, after instantiating a `hatchet` instance.
```
abelanger5, line 3: Love the `hatchet.run`, going to do this on all the clients.

frontend/docs/pages/typescript-sdk/creating-a-worker.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+For example:
+
+```ts
+import Hatchet from '@hatchet/sdk';
+import { Workflow } from '@hatchet/sdk';
+
+const hatchet = Hatchet.init();
+
+# workflow code...
```
abelanger5: This isn't a valid TS comment - also the example should use `dotenv`, no?

frontend/docs/pages/typescript-sdk/creating-a-worker.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+```ts
+const worker = hatchet.run(workflow);
+worker.stop();
+```
+
+Alternatively, you can create a worker object with `hatchet.worker()` and manually start the service.
+
+```ts
+const worker = hatchet.worker(workflow);
```
abelanger5, line 33: It's confusing that both `hatchet.run` and `hatchet.worker` seem to return a worker object - what is the difference? And how can I name workers?

frontend/docs/pages/typescript-sdk/creating-a-workflow.mdx
```diff
@@ -0,0 +1,113 @@
...
+}
+```
+
+Here's an example `Step` which accesses the workflow input:
+
+```ts
+const stepPrintsInput: Step = {
+  name: "step2",
+  parents: ["step1"],
+  run: (input, ctx) => {
```
abelanger5, line 58: Hmm, why do we have the `input` arg? I thought this was a convenience object for `ctx.workflow_input()` but it seems different, so this is confusing me.

frontend/docs/pages/typescript-sdk/creating-a-workflow.mdx
```diff
@@ -0,0 +1,113 @@
...
+    console.log("executed step2!", ctx.workflow_input());
+  },
+};
+```
+
+## Step Outputs
+
+Step outputs should be a of type `Record<string, any>`, should be `JSON` serializable, and are optional. For example:
+
+```py
```
abelanger5:
```suggestion
```ts
```

frontend/docs/pages/typescript-sdk/creating-a-workflow.mdx
```diff
@@ -0,0 +1,113 @@
...
+};
+```
+
+You'll notice that the workflow defines a workflow trigger (in this case, `on_events`), and the workflow definition. The workflow definition includes a series of steps which is simply an array of `Step` objects. Each step has a `run` prop which is a function that takes an `input` and a `context`. The `context` argument is a `Context` object, which contains information about the workflow, such as the input data and the output data of previous steps.
+
+To create multi-step workflows, you can use `parents` to define the steps which the current step depends on. In the example, `step2` will not invoke until after `step1` completes.
+
+## Getting Access to the Input Data
+
+You can get access to the workflow's input data, such as the event data or other specified input data, by using the `context.workflow_input()` method on the `context`. For example, given the following event:
```
abelanger5:
```suggestion
You can get access to the workflow's input data, such as the event data or other specified input data, by using the `context.workflowInput()` method on the `context`. For example, given the following event:
```

As much as I don't like the inconsistency, I think it's important that we camelCase here and snake_case in python. It just feels more natural and will probably more closely align with most people's linters.

typescript-client/hatchet/clients/worker/worker.ts
```diff
@@ -0,0 +1,228 @@
...
+
+    this.logger.info('Gracefully exiting hatchet worker...');
+
+    try {
+      this.listener?.unregister();
+    } catch (e: any) {
+      this.logger.error(`Could not unregister listener: ${e.message}`);
+    }
+
+    // TODO wait for futures to finish
```
abelanger5: address TODO?

typescript-client/hatchet/workflow.ts
```diff
@@ -0,0 +1,30 @@
+import * as z from 'zod';
+
+import { CreateStep, CreateStepSchema } from './step';
+
+const CronConfigSchema = z.object({
+  cron: z.string(), // TODO validate cron
```
abelanger5: This is validated on the backend, frontend would be nice but not strictly necessary

abelanger5 (approved):
Left two minor comments but the rest looks good!

frontend/docs/pages/typescript-sdk/setup.mdx
```diff
@@ -22,13 +22,10 @@ yarn add @hatchet/node-sdk
 
 ## Usage
 
-You should have generated a set of client certificates (or received them through Hatchet cloud). These certificates are `ca.cert`, `client-worker.pem`, and `client-worker.key`. You should also make note of the Hatchet server address and the tenant ID you'd like to connect to.
+Navigate to `https://app.dev.hatchet-tools.com/tenant-settings` to generate an API Token and set the following environment variables.
 
 ```sh
 HATCHET_CLIENT_TENANT_ID=<tenant-id>
```
abelanger5:
```suggestion
```

abelanger5 (approved):

typescript-sdk/hatchet/step.ts
```diff
@@ -0,0 +1,42 @@
...
+    return this.data?.triggered_by_event === 'event';
+  }
+
+  workflowInput(): any {
+    return this.data?.input || {};
+  }
+}
+
+export interface CreateStep<T> extends z.infer<typeof CreateStepSchema> {
+  run: (input: T, ctx: Context) => Promise<NextStep> | NextStep | void;
```
steebchen: I'd maybe switch `(input: T, ctx: Context)`

steebchen (approved):

PR #112, opened 2024-01-21
feat: google sso

abelanger5:
- Many file changes so it becomes much easier to add a new SSO provider in the future (assuming oauth). This includes a server `/api/v1/metadata` endpoint to query for the permitted auth schemes on the Hatchet instance. 
- Frontend is dynamic so it will change the login screen based on the registered providers. This can be changed with the `SERVER_AUTH_GOOGLE_ENABLED` and `SERVER_AUTH_BASIC_AUTH_ENABLED` flags at the moment. Here are the variants:

**Basic + Google Login:**

image

**Just basic:**

image

**Just Google:**

image

### DIFF
--- a/frontend/app/src/pages/authenticated.tsx
+++ b/frontend/app/src/pages/authenticated.tsx
@@ -87,17 +87,3 @@ export default function Authenticated() {
     </div>
   );
 }
-
-// function Authenticated() {
-//   const ctx = useOutletContext<UserContextType>();
-//   const { user } = ctx;
-
-//   return (
-//     <>
-//       <MainNav user={user} />
-//       <Outlet />
-//     </>
-//   );
-// }
-
-// export default Authenticated;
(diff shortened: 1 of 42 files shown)


internal/config/server/server.go
```diff
@@ -133,6 +150,10 @@ func BindAllEnv(v *viper.Viper) {
 	v.BindEnv("auth.cookie.domain", "SERVER_AUTH_COOKIE_DOMAIN")
 	v.BindEnv("auth.cookie.secrets", "SERVER_AUTH_COOKIE_SECRETS")
 	v.BindEnv("auth.cookie.insecure", "SERVER_AUTH_COOKIE_INSECURE")
+	v.BindEnv("auth.google.enabled", "SERVER_AUTH_GOOGLE_ENABLED")
```
steebchen: if we want to ignore explicitly, we should do `_ = v.BindEnv(...)`

frontend/app/src/pages/authenticated.tsx
```diff
@@ -77,5 +78,26 @@ export default function Auth() {
...
+
+//   return (
+//     <>
+//       <MainNav user={user} />
+//       <Outlet />
+//     </>
+//   );
+// }
+
+// export default Authenticated;
```
steebchen: cleanup

steebchen (approved):

PR #114, opened 2024-01-21
feat(events): add workflow filter

steebchen:


frontend/app/src/pages/main/events/components/event-columns.tsx
```diff
@@ -78,6 +78,13 @@ export const columns = ({
         return <div>{relativeDate(row.original.metadata.createdAt)}</div>;
       },
     },
+    {
+      accessorKey: 'workflows',
+      header: () => <></>,
+      cell: () => {
+        return <div></div>;
+      },
+    },
```
steebchen, line 87: Not so proud of this part, when I started I figured it was the easiest way to add it, but not sure if we actually want to show workflow names as it could be a lot. So I just don't show them right now, but the column has to exist here, otherwise the filter won't work with the existing column filter logic.

frontend/app/src/pages/main/events/index.tsx
```diff
@@ -123,6 +123,16 @@ function EventsTable() {
...
+  const workflows = useMemo(() => {
+    const filter = columnFilters.find((filter) => filter.id === 'workflows');
+
+    if (!filter) {
+      return;
+    }
+
+    return filter?.value as Array<string>;
+  }, [columnFilters]);
+
```
steebchen, line 135: Also not sure if this gets too repetitive as I just copied it from above

frontend/app/src/pages/main/events/index.tsx
```diff
@@ -173,7 +184,24 @@ function EventsTable() {
...
+  });
+
+  const workflowKeyFilters = useMemo((): FilterOption[] => {
+    return (
+      workflowKeys?.rows?.map((key) => ({
+        value: key.metadata.id,
+        label: key.name,
+      })) || []
+    );
+  }, [workflowKeys]);
```
steebchen, line 204: same regarding the copying/repetitiveness

internal/repository/prisma/dbsqlc/events.sql
```diff
@@ -3,12 +3,28 @@ SELECT
...
+    events."key" = ANY(sqlc.narg('keys')::text[])
+    ) AND
+  (
+    sqlc.narg('workflows')::uuid[] IS NULL OR
+    workflow."id" = ANY(sqlc.narg('workflows')::uuid[])
+    ) AND
+  (
+    sqlc.narg('search')::text IS NULL OR
+    jsonb_path_exists(events."data", cast(concat('$.** ? (@.type() == "string" && @ like_regex "', sqlc.narg('search')::text, '")') as jsonpath))
+  );
```
steebchen: I figured we need to return the count of the current search, so this should also include all the existing and the new filter. Is this correct? @abelanger5

internal/repository/prisma/event.go
```diff
@@ -79,6 +80,15 @@ func (r *eventRepository) ListEvents(tenantId string, opts *repository.ListEvent
 		countParams.Keys = opts.Keys
 	}
 
+	if opts.Workflows != nil {
+		var wfs []pgtype.UUID
+		for _, wf := range opts.Workflows {
+			wfs = append(wfs, sqlchelpers.UUIDFromStr(wf))
+		}
+		queryParams.Workflows = wfs
+		countParams.Workflows = wfs
```
steebchen, line 89: alternatively we can just do

```
queryParams.Workflows = opts.Workflows
```

but then we need to cast the `sqlc.narg('workflows')::uuid[]` to a string, not sure which one is uglier

internal/repository/prisma/dbsqlc/events.sql
abelanger5: Yep, that's right 👍

frontend/app/src/pages/main/events/components/event-columns.tsx
abelanger5: This seems fine, but would probably make sense to add a comment to the codebase so future maintainers are aware that it shouldn't be removed?

frontend/app/src/pages/main/events/index.tsx
abelanger5: Honestly seems fine, it's pretty readable

abelanger5 (approved):

PR #118, opened 2024-01-22
fix: python sdk graceful shutdown and retry errors

abelanger5:


python-client/hatchet_sdk/clients/dispatcher.py
```diff
@@ -23,8 +24,10 @@ def get_action_listener(self, ctx, req):
     def send_action_event(self, ctx, in_):
         raise NotImplementedError
 
-DEFAULT_ACTION_LISTENER_RETRY_INTERVAL = 5  # seconds
+DEFAULT_ACTION_LISTENER_RETRY_INTERVAL = 1  # seconds
```
grutt, line 27: should this be configurable in the yaml?

python-client/hatchet_sdk/clients/dispatcher.py
```diff
@@ -166,15 +175,15 @@ def get_action_listener(self, req: GetActionListenerRequest) -> ActionListenerIm
...
-        ))
+        ), timeout=DEFAULT_REGISTER_TIMEOUT)
 
         # Subscribe to the worker
-        listener = self.client.Listen(WorkerListenRequest(
-            tenantId=self.tenant_id,
-            workerId=response.workerId,
-        ))
+        # listener = self.client.Listen(WorkerListenRequest(
+        #     tenantId=self.tenant_id,
```
grutt: was this meant to be commented out?

python-client/hatchet_sdk/clients/dispatcher.py
abelanger5: Hmm, definitely eventually - going to leave it as-is for now though, since I'm not sure if we add exponential backoff before making this configurable.

python-client/hatchet_sdk/clients/dispatcher.py
abelanger5: Nice catch - I actually meant to remove it, because the action listener handles its own `.Listen` invocation now.

grutt (approved):

PR #129, opened 2024-01-26
feat: improve versioning to use checksums

abelanger5:
This uses `sha512` to compute checksums for the declared workflow, instead of offloading versioning logic to clients.

### DIFF
--- a/internal/digest/compute_test.go
+++ b/internal/digest/compute_test.go
@@ -1,0 +1,70 @@
+package digest_test
+
+import (
+	"testing"
+
+	"github.com/hatchet-dev/hatchet/internal/digest"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDigestValues(t *testing.T) {
+	tests := []struct {
+		name   string
+		values map[string]interface{}
+	}{
+		{
+			name:   "Empty map",
+			values: map[string]interface{}{},
+		},
+		{
+			name: "Single key-value pair",
+			values: map[string]interface{}{
+				"key": "value",
+			},
+		},
+		{
+			name: "Multiple key-value pairs",
+			values: map[string]interface{}{
+				"a": 1,
+				"b": 2,
+				"c": 3,
+			},
+		},
+		{
+			name: "Nested map",
+			values: map[string]interface{}{
+				"parent": map[string]interface{}{
+					"child1": 1,
+					"child2": 2,
+				},
+			},
+		},
+	}
+
+	for _, tc := range tests {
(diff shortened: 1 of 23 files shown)


internal/digest/compute_test.go
```diff
@@ -0,0 +1,70 @@
...
+			name: "Single key-value pair",
+			values: map[string]interface{}{
+				"key": "value",
+			},
+		},
+		{
+			name: "Multiple key-value pairs",
+			values: map[string]interface{}{
+				"a": 1,
+				"b": 2,
```
grutt, line 30: we talked about a case where keys are in different orders, do we want this to hash to the same or different values and cover?

prisma/migrations/20240126235456_v0_7_0/migration.sql
```diff
@@ -33,6 +37,10 @@ ADD COLUMN     "accessToken" BYTEA NOT NULL,
 DROP COLUMN "refreshToken",
 ADD COLUMN     "refreshToken" BYTEA;
 
+-- AlterTable
+ALTER TABLE "WorkflowVersion" ADD COLUMN     "checksum" TEXT NOT NULL,
```
grutt: why dont we just default to a random string so we can migrate existing volumes and it will just update on next seen version?

python-client/hatchet_sdk/clients/admin.py
```diff
@@ -19,54 +19,16 @@ def __init__(self, client : WorkflowServiceStub, token):
...
-        
-        existing_workflow : Workflow = None
-
-        # Get existing workflow by name
+    def put_workflow(self, workflow: CreateWorkflowVersionOpts):
         try:
-            existing_workflow : Workflow = self.client.GetWorkflowByName(
-                GetWorkflowByNameRequest(
-                    name=workflow.name,
+            self.client.PutWorkflow(
```
grutt, line 24: this makes me happy

grutt (changes requested):
only thing i'd change is the migration so that i don't have to recycle my docker volumes :)

internal/digest/compute_test.go
abelanger5: Keys in different orders should map to the same digest no? There's really no meaning in key ordering in Go maps since they're accessed randomly when iterated anyway, neither should there be meaning behind key ordering in JSON/YAML.

prisma/migrations/20240126235456_v0_7_0/migration.sql
abelanger5: Done.

PR #135, opened 2024-01-28
feat: concurrency groups

abelanger5:
# Description

Feature to add concurrency groups logic to Hatchet. This is a public-facing feature and will be well-documented in `./frontend/docs`. 

## Type of change

- [X] This change requires a documentation update

# What's Changed

- [X] Move `controllers` into their own folder to improve naming, i.e. `controllers/events` instead of `eventscontroller`
- [X]  Add `WorkflowConcurrency` model to the database
- [X]  A new `QUEUED` status for workflows which is used to manage the postgres-backed queue
- [X] A new type of action which retrieves a concurrency group ID from the client. This is referenced on the API side as `getGroupKeyRun`
- [X] A new controller for workflows which handles queueing 

**SDKs - Python**

- [X] Add decorator method called `hatchet.concurrency` which can decorate a workflow function to get the concurrency group based on the input. 
- [X] Test for backwards compatibility

**SDKs - Golang**

- [X] Add `Concurrency` field to `WorkflowJob`
- [X] Test for backwards compatibility

**Not implemented**

- Different queueing strategies. This PR only focuses on `CANCEL_IN_PROGRESS`, which doesn't actually need a queue at all since you just cancel and replace running workflows.


abelanger5:
Improvements before merge:
- [ ] When step runs are cancelled, ensure any associated timeouts are also cancelled
- [ ] Ensure that workflows runs can't be in a queued state forever (i.e. set a max length on the workflow queue). The timeout on the `getGroupKeyRun` should handle this, but there may be situations where the tickers redeploy and we miss the cancellation.

PR #150, opened 2024-02-05
feat: custom claims for setting grpc/server addr in token

abelanger5:
# Description

Removes the requirement of setting `HATCHET_CLIENT_HOST_PORT` as an environment variable in all of the SDKs. This instead adds custom claims of `server_url` and `grpc_broadcast_address` which are used for connections. 

Security implications of this approach were considered - one could consider a MITM attack which rewrites the claims to trick the user into sending their token to their server. Our SDKs require TLS to minimize the risk, and if an attacker has constructed a MITM attack during token generation time, they would likely be able to modify the payload which shows the broadcast address to the user anyway. This is also fairly standard practice, i.e. GCP credentials files have a set of auth URIs set:

```json
"auth_uri": "https://accounts.google.com/o/oauth2/auth",
"token_uri": "https://oauth2.googleapis.com/token",
"auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs"
```

## Type of change

# What's Changed

This is backwards-compatible with older tokens. 

- [X] SDKs config loaders for all 3 SDKs 
- [X] New claims in tokens
- [X] Docs updates to remove requirement of `HATCHET_CLIENT_HOST_PORT`

### DIFF
--- a/internal/config/server/server.go
+++ b/internal/config/server/server.go
@@ -49,6 +49,9 @@ type ConfigFileRuntime struct {
 	// GRPCBindAddress is the address that the grpc server binds to. Should set to 0.0.0.0 if binding in docker container.
 	GRPCBindAddress string `mapstructure:"grpcBindAddress" json:"grpcBindAddress,omitempty" default:"127.0.0.1"`
 
+	// GRPCBroadcastAddress is the address that the grpc server broadcasts to, which is what clients should use when connecting.
+	GRPCBroadcastAddress string `mapstructure:"grpcBroadcastAddress" json:"grpcBroadcastAddress,omitempty" default:"127.0.0.1:7070"`
+
 	// GRPCInsecure controls whether the grpc server is insecure or uses certs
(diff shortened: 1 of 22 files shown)


internal/config/server/server.go
```diff
@@ -49,6 +49,9 @@ type ConfigFileRuntime struct {
 	// GRPCBindAddress is the address that the grpc server binds to. Should set to 0.0.0.0 if binding in docker container.
 	GRPCBindAddress string `mapstructure:"grpcBindAddress" json:"grpcBindAddress,omitempty" default:"127.0.0.1"`
 
+	// GRPCBroadcastAddress is the address that the grpc server broadcasts to, which is what clients should use when connecting.
+	GRPCBroadcastAddress string `mapstructure:"grpcBroadcastAddress" json:"grpcBroadcastAddress,omitempty" default:"127.0.0.1:7070"`
```
grutt, line 53: need to set this on the go-quickstart docker-compose?

PR #156, opened 2024-02-06
feat: add custom user data to each step

abelanger5:
# Description

Adds a `userData` field which can be on the step to customize behavior. This also adds a `registerAction` method to the Typescript SDK.

## Type of change

# What's Changed

- [X] New DB column on the `Step` table called `customUserData`, to clarify that this is set by the end user and not us 
- [X] New param in the `CreateWorkflowVersionOpts` called `user_data` (on the `protoc` side) for each step
- [X] `registerAction` method for the Typescript worker


typescript-sdk/.gitignore
```diff
@@ -1,2 +1,5 @@
 /dist
-/certs
\ No newline at end of file
+/certs
+
+# ignore docs for now, because we manually use the docs output to construct docs
```
grutt, line 4: you mean we have to copy the output into the fe/docs?

typescript-sdk/package.json
```diff
@@ -35,14 +35,16 @@
     "event:concurrency": "npm run exec -- ./examples/concurrency/concurrency-event.ts",
     "api": "npm run exec -- ./examples/api.ts",
     "prepublish": "cp package.json dist/package.json;",
-    "publish:ci": "rm -rf ./dist && npm run tsc:build && npm run prepublish && cd dist && npm publish --access public --no-git-checks"
+    "publish:ci": "rm -rf ./dist && npm run tsc:build && npm run prepublish && cd dist && npm publish --access public --no-git-checks",
+    "generate-docs": "typedoc"
```
grutt, line 39: cp to frontend && rm?

typescript-sdk/src/clients/dispatcher/dispatcher-client.test.ts
```diff
@@ -39,6 +41,8 @@ describe('DispatcherClient', () => {
           ca_file: 'TLS_ROOT_CA_FILE',
           server_name: 'TLS_SERVER_NAME',
         },
+        api_url: 'API_URL',
+        tenant_id: 'tenantId',
```
grutt, line 45: note for future: we should probably put the config in a root fixtures dir or something bc its a pain to propagate these changes

grutt (approved):

typescript-sdk/.gitignore
abelanger5: no, it's much worse than that. we have to copy over only the `AdminClient.ts` file and the `interfaces` folder, and then change all relative imports and remove some sections from the autogenerated file which aren't useful to the user (like the constructor). I imagine this is all configurable but don't have time to fully automate this.

typescript-sdk/package.json
abelanger5: See above, this isn't automated

PR #172, opened 2024-02-18
feat(engine): standalone tests and engine teardown

### LINKED ISSUE #174: add proper teardown functions for engine

steebchen:
# Description

Runs load tests in a separate CI action and runs the engine from within the tests. Also adds proper engine teardown functions.

## Type of change

- [x] Test suite

# What's Changed

- run load tests separately
- run the engine from inside the test
- tear down engine in the correct order via cleanup functions
- eliminate all goroutine leaks


internal/taskqueue/rabbitmq/rabbitmq.go
```diff
@@ -341,6 +346,8 @@ outer:
 		}
 	}
 
+	wg.Wait()
```
steebchen: potentially we could also wait here by returning a cleanup function

abelanger5 (approved):

PR #197, opened 2024-02-26
feat(engine): readiness & liveness probes

### LINKED ISSUE #111: add readiness and liveness probes
Readiness probe should check for database and AMQP connections, liveness should just return 200 when ready.

steebchen:
# Description

Adds routes for readiness and liveness.

Fixes #111

## Type of change

## What's Changed

- [x] adds /live and /ready routes in the engine on port 8733
- [x] checks for db + queue connectivity
- [x] waits for setup and considers shutdown, configurable via env var SERVER_SHUTDOWN_WAIT


internal/config/server/server.go
```diff
@@ -28,7 +28,7 @@ type ServerConfigFile struct {
 
 	TaskQueue TaskQueueConfigFile `mapstructure:"taskQueue" json:"taskQueue,omitempty"`
 
-	Services []string `mapstructure:"services" json:"services,omitempty" default:"[\"ticker\", \"grpc\", \"eventscontroller\", \"jobscontroller\", \"workflowscontroller\", \"heartbeater\"]"`
+	Services []string `mapstructure:"services" json:"services,omitempty" default:"[\"health\", \"ticker\", \"grpc\", \"eventscontroller\", \"jobscontroller\", \"workflowscontroller\", \"heartbeater\"]"`
```
steebchen, line 31: I figured I also create a service for this. Not sure if this is needed but I figured everything else is configured as a service. Also not sure if there is any other place where I need to put this.

internal/services/health/health.go
```diff
@@ -0,0 +1,62 @@
...
+	mux.HandleFunc("/ready", func(w http.ResponseWriter, r *http.Request) {
+		if h.ready {
+			w.WriteHeader(http.StatusOK)
+		} else {
+			w.WriteHeader(http.StatusServiceUnavailable)
+		}
+	})
+
+	server := &http.Server{
+		Addr:         ":8733",
```
steebchen, line 49: randomly used 8733

cmd/hatchet-engine/engine/run.go
```diff
@@ -237,8 +249,18 @@ func Run(ctx context.Context, cf *loader.ConfigLoader) error {
...
+		h.SetReady(true)
+	}
+
 	<-ctx.Done()
 
+	if healthProbes {
+		h.SetReady(false)
+	}
+
+	time.Sleep(10 * time.Second)
```
steebchen: This is not needed, however I was thinking that there is most likely a delay which is at least a few seconds until any orchestrator (k8s or compose) registers this. So I figured it would be best to set the probe false first, give the orchestrator some time to register, and only then start with the actual shutdown of services.

cmd/hatchet-engine/engine/run.go
```diff
@@ -39,6 +41,16 @@ func Run(ctx context.Context, cf *loader.ConfigLoader) error {
 
 	var teardown []Teardown
 
+	h := health.New()
+	healthProbes := sc.HasService("health")
```
steebchen, line 45: I figured I also create a service for this. Not sure if this is needed but I figured everything else is configured as a service. Also not sure if there is any other place where I need to put this.

internal/taskqueue/rabbitmq/rabbitmq.go
```diff
@@ -378,6 +381,18 @@ func redial(ctx context.Context, l *zerolog.Logger, url string) chan chan sessio
 			}
 
 			newSession, err := getSession(ctx, l, url)
+			t.ready = true
+
+			ch := newSession.Connection.NotifyClose(make(chan *amqp.Error, 1))
```
steebchen, line 386: This seems to be the recommended way to check if the connection is broken. I tried killing rabbitmq and apart from running in a big amount (infinite?) of unmarshaling errors, this works (so ready is set to false upon killing rabbitmq).

internal/repository/prisma/health.go
```diff
@@ -0,0 +1,23 @@
...
+
+func NewHealthRepository(client *db.PrismaClient) repository.HealthRepository {
+	return &healthRepository{
+		client: client,
+	}
+}
+
+func (a *healthRepository) IsHealthy() bool {
+	_, err := a.client.User.FindMany().Exec(context.Background())
+	return err == nil
```
steebchen: Figured that this would be the easiest kind of healthcheck for the DB

abelanger5: We probably want to limit the number of finds to 1 so we don't run out of memory with thousands of users. We should also check the `dbsqlc` connection here.

cmd/hatchet-engine/engine/run.go
abelanger5: Yes, we can set the interval of the health check on the k8s side, and we should indeed wait for that interval times the number of retries (in the case that the kube-proxy cannot reach the pod for the number of retries). So if the interval is 5 seconds and retries is 3, we should sleep for 15 seconds. Because we'll probably be changing this value, can we make this a configurable env var?

abelanger5 (changes requested):

internal/repository/prisma/health.go
steebchen: right good catch

steebchen: done

cmd/hatchet-engine/engine/run.go
steebchen: Makes sense! Done.

env var is `SERVER_SHUTDOWN_WAIT` of type time.Duration

abelanger5 (approved):

PR #220, opened 2024-03-01
fix: eliminate remaining race conditions

### LINKED ISSUE #169: Data races
fix all data races

steebchen:
# Description

Fixes #169 

## Type of change

## What's Changed

- [x] fix race conditions in engine
- [x] fix e2e test race conditions

### DIFF
--- a/internal/logger/stderr.go
+++ b/internal/logger/stderr.go
@@ -7,14 +7,19 @@ import (
 	"github.com/rs/zerolog"
 
 	"github.com/hatchet-dev/hatchet/internal/config/shared"
+
+	"time"
 )
+
+func init() {
+	zerolog.TimeFieldFormat = time.RFC3339Nano
+}
 
 func NewDefaultLogger(service string) zerolog.Logger {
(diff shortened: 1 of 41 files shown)


internal/logger/stderr.go
```diff
@@ -15,7 +14,7 @@ func NewDefaultLogger(service string) zerolog.Logger {
 }
 
 func NewStdErr(cf *shared.LoggerConfigFile, service string) zerolog.Logger {
-	zerolog.TimeFieldFormat = time.RFC3339Nano
+	// zerolog.TimeFieldFormat = time.RFC3339Nano
```
steebchen: There doesn't seem to be another way to do this but this results in a data race
cc @abelanger5

abelanger5: Can't we put this in an `init` method?

abelanger5 (approved):

internal/logger/stderr.go
steebchen: yep works, good idea 👍

PR #229, opened 2024-03-03
feat(engine): improve stability of workers with terminations and disconnects

abelanger5:
# Description

- Adds handling of worker connection edge cases where our worker heartbeat goroutine could have leaked. 
- Adds a new feature which reassigns a step run to a different worker if the worker has not heartbeat for 60 seconds, which will almost always mean the worker is dead. 
- Note that this takes into account retries, because we view this as a failure from an execution perspective.

## Type of change


frontend/app/src/components/ui/json-form.tsx
```diff
@@ -15,7 +15,7 @@ import { DynamicSizeInputTemplate } from './form-inputs/dynamic-size-input-templ
 import { createContext, useRef } from 'react';
 
 type JSONPrimitive = string | number | boolean | null;
```
grutt:
```suggestion
type JSONPrimitive = string | number | boolean | null | Array<JSONPrimitive>;
```

grutt (approved):

PR #232, opened 2024-03-04
feat: events list filtering by status

abelanger5:
# Description

Support filtering events by the status of their workflow runs from the dashboard. 

## Type of change


frontend/app/src/pages/main/events/components/event-columns.tsx
```diff
@@ -78,13 +78,21 @@ export const columns = ({
...
       header: () => <></>,
       cell: () => {
         return <div></div>;
       },
     },
+    {
+      accessorKey: 'status',
+      header: () => <></>,
+      cell: () => {
+        return <div></div>;
```
grutt, line 93: empty div?

grutt (approved):

frontend/app/src/pages/main/events/components/event-columns.tsx
grutt: we should consider using the cell output so we can make these tables dynamic at some point

abelanger5: Yep, see comment above.

PR #233, opened 2024-03-04
fix: parse schema in API, not engine

abelanger5:
# Description

We spend some cycles parsing the schema in the engine and usage of `defer` can crash the engine since it's not caught by `panic...recover`. We move this logic to the API because we're guaranteed panic...recover in the handler and parsing the schema is not core to the engine.

## Type of change

### DIFF
--- a/internal/services/controllers/jobs/controller.go
+++ b/internal/services/controllers/jobs/controller.go
@@ -18,7 +18,6 @@ import (
 	"github.com/hatchet-dev/hatchet/internal/repository"
 	"github.com/hatchet-dev/hatchet/internal/repository/prisma/db"
 	"github.com/hatchet-dev/hatchet/internal/repository/prisma/sqlchelpers"
-	"github.com/hatchet-dev/hatchet/internal/schema"
 	"github.com/hatchet-dev/hatchet/internal/services/shared/defaults"
 	"github.com/hatchet-dev/hatchet/internal/services/shared/tasktypes"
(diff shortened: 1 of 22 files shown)


internal/services/controllers/jobs/controller.go
```diff
@@ -474,17 +473,17 @@ func (ec *JobsControllerImpl) handleStepRunRetry(ctx context.Context, task *task
 			inputBytes = mergedInputBytes
 		}
 
-		jsonSchemaBytes, err := schema.SchemaBytesFromBytes(inputBytes)
+		// jsonSchemaBytes, err := schema.SchemaBytesFromBytes(inputBytes)
 
-		if err != nil {
-			return err
-		}
+		// if err != nil {
```
grutt, line 485: rm

internal/services/dispatcher/server.go
```diff
@@ -423,23 +422,23 @@ func (s *DispatcherImpl) PutOverridesData(ctx context.Context, request *contract
...
 
-	jsonSchemaBytes, err := schema.SchemaBytesFromBytes(input)
+	// jsonSchemaBytes, err := schema.SchemaBytesFromBytes(input)
 
-	if err != nil {
-		return nil, err
-	}
+	// if err != nil {
+	// 	return nil, err
+	// }
```
grutt, line 435: rm

grutt (approved):

PR #234, opened 2024-03-05
feat: delete workflow from UI

abelanger5:
# Description

Adds support for deleting a workflow from the UI. 

## Type of change


frontend/app/src/pages/main/workflows/$workflow/components/delete-workflow-form.tsx
```diff
@@ -0,0 +1,55 @@
...
+            onClick={() => {
+              props.onCancel();
+            }}
+          >
+            Cancel
+          </Button>
+          <Button
+            variant="destructive"
+            onClick={() => {
+              props.onSubmit(props.workflow);
```
grutt, line 45: We should probably do the "type the name" pattern here since this can very much break things

grutt (approved):
One UI suggestion

PR #235, opened 2024-03-05
fix: crash on typing json input

abelanger5:
# Description

Fixes JSON input crashing when typing invalid input. 

## Type of change


grutt (changes requested):
Should we disable replay if invalid?

PR #271, opened 2024-03-15
feat(repository): use generic cacheable function

steebchen:
# Description

Fixes # (issue)

## Type of change


abelanger5 (approved):
Yeah this is kinda nice, I like this. I do think we should be careful of using the IDs in the cache and potentially setting a different cache per resource. There are a handful of IDs in the database who are uniquely keyed on tuples like `(tenantId, actionId)` for example. It would be easy to misuse this and start accidentally poisoning the cache.

internal/msgqueue/rabbitmq/rabbitmq_test.go
```diff
@@ -18,6 +18,7 @@ import (
 )
 
 func TestMessageQueueIntegration(t *testing.T) {
+	t.SkipNow()
```
abelanger5: Why are we skipping these tests?

abelanger5 (approved):

internal/msgqueue/rabbitmq/rabbitmq_test.go
steebchen: somehow these tests fail on my machine. pushed this by accident, good catch

Issue #255, opened 2024-03-11
feat(dashboard): token expiration date


grutt:
Unclear affordance for token expiration, and no notification of expiring tokens:
![](https://files.slack.com/files-pri/T0659SG0T89-F06NT251RUN/cleanshot_2024-03-11_at_13.15.37_2x.png)

PR #275, opened 2024-03-19
feat(dashboard): improve dashboard settings for user and tokens

### LINKED ISSUE #274: feat(dashboard): change user password
Problem: Users should be able to change their password from the settings page.

Solve: 

- [ ] update credentials endpoint
- [ ] add form for current password and new password

grutt:
# Description

When using the password auth strategy there is no way for a user to update their password.

Fixes #274, #255 

## Type of change

## What's Changed

- [x] Render token list expiration as date
- [x] Add update password mutation
- [x] Add update password dialog and form


api-contracts/openapi/components/schemas/user.yaml
```diff
@@ -49,6 +49,23 @@ UserLoginRequest:
...
+      description: The password of the user.
+      x-oapi-codegen-extra-tags:
+        validate: "required,password"
+    newPassword:
+      type: string
+      description: The new password for the user.
+      x-oapi-codegen-extra-tags:
+        validate: "required,password"
+  required:
+    - email
```
abelanger5:
```suggestion
    - newPassword
```

api/v1/server/handlers/users/update_password.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (u *UserService) UserUpdatePassword(ctx echo.Context, request gen.UserUpdatePasswordRequestObject) (gen.UserUpdatePasswordResponseObject, error) {
+	// check that the server supports local registration
+	if !u.config.Auth.ConfigFile.BasicAuthEnabled {
+		return gen.UserUpdatePassword405JSONResponse(
+			apierrors.NewAPIErrors("local registration is not enabled"),
+		), nil
+	}
+
+	ctx.Logger().Info("Body: ", request.Body)
```
abelanger5: Remove

api/v1/server/handlers/users/update_password.go
```diff
@@ -0,0 +1,74 @@
...
+	// validate the request
+	if apiErrors, err := u.config.Validator.ValidateAPI(request.Body); err != nil {
+		return nil, err
+	} else if apiErrors != nil {
+		return gen.UserUpdatePassword400JSONResponse(*apiErrors), nil
+	}
+
+	// determine if the user exists before attempting to write the user
+
+	existingUser := ctx.Get("user").(*db.UserModel)
```
abelanger5: It's better if this is at the top of the handler so we panic before starting any business logic

api/v1/server/handlers/users/update_password.go
```diff
@@ -0,0 +1,74 @@
...
+
+	ctx.Logger().Info("existing user: ", existingUser)
```
abelanger5: Remove

api/v1/server/handlers/users/update_password.go
```diff
@@ -0,0 +1,74 @@
...
+
+	user, err := u.config.Repository.User().UpdateUser(existingUser.ID, &repository.UpdateUserOpts{
+		Password: newPass,
+	})
+
+	if err != nil {
+		return nil, fmt.Errorf("could not update user: %w", err)
+	}
+
+	err = authn.NewSessionHelpers(u.config).SaveAuthenticated(ctx, existingUser)
```
abelanger5: Shouldn't need this, this is just resaving the cookie

internal/repository/prisma/user.go
```diff
@@ -123,6 +123,17 @@ func (r *userRepository) UpdateUser(id string, opts *repository.UpdateUserOpts)
 		updateTx,
 	}
 
+	if opts.Password != nil {
+		txs = append(txs, r.client.UserPassword.UpsertOne(
```
abelanger5: Would be good to add an extra check that the user actually has a password set already. This could just be an `Update`. 

I don't think this is exploitable at the moment, but if things upstream change, there's potentially a way for a user to get a new password saved even with a different auth mechanism (like Google SSO) linked to their account. Perhaps by passing in an empty password and saving a new password. This would then allow someone to "reset" a password even if some other auth mechanism was linked.

abelanger5 (changes requested):

internal/repository/prisma/user.go
grutt: good catch.

frontend/app/src/pages/main/tenant-settings/components/members-columns.tsx
```diff
@@ -39,5 +46,40 @@ export const columns = (): ColumnDef<TenantMember>[] => {
...
+function MemberActions({
+  row,
+  onChangePasswordClick,
+}: {
+  row: Row<TenantMember>;
+  onChangePasswordClick: (row: TenantMember) => void;
+}) {
+  const { user } = useOutletContext<UserContextType>();
+
+  if (user.metadata.id === row.original.metadata.id) {
```
grutt, line 79: @abelanger5 we should likely also guard here if its SSO, but currently this is not exposed on the user object. Any toughts?

abelanger5 (approved):

PR #281, opened 2024-03-21
refactor: separate api and engine repositories, change ticker logic

abelanger5:
# Description

This refactor removes the dependency on the Prisma ORM for the engine (hence the large number of changed files). We did this for a number of reasons - to reduce query load on the database and have more control over query latency, and because connections pools were unpredictable and spiking with Prisma. 

We will still be modeling all database changes via the Prisma DSL, and running migrations through Prisma. 

This PR also refactors the `ticker` service and switches the logic of timeouts/cron/scheduling to polling at a certain interval (previously all scheduling was done in-process). This makes it easier to recover from failure of the engine.   

## Type of change


.github/workflows/test.yml
```diff
@@ -243,7 +249,8 @@ jobs:
 
       - name: Generate
         run: |
-          go run github.com/steebchen/prisma-client-go db push
+          go run github.com/steebchen/prisma-client-go migrate deploy
+          task generate-all
```
steebchen, line 253: @abelanger5 why does the test action need this now? the idea was that this just runs unit test which doesn't need all the generated stuff to keep it small and simple

abelanger5: This is the load test job - we at least need to run `generate-go`, but perhaps we don't need to run the other generators?

api-contracts/workflows/workflows.proto
```diff
@@ -69,34 +64,13 @@ message CreateWorkflowStepOpts {
 message ListWorkflowsRequest {}
 
 message ScheduleWorkflowRequest {
-    string workflow_id = 1;
+    string name = 1;
```
grutt, line 67: generate protobufs for go sdk and python sdk

internal/services/controllers/jobs/controller.go
```diff
@@ -694,13 +691,7 @@ func (ec *JobsControllerImpl) scheduleStepRun(ctx context.Context, tenantId, ste
...
 	defer span.End()
 
-	stepRun, err := ec.repo.StepRun().GetStepRunById(tenantId, stepRunId)
-
-	if err != nil {
-		return fmt.Errorf("could not get step run: %w", err)
-	}
-
-	servertel.WithStepRunModel(span, stepRun)
+	// servertel.WithStepRunModel(span, stepRun)
```
grutt: comment?

grutt (approved):

.github/workflows/test.yml
steebchen: ah my mistake. probably fine then

steebchen (approved):

PR #288, opened 2024-03-24
fix(python sdk): adds namespace config

### LINKED ISSUE #279: feat(python sdk): namespace config var
[typescript/#27](https://github.com/hatchet-dev/hatchet-typescript/pull/27) added a namespace configuration prop to improve the local developer experience for multiple devs on a single hatchet tenant.

We need the same namespacing functionality for the Python SDK 

Related #278

aeswibon:
# Description

- Adds new namespace config to `python-sdk`  
- Fixes #279

## Type of change

## What's Changed

### DIFF
--- a/python-sdk/hatchet_sdk/clients/events.py
+++ b/python-sdk/hatchet_sdk/clients/events.py
@@ -1,18 +1,22 @@
-from ..events_pb2_grpc import EventsServiceStub
-from ..events_pb2 import PushEventRequest, PutLogRequest
-
 import datetime
-from ..loader import ClientConfig
 import json
+
 import grpc
 from google.protobuf import timestamp_pb2
+
+from ..events_pb2 import PushEventRequest, PutLogRequest
+from ..events_pb2_grpc import EventsServiceStub
+from ..loader import ClientConfig
 from ..metadata import get_metadata
+
 
 def new_event(conn, config: ClientConfig):
     return EventClientImpl(
         client=EventsServiceStub(conn),
         token=config.token,
+        namespace=config.namespace or "",
     )
+
 
 def proto_timestamp_now():
     t = datetime.datetime.now().timestamp()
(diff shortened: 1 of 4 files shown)


python-sdk/hatchet_sdk/clients/events.py
```diff
@@ -1,37 +1,44 @@
...
+        self.namespace = namespace
 
     def push(self, event_key, payload):
         try:
-            payload_bytes = json.dumps(payload).encode('utf-8')
+            payload_bytes = json.dumps(payload).encode("utf-8")
         except json.UnicodeEncodeError as e:
             raise ValueError(f"Error encoding payload: {e}")
 
+        event_key = f"{self.namespace}:{event_key}"
```
abelanger5, line 41: This doesn't seem to match the implementation of the TS SDK: https://github.com/hatchet-dev/hatchet-typescript/blob/eaca1fcfd416665467a6fa8bcc0ac0989dd7512f/src/clients/event/event-client.ts#L33.

python-sdk/hatchet_sdk/worker.py
```diff
@@ -2,49 +2,75 @@
...
 class Worker:
-    def __init__(self, name: str, max_runs: int | None = None, debug=False, handle_kill=True):
+    def __init__(
+        self,
+        name: str,
+        max_runs: int | None = None,
+        debug=False,
+        handle_kill=True,
+    ):
         self.name = name
```
abelanger5, line 46: We should be modifying the `self.name` to be namespaced as well, to match the TS SDK: https://github.com/hatchet-dev/hatchet-typescript/pull/27/files#diff-2bba1bd36b8922955278a73eeaca7a34816e0ea0a63c498f0f6cfc028010f40d

abelanger5: Note that the `event_triggers` when creating the workflow should also be modified.

abelanger5 (changes requested):
Hey @aeswibon thanks for the PR! There are a few things which don't match up with the typescript SDK that I flagged. 

It might be easier if you make a separate PR for Python formatting and format all `.py` files via that PR. We can then merge that in ahead of this PR so it's easier to see what's changed.

aeswibon:
Sure, I will open a PR to format all the `.py` files, then fixed all the changes suggested.

grutt:
tracking in #354

PR #310, opened 2024-03-31
fix: update `Requeue` and `Reassign` logic to fix performance degradation when many events are queued

abelanger5:
# Description

Logic for requeueing and reassigning did not limit the number of step runs to requeue, so when events accumulate with no worker present it causes memory to spike along with a very high query latency on the database. This PR primarily limits the number of step runs returned in the requeue and reassign queries, and also properly locks step run rows for these queries so only a step run in a `PENDING` or `PENDING_ASSIGNMENT` state can be requeued. 

This PR also improves the logging for the load test suite as there were simply too many logs generated to figure out the source of duplicated step runs. 

## Type of change


internal/repository/prisma/dbsqlc/step_runs.sql
```diff
@@ -274,105 +275,186 @@ FROM step_run_data
...
+    FROM
+        step_runs sr
+    FOR UPDATE SKIP LOCKED
+)
+UPDATE
+    "StepRun"
+SET
+    "status" = 'PENDING_ASSIGNMENT',
+    -- requeue after now plus 4 seconds
+    "requeueAfter" = CURRENT_TIMESTAMP + INTERVAL '4 seconds',
```
steebchen, line 355: not really related to this PR but I'm wondering if we should make this configurable so we can adapt this in the load test to see if it has any influence on the general reliability and performance?

internal/repository/prisma/step_run.go
```diff
@@ -658,6 +700,21 @@ func (s *stepRunEngineRepository) updateStepRunCore(
...
+			Tenantid:  sqlchelpers.UUIDFromStr(tenantId),
+		})
+
+		if err != nil {
+			// not a fatal error if there's not a semaphore to update
+			if !errors.Is(err, pgx.ErrNoRows) {
+				return nil, fmt.Errorf("could not upsert worker semaphore: %w", err)
+			}
+		}
+	}
```
steebchen:
```suggestion
		// not a fatal error if there's not a semaphore to update
		if err != nil && !errors.Is(err, pgx.ErrNoRows) {
			return nil, fmt.Errorf("could not upsert worker semaphore: %w", err)
		}
	}
```

steebchen (approved):

internal/services/controllers/jobs/controller.go
```diff
@@ -556,30 +547,38 @@ func (ec *JobsControllerImpl) runStepRunReassignTenant(ctx context.Context, tena
...
-			var innerStepRun *dbsqlc.GetStepRunForEngineRow
-
 			ctx, span := telemetry.NewSpan(ctx, "handle-step-run-reassign-step-run")
 			defer span.End()
 
-			stepRunId := sqlchelpers.UUIDToStr(stepRunCp.ID)
+			stepRunId := sqlchelpers.UUIDToStr(stepRunCp.StepRun.ID)
+
+			// if the current time is after the scheduleTimeoutAt, then mark this as timed out
+			now := time.Now().UTC().UTC()
```
grutt, line 556: double UTC?

grutt (approved):

internal/repository/prisma/dbsqlc/step_runs.sql
abelanger5: Yep will file an issue

PR #324, opened 2024-04-01
feat(engine): initial rate-limiting engine implementation

abelanger5:
# Description

Adds an initial implementation for global rate limiting keys based on https://github.com/hatchet-dev/hatchet/discussions/305. 

Note that keys are not dynamic, only supported are static string values for now. 

## Type of change


prisma/migrations/20240402034010_v0_19_0/migration.sql
```diff
@@ -0,0 +1,67 @@
...
+-- AddForeignKey
+ALTER TABLE "RateLimit" ADD CONSTRAINT "RateLimit_tenantId_fkey" FOREIGN KEY ("tenantId") REFERENCES "Tenant"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "StreamEvent" ADD CONSTRAINT "StreamEvent_tenantId_fkey" FOREIGN KEY ("tenantId") REFERENCES "Tenant"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "StreamEvent" ADD CONSTRAINT "StreamEvent_stepRunId_fkey" FOREIGN KEY ("stepRunId") REFERENCES "StepRun"("id") ON DELETE SET NULL ON UPDATE CASCADE;
+
+CREATE OR REPLACE FUNCTION get_refill_value(rate_limit "RateLimit")
```
grutt, line 55: just confirming, this is only defined in the migration?

internal/repository/prisma/step_run.go
```diff
@@ -291,18 +291,35 @@ func (s *stepRunEngineRepository) AssignStepRunToWorker(ctx context.Context, ten
...
+		Steprunid: stepRun.StepRun.ID,
+		Tenantid:  stepRun.StepRun.TenantId,
 	})
 
 	if err != nil {
 		return "", "", fmt.Errorf("could not upsert worker semaphore: %w", err)
 	}
 
-	if semaphore.Slots <= 0 {
+	if semaphore.Slots < 0 {
```
grutt, line 302: intentional change?

prisma/migrations/20240402034010_v0_19_0/migration.sql
```diff
@@ -0,0 +1,67 @@
+-- CreateTable
```
grutt, line 1: rename 0_20_0

grutt (changes requested):
just migration version name

internal/repository/prisma/step_run.go
abelanger5: Yes, since the semaphore decrements by default (and rolls back when full), we were losing 1 run when it reached 0.

prisma/migrations/20240402034010_v0_19_0/migration.sql
abelanger5: No since we haven't built `v0.19.0` yet

prisma/migrations/20240402034010_v0_19_0/migration.sql
abelanger5: Yes

grutt (approved):

PR #335, opened 2024-04-03
fix: cleanup subscribe

grutt:
# Description

Cleanup was never reached on send error

## Type of change


internal/services/dispatcher/server.go
```diff
@@ -406,15 +407,14 @@ func (s *DispatcherImpl) SubscribeToWorkflowEvents(request *contracts.SubscribeT
...
-
-		// drain the existing connections
-		wg.Wait()
+	<-ctx.Done()
+	if err := cleanupQueue(); err != nil {
+		return fmt.Errorf("could not cleanup queue: %w", err)
 	}
 
+	// drain the existing connections
+	wg.Wait()
```
abelanger5: We should set a max time that we're waiting on tasks to call `wg.Done()`. Let's set a timer here and exit either when the wg is done or after 60 seconds?

abelanger5 (changes requested):

abelanger5 (approved):

PR #344, opened 2024-04-06
test(timeout): add test for timeout

steebchen:
# Description

Fixes # (issue)

## Type of change

## What's Changed

- [x] add test for workflow timeout (skipped as it's broken)
- [x] add test for step timeout


examples/timeout/main_e2e_test.go
```diff
@@ -0,0 +1,70 @@
...
+		job  worker.WorkflowJob
+	}{
+		{
+			name: "step timeout",
+			job: worker.WorkflowJob{
+				Name:        "timeout",
+				Description: "timeout",
+				Steps: []*worker.WorkflowStep{
+					worker.Fn(func(ctx worker.HatchetContext) (result *stepOneOutput, err error) {
+						time.Sleep(time.Second * 60)
```
abelanger5: I think it'd be better if we made sure the timeout signal propagates to the step over `ctx.Done()`, rather than just sleeping

abelanger5 (changes requested):

examples/timeout/main_e2e_test.go
steebchen: Good point, I thought we already tested this in the cancellation but indeed it's not quite the same thing. I'll adapt.

abelanger5 (approved):

PR #366, opened 2024-04-10
fix(workflow): remove workflow timeout in favor of step timeout

steebchen:
# Description

Removes the unused workflow timeout option in favor of step timeouts

Fixes # (issue)

## Type of change


api-contracts/workflows/workflows.proto
```diff
@@ -42,15 +42,15 @@ message WorkflowConcurrencyOpts {
...
     ConcurrencyLimitStrategy limit_strategy = 3; // (optional) the strategy to use when the concurrency limit is reached, default CANCEL_IN_PROGRESS
 }
-  
+
 // CreateWorkflowJobOpts represents options to create a workflow job.
 message CreateWorkflowJobOpts {
     string name = 1; // (required) the job name
     string description = 2; // (optional) the job description
-    string timeout = 3; // (optional) the job timeout
+    reserved 3; // (deprecated) timeout
```
steebchen, line 50: figured `reserved` is correct in this case instead of deprecating the field

abelanger5 (approved):

PR #368, opened 2024-04-10
feat: add cancellation reason to

grutt:
# Description

Add cancelation reason to the error output object.

## Type of change

## What's Changed

- [x] Add cancelation reason to the error output object.

### DIFF
--- a/frontend/app/src/pages/main/workflow-runs/$run/components/step-run-playground.tsx
+++ b/frontend/app/src/pages/main/workflow-runs/$run/components/step-run-playground.tsx
@@ -434,6 +434,7 @@ export function StepRunPlayground({
                     isLoading={isLoading}
                     errors={
                       [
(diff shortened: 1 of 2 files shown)


frontend/app/src/pages/main/workflow-runs/$run/components/step-run-playground.tsx
```diff
@@ -434,6 +434,7 @@ export function StepRunPlayground({
                     isLoading={isLoading}
                     errors={
                       [
+                        stepRun.cancelledReason,
```
abelanger5: Let's create a map of cancelledReason to readable reasons? `TIMED_OUT`=`Step timed out.`

And fallback to the uppercase reason only if absolutely necessary?

abelanger5 (changes requested):

abelanger5 (approved):

PR #369, opened 2024-04-10
fix: retry rabbitmq connections properly and retry published messages

abelanger5:
# Description

RabbitMQ sessions are not properly retried in certain cases. Messages which are published are also not retried. 

## Type of change


grutt (approved):
verified by restarting rabbitmq

PR #373, opened 2024-04-11
feat: add github sso to dashboard

abelanger5:
# Description

Adds Github SSO to the login page.

Fixes # (issue)

## Type of change

### DIFF
--- a/api/v1/server/handlers/users/github_oauth_callback.go
+++ b/api/v1/server/handlers/users/github_oauth_callback.go
@@ -1,0 +1,186 @@
+package users
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	githubsdk "github.com/google/go-github/v57/github"
+	"github.com/labstack/echo/v4"
+	"golang.org/x/oauth2"
+
+	"github.com/hatchet-dev/hatchet/api/v1/server/authn"
+	"github.com/hatchet-dev/hatchet/api/v1/server/oas/gen"
+	"github.com/hatchet-dev/hatchet/internal/config/server"
+	"github.com/hatchet-dev/hatchet/internal/repository"
+	"github.com/hatchet-dev/hatchet/internal/repository/prisma/db"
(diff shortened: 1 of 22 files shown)


api/v1/server/handlers/users/github_oauth_callback.go
```diff
@@ -0,0 +1,186 @@
...
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to encrypt refresh token: %s", err.Error())
+	}
+
+	oauthOpts := &repository.OAuthOpts{
+		Provider:       "github",
+		ProviderUserId: gInfo.ID,
+		AccessToken:    accessTokenEncrypted,
+		RefreshToken:   &refreshTokenEncrypted,
```
grutt, line 94: why are we passing a pointer for one but not the other?

frontend/app/src/pages/auth/login/index.tsx
```diff
@@ -22,19 +23,27 @@ export default function Login() {
...
   let prompt = 'Enter your email and password below.';
 
-  if (basicEnabled && googleEnabled) {
-    prompt = 'Enter your email and password below, or continue with Google.';
-  } else if (googleEnabled) {
-    prompt = 'Continue with Google.';
+  if (basicEnabled && (googleEnabled || githubEnabled)) {
+    prompt =
+      'Enter your email and password below, or continue with a supported provider.';
+  } else if (googleEnabled || githubEnabled) {
```
grutt, line 33: not critical; 
maybe write this as `schemas.filter((x) => x != 'basic').length > 1` so we don't have to rewrite for all new providers?

grutt (approved):

api/v1/server/handlers/users/github_oauth_callback.go
abelanger5: To mark it as optional vs not optional

PR #386, opened 2024-04-16
feat(docs): automatic theme

### LINKED ISSUE #345: light theme
For web UI and documentation, could you add a switch to use light theme? Thanks!

steebchen:
# Description

fix #345

## Type of change


grutt:
Thanks @steebchen . I tried something similar but it looks like the light theme has limited coverage:

Screenshot 2024-04-16 at 7 19 19 AM

steebchen:
@grutt fixed by removing the fixed bg color in the main tag; I tried to set it only to dark but then it doesn't work when the user manually switches the theme. So I removed it completely for now.

Alternatively, we can opt in for the slightly different background color for the dark mode, but would then need to disable manual theme switching.

I think the best thing to do is what's in this PR: leave the default colors as is set by the default nextjs theme and detect the user's preferred theme. And optionally the user can switch it by using the integrated dark/light switcher in the bottom left.

grutt:
Almost there! Almost there, code blocks are now unreadable.

Screenshot 2024-04-16 at 1 28 46 PM

steebchen:
dammit, will take another look tomorrow in all places 😅

steebchen:
ok done @grutt

grutt (approved):

PR #397, opened 2024-04-18
fix: worker assignment race

grutt:
# Description

Workers with max run of 1 would be assigned multiple step runs which subsequently failed due to slots being full.

Depends on #394 

## Type of change

## What's Changed

- [x] refactored retrier as generic
- [x] added unassigned retrier
- [x] separated worker semafore increment on retry into separate transaction


internal/repository/prisma/get_group_key_run.go
```diff
@@ -46,8 +46,8 @@ func (s *getGroupKeyRunRepository) AssignGetGroupKeyRunToWorker(ctx context.Cont
 	// var assigned
 	var assigned *dbsqlc.AssignGetGroupKeyRunToWorkerRow
 
-	err = retrier(s.l, func() (err error) {
-		assigned, err = s.queries.AssignGetGroupKeyRunToWorker(ctx, s.pool, dbsqlc.AssignGetGroupKeyRunToWorkerParams{
+	err = deadlockRetry(s.l, func() (err error) {
+		assigned, err = s.queries.AssignGetGroupKeyRunToWorker(context.Background(), s.pool, dbsqlc.AssignGetGroupKeyRunToWorkerParams{
```
abelanger5:
```suggestion
		assigned, err = s.queries.AssignGetGroupKeyRunToWorker(ctx, s.pool, dbsqlc.AssignGetGroupKeyRunToWorkerParams{
```

Why is this `context.Background()`?

abelanger5 (changes requested):

PR #425, opened 2024-04-25
Feat  improved visibility

grutt:
# Description

Several metrics and states were not exposed making debugging challening.

Fixes HAT-145 HAT-125 HAT-142

## Type of change

## What's Changed

- [x] expose worker state
- [x] improve worker view
- [x] expose workflow run metrics counts
- [x] add workflow run summary
- [x] improve output with \n


grutt:
image
image
image

grutt:
image

api-contracts/openapi/paths/workflow/workflow.yaml
```diff
@@ -366,6 +366,79 @@ workflowRuns:
     summary: Get workflow runs
     tags:
       - Workflow
+workflowRunsMetrics:
+  get:
+    x-resources: ["tenant"]
+    description: Get all workflow runs for a tenant
+    operationId: workflow-run:metrics
```
abelanger5: This doesn't really matter at the moment, and isn't documented, but this should be `workflow-run:get:metrics`, the second part of the operation is always a verb in the other endpoints. This will eventually be important when doing RBAC.

internal/repository/prisma/dbsqlc/workflow_runs.sql
```diff
@@ -46,6 +46,41 @@ WHERE
         "status" = ANY(cast(sqlc.narg('statuses')::text[] as "WorkflowRunStatus"[]))
     );
 
+-- name: WorkflowRunsMetricsCount :one
+SELECT
+    COUNT(CASE WHEN runs."status" = 'PENDING' THEN 1 END) AS "PENDING",
```
abelanger5, line 51: Missing a `QUEUED` status here?

internal/repository/prisma/workflow_run.go
```diff
@@ -348,6 +356,65 @@ func listWorkflowRuns(ctx context.Context, pool *pgxpool.Pool, queries *dbsqlc.Q
...
+		queryParams.ParentStepRunId = pgParentStepRunId
+	}
+
+	if opts.EventId != nil {
+		pgEventId := sqlchelpers.UUIDFromStr(*opts.EventId)
+
+		queryParams.EventId = pgEventId
+	}
+
+	tx, err := pool.Begin(ctx)
```
abelanger5: You don't need to do `pool.Begin` when only doing 1 query (I've done this in other places, but shouldn't have). You can just do:

```
workflowRunsCount, err := queries.WorkflowRunsMetricsCount(ctx, pool, queryParams)

if err != nil {
    return nil, err
}
```

Without any of the begin/rollback/commit stuff.

internal/repository/workflow_run.go
```diff
@@ -233,6 +233,29 @@ type ListWorkflowRunsOpts struct {
...
+	ParentId *string `validate:"omitempty,uuid"`
+
+	// (optional) the parent step run id
+	ParentStepRunId *string `validate:"omitempty,uuid"`
+
+	// (optional) the event id that triggered the workflow run
+	EventId *string `validate:"omitempty,uuid"`
+
+	// (optional) the group key for the workflow run
+	GroupKey *string
```
abelanger5: I don't see this group key used in the query, or the `IDs` param either.

internal/repository/prisma/dbsqlc/workflow_runs.sql
grutt: Added, but I'm still confused by `QUEUED` and `PENDING`... it looks like we're not using these states for workflow runs and setting all runs directly to `RUNNING`?

api-contracts/openapi/paths/workflow/workflow.yaml
grutt: good catch

abelanger5 (approved):

PR #445, opened 2024-05-02
add jobs which always run on failure

abelanger5:
# Description

Adds Go SDK and backend support for `on_failure` jobs. These jobs will always run if the workflow fails (but will not run at the moment if the workflow gets cancelled - it's unclear if we should support this behavior or not). 

## Type of change


examples/on-failure/main.go
```diff
@@ -0,0 +1,92 @@
...
+
+	err = w.On(
+		worker.NoTrigger(),
+		&worker.WorkflowJob{
+			Name:        "on-failure-workflow",
+			Description: "This runs at a scheduled time.",
+			Steps: []*worker.WorkflowStep{
+				worker.Fn(StepOne).SetName("step-one"),
+			},
+			OnFailure: &worker.WorkflowJob{
```
grutt, line 59: should we abstract this so its just a step in the sdks...? i feel like it might be confusing to introduce a new "jobs" concept out of no where

internal/repository/prisma/dbsqlc/workflow_runs.sql
```diff
@@ -242,13 +242,16 @@ WITH jobRuns AS (
...
     WHERE
         "workflowRunId" = (
             SELECT "workflowRunId"
             FROM "JobRun"
             WHERE "id" = @jobRunId::uuid
         ) AND
-        "tenantId" = @tenantId::uuid
+        runs."tenantId" = @tenantId::uuid AND
+        -- we should not include onFailure jobs in the calculation
+        job."kind" != 'ON_FAILURE'
```
grutt: Should we make this only including defaults i would think future cases will also need to filtered?

grutt (changes requested):
minor things, i also want to test locally

examples/on-failure/main.go
abelanger5: This isn't out of nowhere, it's always been on the Go SDK. It's just that `worker.Fn` implements `WorkflowJob` as well.

internal/repository/prisma/dbsqlc/workflow_runs.sql
abelanger5: Good catch, will fix

PR #446, opened 2024-05-02
feat: event and workflow run metadata

grutt:
# Description

Often arbitrary metadata tags or labels are desirable to organize events and workflow runs. 

## Type of change

## What's Changed

- [x] Adds additionalMetadata columns to event and workflow runs
- [x] Revise list queries to filter by key-value pairs
- [x] add supporting ui
- [x] adds metadata to go sdk event push 

References: https://github.com/hatchet-dev/hatchet-typescript/pull/86


internal/repository/event.go
```diff
@@ -48,6 +51,9 @@ type ListEventOpts struct {
 
 	// (optional) the order direction
 	OrderDirection *string `validate:"omitempty,oneof=ASC DESC"`
+
+	// (optional) the event metadata
+	AdditionalMetadata *[]byte
```
abelanger5:
```suggestion
	AdditionalMetadata []byte
```

Arrays are nil by default, so `*` is unnecessary

internal/repository/workflow_run.go
```diff
@@ -231,6 +253,9 @@ type ListWorkflowRunsOpts struct {
 
 	// (optional) the order direction
 	OrderDirection *string `validate:"omitempty,oneof=ASC DESC"`
+
+	// (optional) exact metadata to filter by
+	AdditionalMetadata *map[string]interface{} `validate:"omitempty"`
```
abelanger5:
```suggestion
	AdditionalMetadata map[string]interface{} `validate:"omitempty"`
```

internal/repository/workflow_run.go
```diff
@@ -248,6 +273,9 @@ type WorkflowRunsMetricsOpts struct {
 
 	// (optional) the event id that triggered the workflow run
 	EventId *string `validate:"omitempty,uuid"`
+
+	// (optional) exact metadata to filter by
+	AdditionalMetadata *map[string]interface{} `validate:"omitempty"`
```
abelanger5:
```suggestion
	AdditionalMetadata map[string]interface{} `validate:"omitempty"`
```

internal/services/controllers/events/controller.go
```diff
@@ -146,10 +147,18 @@ func (ec *EventsControllerImpl) handleTask(ctx context.Context, task *msgqueue.M
 		return fmt.Errorf("could not decode task metadata: %w", err)
 	}
 
-	return ec.processEvent(ctx, metadata.TenantId, payload.EventId, payload.EventKey, []byte(payload.EventData))
+	additionalMetadata := make(map[string]interface{})
+
+	err = json.Unmarshal([]byte(payload.EventAdditionalMetadata), &additionalMetadata)
```
abelanger5: Need nil check

prisma/migrations/20240426175227_v0_22_1/migration.sql
```diff
@@ -0,0 +1,12 @@
+/*
```
abelanger5: Collapse into 1 migration

api/v1/server/handlers/workflows/trigger.go
```diff
@@ -17,6 +17,8 @@ import (
 )
 
 func (t *WorkflowService) WorkflowRunCreate(ctx echo.Context, request gen.WorkflowRunCreateRequestObject) (gen.WorkflowRunCreateResponseObject, error) {
+	// TODO metadata
```
abelanger5: Remove todo?

abelanger5 (changes requested):

api/v1/server/handlers/workflows/trigger.go
```diff
@@ -57,8 +59,26 @@ func (t *WorkflowService) WorkflowRunCreate(ctx echo.Context, request gen.Workfl
 		), nil
 	}
 
-	createOpts, err := repository.GetCreateWorkflowRunOptsFromManual(workflowVersion, inputBytes)
+	additionalMetadata := make(map[string]interface{})
```
steebchen:
```suggestion
	var additionalMetadata map[string]interface{}
```

api/v1/server/oas/transformers/workflow_run.go
```diff
@@ -296,16 +296,28 @@ func ToWorkflowRunFromSQLC(row *dbsqlc.ListWorkflowRunsRow) *gen.WorkflowRun {
 
 	workflowRunId := sqlchelpers.UUIDToStr(run.ID)
 
+	var additionalMetadata (map[string]interface{})
```
steebchen:
```suggestion
	var additionalMetadata map[string]interface{}
```

prisma/schema.prisma
```diff
@@ -226,6 +226,9 @@ model Event {
   // data stored in the event
   data Json?
 
+  // metadata stored in the event
+  additionalMetadata Json?
```
steebchen, line 233: can we not name this `metadata`? Would be a little cleaner. Or is this conflicting with existing metadata names?

abelanger5: I'm going to leave this as-is, we can certainly change the column name later. I imagine this is preventing `metadata` from being an overloaded field in the API. I think a more apt name would be `userMetadata`?

abelanger5 (approved):

PR #452, opened 2024-05-03
feat: improved optional product analytics

grutt:
# Description

We're adding opt-out frontend analytics to help improve our understanding of how users are interacting with the hatchet dashboard to provide a better user experience. We're extremely privacy and security conscious and have made this change with three key decisions 1) the feature is disabled on self-host 2) the feature can easily be disabled for any cloud tenant and 3) we mask as much as possible so we can understand behavior without leaking state.

## Type of change

## What's Changed

- [x] Added opt out attribute to the tenant model and update endpoint
- [x] Added FE Posthog Config variables and exposed through metadata endpoint
- [x] Added FE Posthog script which is loaded _after_ the tenant object is resolved


api/v1/server/handlers/tenants/update.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/labstack/echo/v4"
+
+	"github.com/hatchet-dev/hatchet/api/v1/server/oas/gen"
+	"github.com/hatchet-dev/hatchet/api/v1/server/oas/transformers"
+	"github.com/hatchet-dev/hatchet/internal/repository"
+	"github.com/hatchet-dev/hatchet/internal/repository/prisma/db"
+)
+
+func (t *TenantService) TenantUpdate(ctx echo.Context, request gen.TenantUpdateRequestObject) (gen.TenantUpdateResponseObject, error) {
+	tenant := ctx.Get("tenant").(*db.TenantModel)
```
grutt, line 13: @abelanger5 do we need to do any further permission checks here? 

Do we want anyone in a tenant to be able to update or only Owners?

abelanger5: Let's just allow anyone to update this for now, will do a more fine-grained permissions sweep once we have more to differentiate members from admins/owners.

abelanger5 (approved):

PR #461, opened 2024-05-08
feat: alerting from slack and email

abelanger5:
# Description

Implements alerting for Slack and email-based alerts. Exposes an alerting frequency on the dashboard for users to set maximum alerting frequencies. Currently only exposes 3 options for frequencies: 5 minutes, 1 hour, or 24 hours. 

Also includes a very hefty refactor of tenant settings, as they were getting very clunky contained in the same view. 

## Type of change


internal/integrations/alerting/alerter.go
```diff
@@ -0,0 +1,152 @@
...
+	// read in all failed workflow runs since the last alerted time, ordered by the most recent runs first
+	statuses := []db.WorkflowRunStatus{
+		db.WorkflowRunStatusFailed,
+	}
+
+	limit := 5
+
+	tenantId := sqlchelpers.UUIDToStr(tenantAlerting.Settings.TenantId)
+
+	failedWorkflowRuns, err := t.repo.WorkflowRun().ListWorkflowRuns(
```
grutt, line 91: we'll likely want to filter by workflow/namespace?

grutt (approved):

internal/integrations/alerting/alerter.go
abelanger5: eventually yes, might be a more general alerting setting that we expose to the user. right now alerts are opt-in and users are typically creating multiple tenants to split up dev/prod.

grutt: i think we'll want to put this on the alert group object since users will likely want different "routing" for different workflows

PR #468, opened 2024-05-08
fix: handle last alerted null case + some logging/cleanup improvements

abelanger5:
# Description

We previously were alerting on all failed workflows that had ever been run when `lastAlertedAt = null` - this refactors to only look at failed workflows within the max frequency duration time. 

Also improves logging for the API. 

## Type of change

### DIFF
--- a/internal/repository/prisma/dbsqlc/tickers.sql
+++ b/internal/repository/prisma/dbsqlc/tickers.sql
@@ -223,9 +223,6 @@ WITH active_tenant_alerts AS (
         alerts.*
     FROM
         "TenantAlertingSettings" as alerts
-    -- only return alerts which have a slack webhook enabled
-    JOIN
-        "SlackAppWebhook" as webhooks ON webhooks."tenantId" = alerts."tenantId"
     WHERE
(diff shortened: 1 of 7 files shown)


internal/repository/prisma/dbsqlc/tickers.sql
```diff
@@ -242,7 +239,10 @@ failed_run_count_by_tenant AS (
     WHERE
         "status" = 'FAILED'
         AND (
-            "lastAlertedAt" IS NULL OR
+            (
+                "lastAlertedAt" IS NULL AND
+                workflowRun."finishedAt" >= NOW() - convert_duration_to_interval(active_tenant_alerts."maxFrequency")
```
grutt, line 244: quick check, we're validating active_tenant_alerts."maxFrequency"? on input?

grutt (approved):

internal/repository/prisma/dbsqlc/tickers.sql
abelanger5: yeah we validate it's a duration string

PR #477, opened 2024-05-09
feat: worker semaphore slot resolver

grutt:
# Description

Adds an interval based query to ensure semaphore values are accurate.

Depends on #476 

## Type of change

## What's Changed

- [x] add a query to update semaphores based on running step runs
- [x] add a task to the ticker to run this query every minute


internal/repository/prisma/dbsqlc/workers.sql
```diff
@@ -88,6 +88,25 @@ WHERE
...
 RETURNING *;
 
+
+-- name: ResolveWorkerSemaphoreSlots :execrows
+WITH stepRuns AS (
+  SELECT 
+    sr."workerId",
+    COUNT(*) AS runningRuns
+  FROM "StepRun" sr
+  WHERE sr."status" IN ('RUNNING', 'ASSIGNED') AND
```
grutt: @abelanger5 can you confirm if an ASSIGNED task takes a slot?

internal/repository/prisma/dbsqlc/workers.sql
```diff
@@ -88,6 +88,31 @@ WHERE
...
+        SELECT COUNT(*)
+        FROM "StepRun" sr
+        WHERE sr."status" IN ('RUNNING', 'ASSIGNED')
+            AND sr."semaphoreReleased" = FALSE
+            AND sr."workerId" = w."id"
+    ), 0
+)
+FROM "Worker" w
+WHERE ws."workerId" = w."id"
+    AND "slots" != COALESCE(w."maxRuns", 100) - COALESCE(
```
grutt, line 105: I think this should be fine for now, its not guaranteed to update on every interval but it will update to the correct value

PR #484, opened 2024-05-10
feat: improve reassign and timeout behavior and visibility

grutt:
# Description

Improves consistency for how we handle reassignments and retires. 

## Type of change

## What's Changed

- [x] added step run events for reassignments and timeouts
- [x] runtime timeouts are treated as failures, meaning they are handled the same as an error and will exhibit the same retry behavior
- [x] reduced last heartbeat check from 60s to 30s to improve reassign performance
- [x] updated docs to reflect timeout behavior


internal/repository/prisma/dbsqlc/step_runs.sql
```diff
@@ -340,11 +340,11 @@ step_runs AS (
         sr."tenantId" = @tenantId::uuid
         AND ((
             sr."status" = 'RUNNING'
-            AND w."lastHeartbeatAt" < NOW() - INTERVAL '60 seconds'
+            AND w."lastHeartbeatAt" < NOW() - INTERVAL '30 seconds'
```
abelanger5, line 344: For consistency, if we are enforcing 30 seconds as a global worker heartbeat timeout, then we should fail steps with no retries left which belong to a timed-out worker. So shouldn't we remove `AND s."retries" > sr."retryCount"`?

abelanger5 (changes requested):

internal/repository/prisma/dbsqlc/step_runs.sql
grutt: good catch

abelanger5 (approved):

PR #487, opened 2024-05-11
use consistent logger name 

RomanMIzulin:
# Description

One line.  Use consistent logger name at client.go file.

Fixes # (issue)

## improve unambiguity

## What's Changed

- [ ] Make sure previous name was **not** used intentionally


abelanger5 (approved):

abelanger5:
Thanks for the PR @RomanMIzulin! This makes sense since there are services which are not the worker which can call the API. Just merged and this will be available in the next release.

PR #489, opened 2024-05-13
feat: workflow run cancel

grutt:
# Description

can cancel workflow runs via api and dashboard. previously only individual step runs could be cancelled

## Type of change

## What's Changed

- [x] add workflow run cancel endpoint
- [x] add cancel workflow run button to workflow run page
- [x] add select to workflow run list
- [x] add cancel selected button to workflow run list page


grutt:
Screenshot 2024-05-13 at 10 09 04 AM
Screenshot 2024-05-13 at 10 09 19 AM

api/v1/server/handlers/workflows/cancel.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+
+		canceledStepRunsResponse[i] = *res
+	}
+
+	// Return the list of canceled step runs in the response
+	return gen.WorkflowRunCancel200JSONResponse(canceledStepRunsResponse), nil
+}
+
+func pgUUIDToStr(uuid pgtype.UUID) string {
```
abelanger5: Use the `sqlchelpers` package instead

api/v1/server/handlers/workflows/cancel.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/hatchet-dev/hatchet/internal/repository/prisma/db"
+	"github.com/hatchet-dev/hatchet/internal/repository/prisma/dbsqlc"
+	"github.com/hatchet-dev/hatchet/internal/services/shared/tasktypes"
+)
+
+func (t *WorkflowService) WorkflowRunCancel(ctx echo.Context, request gen.WorkflowRunCancelRequestObject) (gen.WorkflowRunCancelResponseObject, error) {
+	tenant := ctx.Get("tenant").(*db.TenantModel)
+	runIds := request.Body.WorkflowRunIds
+
+	canceledStepRunsChannel := make(chan []*db.StepRunModel, len(runIds))
```
abelanger5: While a channel can be used here, I think a `sync.WaitGroup` together with a result object like `sync.Map` would be more readable and easier to maintain.  With this implementation, you need to ensure that a message is sent over `canceledStepRunsChannel` on every branch of the goroutine. With a shared result object like `sync.Map` (or thread-safe slice implementation) together with `sync.WaitGroup`, you only need to write a single result to the result object.

api/v1/server/handlers/workflows/cancel.go
```diff
@@ -0,0 +1,103 @@
...
+				status := stepRun.StepRun.Status
+
+				// If the step run is not in a final state, send a task to the taskqueue to cancel it
+				if status != dbsqlc.StepRunStatusSUCCEEDED && status != dbsqlc.StepRunStatusFAILED && status != dbsqlc.StepRunStatusCANCELLED {
+					var reason = "CANCELLED_BY_USER"
+					// send a task to the taskqueue
+					err = t.config.MessageQueue.AddMessage(
+						ctx.Request().Context(),
+						msgqueue.JOB_PROCESSING_QUEUE,
+						tasktypes.StepRunCancelToTask(stepRun, reason),
```
abelanger5: I would modify the implementation of `JobRunCancelledToTask` to include a reason and cancel the job run with that task, instead of the step run. We don't want the API to be responsible for understanding the internals of looping over step runs and deciding which ones to cancel in a way that could eventually conflict with the jobs controller.

api/v1/server/handlers/workflows/cancel.go
```diff
@@ -0,0 +1,103 @@
...
+
+		if err != nil {
+			return nil, err
+		}
+
+		canceledStepRunsResponse[i] = *res
+	}
+
+	// Return the list of canceled step runs in the response
+	return gen.WorkflowRunCancel200JSONResponse(canceledStepRunsResponse), nil
```
abelanger5: The API design here isn't very intuitive. It doesn't make sense to me that the resource which is cancelled is a workflow run, but a list of step runs are returned from this endpoint. It might be better to use `ListWorkflowRuns` and pass the array of workflow run IDs, and just return the list of workflow runs which were cancelled.

internal/repository/step_run.go
```diff
@@ -140,6 +140,8 @@ type StepRunEngineRepository interface {
 
 	GetStepRunForEngine(ctx context.Context, tenantId, stepRunId string) (*dbsqlc.GetStepRunForEngineRow, error)
 
+	ListStepRunsByWorkflowRunId(ctx context.Context, tenantId, workflowRunId string) ([]*dbsqlc.GetStepRunForEngineRow, error)
```
abelanger5: This looks like a duplicate of `ListStepRuns` which takes the following options:

```
type ListStepRunsOpts struct {
	JobRunId *string `validate:"omitempty,uuid"`

	WorkflowRunIds []string `validate:"dive,uuid"`

	Status *dbsqlc.StepRunStatus
}
```

abelanger5 (changes requested):

api/v1/server/handlers/workflows/cancel.go
grutt: not sure i'm following here, would we not still have to loop step runs to get the job ID?

abelanger5: I'm basically suggesting to invoke this function instead: https://github.com/hatchet-dev/hatchet/blob/615ed4296b2d235975f4a48ceb0f1700b76ed6a0/internal/services/controllers/jobs/controller.go#L291

We would need to list job runs by a list of workflow run IDs to avoid an `n+1` query on the database.

frontend/app/src/pages/main/workflow-runs/$run/index.tsx
```diff
@@ -105,7 +150,34 @@ export default function ExpandedWorkflowRun() {
...
+                </Button>
+              </DropdownMenuTrigger>
+              <DropdownMenuContent>
+                <DropdownMenuItem
+                  disabled={WORKFLOW_RUN_TERMINAL_STATUSES.includes(run.status)}
+                  onClick={() => {
+                    cancelWorkflowRunMutation.mutate();
+                  }}
+                >
+                  Attempt Cancel - All Running Steps
```
abelanger5:
```suggestion
                  Cancel all running steps
```

abelanger5 (approved):

PR #507, opened 2024-05-16
feat: make step run replays more intuitive

abelanger5:
# Description

When step runs are currently replayed, they only reset the state of the target step run. This PR modifies this behavior so that all subsequent step runs in the DAG will get reset when the parent step run is replayed. This also resets the workflow run's state. 

## Type of change


api-contracts/openapi/components/schemas/workflow_run.yaml
```diff
@@ -287,6 +287,7 @@ StepRunEventReason:
     - REASSIGNED
     - TIMED_OUT
     - SLOT_RELEASED
+    - MANUAL_RETRY
```
grutt: should we keep this somewhat consistent with CANCELLED_BY_USER? RETRY_BY_USER?

internal/repository/prisma/step_run.go
```diff
@@ -769,6 +765,111 @@ func (s *stepRunEngineRepository) UpdateStepRun(ctx context.Context, tenantId, s
...
+		if err != nil {
+			return err
+		}
+
+		// archive each of the later step run results
+		for _, laterStepRun := range laterStepRuns {
+			laterStepRunCp := laterStepRun
+			laterStepRunId := sqlchelpers.UUIDToStr(laterStepRun.ID)
+
+			err = s.archiveStepRunResult(ctx, tx, tenantId, laterStepRunId)
```
grutt, line 835: case on step run running state and cancel?

internal/repository/prisma/step_run.go
```diff
@@ -769,6 +765,111 @@ func (s *stepRunEngineRepository) UpdateStepRun(ctx context.Context, tenantId, s
...
+				laterStepRunCp.ID,
+				dbsqlc.StepRunEventReasonMANUALRETRY,
+				dbsqlc.StepRunEventSeverityINFO,
+				fmt.Sprintf("Parent step run %s was replayed, resetting step run result", stepRun.StepReadableId.String),
+				nil,
+			)
+		}
+
+		// reset all later step runs to a pending state
+		_, err = s.queries.ReplayStepRunResetLaterStepRuns(ctx, tx, dbsqlc.ReplayStepRunResetLaterStepRunsParams{
```
grutt, line 852: Create archives for each later step run?

grutt (changes requested):
- [x] handle cancelation of downstream and 
- [x] handle archive of downstream replays

grutt (approved):

PR #540, opened 2024-05-29
feat: worker semaphore v2

grutt:
# Description

Previous worker semaphore approach led to blocking table level isolation #535.

## Type of change

## What's Changed

- [ ] Replaced Worker Semaphore object with table of slots


internal/repository/prisma/step_run.go
```diff
@@ -520,17 +524,25 @@ func (s *stepRunEngineRepository) assignStepRunToWorkerAttempt(ctx context.Conte
 
 	defer deferRollback(ctx, s.l, tx.Rollback)
 
-	_, err = tx.Exec(ctx, "SET TRANSACTION ISOLATION LEVEL SERIALIZABLE")
```
abelanger5, line 523: Along with this, remove the `serializeRetry` wrapper

prisma/schema.prisma
```diff
@@ -1203,16 +1205,19 @@ model Worker {
   // the runs which retrieve the group keys
   groupKeyRuns GetGroupKeyRun[]
 
-  semaphore WorkerSemaphore?
+  slots WorkerSemaphoreSlot[]
 }
 
-model WorkerSemaphore {
-  // the parent worker
+model WorkerSemaphoreSlot {
```
abelanger5, line 1211: I think an index on the `workerId` would make sense here, because we're not writing semaphore slots very often (relative to the amount of step runs we're assigning), and we don't want the acquire query to slow down as we put more slots in the database.

internal/repository/prisma/dbsqlc/step_runs.sql
```diff
@@ -306,18 +306,24 @@ WITH valid_workers AS (
...
         DISTINCT ON (w."id")
         w."id",
-        COALESCE(SUM(ws."slots"), 100) AS "slots"
+        slots."remainingSlots" AS "slots"
     FROM
         "Worker" w
-    LEFT JOIN
-        "WorkerSemaphore" ws ON ws."workerId" = w."id"
+  LEFT JOIN (
+    SELECT wss."workerId", (COALESCE(w."maxRuns", 100) - COUNT(wss."id")) AS "remainingSlots"
```
abelanger5: This inner select is joining all worker semaphores with workers in the entire table, which is going to get very slow. Why not just return the worker ids from `valid_workers` in the first CTE and use the result to get all worker semaphores which match condition `"workerId" IN (valid_workers)`?

internal/repository/prisma/dbsqlc/step_runs.sql
```diff
@@ -393,18 +399,24 @@ WITH valid_workers AS (
     SELECT
         DISTINCT ON (w."id")
         w."id",
-        COALESCE(SUM(ws."slots"), 100) AS "slots"
+        slots."remainingSlots" AS "slots"
     FROM
         "Worker" w
-    LEFT JOIN
-        "WorkerSemaphore" ws ON w."id" = ws."workerId"
+  LEFT JOIN (
```
abelanger5: Same comment as above

internal/repository/prisma/dbsqlc/step_runs.sql
abelanger5: I noticed we're doing something similar below in the `Acquire` query:

```sql
    -- (in a second CTE)
    SELECT wss."id" AS "slotId", wss."workerId" AS "workerId"
    FROM "WorkerSemaphoreSlot" wss
    JOIN valid_workers w ON wss."workerId" = w."id"
    WHERE wss."stepRunId" IS NULL
```

internal/repository/prisma/step_run.go
```diff
@@ -381,6 +381,20 @@ var genericRetry = func(l *zerolog.Event, maxRetries int, f func() error, msg st
 	return nil
 }
 
+func (s *stepRunEngineRepository) ReleaseWorkerSemaphoreSlot(ctx context.Context, tenantId, stepRunId string) (*dbsqlc.WorkerSemaphoreSlot, error) {
```
abelanger5: This method seems to be unused? And seems like a duplicate of `ReleaseStepRunSemaphore(ctx context.Context, tenantId, stepRunId string) error` which is used?

abelanger5 (changes requested):

internal/repository/prisma/dbsqlc/step_runs.sql
grutt: good catch

abelanger5 (approved):

PR #542, opened 2024-05-30
feat: webhook workers

### LINKED ISSUE #347: Serverless support
Tracking issue for serverless support. For implementation and details, see the [discussion](https://github.com/hatchet-dev/hatchet/discussions/346).

steebchen:
# Description

fix #347 

## Type of change

## TODO

- [x] webhook request check (mark step as failed if webhook returns non-200)
- [x] basic healthcheck (call API route before starting)
- [x] advanced healtcheck (ping continuously and take down worker if more than 3 failed healthchecks)
- [x] change action names to get from healthcheck instead of query
- [x] return response only once stuff is processed
- [x] do not require workflow IDs to be specified in registerWorkflow (or UI), do that automatically instead
- [x] think about: changing workflows (with their actions) or assigning new workflows to webhook workers, and killing existing ones/re-creating (internal note: check both worfklows and actions (via workflow hash?) and re-create if needed)
- [x] !!! the healthcheck needs to automatically update/register the workflow (it doesn't seem to be registering it the first time?)
	- [x] Go client
		- [x] requires refactor of .On->.RegisterWorkflow methods and signature
	- [x] TS client
- [x] TS client: check for signature as middleware
- [x] !!! verify webhook secret in health endpoint to make sure it is set correctly before sending any events
- [x] !! cleanup hardcoded env vars (token & tlsStrategy)
- [x] !! do not re-create token every time
- [x] investigate bug where no healthcheck requests are sent (might be invalid now, not sure)
- [x] encrypt token
- [x] rewrite queries to sqlc x)
- [ ] test services
- [ ] test register action
- [x] delete webhook workers
	- [x] (API & UI)
	- [x] stop currently running workers
- [x] add webhook worker name to DB and UI
- [ ] optional: vercel streaming or waitUntil


api-contracts/openapi/components/schemas/webhook_worker.yaml
```diff
@@ -0,0 +1,68 @@
+WebhookWorker:
+  x-resources: ["tenant", "event"]
```
abelanger5: Don't need `x-resources` defined for components, only paths.

api-contracts/openapi/components/schemas/webhook_worker.yaml
```diff
@@ -0,0 +1,68 @@
...
+  properties:
+    metadata:
+      $ref: "./metadata.yaml#/APIResourceMeta"
+    name:
+      type: string
+      description: The name of the webhook worker.
+    url:
+      type: string
+      description: The webhook url.
+    secret:
```
abelanger5, line 27: We probably don't want to pass this secret key with every webhook worker response. We probably only want to send it after the initial create endpoint is called with a separate component definition.

api-contracts/openapi/components/schemas/webhook_worker.yaml
```diff
@@ -0,0 +1,68 @@
...
+      description: The name of the webhook worker.
+    url:
+      type: string
+      description: The webhook url.
+    workflows:
+      type: array
+      description: The workflow IDs or names to register for this webhook worker. If not provided, workflows will be automatically detected.
+      items:
+        type: string
+    secret:
```
abelanger5, line 45: Is this used/does it make sense for the user to set the secret value?

api-contracts/openapi/components/schemas/webhook_worker.yaml
```diff
@@ -0,0 +1,68 @@
...
+      description: The secret key for validation. If not provided, a random secret will be generated.
+      minLength: 32
+  required:
+    - name
+    - url
+  type: object
+
+WebhookWorkerDeleteRequest:
+  properties:
+    id:
```
abelanger5: Instead of the ID being in the request body, we should place it as a URI param to be consistent with the rest of the API.

api-contracts/openapi/openapi.yaml
```diff
@@ -160,3 +160,9 @@ paths:
     $ref: "./paths/github-app/github-app.yaml#/repos"
   /api/v1/github-app/installations/{gh-installation}/repos/{gh-repo-owner}/{gh-repo-name}/branches:
     $ref: "./paths/github-app/github-app.yaml#/branches"
+  /api/v1/webhook-workers/{tenant}/list:
```
abelanger5:
```suggestion
  /api/v1/tenants/{tenant}/webhook-workers:
```

(for consistency)

api-contracts/openapi/openapi.yaml
```diff
@@ -160,3 +160,9 @@ paths:
...
+    $ref: "./paths/webhook-worker/webhook-worker.yaml#/list"
+  /api/v1/webhook-workers/{tenant}/create:
```
abelanger5: For consistency, this should be part of the `/api/v1/tenants/{tenant}/webhook-workers` endpoint.

api-contracts/openapi/openapi.yaml
```diff
@@ -160,3 +160,9 @@ paths:
...
+    $ref: "./paths/webhook-worker/webhook-worker.yaml#/create"
+  /api/v1/webhook-workers/{tenant}/delete:
```
abelanger5:
```suggestion
  /api/v1/webhook-workers/{webhook}:
```

(for consistency)

api-contracts/openapi/paths/webhook-worker/webhook-worker.yaml
```diff
@@ -0,0 +1,129 @@
...
+        description: Unauthorized
+      "405":
+        content:
+          application/json:
+            schema:
+              $ref: "../../components/schemas/_index.yaml#/APIErrors"
+        description: Method not allowed
+    summary: Create a webhook
+delete:
+  post:
```
abelanger5, line 42:
```suggestion
  delete:
```

api-contracts/openapi/paths/webhook-worker/webhook-worker.yaml
```diff
@@ -0,0 +1,129 @@
...
+    description: Deletes a webhook
+    operationId: webhook:delete
+    x-resources: ["tenant"]
```
abelanger5, line 46:
```suggestion
    x-resources: ["tenant", "webhook"]
```

Note that when you add `webhook` to the path param here, you'll want to specify `webhook` as part of the resource hierarchy as well, which will automatically populate the tenant resource.

api/v1/server/run/run.go
abelanger5, line 1: You will need to add a `getter` method to this file when you include `webhook` in the path URI. Something like:

```go
populatorMW.RegisterGetter("webhook", func(config *server.ServerConfig, parentId, id string) (result interface{}, uniqueParentId string, err error) {
		webhook, err := ... // TODO: method
	        if err != nil {
			return nil, "", err
		}
		return webhook, webhook.TenantId, nil
	})
```
	
Note that this getter returns the resource and also a parent id for the resource, which is the parent in the `x-resources` array (which is what defines the resource tree).

examples/webhook/main_e2e_test.go
abelanger5, line 1: 💯 awesome to see an e2e test for this

internal/randstr/randstr.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	return secret
+}
+
+func GenerateWebhookSecret() (string, error) {
+	return GenerateRandomString(32)
+}
+
+func MustGenerateRandomString(n int) string {
```
abelanger5: The `encryption` service has a `GenerateRandomBytes` method. It would be great to consolidate using one method or the other. You are free to replace all usage of `GenerateRandomBytes` with this package (would be great to add a simple unit test as well)

internal/services/webhooks/webhooks.go
```diff
@@ -0,0 +1,317 @@
...
+		if err != nil {
+			return fmt.Errorf("could not get webhook workers: %w", err)
+		}
+
+		for _, ww := range wws {
+			id := sqlchelpers.UUIDToStr(ww.ID)
+			if _, ok := c.registeredWorkerIds[id]; ok {
+				if ww.Deleted {
+					c.sc.Logger.Debug().Msgf("webhook worker %s of tenant %s has been deleted", id, tenantId)
+					err := c.sc.EngineRepository.Worker().DeleteWorkersByName(context.Background(), tenantId, "Webhook_"+id)
```
abelanger5: Why aren't we just using the `DeleteWorker` method for this? Don't we have access to the associated worker id?

internal/services/webhooks/webhooks.go
```diff
@@ -0,0 +1,317 @@
...
+
+	for _, tenant := range tenants {
+		tenantId := sqlchelpers.UUIDToStr(tenant.ID)
+
+		wws, err := c.sc.EngineRepository.WebhookWorker().ListWebhookWorkers(context.Background(), tenantId)
+		if err != nil {
+			return fmt.Errorf("could not get webhook workers: %w", err)
+		}
+
+		for _, ww := range wws {
```
abelanger5, line 80: We have a couple of blocking operations in this for loop. Should we consider using an `errgroup` or some other `sync` primitive to run sets of these registrations concurrently?

internal/services/webhooks/webhooks.go
```diff
@@ -0,0 +1,317 @@
...
+
+						isActive := false
+						_, err := c.sc.EngineRepository.Worker().UpdateWorker(context.Background(), tenantId, id, &repository.UpdateWorkerOpts{
+							IsActive: &isActive,
+						})
+						if err != nil {
+							c.sc.Logger.Err(err).Msgf("could not update worker")
+						}
+					} else {
+						c.sc.Logger.Printf("webhook worker %s of tenant %s failed one health check, retrying...", id, tenantId)
```
abelanger5: scope this to `Warn` level?

internal/services/webhooks/webhooks.go
```diff
@@ -0,0 +1,317 @@
...
+				if ww.Deleted {
+					c.sc.Logger.Debug().Msgf("webhook worker %s of tenant %s has been deleted", id, tenantId)
+					err := c.sc.EngineRepository.Worker().DeleteWorkersByName(context.Background(), tenantId, "Webhook_"+id)
+					if err != nil {
+						return fmt.Errorf("could not delete worker: %w", err)
+					}
+
+					delete(c.registeredWorkerIds, id)
+
+					continue
```
abelanger5: This `continue` seems unnecessary

pkg/repository/prisma/dbsqlc/api_tokens.sql
```diff
@@ -22,3 +22,21 @@ INSERT INTO "APIToken" (
     sqlc.narg('name')::text,
     @expiresAt::timestamp
 ) RETURNING *;
+
+-- name: UpsertAPIToken :one
```
abelanger5: Where is this used? Not finding it

prisma/schema.prisma
```diff
@@ -91,6 +91,41 @@ model UserSession {
...
+  secret     String
+  url        String  @unique
+  // stores the encrypted hatchet auth token
+  tokenValue String?
+  deleted    Boolean @default(false)
+
+  tokenId String?   @db.Uuid
+  token   APIToken? @relation(fields: [tokenId], references: [id], onDelete: Cascade, onUpdate: Cascade)
+
+  // TODO remove tenant ID as it is not needed
```
abelanger5: address TODO?

pkg/worker/webhook_handler.go
```diff
@@ -0,0 +1,158 @@
...
+
+		if r.Method != http.MethodPost {
+			writer.WriteHeader(http.StatusMethodNotAllowed)
+			_, _ = writer.Write([]byte("Method not allowed"))
+			return
+		}
+
+		data, err := io.ReadAll(r.Body)
+		if err != nil {
+			panic(err)
```
abelanger5: Should we be panicking in this method?

api/v1/server/handlers/webhook-worker/create.go
```diff
@@ -0,0 +1,44 @@
...
+	var wfs []string
+	if request.Body.Workflows != nil {
+		wfs = append(wfs, *request.Body.Workflows...)
+	}
+
+	ww, err := i.config.EngineRepository.WebhookWorker().UpsertWebhookWorker(ctx.Request().Context(), &repository.UpsertWebhookWorkerOpts{
+		TenantId:  tenant.ID,
+		Name:      request.Body.Name,
+		URL:       request.Body.Url,
+		Secret:    secret,
```
abelanger5: We should be storing the webhook secret encrypted as well.

api/v1/server/handlers/webhook-worker/create.go
```diff
@@ -0,0 +1,44 @@
...
+		if err != nil {
+			return nil, err
+		}
+		secret = s
+	} else {
+		secret = *request.Body.Secret
+	}
+
+	var wfs []string
+	if request.Body.Workflows != nil {
```
abelanger5: This is sent through the healthcheck now, right?

pkg/worker/webhook_handler.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/hatchet-dev/hatchet/pkg/client"
+)
+
+type WebhookHandlerOptions struct {
+	Secret string
+}
+
+type HealthCheckResponse struct {
+	Actions   []string `json:"actions"`
+	Workflows []string `json:"workflows"`
```
abelanger5, line 21: Shouldn't `Workflows` be a complete workflows object? We'd want to register the correct concurrency settings, timeout settings, etc, right?

abelanger5 (changes requested):

pkg/worker/webhook_handler.go
steebchen: This just exists to register the actions for the events – the actual workflow is registered separately and thus all settings like concurrency etc should be registered correctly

api/v1/server/handlers/webhook-worker/create.go
steebchen: yes, in theory we can remove this part, not sure if this would be useful in the future or not

internal/services/webhooks/webhooks.go
steebchen: we don't have access to the worker ID, no. We would need to track it in the DB and while using the worker name does not seem perfect, neither seems tracking this additionaly in the DB because you mentioned that you do not want too much webhook workers logic as this ideally could be done outside of the engine as well

api-contracts/openapi/components/schemas/webhook_worker.yaml
steebchen: I added it so we can set it in our tests, if we really don't want to allow users to set it ever I could workaround it by editing the DB in the tests, but it seemed like to much trouble. So technically a user can set this via the API, but it's not exposed in the UI.

pkg/repository/prisma/dbsqlc/api_tokens.sql
steebchen: was unused from an old implementation 👍

internal/services/webhooks/webhooks.go
abelanger5: Ok, that makes sense - but does it even make sense to delete the worker vs mark the worker as inactive? We don't currently delete workers, because it's useful to look back at the run history sometimes.

api/v1/server/handlers/webhook-worker/create.go
abelanger5: I think it would make sense to remove it if it's unused, otherwise it gets confusing to track where workflows are getting registered

pkg/worker/webhook_handler.go
abelanger5: Got it, where is the actual workflow registered from?

api-contracts/openapi/components/schemas/webhook_worker.yaml
abelanger5: Makes sense

pkg/worker/webhook_handler.go
steebchen: In the respective SDKs in the healthcheck response right now

api-contracts/openapi/openapi.yaml
```diff
@@ -138,15 +138,7 @@ paths:
...
-    $ref: "./paths/github-app/github-app.yaml#/branches"
-  /api/v1/webhook-workers/{tenant}/list:
-    $ref: "./paths/webhook-worker/webhook-worker.yaml#/list"
-  /api/v1/webhook-workers/{tenant}/create:
-    $ref: "./paths/webhook-worker/webhook-worker.yaml#/create"
-  /api/v1/webhook-workers/{tenant}/delete:
-    $ref: "./paths/webhook-worker/webhook-worker.yaml#/delete"
+  /api/v1/tenants/{tenant}/webhook-workers:
+    $ref: "./paths/webhook-worker/webhook-worker.yaml#/webhookworkers"
+  /api/v1/tenants/webhook-workers/{webhook}:
```
abelanger5:
```suggestion
  /api/v1/webhook-workers/{webhook}:
```

(for consistency)

internal/services/webhooks/webhooks.go
```diff
@@ -76,88 +79,108 @@ func (c *WebhooksController) check() error {
...
-			cleanup, err := c.run(tenantId, ww, token, h)
-			if err != nil {
-				c.sc.Logger.Error().Err(err).Msgf("error running webhook worker %s of tenant %s healthcheck", id, tenantId)
-				continue
-			}
-			if cleanup != nil {
-				c.cleanups = append(c.cleanups, cleanup)
-			}
+		if err = ec.Wait(); err != nil {
+			panic(err) // the error group above is not expected to error
```
abelanger5: I'm still not sure we should panic here, or the entire engine will restart right? Unless there's somewhere you're handling panic/recover.

abelanger5 (changes requested):

internal/services/webhooks/webhooks.go
steebchen: this line will never happen, I could have written _  = ec.Wait() as well

steebchen: I will rewrite this

steebchen: done, it's just using a goroutine now

api-contracts/openapi/components/schemas/webhook_worker.yaml
```diff
@@ -0,0 +1,69 @@
...
+      type: string
+      description: The name of the webhook worker.
+    url:
+      type: string
+      description: The webhook url.
+  required:
+    - metadata
+    - name
+    - url
+    - secret
```
grutt:
```suggestion

```

grutt: looks like this is unused?

api-contracts/openapi/paths/webhook-worker/webhook-worker.yaml
```diff
@@ -0,0 +1,123 @@
...
+          format: uuid
+          minLength: 36
+          maxLength: 36
+    responses:
+      "200":
+        content:
+          application/json:
+            schema:
+              $ref: "../../components/schemas/_index.yaml#/WebhookWorkerListResponse"
+        description: Successfully changed password
```
grutt:
```suggestion
        description: The list of webhooks
```

api/v1/server/handlers/webhook-worker/list.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/hatchet-dev/hatchet/api/v1/server/oas/gen"
+	"github.com/hatchet-dev/hatchet/api/v1/server/oas/transformers"
+	"github.com/hatchet-dev/hatchet/pkg/repository/prisma/db"
+)
+
+func (i *WebhookWorkersService) WebhookList(ctx echo.Context, request gen.WebhookListRequestObject) (gen.WebhookListResponseObject, error) {
+	tenant := ctx.Get("tenant").(*db.TenantModel)
+
+	webhooks, err := i.config.EngineRepository.WebhookWorker().ListActiveWebhookWorkers(context.TODO(), tenant.ID)
```
grutt:
```suggestion
	webhooks, err := i.config.EngineRepository.WebhookWorker().ListActiveWebhookWorkers(context.Background(), tenant.ID)
```

grutt: unless there's a reason

frontend/app/src/pages/main/tenant-settings/webhooks/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+              </div>
+            </div>
+          ))}
+
+          <CreateWebhookWorker
+            tenant={tenant.metadata.id}
+            showDialog={showCreateTokenDialog}
+            setShowDialog={setShowCreateTokenDialog}
+            onSuccess={() => {
+              void listWebhookWorkersQuery.refetch();
```
grutt, line 105: nit: invalidate the query instead of refetching

frontend/app/src/router.tsx
```diff
@@ -209,6 +209,17 @@ const routes: RouteObject[] = [
                     },
                   ),
               },
+              {
+                path: '/tenant-settings/webhooks',
```
grutt, line 213: nit: call the page webhook-workers since at some point we'll likely add webhooks for workflow run events?

internal/services/webhooks/webhooks.go
```diff
@@ -0,0 +1,342 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not start webhook worker: %w", err)
+	}
+
+	cleanups = append(cleanups, cleanup)
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	go func() {
+		ticker := time.NewTicker(5 * time.Second)
```
grutt, line 245: less frequent? maybe once per minute?

at some point it would be nice to "invalidate" on code changes

internal/whrequest/request.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+	req.Header.Set("X-Hatchet-Signature", sig)
+
+	for _, h := range headers {
+		h(req)
+	}
+
+	httpClient := &http.Client{
+		// use 10 minutes timeout
+		Timeout: time.Second * 600,
```
grutt, line 38: why 10m?

prisma/schema.prisma
```diff
@@ -90,6 +90,40 @@ model UserSession {
...
 }
 
+// WebhookWorker can be used to handle workflows via webhooks
+model WebhookWorker {
+  id        String   @id @unique @default(uuid()) @db.Uuid
+  createdAt DateTime @default(now())
+  updatedAt DateTime @default(now()) @updatedAt
+
+  name       String
+  secret     String
```
grutt, line 100: check: secret encrypted

grutt: looks good

prisma/schema.prisma
```diff
@@ -90,6 +90,40 @@ model UserSession {
...
+  url        String  @unique
+  // stores the encrypted hatchet auth token
+  tokenValue String?
```
grutt, line 103: why are we storing this here and also in token?

grutt (approved):
good work.
a few minor nits and questions but probably nothing show stopping

internal/whrequest/request.go
steebchen: Good point – I chose this randomly because the default http client doesn't have a timeout. I figured workflows which run over http should not really take that long, but it's set completely arbitrary and if it says we should document it

internal/services/webhooks/webhooks.go
steebchen: good catch I decreased it for testing

steebchen: Whenever it runs this it compares the hash of the actions and workflows, so it kinda invalidates on code changes if that's what you meant

frontend/app/src/router.tsx
steebchen: hmm yeah, though I'm not sure if we want to expose the name Webhook Workers to the users as I was thinking it might be confused with regular Workers. Maybe "Webhook Apps" although that also sounds weird (Inngest uses the term app)

frontend/app/src/pages/main/tenant-settings/webhooks/index.tsx
steebchen: I do need to refetch the data so that the newly added item shows, no? Or what am I missing?

api-contracts/openapi/components/schemas/webhook_worker.yaml
steebchen: good catch, interesting that this doesn't error as the field doesn't exist

prisma/schema.prisma
steebchen: Where in `token` do you mean?

frontend/app/src/pages/main/tenant-settings/webhooks/index.tsx
grutt:
```
queryClient.invalidateQueries({
        queryKey: queries.workflowRuns.get(
          tenant.metadata.id,
          workflowRun.metadata.id,
        ).queryKey,
      });

```

Something like this will invalidate all requests with that query and refetch accordingly, i think refetch will just refetch that specific request

grutt: minor thing and not required for this pr

frontend/app/src/router.tsx
grutt: we can change it later 🫠

internal/services/webhooks/webhooks.go
grutt: nice, i meant more for adding actions but i dont think its critical

prisma/schema.prisma
grutt: my fault

frontend/app/src/router.tsx
steebchen: aka never 😂

frontend/app/src/pages/main/tenant-settings/webhooks/index.tsx
steebchen: if the new response is cached it would not matter though I think, is it not cached atm?

abelanger5 (approved):

PR #547, opened 2024-05-31
feat:  default email alert group

grutt:
# Description

When new users join, there is no alerting configured by default. While this is the intended behavior for things like workflow runs which may be noisy, it may be problematic for alerts for token expiration.

This change adds additional configuration settings for alerts and exposes the tenant member list as a special email group.

## Type of change

## What's Changed

- [x] add tenant alert setting configuration to enable workflow run failure and token expiration alerts
- [x] add tenant configuration to enable tenant member email alert group
- [x] add query to get emails for a tenant
- [x] update alerters with config
- [x] expose config on ui


grutt:
image

internal/repository/tenant_alerting.go
```diff
@@ -40,12 +44,17 @@ type TenantAlertingAPIRepository interface {
 	DeleteTenantAlertGroup(tenantId string, id string) error
 }
 
+type TenantAlertEmailGroupForSend struct {
+	TenantId pgtype.UUID `json:"tenantId"`
+	Emails   string      `json:"emails"`
```
abelanger5:
```suggestion
	Emails []string `validate:"required,dive,email,max=255"`
```

(assuming this is meant to be a list of emails) - better to validate with the string array and concat in the repository.

prisma/schema.prisma
```diff
@@ -134,6 +134,7 @@ model Tenant {
   rateLimits                RateLimit[]
   stepRateLimits            StepRateLimit[]
   alertEmailGroups          TenantAlertEmailGroup[]
+  alertMemberEmails         Boolean @default(true)
```
abelanger5, line 138: I don't know at first glance what this does - could you add a comment?

internal/repository/prisma/tenant_alerting.go
```diff
@@ -142,22 +147,48 @@ func (r *tenantAlertingEngineRepository) GetTenantAlertingSettings(ctx context.C
...
+	}
+
 	tenant, err := r.queries.GetTenantByID(ctx, tx, pgTenantId)
 
 	if err != nil {
 		return nil, err
 	}
 
+	if tenant.AlertMemberEmails {
+		emails, err := r.queries.GetMemberEmailGroup(ctx, tx, pgTenantId)
```
abelanger5, line 174: I actually looks like this tx is never committed, is that right? Should fix ASAP.

abelanger5 (changes requested):

internal/repository/prisma/tenant_alerting.go
grutt: good catch

abelanger5 (approved):

PR #559, opened 2024-06-05
feat: limits

grutt:
# Description

This PR introduces a new feature to impose and manage tenant-specific limits within our system. The key metrics being tracked include task executions per day, concurrent workers (burstable), and users. The feature includes mechanisms for alerting and blocking based on predefined usage thresholds.

## Type of change

## What's Changed

- [x] Add db models for tracking tenant resource limits
- [x] Add entitlement checks and metering queries to relevant requests   
- [x] Expose Tenant Resource Limit API
- [x] Add new Tenant Resource Limit Alert Ticker


internal/integrations/alerting/email.go
```diff
@@ -59,3 +59,42 @@ func (t *TenantAlertManager) sendEmailExpiringTokenAlert(tenant *dbsqlc.Tenant,
...
+func (t *TenantAlertManager) sendEmailTenantResourceLimitAlert(tenant *dbsqlc.Tenant, emailGroup *repository.TenantAlertEmailGroupForSend, payload *alerttypes.ResourceLimitAlert) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	var subject string
+	var summary string
+	var summary2 string
+
+	if payload.AlertType == string(dbsqlc.TenantResourceLimitAlertTypeAlarm) {
+		subject = fmt.Sprintf("Hatchet Limit Alarm! `%s` resource is at %d%% of its limit (%d/%d)", payload.Resource, payload.Percentage, payload.CurrentValue, payload.LimitValue)
```
abelanger5:
```suggestion
		subject = fmt.Sprintf("%s has exhausted %d%% of its limit (%d/%d)", payload.Resource, payload.Percentage, payload.CurrentValue, payload.LimitValue)
```

Also, if possible, can we map things like `WORKFLOW_RUN` -> `Workflow Run`?

internal/integrations/alerting/email.go
```diff
@@ -59,3 +59,42 @@ func (t *TenantAlertManager) sendEmailExpiringTokenAlert(tenant *dbsqlc.Tenant,
...
+		summary = "We're sending you this alert because a resource on your Hatchet tenant is approaching its usage limit."
+		summary2 = "Once the limit is reached, any further resource usage will be denied."
+	}
+
+	if payload.AlertType == string(dbsqlc.TenantResourceLimitAlertTypeExhausted) {
+		subject = fmt.Sprintf("Hatchet Limit Exhausted! `%s` resource is at 100%% of its limit (%d/%d)", payload.Resource, payload.CurrentValue, payload.LimitValue)
```
abelanger5: Same as above

internal/services/ingestor/ingestor.go
```diff
@@ -92,14 +102,18 @@ func NewIngestor(fs ...IngestorOptFunc) (Ingestor, error) {
...
+		streamEventRepository:  opts.streamEventRepository,
+		entitlementsRepository: opts.entitlementsRepository,
+
+		logRepository: opts.logRepository,
+		mq:            opts.mq,
+		v:             validator.NewDefaultValidator(),
 	}, nil
 }
 
+var ErrResourceExhausted = fmt.Errorf("resource exhausted: cannot create event for tenant at this time")
```
abelanger5: Why do we have a new `ErrResourceExhausted` within the service instead of using a shared error?

internal/repository/prisma/tenant_limit.go
```diff
@@ -0,0 +1,249 @@
...
+			}
+			limit.Value = int32(workerCount)
+		}
+
+	}
+
+	return limits, nil
+}
+
+func (t *tenantLimitRepository) CanCreate(ctx context.Context, resource dbsqlc.LimitResource, tenantId string) (bool, error) {
```
abelanger5: We probably want to use the cache here, it's not ideal that every single lookup to determine whether to create an event/workflow run involves a database lookup, part of the design here would be to reduce the database load. I'd suggest at least caching for a minute here.

internal/repository/repository.go
```diff
@@ -40,6 +40,10 @@ type EngineRepository interface {
 	RateLimit() RateLimitEngineRepository
 }
 
+type EntitlementsRepository interface {
```
abelanger5, line 43: Why do we have the `EntitlementsRepository`? Why not have the `TenantLimitRepository` as a direct call in the `ApiRepository` and `EngineRepository` which also gets passed as an argument to the metering struct?

internal/repository/prisma/tenant_limit.go
```diff
@@ -0,0 +1,249 @@
...
+		Resource: dbsqlc.NullLimitResource{
+			LimitResource: resource,
+			Valid:         true,
+		},
+	})
+
+	if err == pgx.ErrNoRows {
+		t.l.Warn().Msgf("no %s tenant limit found, creating default limit", string(resource))
+
+		err = t.CreateTenantDefaultLimits(ctx, tenantId)
```
abelanger5, line 199: Just to clarify, this means that all existing tenants will get a limit set to the default?

abelanger5 (changes requested):

internal/repository/prisma/tenant_limit.go
grutt: yes, on first meter it'll insert if none exists

internal/integrations/alerting/email.go
grutt: added some string transforms

internal/services/ingestor/ingestor.go
grutt: good catch

internal/repository/repository.go
grutt: I figured it make more sense to separate this responsibility and share it across the two repositories. in the future we'll likely want to add additional things on the entitlement repositories (i.e. feature flags) so i think having it as its own thing make sense

abelanger5 (approved):

PR #561, opened 2024-06-06
fix: ensure auth redirects throw an error

abelanger5:
# Description

Auth redirects aren't currently throwing an error, which means there are situations where security should be disabled `security: []` causing an authenticated user to be able to call a handler. 

## Type of change


steebchen (approved):
confirmed it resolves the issue

PR #568, opened 2024-06-10
fix: only join latest workflow version

grutt:
# Description

Old workflow version event triggers would trigger new workflow runs.

## Type of change

## What's Changed

- Only filter Workflows to Run from the latest versions of workflows

### DIFF
--- a/internal/repository/prisma/dbsqlc/workflows.sql
+++ b/internal/repository/prisma/dbsqlc/workflows.sql
@@ -364,10 +364,16 @@ INSERT INTO "WorkflowTriggerScheduledRef" (
 ) RETURNING *;
 
 -- name: ListWorkflowsForEvent :many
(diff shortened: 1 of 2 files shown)


internal/repository/prisma/dbsqlc/workflows.sql
```diff
@@ -364,10 +364,16 @@ INSERT INTO "WorkflowTriggerScheduledRef" (
 ) RETURNING *;
 
 -- name: ListWorkflowsForEvent :many
+WITH LatestWorkflowVersions AS (
```
abelanger5: This queries every workflow in the database. It would be more performant to track down the candidate workflow versions first, and then make sure the candidate workflow versions are the latest version.

abelanger5 (changes requested):

abelanger5 (approved):

PR #582, opened 2024-06-12
fix: don't requeue step runs immediately

abelanger5:
# Description

When workflow runs sit in the queue for a long time, any step runs which are started are immediately requeued, because we only set `requeueAfter` in when the step run is first created. This is causing collisions in the `AssignStepRunToWorker` method. 

This fix simply updates the `requeueAfter` time in the `queueStepRun` query, which occurs when the step run is placed into the `PENDING_ASSIGNMENT` state for the first time. 

## Type of change

### DIFF
--- a/internal/repository/prisma/step_run.go
+++ b/internal/repository/prisma/step_run.go
@@ -527,7 +527,7 @@ func (s *stepRunEngineRepository) assignStepRunToWorkerAttempt(ctx context.Conte
 
 	if err != nil {
 		if errors.Is(err, pgx.ErrNoRows) {
-			return nil, repository.ErrNoWorkerAvailable
+			s.l.Warn().Err(err).Msg("no rows returned from worker assign")
 		}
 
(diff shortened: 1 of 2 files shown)


internal/repository/prisma/step_run.go
```diff
@@ -527,7 +527,7 @@ func (s *stepRunEngineRepository) assignStepRunToWorkerAttempt(ctx context.Conte
 
 	if err != nil {
 		if errors.Is(err, pgx.ErrNoRows) {
-			return nil, repository.ErrNoWorkerAvailable
+			s.l.Warn().Err(err).Msg("no rows returned from worker assign")
```
grutt, line 530: do we still want to return, we're just going to err on ln 533

grutt (approved):

internal/repository/prisma/step_run.go
abelanger5: Right, we'll error but I think that's somewhat expected - we should bubble up this error

PR #595, opened 2024-06-14
fix: allow key generation script to run as non-root user

### LINKED ISSUE #538: docs: Kubernetes Quickstart script only works as root user
When run as a regular user, the script found at https://docs.hatchet.run/self-hosting/kubernetes#generate-encryption-keys fails because the key files are created by root and are marked as read-only.

abelanger5:
# Description

Fixes #538 

## Type of change


steebchen (approved):
lgtm

PR #598, opened 2024-06-14
fix: add delay to requeue and reassign to allow workers to reconnect

### LINKED ISSUE #552: bug: after restarting the engine, the same step gets executed twice, leading to unexpected behavior
After restarting the engine, the same step gets executed twice, leading to unexpected behavior. Below are the steps to reproduce this issue and the observed logs highlighting the problem.

1. Run the engine: In Terminal 1, execute docker compose up.
2. Start the worker script: In Terminal 2, run the worker script.
3. Start the client script: In Terminal 3, run the client script (just call run_workflow).
4. Wait for one minute, then kill the engine with docker compose down.
5. Wait another minute and start the engine again using docker compose up.

Upon restarting the engine, the logs indicate a doubled execution of the same step:

```
2024-06-04T12:33:50.018Z: Step 1
2024-06-04T12:33:50.036Z: Input: {'items': ['c']}
2024-06-04T12:33:50.05Z: Processing 0
2024-06-04T12:33:55.023Z: Processing 1
2024-06-04T12:34:00.025Z: Processing 2
2024-06-04T12:34:05.026Z: Processing 3
2024-06-04T12:34:10.027Z: Processing 4
2024-06-04T12:34:15.027Z: Processing 5
2024-06-04T12:34:20.029Z: Processing 6
2024-06-04T12:34:25.032Z: Processing 7
2024-06-04T12:34:30.033Z: Processing 8
2024-06-04T12:34:35.036Z: Processing 9
2024-06-04T12:34:40.036Z: Processing 10
2024-06-04T12:35:30.05Z: Processing 20
2024-06-04T12:35:34.453Z: Step 1
2024-06-04T12:35:34.464Z: Input: {'items': ['c']}
2024-06-04T12:35:34.471Z: Processing 0
2024-06-04T12:35:35.055Z: Processing 21
2024-06-04T12:35:39.455Z: Processing 1
2024-06-04T12:35:40.058Z: Processing 22
```

The workflow simulates a long running operation:

```
@hatchet.workflow(name="long-workflow")
class LongWorkflow:
    @hatchet.step(name="step1", timeout="1h")
    def step1(self, context):
        context.log("Step 1")
        context.log(f'Input: {context.workflow_input()}')
        for i in range(60):
            context.log(f'Processing {i}')
            context.put_stream(f'Processing item: {i}')
            context.sleep(5)
        context.log('Done sleeping, returning')
        return {"result": "success"}
```

python sdk: hatchet_sdk-0.24.0

Question

Is this behavior intentional, or is it a bug? The expectation is that after restarting, the steps should either continue from where they left off or start over completely, not both.

abelanger5:
# Description

Fixes #552 

## Type of change

### DIFF
--- a/internal/services/controllers/jobs/controller.go
+++ b/internal/services/controllers/jobs/controller.go
@@ -134,10 +134,12 @@ func (jc *JobsControllerImpl) Start() (func() error, error) {
 
 	wg := sync.WaitGroup{}
 
+	startedAt := time.Now()
+
 	_, err := jc.s.NewJob(
 		gocron.DurationJob(time.Second*1),
 		gocron.NewTask(
-			jc.runStepRunRequeue(ctx),
+			jc.runStepRunRequeue(ctx, &startedAt),
 		),
 	)
 
@@ -149,7 +151,7 @@ func (jc *JobsControllerImpl) Start() (func() error, error) {
 	_, err = jc.s.NewJob(
 		gocron.DurationJob(time.Second*1),
 		gocron.NewTask(
-			jc.runStepRunReassign(ctx),
+			jc.runStepRunReassign(ctx, &startedAt),
 		),
 	)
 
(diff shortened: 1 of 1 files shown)


internal/services/controllers/jobs/controller.go
```diff
@@ -553,8 +555,13 @@ func (ec *JobsControllerImpl) handleStepRunQueued(ctx context.Context, task *msg
 	return ec.queueStepRun(ctx, metadata.TenantId, metadata.StepId, payload.StepRunId)
 }
 
-func (jc *JobsControllerImpl) runStepRunRequeue(ctx context.Context) func() {
+func (jc *JobsControllerImpl) runStepRunRequeue(ctx context.Context, startedAt *time.Time) func() {
```
steebchen: any reason this is a pointer when the single execution always provides a value?

abelanger5: because we're saving TONs of memory :)

... there's no real reason, I usually default to passing time.Time as a pointer

steebchen: not a big deal but it seems like if the value is required it could make more sense to use a value, especially as you explicitly need to check for != nil.

steebchen (approved):

PR #615, opened 2024-06-19
feat: add endpoint for listing queue metrics

### LINKED ISSUE #608: Add metrics API and document usage with Keda
Add a metrics API to return the number of queued step runs and workflow runs in various states. 

This should support filtering by metadata and workflow. We should document how Keda can be used to autoscale based on queue depth.

abelanger5:
# Description

Adds an endpoint at `GET /api/v1/tenants/:tenant/queue-metrics` for displaying queue metrics.

Fixes #608 

## Type of change


pkg/repository/prisma/tenant.go
```diff
@@ -169,6 +179,82 @@ func (r *tenantAPIRepository) DeleteTenantMember(memberId string) (*db.TenantMem
...
+		totalParams.WorkflowIds = uuids
+	}
+
+	tx, err := r.pool.Begin(context.Background())
+
+	if err != nil {
+		return nil, err
+	}
+
+	defer deferRollback(context.Background(), r.l, tx.Rollback)
```
grutt, line 222: why are we deferring a rollback for read only queries?

grutt (approved):

pkg/repository/prisma/tenant.go
abelanger5: doesn't seem to make a difference: https://www.postgresql.org/message-id/flat/07FDEE0ED7455A48AC42AC2070EDFF7C67EBDF%40corpsrv2.tazznetworks.com

the important part is that we're either committing or rolling back

PR #623, opened 2024-06-19
style(navbar): use cursor pointer for navbar links

steebchen:
# Description

Uses a pointer cursor for these nav items

image

## Type of change

- [x] Style change


abelanger5 (approved):
This was an intentional change as I liked Linear's sidebar...but this is probably better for accessibility.

PR #627, opened 2024-06-20
fix(worker): deprecate On in favor of RegisterWorkflow

steebchen:
# Description

## Type of change


pkg/worker/worker.go
```diff
@@ -213,6 +213,14 @@ func (w *Worker) NewService(name string) *Service {
 	return svc
 }
 
+func (w *Worker) RegisterWorkflow(workflow workflowConverter) error {
+	wf, ok := workflow.(*WorkflowJob)
```
steebchen: seems like of the possible values of `workflowConverter` only `WorkflowJob` can have triggers

abelanger5 (approved):

PR #631, opened 2024-06-24
feat(frontend): cloud api integration

abelanger5:
# Description

Adds additional API client to the frontend for the Hatchet Cloud API. 

## Type of change

- [X] New feature (non-breaking change which adds functionality)

PR #633, opened 2024-06-24
refactor(random): refactor random string generation

steebchen:
# Description

Refactor random string generation

## Type of change

## What's Changed

- move method to new random package
- rename GenerateToRandomBytes to GenerateRandomString
- use specific characters for random string
- add tests


pkg/repository/github.go
abelanger5, line 1: Just note, we're removing this file in #631

abelanger5 (approved):

PR #639, opened 2024-06-25
feat: check security service

grutt:
# Description

It is desirable to alert users running vulnerable instances

## Type of change

## What's Changed

- [x] On api start, check the security.onhatchet.run instance for messages relevant to the running version

## To Do

- [ ] get the current running version


pkg/config/server/server.go
```diff
@@ -79,6 +81,11 @@ type ConfigFileRuntime struct {
 	Limits LimitConfigFile `mapstructure:"limits" json:"limits,omitempty"`
 }
 
+type SecurityCheckConfigFile struct {
+	Enabled  bool   `mapstructure:"enabled" json:"enabled,omitempty" default:"true"`
+	Endpoint string `mapstructure:"endpoint" json:"endpoint,omitempty" default:"https://security.onhatchet.run"`
```
abelanger5:
```suggestion
	Endpoint string `mapstructure:"endpoint" json:"endpoint,omitempty" default:"https://security.hatchet.run"`
```

prisma/schema.prisma
abelanger5, line 1: Need to run `task generate-go`

pkg/security/security.go
```diff
@@ -0,0 +1,73 @@
...
+		Enabled:  opts.Enabled,
+		Endpoint: opts.Endpoint,
+		Logger:   opts.Logger,
+		Version:  opts.Version,
+		Repo:     repo,
+	}
+}
+
+func (a DefaultSecurityCheck) Check() {
+	if !a.Enabled {
```
abelanger5, line 36: Add a `panic...recover` deferred method here so even if a random dependency panics we don't take down the engine.

abelanger5 (changes requested):

abelanger5 (approved):

PR #649, opened 2024-06-26
feat: tenant partitioning

abelanger5:
# Description

Adds partitioning logic across engine instances that perform list operations on tenants. 

This isn't a true database partition - each engine version generates a partition id and rebalances tenants across partitions on startup. The logic relies on partitions being active (i.e. sitting on a hot loop) as partition heartbeats are updated when the partition is used. Inactive partitions are automatically removed and tenants on those partitions are rebalanced. 

## Type of change


Taskfile.yaml
```diff
@@ -67,7 +66,7 @@ tasks:
       - task: generate-sqlc
   seed-dev:
     cmds:
-      - sh ./hack/dev/run-go-with-env.sh run github.com/steebchen/prisma-client-go migrate deploy
+      - sh ./hack/dev/run-go-with-env.sh run github.com/steebchen/prisma-client-go migrate dev --skip-generate
```
steebchen, line 69: I would prefer not to do this as otherwise this will ask you to create a migration if there are new changes which is then however not synced with atlas. Why do you need to run dev here? Alternatively we can also do db push?

abelanger5: Without this, I believe the prisma command which is generating `schema.sql` wasn't generating the correct `schema.sql` as it looks at the migration directory, which meant the atlas migrations weren't being created properly. On the other hand, if we actually deploy from the prisma schema using prisma, the atlas migrations won't succeed. This is the only configuration that seems to account for both cases (both @grutt and I had issues with migrations until we reverted).

steebchen: hmm I see, I had issues the other way round but as long as it works for you now that's fine then

steebchen (approved):
nice!

grutt (approved):

PR #677, opened 2024-07-01
feat: worker paused state

### LINKED ISSUE #663: feat: manually mark workers as inactive
It can sometimes be important to manually "lock" a worker so new step runs aren't assigned but it continues to process old step runs.

grutt:
# Description

Adds a Paused state to worker and UI affordance to update this state. When a worker is Paused work will not be assigned to the worker but all connections will not be interrupted.

Fixes #663 

## Type of change

## What's Changed

- [x] adds isPaused state to worker model
- [x] adds update worker api endpoint
- [x] adds toggle button to pause/resume work for worker
- [x] adds paused state to isActive worker api model


api/v1/server/oas/transformers/worker.go
```diff
@@ -63,19 +67,27 @@ func ToWorker(worker *db.WorkerModel) *gen.Worker {
...
 
 	dispatcherId := uuid.MustParse(pgUUIDToStr(worker.DispatcherId))
 
 	maxRuns := int(worker.MaxRuns)
 
 	status := gen.ACTIVE
 
+	if worker.IsPaused {
+		status = gen.PAUSED
+	}
```
steebchen, line 80: I think we might need to move this below the next if, so that it shows paused as long as it is paused, even when it is offline

frontend/app/src/pages/main/workers/$worker/index.tsx
```diff
@@ -82,6 +104,19 @@ export default function ExpandedWorkflowRun() {
 
   const healthy = isHealthy(worker);
 
+  const updateWorker = useMutation({
+    mutationKey: ['worker:update', worker?.metadata.id],
+    mutationFn: async (data: UpdateWorkerRequest) =>
+      (await api.workerUpdate(worker!.metadata.id, data)).data,
+    onSuccess: async () => {
+      // Handle success logic here
```
steebchen: unnecessary comment

steebchen (approved):
lgtm with comments

api/v1/server/oas/transformers/worker.go
grutt: i'm not sure, i think we want to surface inactive even if it is paused since a paused worker may go offline and never come back online

abelanger5 (approved):

PR #681, opened 2024-07-02
Fix: #285

### LINKED ISSUE #285: bug(frontend): copy to clipboard doesn't work on firefox (within a secure context)
Error message: 

image

We are directly using `navigator.clipboard` in [`copy-to-clipboard.tsx`](https://github.com/hatchet-dev/hatchet/blob/82cb1c093641f815aad0500a8e4ba824b8aeb5fa/frontend/app/src/components/ui/copy-to-clipboard.tsx#L26), this should be a very easy switch to something like `clipboardjs`.

shiv4nk4r:
# Description

Fixes #285

## Type of change

## What's Changed

- [ ] [react-copy-to-clipboard](https://www.npmjs.com/package/react-copy-to-clipboard) to change the behavior of `copy-to-clipboard.tsx` file.

### DIFF
--- a/frontend/app/package.json
+++ b/frontend/app/package.json
@@ -87,6 +87,7 @@
     "@types/node": "^20.10.1",
     "@types/qs": "^6.9.10",
     "@types/react": "^18.2.37",
+    "@types/react-copy-to-clipboard": "^5.0.7",
     "@types/react-dom": "^18.2.15",
     "@types/react-syntax-highlighter": "^15.5.11",
     "@typescript-eslint/eslint-plugin": "^6.10.0",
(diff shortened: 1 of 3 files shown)


frontend/app/package.json
```diff
@@ -69,6 +69,7 @@
     "prism-react-renderer": "^2.3.0",
     "qs": "^6.11.2",
     "react": "^18.2.0",
+    "react-copy-to-clipboard": "^5.1.0",
```
abelanger5, line 72: There should be a corresponding change in `pnpm-lock.yaml`, no?

shiv4nk4r: sure I will update the lock file aswell

shiv4nk4r:
I have updated the file

PR #696, opened 2024-07-06
fix: step run queries

grutt:
# Description

Improves performance under high load of wfrs

## Type of change


sql/migrations/20240706225700_v0.36.0.sql
```diff
@@ -0,0 +1,30 @@
+-- Create index "Action_tenantId_idx" to table: "Action"
+CREATE INDEX "Action_tenantId_idx" ON "Action" ("tenantId");
```
abelanger5: These will all need to be modified to `CREATE INDEX CONCURRENTLY IF NOT EXISTS` (see v0.35.1 for an example)

abelanger5 (changes requested):

abelanger5 (approved):

PR #701, opened 2024-07-08
fix: batched reassign and requeue

grutt:
# Description

Addresses db load under high throughput

## Type of change

## What's Changed

- [x] batches assignment of step runs
- [x] limits total number of reassigns per poll

### DIFF
--- a/internal/services/controllers/jobs/controller.go
+++ b/internal/services/controllers/jobs/controller.go
@@ -634,7 +634,7 @@ func (ec *JobsControllerImpl) runStepRunRequeueTenant(ctx context.Context, tenan
 		batch := stepRuns[start:end]
 
 		g.Go(func() error {
-			scheduleCtx, cancel := context.WithTimeout(ctx, 1*time.Second)
+			scheduleCtx, cancel := context.WithTimeout(ctx, 3*time.Second)
 			defer cancel()
 
 			for i := range batch {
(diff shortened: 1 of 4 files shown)


internal/services/controllers/jobs/controller.go
```diff
@@ -708,44 +727,63 @@ func (ec *JobsControllerImpl) runStepRunReassignTenant(ctx context.Context, tena
...
 
-		// wrap in func to get defer on the span to avoid leaking spans
 		g.Go(func() error {
-			ctx, span := telemetry.NewSpan(ctx, "handle-step-run-reassign-step-run")
-			defer span.End()
+			for i := range batch {
+				stepRunCp := batch[i]
 
-			stepRunId := sqlchelpers.UUIDToStr(stepRunCp.StepRun.ID)
+				// wrap in func to avoid leaking spans
```
grutt, line 749: im not sure if this is needed

internal/services/controllers/jobs/controller.go
```diff
@@ -623,29 +623,50 @@ func (ec *JobsControllerImpl) runStepRunRequeueTenant(ctx context.Context, tenan
...
+		if end > len(stepRuns) {
+			end = len(stepRuns)
+		}
+		batch := stepRuns[start:end]
 
-		// wrap in func to get defer on the span to avoid leaking spans
 		g.Go(func() error {
-			ctx, span := telemetry.NewSpan(ctx, "handle-step-run-requeue-step-run")
-			defer span.End()
+			scheduleCtx, cancel := context.WithTimeout(ctx, 1*time.Second)
```
grutt: not sure about this. it should be fine considering requeueAfter is set to now + 4s and initial context takes up 1 second and sets sr into a state so they should not be re-re-assigned. but writing 100 sr updates maybe shouldnt take more than 1s?

PR #717, opened 2024-07-12
feat: soft delete

grutt:
# Description

Changes the delete strategy from hard to soft delete for relations with deep cascades

## Type of change


pkg/repository/prisma/dbsqlc/step_runs.sql
```diff
@@ -846,3 +863,51 @@ FROM
...
+    "StepRun"
+SET
+    "input" = NULL,
+    "output" = NULL,
+    "error" = NULL
+WHERE
+    "id" IN (SELECT "id" FROM deleted_with_limit)
+RETURNING
+    (SELECT count FROM deleted_count) as total,
+    (SELECT count FROM deleted_count) - (SELECT COUNT(*) FROM deleted_with_limit) as remaining,
```
abelanger5: I'm concerned about the logic of returning remaining - I'm curious if there's a chance that the counts might not align if new data is pushed while executing this query. Are we guaranteed that both counts are seeing the same snapshot of the data?

Additionally, the count across all `deleted_count` is not ideal. Could we modify these queries so that we return `has_more` as a boolean, call `LIMIT sqlc.arg('limit') + 1`, and just case on whether there's at least 1 more step run than the limit? Since we're not using `remaining` for anything other than signifying whether to continue the deletion loop.

grutt: good call, revised

pkg/repository/prisma/db/db_gen.go
```diff
@@ -946,14 +949,15 @@ model WorkflowRun {
...
+  @@index([deletedAt])
 }
 
 model GetGroupKeyRun {
   // base fields
   id        String    @id @unique @default(uuid()) @db.Uuid
   createdAt DateTime  @default(now())
   updatedAt DateTime  @default(now()) @updatedAt
-  deletedAt DateTime?
+  deletedAt DateTime? // TODO verify we're setting this
```
abelanger5, line 960: Address todo?

pkg/repository/prisma/dbsqlc/events.sql
```diff
@@ -147,27 +150,77 @@ SELECT
...
 WHERE
+    events."deletedAt" IS NOT NULL AND
     "tenantId" = @tenantId::uuid AND
     "id" = ANY (sqlc.arg('ids')::uuid[]);
 
--- name: DeleteExpiredEvents :one
-WITH expired_events_count AS (
-    SELECT COUNT(*) as count
+-- name: SoftDeleteExpiredEvents :one
+WITH has_more AS (
```
abelanger5: I think we are still risking having the count returned within `has_more` be different from the count within `expired_events_with_limit` since we're skipping locked in 1 case but not the other. 

Perhaps place the lock on all rows (the `+ 1` row included), and just use a row_number to filter by `<= limit` from the CTE later on?

pkg/config/server/server.go
```diff
@@ -37,7 +37,7 @@ type ServerConfigFile struct {
 
 	MessageQueue MessageQueueConfigFile `mapstructure:"msgQueue" json:"msgQueue,omitempty"`
 
-	Services []string `mapstructure:"services" json:"services,omitempty" default:"[\"health\", \"ticker\", \"grpc\", \"eventscontroller\", \"queue\", \"webhookscontroller\", \"heartbeater\"]"`
+	Services []string `mapstructure:"services" json:"services,omitempty" default:"[\"health\", \"ticker\", \"grpc\", \"eventscontroller\", \"queue\", \"webhookscontroller\", \"heartbeater\", \"retention\"]"`
```
abelanger5, line 40: As a heads up, this is not going to be enabled for existing installations that have a generated config file (via docker compose). We should support regenerating the config file with changes like this eventually.

abelanger5 (approved):

PR #735, opened 2024-07-20
feat: deduplicated enqueue

### LINKED ISSUE #666: feat: Deduplicated enqueue
I'm wondering if hatchet has any built in support for any sort of deduplicated enqueue, where a task/step/workflow could be enqueued in an idempotent way i.e. deduplicated based on its parameters.

I realize that there are some tricky details here, but this would be super nice.

abelanger5:
# Description

Adds support for deduplicated enqueueing of workflow runs based on a custom `dedupe` key in the additional metadata of a workflow run. 

Docs to come. 

Fixes #666 

## Type of change


pkg/repository/prisma/workflow_run.go
```diff
@@ -657,6 +658,37 @@ func createNewWorkflowRun(ctx context.Context, pool *pgxpool.Pool, queries *dbsq
 				return nil, err
 			}
 			createParams.Additionalmetadata = additionalMetadataBytes
+
+			// if additional metadata contains a "dedupe" key, use it as the dedupe value
+			if dedupeValue, ok := opts.AdditionalMetadata["dedupe"]; ok {
+				if dedupeStr, ok := dedupeValue.(string); ok {
```
grutt, line 664: should we cast int to string or just force string?

pkg/repository/prisma/workflow_run.go
```diff
@@ -865,3 +905,12 @@ func defaultWorkflowRunPopulator() []db.WorkflowRunRelationWith {
 		),
 	}
 }
+
+func isUniqueViolationOnDedupe(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	return strings.Contains(err.Error(), "WorkflowRunDedupe_tenantId_workflowId_value_key") &&
```
grutt, line 920: put a note on this index that we're casing on it

pkg/repository/prisma/dbsqlc/workflow_runs.sql
```diff
@@ -437,6 +437,26 @@ INSERT INTO "WorkflowRun" (
     @additionalMetadata::jsonb
 ) RETURNING *;
 
+-- name: CreateWorkflowRunDedupe :one
+WITH workflow_id AS (
+    SELECT w."id" FROM "Workflow" w
+    JOIN "WorkflowVersion" wv ON wv."workflowId" = w."id"
+    WHERE wv."id" = @workflowVersionId::uuid
+)
+INSERT INTO "WorkflowRunDedupe" (
```
grutt, line 446: do we want to track the workflow run id that was actually run?

grutt (approved):

abelanger5:
As a future note - been playing around with DB models with the following spec:

```
enum DedupeLifecycle {
  // the dedupe value is only used once
  ONCE

  // the dedupe value is used for the lifetime of the workflow run, and
  // removed when the workflow run enters a terminal state
  WORKFLOW_RUN

  // the dedupe value is used for a specific time period
  TIME_PERIOD
}

model WorkflowRunDedupe {
  id        BigInt   @unique @default(autoincrement()) @db.BigInt
  createdAt DateTime @default(now())
  updatedAt DateTime @default(now()) @updatedAt

  // the parent tenant
  tenant   Tenant @relation(fields: [tenantId], references: [id], onDelete: Cascade, onUpdate: Cascade)
  tenantId String @db.Uuid

  // the parent workflow
  workflowId String @db.Uuid

  // the workflow run id which used this dedupe value
  workflowRunId String @db.Uuid

  // the dedupe value
  value String

  // the lifecycle configuration for the dedupe value
  lifecycle DedupeLifecycle @default(ONCE)

  // if the dedupe lifecycle is TIME_PERIOD, when the dedupe value expires
  expiresAt DateTime?

  // DO NOT REMOVE - this uniqueness constraint is cased on in code
  @@unique([tenantId, workflowId, value])
  @@index([tenantId, value])
}
```

PR #772, opened 2024-08-08
feat: Add Proxied File Uploads

macwilk:
# Description

Draft of the work so far!

## Type of change

## What's Changed


pkg/repository/prisma/file.go
```diff
@@ -0,0 +1,88 @@
...
+		return nil, fmt.Errorf("could not create event: %w", err)
+	}
+
+	return e, nil
+}
+
+func (r *fileAPIRepository) ListFiles(ctx context.Context, tenantId string, ids []string) ([]*dbsqlc.File, error) {
+	pgIds := make([]pgtype.UUID, len(ids))
+
+	for i, id := range ids {
```
grutt, line 72: interesting choice, idea is we have a list of ids (i.e. in payload) to batch the request?

prisma/schema.prisma
```diff
@@ -1649,3 +1650,29 @@ model SNSIntegration {
...
+  id        String    @id @unique @default(uuid()) @db.Uuid
+  createdAt DateTime  @default(now())
+  updatedAt DateTime  @default(now()) @updatedAt
+  deletedAt DateTime?
+
+  // the name of the file
+  fileName String
+
+  // the relative path or URL pointing to the file
+  filePath String
```
grutt, line 1666: possible uniqueness constraint needed? tenant-file

api/v1/server/handlers/files/create.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	// marshal the data object to bytes
+	dataBytes, err := json.Marshal(request.Body.Data)
+	if err != nil {
+		return nil, err
+	}
+
+	context := ctx.Request().Context()
+	key := *request.Body.Filename
```
grutt, line 29: do we trust filenames?
should we tenant scope bucket paths?
how do we handle collisions?

pkg/repository/prisma/file.go
macwilk: yeah I figured we could either batch or not a file upload request. Always game for more flexibility where possible

prisma/schema.prisma
macwilk: yeah, though functionally since we're going to be the main consumer of this API, we should be using UUIDs as filenames so there should be functionally no collisions but I see no harm in double checking that here.

api/v1/server/handlers/files/create.go
macwilk: agreed. we shouldn't be trusting them and should be:
- tenant scoping bucket paths
- setting fileName as a random UUID
- date scoping files to a folder to a certain time frame, probably days.

PR #826, opened 2024-08-27
refactor: consolidate partition logic

abelanger5:
# Description

Consolidates partitioner logic to a single package. Also creates a new partition in the existing instance if the old partition was removed, which can happen if the engine is active without a database connection for a considerable period of time.

## Type of change


pkg/repository/prisma/tenant.go
```diff
@@ -301,14 +301,42 @@ func (r *tenantEngineRepository) GetTenantByID(ctx context.Context, tenantId str
...
 }
 
-func (r *tenantEngineRepository) UpdatePartitionHeartbeat(ctx context.Context, partitionId string) error {
+func (r *tenantEngineRepository) UpdateControllerPartitionHeartbeat(ctx context.Context, partitionId string) (string, error) {
 	_, err := r.queries.ControllerPartitionHeartbeat(ctx, r.pool, partitionId)
 
 	if err == pgx.ErrNoRows {
-		return nil
+		// create a new partition
+		newPartitionId := uuid.New().String()
```
grutt: why are we doing this here an not in the sqlc query?

grutt (approved):

PR #836, opened 2024-08-29
chore: more visibility and debug lines for queues

abelanger5:
# Description

Adds more visibility and debug lines into the queueing logic. 

## Type of change

- [X] Chore (changes which are not directly related to any business logic)


pkg/repository/prisma/tenant.go
```diff
@@ -395,3 +441,13 @@ func (r *tenantEngineRepository) RebalanceAllTenantWorkerPartitions(ctx context.
 func (r *tenantEngineRepository) RebalanceInactiveTenantWorkerPartitions(ctx context.Context) error {
 	return r.queries.RebalanceInactiveTenantWorkerPartitions(ctx, r.pool)
 }
+
+func getPartitionName() pgtype.Text {
+	hostname, ok := os.LookupEnv("HOSTNAME")
```
grutt, line 446: do we want to put this on the server config file?

grutt (approved):

pkg/repository/prisma/tenant.go
abelanger5: not at the moment, as this will only be used internally for multi-container setups where hostname is almost always set.

PR #842, opened 2024-09-04
feat: queue v4

abelanger5:
# Description

Implements a queue without updates on worker semaphore slots, instead maintaining a count parameter which gets computed through a separate `InternalQueue` that updates a row `WorkerSemaphoreCount` per worker. This allows semaphore counts to effectively be read-only on the queue which allows fewer updates in the core queueing tx. 

The introduction of the `InternalQueueItem` table is also designed for future use of batch operations, such as aggregating and pushing metrics data or bulk-updating step runs (to reduce the amount of txs on the StepRun table).


prisma/migrations/20240904120320_v0_44_0/migration.sql
```diff
@@ -0,0 +1,53 @@
...
+-- CreateTable
+CREATE TABLE "WorkerSemaphoreCount" (
+    "workerId" UUID NOT NULL,
+    "count" INTEGER NOT NULL,
+
+    CONSTRAINT "WorkerSemaphoreCount_pkey" PRIMARY KEY ("workerId")
+);
+
+-- CreateTable
+CREATE TABLE "WorkerAssignEvent" (
```
grutt, line 26: do we want to make this more generic for worker event?

grutt (approved):

PR #887, opened 2024-09-18
feat: bulk inserts of events

reillyse:
## Type of change Feature

## What's Changed

new endpoint for bulk inserts


api-contracts/openapi/openapi.yaml
```diff
@@ -94,6 +94,8 @@ paths:
     $ref: "./paths/tenant/tenant.yaml#/getQueueMetrics"
   /api/v1/tenants/{tenant}/events:
     $ref: "./paths/event/event.yaml#/withTenant"
+  /api/v1/tenants/{tenant}/events/bulk-create:
```
grutt: i think this should just be POST .../events/bulk

grutt: also do we have a POST /events for non-bulk insert? im pretty sure we only have a grpc for this... would be good to add a rest endpoint

api-contracts/openapi/paths/event/event.yaml
```diff
@@ -159,6 +159,59 @@ withTenant:
     tags:
       - Event
 
+
+bulkCreateEvents:
+  post:
+    x-resources: ["tenant"]
+    description: Bulk creates new events.
+    operationId: event:bulk:create
```
grutt: nit `noun:verb:noun` so `event:create:bulk`

internal/services/ingestor/ingestor.go
```diff
@@ -145,6 +146,42 @@ func (i *IngestorImpl) IngestEvent(ctx context.Context, tenantId, key string, da
...
+		Events: eventOpts,
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	return events, nil
+
+	// if err == metered.ErrResourceExhausted {
```
grutt: rm?

pkg/repository/prisma/event.go
```diff
@@ -283,6 +283,74 @@ func (r *eventEngineRepository) CreateEvent(ctx context.Context, opts *repositor
...
+
+	for _, e := range events {
+
+		for _, cb := range r.callbacks {
+			cb.Do(e) // nolint: errcheck
+		}
+	}
+
+	return events, nil
+	// })
```
grutt: rm?

internal/services/ingestor/ingestor.go
grutt: looks like we just dont have metering, etc enabled so i assume this is wip

grutt (changes requested):

internal/services/ingestor/ingestor.go
reillyse: yes, hadn't figured out the metering piece yet

api-contracts/openapi/openapi.yaml
reillyse: I thought we did 
  /api/v1/tenants/{tenant}/events:
    $ref: "./paths/event/event.yaml#/withTenant"
    
and in event.yaml in "WithTenant" a post looks like it creates
      post:
    x-resources: ["tenant"]
    description: Creates a new event.
    operationId: event:create
    parameters:
      - description: The tenant id
        in: path
        name: tenant
        required: true
        schema:
          type: string
          format: uuid
          minLength: 36
          maxLength: 36
    requestBody:
      content:
        application/json:

examples/bulk_imports/main.go
```diff
@@ -0,0 +1,119 @@
...
+				},
+				),
+			},
+		},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error registering workflow: %w", err)
+	}
+
+	var events []client.EventWithMetadata
```
reillyse, line 84: @grutt just wanted to get your input on this. I added a new type here to simplify things on the ingest - again I'm a little unsure of the general interface here how this SDK is used.

internal/services/ingestor/ingestor.go
```diff
@@ -145,6 +146,41 @@ func (i *IngestorImpl) IngestEvent(ctx context.Context, tenantId, key string, da
...
+
+	if err == metered.ErrResourceExhausted {
+		return nil, metered.ErrResourceExhausted
+	}
+
+	if err != nil {
+		return nil, fmt.Errorf("could not create events: %w", err)
+	}
+
+	// TODO any attributes we want to add here? could jam in all the event ids? but could be a lot
```
reillyse, line 166: @grutt Unsure about what attributes we want to add to telemetry if any ?

pkg/repository/metered/metered.go
```diff
@@ -32,7 +32,7 @@ func NewMetered(entitlements repository.EntitlementsRepository, l *zerolog.Logge
 
 var ErrResourceExhausted = fmt.Errorf("resource exhausted")
 
-func MakeMetered[T any](ctx context.Context, m *Metered, resource dbsqlc.LimitResource, tenantId string, f func() (*string, *T, error)) (*T, error) {
+func MakeMetered[T any](ctx context.Context, m *Metered, resource dbsqlc.LimitResource, tenantId string, numberOfResources int32, f func() (*string, *T, error)) (*T, error) {
```
reillyse, line 35: @grutt I needed to modify the metering because it was all based on 1 resource (from my reading of it) - modified it to now take a number of resources we want to meter. Tested it by hardcoding enforce limits -seems to work.

examples/bulk_imports/main.go
grutt: This looks good to me, the existing `Push` method takes args with optional `client.PushOpFunc`s  to build the object internally. Personally,  not a fan of this old pattern and prefer how you implemented.

```
		err := c.Event().Push(
			context.Background(),
			"user:create:simple",
			testEvent,
			client.WithEventMetadata(map[string]string{
				"hello": "world",
			}),
		)
```

frontend/app/src/lib/api/generated/data-contracts.ts
```diff
@@ -434,6 +434,18 @@ export interface CreateEventRequest {
...
+export interface BulkCreateEventRequest {
+  events: CreateEventRequest[];
+}
+
+export interface BulkCreateEventResponse {
+  metadata: APIResourceMeta;
+  /** The events. */
+  events: Event[];
+  /** The pagination information. */
+  pagination: PaginationResponse;
```
grutt: we're paginating the create response?

hack/dev/atlas-migrate.sh
```diff
@@ -1,14 +1,15 @@
 #!/bin/bash
 
+MIGRATION_NAME=$1
+
 # check if the first argument is empty
-if [ -z "$1" ]; then
-  echo "Usage: $0 <version>"
-  exit 1
+if [ -z "$MIGRATION_NAME" ]; then
+  MIGRATION_NAME="temp"
```
grutt: lol eventually this merge will make it into main 😅

pkg/client/event.go
```diff
@@ -14,14 +14,24 @@ import (
...
 	Push(ctx context.Context, eventKey string, payload interface{}, options ...PushOpFunc) error
 
+	BulkPush(ctx context.Context, payloads []EventWithMetadata, options ...BulkPushOpFunc) error
+
 	PutLog(ctx context.Context, stepRunId, msg string) error
 
 	PutStreamEvent(ctx context.Context, stepRunId string, message []byte) error
 }
 
+type EventWithMetadata struct {
```
grutt, line 29: not pretty, but `AdditionalMetadata` is named as such because of a conflict on all our API resources where `metadata` is reserved for `createdAt`, `updatedAt`, and `ID`s...

pkg/repository/metered/metered.go
grutt: ahh i thought i thought ahead here... i probably did think ahead but didnt implement ahead.

pkg/repository/prisma/dbsqlc/events.sql
```diff
@@ -82,6 +82,32 @@ INSERT INTO "Event" (
...
+    $4,
+    $5,
+    $6
+);
+
+
+-- name: GetInsertedEvents :many
+
+SELECT * FROM "Event"
+WHERE xmin::text = (txid_current() % (2^32)::bigint)::text
```
grutt, line 107: why are we using `xmin` here?

pkg/repository/prisma/dbsqlc/tenant_limits.sql
```diff
@@ -78,7 +78,7 @@ SET
         WHEN ("customValueMeter" = true OR ("window" IS NOT NULL AND "window" != '' AND NOW() - "lastRefill" >= "window"::INTERVAL)) THEN
             0 -- Refill to 0 since the window has passed
         ELSE
-            "value" + 1 -- Increment the current value within the window
+            "value" + @numResources -- Increment the current value within the window by the number of resources
```
grutt: typecast?
```suggestion
            "value" + @numResources::INT -- Increment the current value within the window by the number of resources
```

pkg/repository/prisma/dbsqlc/events.sql
reillyse: This is how we are retrieving the inserted events. Copyfrom doesn't return anything and so in order to find out what events where changed we are getting the transaction id and finding out what events were last modified using our current transaction id. Also why I needed to wrap in a txn. 

We need to retrieve the events so we can call the callbacks on them once inserted.

frontend/app/src/lib/api/generated/data-contracts.ts
reillyse: I wasn't sure about this. If I include that does it magically paginate the response ? For many events maybe it would make sense to paginate (like 1000s) but probably not required.

grutt: no, we still need the query to page the results but the bigger problem is the ids generated are ephemeral and we dont have a way to GET the next page. I'd just return a list of ids probably

pkg/config/loader/loader.go
```diff
@@ -418,6 +418,7 @@ func GetServerConfigFromConfigfile(dc *database.Config, cf *server.ServerConfigF
 		}
 		return nil
 	}
+	cf.Runtime.EnforceLimits = true
```
grutt: we dont want this set here

reillyse: should already be gone

frontend/app/src/lib/api/generated/data-contracts.ts
reillyse: Which ids are ephemeral?

grutt: the ids returned from the copy... in other words we cant return the next page unless we store it somewhere

grutt (approved):

PR #889, opened 2024-09-18
feat: expression-based concurrency keys

abelanger5:
# Description

Adds support for expression-based parsing of concurrency keys which are computed server-side. For example, `input.user_id` will evaluate the input data to the workflow and look at the `user_id` field passed in the input. 

## Type of change


examples/simple/main.go
```diff
@@ -70,7 +70,7 @@ func run(events chan<- string) (func() error, error) {
 			On:          worker.Events("user:create:simple"),
 			Name:        "simple",
 			Description: "This runs after an update to the user model.",
-			Concurrency: worker.Concurrency(getConcurrencyKey),
+			Concurrency: worker.Expression("input.user_id"),
```
grutt, line 73: i liked the `{{input.user_id}}` pattern because we can have composite keys

pkg/repository/prisma/workflow.go
```diff
@@ -593,24 +593,31 @@ func (r *workflowEngineRepository) createWorkflowVersionTxs(ctx context.Context,
 
 	// create concurrency group
 	if opts.Concurrency != nil {
+		params := dbsqlc.CreateWorkflowConcurrencyParams{
+			ID:                sqlchelpers.UUIDFromStr(uuid.New().String()),
```
grutt: generate in sql?

grutt (approved):

reillyse (commented):
it's a lot, can't say as I understand it all but looks good

grutt (approved):

PR #903, opened 2024-09-24
fix: trunc large payloads

grutt:
# Description

Prevents sending of large workflow run events over the stream listener

## Type of change

### DIFF
--- a/internal/services/dispatcher/server.go
+++ b/internal/services/dispatcher/server.go
@@ -11,7 +11,6 @@ import (
 	"strconv"
 	"sync"
 	"time"
-	"unsafe"
 
 	"github.com/google/uuid"
 	"github.com/jackc/pgx/v5"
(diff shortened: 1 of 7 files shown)


internal/services/dispatcher/server.go
```diff
@@ -754,6 +755,25 @@ func (s *sendTimeFilter) canSend() bool {
 	return true
 }
 
+func calculateResultsSize(results []*contracts.StepRunResult) int64 {
+	var totalSize int64
+
+	for _, result := range results {
+		// Size of the pointer
+		totalSize += int64(unsafe.Sizeof(result))
```
abelanger5: I'm not sure why we're using `unsafe` here -- this doesn't seem to correspond precisely to the size that we'd see when sent with grpc anyway, so why not just estimate using `len(*result.Output)` anyway?

abelanger5 (changes requested):

abelanger5 (approved):

reillyse (approved):
looks good

PR #912, opened 2024-09-26
feat: workflow run metrics view

abelanger5:
# Description

Adds a view for workflow run metrics from the workflow runs tab. Currently depends on a cloud endpoint implementation with a goal to support a more lightweight version of this query in the repo soon. If metrics aren't available on the API, it just hides the view. 

## Type of change


pkg/repository/prisma/job_run.go
```diff
@@ -35,8 +38,26 @@ func NewJobRunAPIRepository(client *db.PrismaClient, pool *pgxpool.Pool, v valid
 	}
 }
 
+func (j *jobRunAPIRepository) RegisterWorkflowRunRunningCallback(callback repository.Callback[pgtype.UUID]) {
```
reillyse, line 41: Is this Workflow Run Callback stuff related to the metrics ?

abelanger5: Yes, we register callbacks so that when we wrap the engine with our own custom code, we can get events into the metrics database.

PR #913, opened 2024-09-26
bulk insert buffering

reillyse:
# Description

A generic way to do buffered writes

## Type of change

## What's Changed

Buffering up single events into a bulk update


internal/services/ingestor/buffered.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	b.lastFlush = time.Now()
+	// TODO: if we have additional events in the buffer, that haven't been flushed, we will not be counting their size
+	b.sizeOfData = 0
+
+	fmt.Println("============================================  Done sending to done chans")
+
+}
+
+func (b *IngestBuf) Start(ctx context.Context) {
```
abelanger5: Most of our services/methods return a cleanup function with signature `func() error` that gets called on shutdown. This makes it much easier to gracefully shut down (rather than just relying on context + a timeout). In this case, the cleanup method should probably flush all existing messages?

internal/services/ingestor/buffered.go
```diff
@@ -0,0 +1,131 @@
...
+func (b *IngestBuf) buffEventWorker(ctx context.Context) {
+	for {
+		select {
+		case e := <-b.eventOpsChan:
+			b.internalArr = append(b.internalArr, e)
+			b.sizeOfData += len(e.eventOps.Data)
+			b.sizeOfData += len(e.eventOps.AdditionalMetadata)
+
+			if len(b.internalArr) >= b.maxCapacity {
+				b.flush()
```
abelanger5: It seems the `flush` is blocking and waits for writes to succeed before exiting. Does it make sense to make `flush` non-blocking?

In that case we will need to ensure that `internalArr` is rw concurrency-safe, or copy the contents of `internalArr` over to the flush method.

internal/services/ingestor/buffered.go
```diff
@@ -0,0 +1,131 @@
...
+			}
+			if b.sizeOfData >= b.maxDataSizeInQueue {
+				b.flush()
+			}
+
+		case <-time.After(time.Until(b.lastFlush.Add(b.flushPeriod))):
+			if len(b.internalArr) > 0 {
+				b.flush()
+			} else {
+				b.lastFlush = time.Now()
```
abelanger5: I see that we set `lastFlush` here, should we be setting `lastFlush` when we hit the `case e := <-b.eventOpsChan` case as well?

internal/services/ingestor/buffered.go
```diff
@@ -0,0 +1,131 @@
...
+	// bulk create events
+	writtenEvents, err := b.BulkCreateFunc(ctx, &repository.BulkCreateEventSharedTenantOpts{
+		Events: eventOpts,
+	})
+
+	fmt.Printf("============================================  BulkCreateEventSharedTenant flushed %+v events\n", len(writtenEvents.Events))
+
+	if err != nil {
+		for _, doneChan := range doneChans {
+			doneChan <- &flushResponse{event: nil, err: nil}
```
abelanger5:
```suggestion
			doneChan <- &flushResponse{event: nil, err: err}
```

shouldn't we be propagating the error?

internal/services/ingestor/buffered.go
```diff
@@ -0,0 +1,131 @@
...
+		internalArr:        make([]*eventBuffWrapper, 0),
+		sizeOfData:         0,
+		maxDataSizeInQueue: maxDataSizeInQueue,
+		BulkCreateFunc:     bulkCreateFunc,
+	}
+}
+
+type eventBuffWrapper struct {
+	eventOps *repository.CreateEventOpts
+	doneChan chan *flushResponse
```
abelanger5:
```suggestion
	doneChan chan<- *flushResponse
```

write-only channel?

internal/services/ingestor/buffered.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		for _, doneChan := range doneChans {
+			doneChan <- &flushResponse{event: nil, err: nil}
+		}
+		return
+	}
+
+	for i, d := range doneChans {
+		// FIXME: these events are not in the correct order - need to fix cause we could send the wrong event back to the wrong tenant
+		d <- &flushResponse{event: writtenEvents.Events[i], err: nil}
```
abelanger5: Should this be a non-blocking send? What happens when the subscriber times out or exits?

internal/services/ingestor/buffered.go
reillyse: good idea, was thinking that we wanted some back pressure to control writes to the DB, but in the case where the max buffer size is one (or low) we want to write straight away - and what we are replacing is a write straight away - so it should not be too much for the DB to handle.

internal/services/ingestor/buffered.go
reillyse: the case where we read from the channel should just buffer. We are setting lastFlush here because we are basically doing a no-op flush and otherwise we'd trigger again too soon in the after above as soon as we get a single item.

internal/services/ingestor/buffered.go
reillyse: good catch thanks

internal/services/ingestor/buffered.go
reillyse: the done channel is buffered so it shouldn't block but I think it makes sense to be explicit

internal/services/ingestor/buffered.go
reillyse: makes sense

pkg/repository/prisma/buffered.go
```diff
@@ -0,0 +1,276 @@
...
+		maxDataSizeInQueue: opts.maxDataSizeInQueue,
+		outputFunc:         opts.outputFunc,
+		sizeFunc:           opts.sizeFunc,
+		l:                  opts.l,
+		ctx:                ctx,
+		cancel:             cancel,
+	}
+}
+
+func (b *IngestBuf[T, U]) validate() error {
```
grutt: nit we generally validate with struct tags:

```
type IngestBufOpts[T any, U any] struct {
    MaxCapacity        int                                                `validate:"required,gt=0"`
    FlushPeriod        time.Duration                                      `validate:"required,gt=0"`
    MaxDataSizeInQueue int                                                `validate:"required,gt=0"`
    OutputFunc         func(ctx context.Context, items []T) ([]U, error)  `validate:"required"`
    SizeFunc           func(T) int                                        `validate:"required"`
    Logger             *zerolog.Logger                                    `validate:"required"`
}
```

pkg/repository/prisma/buffered.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	b.cancel()
+
+	return nil
+}
+
+func (b *IngestBuf[T, U]) Start() (func() error, error) {
+	b.l.Debug().Msg("Starting buffer")
+
+	b.lock.Lock()
```
grutt, line 276: nit: `defer b.lock.Unlock()` instead of manually releasing the locks to catch both cases

pkg/repository/prisma/buffered.go
```diff
@@ -0,0 +1,276 @@
...
+
+func (b *IngestBuf[T, U]) calcSizeOfData(items []T) int {
+	size := 0
+	for _, item := range items {
+		size += b.sizeFunc(item)
+	}
+	return size
+}
+
+func (b *IngestBuf[T, U]) flush(items []*inputWrapper[T, U]) {
```
grutt, line 189: maybe we could use some unit tests on flush to ensure we're handling all cases of setting state post-flush. 
this looks good to me for now, but concerned this will be fragile

pkg/repository/prisma/tenant_buffer.go
```diff
@@ -0,0 +1,121 @@
...
+	if opts.l == nil {
+		return nil, fmt.Errorf("logger is required")
+	}
+
+	megabyte := 1024 * 1024
+
+	defaultOpts := IngestBufOpts[T, U]{
+		maxCapacity:        1000,
+		flushPeriod:        50 * time.Millisecond,
+		maxDataSizeInQueue: 4 * megabyte,
```
grutt: what happens if we put a single event > 4mb (or reduce this to 3 and put a 4 mb)

prisma/schema.prisma
```diff
@@ -424,6 +424,9 @@ model Event {
   // metadata stored in the event
   additionalMetadata Json?
 
+  // this is used to order the events on bulk inserts
+  insertOrder Int?
```
grutt, line 428: note: not for this PR, but putting it into the universe. 
i think we may want to set something like an int unix time sdk side to expose more implicit order control for our customers so this is a step in the right direction

prisma/migrations/atlas.sum
```diff
@@ -0,0 +1 @@
+h1:47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=
```
grutt: confused by this, probably need to delete?

it doesn't match the hash in `sql/migrations/atlas.sum` and i'm pretty sure we don't usually compute this here

grutt (changes requested):

pkg/repository/prisma/buffered.go
reillyse: done

pkg/repository/prisma/buffered.go
reillyse: Added a bunch of Unit tests.

pkg/repository/prisma/buffered.go
reillyse: done

pkg/repository/prisma/tenant_buffer.go
reillyse: A single event greater than 4MB will flush straight away. If we add a 3mb first and then a 4mb we will be over the size and we will flush.

prisma/migrations/atlas.sum
reillyse: Sounds good, probably something janky with me running atlas myself to try and fix things

grutt (approved):
nice

internal/services/controllers/events/controller.go
```diff
@@ -194,6 +198,13 @@ func (ec *EventsControllerImpl) processEvent(ctx context.Context, tenantId, even
 
 		g.Go(func() error {
 
+			if additionalMetadata["hatchet__event_id"] == nil {
```
abelanger5: Do we restrict the user from setting/viewing this? If not, we definitely should.

pkg/repository/prisma/buffered.go
```diff
@@ -0,0 +1,290 @@
...
+	for {
+		select {
+		case <-b.ctx.Done():
+			return
+		case e := <-b.inputChan:
+			b.safeAppendInternalArray(e)
+			b.safeIncSizeOfData(b.calcSizeOfData([]T{e.item}))
+
+			if b.safeCheckSizeOfBuffer() >= b.maxCapacity {
+				go b.flush(b.sliceInternalArray())
```
abelanger5: I think it would be cleaner to create a goroutine only after all concurrent writes w/ the mutex have been performed. With the way this is written, I think we may call `go b.flush` more than once because it calls `b.safeSetLastFlush` and `b.safeDecSizeOfData`. So instead of calling `go b.flush`, I think we want to refactor the `flush` method to something like:

```go
func (b *IngestBuf[T, U]) flush(items []*inputWrapper[T, U]) {
        b.safeSetLastFlush(time.Now())

	numItems := len(items)
	
	var doneChans []chan<- *flushResponse[U]
	opts := make([]T, numItems)

	for i := 0; i < numItems; i++ {
		opts[i] = items[i].item
		doneChans = append(doneChans, items[i].doneChan)
	}

	b.safeDecSizeOfData(b.calcSizeOfData(opts))

        // SPAWN GOROUTINE HERE
        go func() {
	ctx := context.Background()
	result, err := b.outputFunc(ctx, opts)

	if err != nil {
		for _, doneChan := range doneChans {
			select {
			case doneChan <- &flushResponse[U]{err: err}:
			default:
				b.l.Error().Msgf("could not send error to done chan: %v", err)

			}
		}
		return
	}

	for i, d := range doneChans {
		select {
		case d <- &flushResponse[U]{result: result[i], err: nil}:
		default:
			b.l.Error().Msg("could not send done to done chan")
		}
	}

	b.l.Debug().Msgf("Flushed %d items", numItems)
        }()
}
```

And then just call `b.flush` so that we can set the correct size and last flush time synchronously (which should be extremely fast anyway)

pkg/repository/prisma/event.go
```diff
@@ -194,6 +194,24 @@ func (r *eventAPIRepository) ListEventKeys(tenantId string) ([]string, error) {
 	return keys, nil
 }
 
+func sizeOfEvent(item *repository.CreateEventOpts) int {
+	return len(item.Data) + len(item.AdditionalMetadata)
+}
+
+func (e *eventEngineRepository) StartBufferLoop() (func() error, error) {
```
abelanger5: If the buffer loop is absolutely necessary for the event repository to function as expected, then we shouldn't expose it as a function to any callers. We should start it on `NewEventEngineRepository` and return a cleanup method out of the EngineRepository, otherwise we're doing a lot of `StartBufferLoop` between packages when it doesn't seem necessary.

abelanger5 (changes requested):

pkg/repository/prisma/buffered.go
reillyse: sounds good, I'll simplify the logic in the calling for select too

pkg/repository/prisma/event.go
reillyse: makes sense, passing the cleanup and errors back up the chain is the annoying bit, I've modified the NewEventRepo calling to return the cleanup as part of it's cleanup and the error as an additional return value

reillyse: added a Cleanup method to EventEngineRepository to take care of cleanup

internal/services/controllers/events/controller.go
reillyse: @abelanger5 @grutt I added this here as an initial pass at the stream by additional metadata thing we talked about gabe. It sounds like we aren't going down that route? Should I just remove it ? Also I'm not clear on the problems with doing this, is it that we might be allowing users to change parent ids?

grutt: i could see a potential exfiltration issue if users can set hatchet__event_id and then we later fetch a different tenant's data for a stream event for example

i dont see an issue with letting users read this though unless im missing something... i think this would actually be kind of convenient

maybe we just set any keys prefixed with `hatchet__` to nil first?

abelanger5 (approved):

PR #918, opened 2024-09-27
refactor: remove foreign keys from unchanged/non-cascading parent tables

abelanger5:
# Description

Removes foreign key constraints from all tables which have an immutable parent key. There are two reasons for this:
- With extremely high tx volume, we start running into `multixact members limit exceeded` errors which starts to drop inserts
- On moderate insert/update volume, we see locking when we don't expect to. If this were the only issue, we could consider `FOR NO KEY UPDATE` -- but given the issue above, it makes sense to remove these FK constraints entirely. 

## Type of change


prisma/schema.prisma
```diff
@@ -205,39 +205,25 @@ model Tenant {
...
+  triggers          WorkflowTriggers[]
+  jobRunLookupDatas JobRunLookupData[]
+  members           TenantMember[]
+  workflowTags      WorkflowTag[]
+  actions           Action[]
+  services          Service[]
+  invites           TenantInviteLink[]
+  apiTokens         APIToken[]
+  vcsProviders      TenantVcsProvider[]
+  streamEvents      StreamEvent[]
```
grutt: missed? potentially high throughput

grutt (approved):

reillyse (commented):
Nice, excited to see how this changes things

PR #940, opened 2024-10-04
Feat bulk workflows

reillyse:
- [X ] New feature (non-breaking change which adds functionality)

- [ X] Refactor (non-breaking changes to code which doesn't change any behaviour)

## What's Changed

Added in a new path for BulkWorkflows

Changed the workflow creation for single workflows to use this path.


pkg/repository/prisma/workflow_run.go
```diff
@@ -725,354 +753,432 @@ func workflowRunMetricsCount(ctx context.Context, pool *pgxpool.Pool, queries *d
 	return workflowRunsCount, nil
 }
 
-func createNewWorkflowRun(ctx context.Context, pool *pgxpool.Pool, queries *dbsqlc.Queries, l *zerolog.Logger, tenantId string, opts *repository.CreateWorkflowRunOpts) (*dbsqlc.WorkflowRun, error) {
-	ctx, span := telemetry.NewSpan(ctx, "db-create-new-workflow-run")
+func createNewWorkflowRuns(ctx context.Context, pool *pgxpool.Pool, queries *dbsqlc.Queries, l *zerolog.Logger, tenantId string, inputOpts []*repository.CreateWorkflowRunOpts) ([]*dbsqlc.WorkflowRun, error) {
```
reillyse: @grutt @abelanger5 

I'd appreciate some eyes on this. I'm still unsure about a couple of parts of it. The ParentId stuff I'm still confused by so.
I've removed the dedupe stuff.

I can run it locally and I can create workflows and whatnot. I'd appreciate help creating test cases to verify things are working well.
I can get a bunch of deadlocks when I try to increase past 200 workflow runs but I'm guessing it could just be contention in my db because smaller sizes seem to work fine.

pkg/repository/prisma/workflow_run.go
```diff
@@ -725,354 +753,432 @@ func workflowRunMetricsCount(ctx context.Context, pool *pgxpool.Pool, queries *d
...
-					Tenantid:          pgTenantId,
-					Workflowversionid: sqlchelpers.UUIDFromStr(opts.WorkflowVersionId),
-					Value:             sqlchelpers.TextFromStr(*opts.DedupeValue),
-					Workflowrunid:     sqlchelpers.UUIDFromStr(workflowRunId),
-				},
+			createRunsParams = append(createRunsParams, crp)
+
+			var desiredWorkerId *string
+
+			if opt.DesiredWorkerId != nil {
```
abelanger5, line 1486: I thought we discussed this casing yesterday, but perhaps I missed something so would be good to get it down in writing -- this isn't the only case where we want to write sticky assignment, I believe to keep the behavior consistent with the current behavior we'll need to write it every single time?

pkg/repository/prisma/workflow_run.go
```diff
@@ -725,354 +753,432 @@ func workflowRunMetricsCount(ctx context.Context, pool *pgxpool.Pool, queries *d
...
+			if opt.Cron != nil {
+				cronSchedule = sqlchelpers.TextFromStr(*opt.Cron)
+			}
+
+			if opt.ScheduledWorkflowId != nil {
+				scheduledWorkflowId = sqlchelpers.UUIDFromStr(*opt.ScheduledWorkflowId)
+			}
+
+			// so I can't pass a NULL in for parentID so I need to set it to something
+			// setting it to workflowRunId for now but maybe thats not correct
```
abelanger5: Ah yeah, I think we got confused yesterday, and the parent id should just be the existing workflowRunId (too many fields named `parentId` :)). The point of this is to uniquely identify the relation of the WorkflowRunTriggeredBy, it doesn't refer to the parent workflow, which is confusing as hell. So just set it to the generated UUID of the workflow run!

pkg/repository/prisma/workflow_run.go
```diff
@@ -725,354 +753,432 @@ func workflowRunMetricsCount(ctx context.Context, pool *pgxpool.Pool, queries *d
...
-				)
+				stepActionIds = append(stepActionIds, step.ActionId)
+				newTenantIds = append(newTenantIds, pgTenantId)
+			}
 
-				if err != nil {
-					return nil, err
-				}
+			// think about doing this all in one query
+			err = queries.UpsertQueues(
```
abelanger5: I actually think this could be where we're seeing deadlocks (not sure though), let's move UpsertQueues to the QueueStepRun method in the step run engine repository, since we'd like to make the name of the queues dynamic anyway.

pkg/repository/prisma/workflow_run.go
```diff
@@ -725,354 +753,432 @@ func workflowRunMetricsCount(ctx context.Context, pool *pgxpool.Pool, queries *d
...
 			)
 
 			if err != nil {
+				l.Err(err).Msg("failed to link step run parents")
 				return nil, err
 			}
+
+		}
+
+		createdWorkflows, err := queries.GetInsertedWorkflowRuns(tx1Ctx, tx)
```
abelanger5: I think we already do this above?

```
workflowRuns, err := queries.GetInsertedWorkflowRuns(tx1Ctx, tx)
```

pkg/repository/prisma/workflow_run.go
reillyse: ah yes, I thought it was barfing cause we weren't providing a desiredWorkerId, I think the problem was a unique constraint on workflowRunId - should be fixed now

pkg/repository/prisma/workflow_run.go
reillyse: yes we do, good catch

pkg/repository/prisma/workflow_run.go
reillyse: I'm not really sure what is involved in moving this to QueueStepRun

reillyse: Does Queue Step Run get called further down during the execution?  We lose the concurrency if we do that right ? It's possible we could just do it after this txn commits if that makes things easier? It seems pretty decoupled from the other data?

internal/services/admin/server.go
```diff
@@ -203,6 +206,207 @@ func (a *AdminServiceImpl) TriggerWorkflow(ctx context.Context, req *contracts.T
...
+		}
+
+		if req.ChildIndex == nil {
+			return nil, status.Error(
+				codes.InvalidArgument,
+				"child index is required when parent id is provided",
+			)
+		}
+
+		// Unsure about this piece
```
reillyse: @abelanger5 @grutt 

Here we seem to be checking to see if we have already created a child workflow run and returning that if we have? 

How should this work for bulk workflow run triggers? 

Should I just return new workflow runs
Should I record this workflow_run_id and return it with the others?

abelanger5: Yes, I'd record and return it with the others, that seems sensible to me. We'll need to find a way to associate it properly on the client-side, but we'll deal with that when we get there.

reillyse: sounds good

pkg/repository/prisma/workflow_run.go
abelanger5: This is quite decoupled, so yeah we could do this in a separate tx, but I'd still prefer to move it to QueueStepRun for now so that we can make the queue names dynamic.

internal/services/admin/server.go
```diff
@@ -646,3 +705,224 @@ func toWorkflowVersion(workflowVersion *dbsqlc.GetWorkflowVersionForEngineRow) *
...
+			}
+
+			if req.ChildIndex == nil {
+				return nil, nil, status.Error(
+					codes.InvalidArgument,
+					"child index is required when parent id is provided",
+				)
+			}
+
+			workflowRun, err := a.repo.WorkflowRun().GetChildWorkflowRun(
```
abelanger5: This is an `n+1` style query. This should probably be a bulk list operation as well with `(parentId, parentStepRunId, index, key)` tuples to match the where condition.

internal/services/admin/server.go
```diff
@@ -646,3 +705,224 @@ func toWorkflowVersion(workflowVersion *dbsqlc.GetWorkflowVersionForEngineRow) *
...
+		if req.AdditionalMetadata != nil {
+			err := json.Unmarshal([]byte(*req.AdditionalMetadata), &additionalMetadata)
+			if err != nil {
+				return nil, nil, fmt.Errorf("could not unmarshal additional metadata: %w", err)
+			}
+		}
+		isParentTriggered := req.ParentId != nil
+
+		if isParentTriggered {
+			parent, err := a.repo.WorkflowRun().GetWorkflowRunById(createContext, tenantId, sqlchelpers.UUIDToStr(sqlchelpers.UUIDFromStr(*req.ParentId)))
```
abelanger5: This is another `n + 1` style query that we should avoid

pkg/client/admin.go
```diff
@@ -27,13 +27,21 @@ type ChildWorkflowOpts struct {
 	AdditionalMetadata *map[string]string
 }
 
+type Workflow struct {
```
abelanger5:
```suggestion
type WorkflowRun struct {
```

To keep naming consistent, instances of workflows should be `WorkflowRun` or we make it clear that it's a run when triggering via `RunWorkflow`.

internal/services/controllers/workflows/queue.go
```diff
@@ -23,6 +23,10 @@ import (
 )
 
 func (wc *WorkflowsControllerImpl) handleWorkflowRunQueued(ctx context.Context, task *msgqueue.Message) error {
+
+	// TODO remove de dupes and fail if they are clashing
```
abelanger5, line 27: Just making sure we address this TODO before merge

pkg/repository/prisma/dbsqlc/workflow_runs.sql
```diff
@@ -519,6 +519,43 @@ INSERT INTO "WorkflowRun" (
...
+    $10,
+    $11
+
+);
+
+
+-- name: GetInsertedWorkflowRuns :many
+
+SELECT * FROM "WorkflowRun"
+WHERE xmin::text = (txid_current() % (2^32)::bigint)::text
```
abelanger5, line 556: Let's not use the `xmin` hack unless we have a good way of avoiding seq scans on the whole table?

pkg/repository/prisma/workflow.go
```diff
@@ -499,6 +500,35 @@ func (r *workflowEngineRepository) CreateSchedules(
 	return r.queries.CreateSchedules(ctx, r.pool, createParams)
 }
 
+func (r *workflowEngineRepository) GetLatestWorkflowVersions(ctx context.Context, tenantId string, workflowIds []string) ([]*dbsqlc.GetLatestWorkflowVersionForWorkflowsRow, error) {
+
+	fmt.Println("the size of the workflowids are ", len(workflowIds))
+	var workflowVersionIds = make([]pgtype.UUID, len(workflowIds))
+
+	fmt.Println("the workflowids are ", strings.Join(workflowIds, ","))
```
abelanger5: Rm print statements before merge

internal/services/admin/server.go
```diff
@@ -204,6 +207,62 @@ func (a *AdminServiceImpl) TriggerWorkflow(ctx context.Context, req *contracts.T
 	}, nil
 }
 
+func (a *AdminServiceImpl) BulkTriggerWorkflow(ctx context.Context, req *contracts.BulkTriggerWorkflowRequest) (*contracts.BulkTriggerWorkflowResponse, error) {
```
abelanger5, line 89: I think we should also validate that the length of opts does not exceed 1000.

pkg/repository/prisma/dbsqlc/workflow_runs.sql
reillyse: sorry was rushing to get in in on Friday and overlooked this, we happy with the following? We have an index on createdAt so we should just be using the xmin to filter a small result set. 

-- name: GetInsertedWorkflowRuns :many

SELECT * FROM "WorkflowRun"
WHERE xmin::text = (txid_current() % (2^32)::bigint)::text
AND "createdAt" >= @ createdAt::timestamp
ORDER BY id;

internal/services/admin/server.go
reillyse: ok, wasn't clear to me how often we were hitting these paths

reillyse: also this is quite a bit more complicated - if we don't find the child workflow we have to trigger the workflow, I could key a hash by the params we use - but it's starting to get even more messy so I'm not sure if the juice is worth the squeeze.

reillyse: ok added a fix for this

internal/services/admin/server.go
```diff
@@ -204,6 +207,69 @@ func (a *AdminServiceImpl) TriggerWorkflow(ctx context.Context, req *contracts.T
...
+	}
+
+	opts, existingWorkflows, err := getOpts(ctx, req.Workflows, a)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(opts) == 0 {
+		if len(existingWorkflows) == 0 {
+			return &contracts.BulkTriggerWorkflowResponse{}, status.Error(codes.InvalidArgument, "no suitable new workflows provided")
```
grutt:
```suggestion
			return nil, status.Error(codes.InvalidArgument, "no suitable new workflows provided")
```

nit, why are we returning this?

internal/services/admin/server.go
```diff
@@ -204,6 +207,69 @@ func (a *AdminServiceImpl) TriggerWorkflow(ctx context.Context, req *contracts.T
...
+
+	for _, workflowRunId := range workflowRunIds {
+		err = a.mq.AddMessage(
+			context.Background(),
+			msgqueue.WORKFLOW_PROCESSING_QUEUE,
+			tasktypes.WorkflowRunQueuedToTask(tenantId, workflowRunId),
+		)
+
+		if err != nil {
+			return nil, fmt.Errorf("could not queue workflow run: %w", err)
```
grutt, line 141: open question: do we want to fail all if one fails?

internal/services/admin/server.go
```diff
@@ -646,3 +712,283 @@ func toWorkflowVersion(workflowVersion *dbsqlc.GetWorkflowVersionForEngineRow) *
 
 	return version
 }
+
+func getOpts(ctx context.Context, requests []*contracts.TriggerWorkflowRequest, a *AdminServiceImpl) ([]*repository.CreateWorkflowRunOpts, []string, error) {
```
grutt, line 600: note to self, reread this one more time

internal/services/admin/server.go
```diff
@@ -646,3 +712,283 @@ func toWorkflowVersion(workflowVersion *dbsqlc.GetWorkflowVersionForEngineRow) *
...
+		workflowNames[i] = req.Name
+	}
+
+	workflows, err := a.repo.Workflow().GetWorkflowsByNames(
+		ctx,
+		tenantId,
+		workflowNames,
+	)
+
+	if err == metered.ErrResourceExhausted {
```
grutt: i don't think this is metered?

internal/services/admin/server.go
```diff
@@ -646,3 +712,283 @@ func toWorkflowVersion(workflowVersion *dbsqlc.GetWorkflowVersionForEngineRow) *
...
+	childWorkflowMap := make(map[repository.ChildWorkflowRun]*dbsqlc.WorkflowRun)
+
+	for _, req := range requests {
+		isParentTriggered := req.ParentId != nil
+
+		if isParentTriggered {
+			if req.ParentStepRunId == nil {
+				return nil, nil, status.Error(
+					codes.InvalidArgument,
+					"parent step run id is required when parent id is provided",
```
grutt: general nit: again we fail all if one fails. realizing we also dont really indicate which thing is failing. probably some room for improvement here

pkg/repository/prisma/workflow_run.go
```diff
@@ -512,8 +554,55 @@ func (w *workflowRunEngineRepository) GetChildWorkflowRun(ctx context.Context, p
 	return w.queries.GetChildWorkflowRun(ctx, w.pool, params)
 }
 
+func (w *workflowRunEngineRepository) GetChildWorkflowRuns(ctx context.Context, childWorkflowRuns []repository.ChildWorkflowRun) ([]*dbsqlc.WorkflowRun, error) {
+
+	var childParams []dbsqlc.GetChildWorkflowRunParams
+	for _, childWorkflowRun := range childWorkflowRuns {
+
+		if childWorkflowRun.ChildIndex <= math.MinInt32 || childWorkflowRun.ChildIndex >= math.MaxInt32 {
```
grutt, line 695: do we want min int or 0 here?

pkg/repository/prisma/workflow_run.go
```diff
@@ -523,10 +612,10 @@ func (w *workflowRunEngineRepository) GetScheduledChildWorkflowRun(ctx context.C
 	}
 
 	if childkey != nil {
```
grutt, line 768: open question, do we want to simplify and merge child key + index? i.e. hash index => key

pkg/client/admin.go
```diff
@@ -27,13 +27,21 @@ type ChildWorkflowOpts struct {
...
+	Name  string
+	Input interface{}
+	// opts  *RunOptFunc
```
grutt: rm unused opts?

pkg/client/event.go
```diff
@@ -19,14 +19,14 @@ type BulkPushOpFunc func(*eventcontracts.BulkPushEventRequest) error
 type EventClient interface {
 	Push(ctx context.Context, eventKey string, payload interface{}, options ...PushOpFunc) error
 
-	BulkPush(ctx context.Context, payloads []EventWithMetadata, options ...BulkPushOpFunc) error
+	BulkPush(ctx context.Context, payloads []EventWithAdditionalMetadata, options ...BulkPushOpFunc) error
```
grutt, line 22: why did this make its way into this PR?

pkg/repository/prisma/dbsqlc/workflow_runs.sql
```diff
@@ -738,6 +945,43 @@ SELECT
...
+    $6,
+    $7,
+    $8
+
+);
+-- name: GetInsertedStepRuns :many
+
+SELECT id FROM "StepRun"
+WHERE xmin::text = (txid_current() % (2^32)::bigint)::text
+AND "createdAt" >= @createdAt::timestamp
```
grutt: note: confirmed index on createdAt

pkg/repository/prisma/dbsqlc/workflow_runs.sql
```diff
@@ -738,6 +945,43 @@ SELECT
...
+ORDER BY id;
+
+-- name: GetStepsForWorkflowVersion :many
+
+SELECT
+    "Step".*  from "Step"
+JOIN "Job" ON "Step"."jobId" = "Job"."id"
```
grutt: why are we joining job it looks like its not needed for the select or where?

grutt: ah workflowVersionId is on Job? 

lets disambiguate with "Job". (or shorthand "Job" j, j."workflowVersionId")

pkg/repository/workflow_run.go
```diff
@@ -49,7 +49,8 @@ type CreateWorkflowRunOpts struct {
 	ChildKey *string
 
 	// (optional) the child index of the workflow run, if this is a child run of a different workflow
-	ChildIndex *int
+	// ChildIndex *int `validate:"omitempty,min=0"`
+	ChildIndex *int // I feel like I should be able to tighten this up but it seems a -1 for child index is coming from somewhere
```
grutt: `self.spawn_index = -1`

tracked this down to the pythonsdk.. defaults to -1 for some reason

grutt: `hatchet_sdk/context/context.py`

pkg/repository/prisma/workflow_run.go
```diff
@@ -1043,323 +1187,426 @@ func workflowRunMetricsCount(ctx context.Context, pool *pgxpool.Pool, queries *d
...
+			}
+			if opt.Cron != nil {
+				cronSchedule = sqlchelpers.TextFromStr(*opt.Cron)
+			}
+
+			if opt.ScheduledWorkflowId != nil {
+				scheduledWorkflowId = sqlchelpers.UUIDFromStr(*opt.ScheduledWorkflowId)
+			}
+
+			// so we are ignoring opt.ParentId
```
grutt: update comment... different parent id

pkg/repository/prisma/workflow_run.go
```diff
@@ -1043,323 +1187,426 @@ func workflowRunMetricsCount(ctx context.Context, pool *pgxpool.Pool, queries *d
...
+				ParentId:     sqlchelpers.UUIDFromStr(workflowRunId),
+				EventId:      eventId,
+				CronParentId: cronParentId,
+				ScheduledId:  scheduledWorkflowId,
+				CronSchedule: cronSchedule,
 			}
+
+			triggeredByParams = append(triggeredByParams, cp)
+
+			// }
```
grutt: rm
```suggestion
			
```

pkg/repository/prisma/workflow_run.go
grutt: i want to reread one more time, but this looks reasonable

grutt (changes requested):

pkg/client/event.go
reillyse: I think it's just something weird with the merge, I'm pretty sure this is all in main already

internal/services/admin/server.go
reillyse: Yea I don't think so either, I just copied it from the single trigger function. I'll remove it.

internal/services/admin/server.go
reillyse: I feel it is probably the cleanest failure state, otherwise we need to support partial errors and a partially failed transaction, which is going to increase the complexity a lot.

internal/services/admin/server.go
reillyse: definitely think it makes sense to include the first failed request in the error

pkg/repository/prisma/workflow_run.go
reillyse: @grutt  I think we want -1 from the convo about the python SDK?

pkg/repository/prisma/workflow_run.go
reillyse: @abelanger5 I definitely don't understand enough about what is happening here to make the change right now. tbh I'm not 100% sure how the queues work, maybe we can have a quick chat about it?

reillyse: ok moved it to *stepRunEngineRepository) QueueStepRun

internal/services/controllers/workflows/queue.go
reillyse: added some dedupe logic here

pkg/repository/prisma/workflow_run.go
```diff
@@ -1553,20 +1571,20 @@ func createNewWorkflowRuns(ctx context.Context, pool *pgxpool.Pool, queries *dbs
...
-				},
-			)
-
-			if err != nil {
-				l.Error().Msgf("trying to upsert queues with names %+v and tenantIds %+v ", stepActionIds, newTenantIds)
-				l.Error().Err(err).Msg("failed to upsert queues")
-				return nil, err
-			}
+			// err = queries.UpsertQueues(
+			// 	tx1Ctx,
```
grutt: intentionally removed?

grutt (changes requested):

pkg/repository/prisma/buffered.go
```diff
@@ -284,9 +289,24 @@ func (b *IngestBuf[T, U]) Start() (func() error, error) {
 	b.state = started
 
 	go b.buffWorker()
+	// go b.startDebugLoop()
```
grutt: rm comment

sql/migrations/atlas.sum
```diff
@@ -65,3 +65,4 @@ h1:4CduYOxaYNUq+L7PFTvWFiMncxTz5b43B6rgB/ryJUA=
 20240930233257_v0.49.0.sql h1:B+JMbME62DxaCnesydvQXPg+ZNB0kB/V8gSclh1VdY4=
 20241004122206_v0.49.1.sql h1:Fas5TXOp4a2g+y5sGBJG9wTaVL/WCaVJ9+ZlASN9Md4=
 20241008124038_v0.49.2.sql h1:YT40sN8Wtqh21emrzDMZIcvcOkipw+4fdwIoBpF+Dek=
+20241009051224_temp.sql h1:N5BkdZ2TK6hpORy53v6XYRrwHn2TNJw7lUld1yKbpho=
```
grutt: migration name

grutt (changes requested):
looks good, small cleanup changes

pkg/repository/prisma/workflow_run.go
reillyse: Yes, moved to the step run

grutt (approved):

grutt (approved):

PR #941, opened 2024-10-04
refactor: buffered writes of step run statuses

abelanger5:
# Description

Refactors step run updates so that they're written via the new buffering method. There are lots of moving parts here: 
- Step run updates are now written via a buffer
- Step run events are now written via a buffer
- Job run and workflow run status resolutions are still processed via a serialized `InternalQueueItem` queue. The reason for this is to avoid race conditions on job run or workflow run statuses. These operations are also very fast (we have benchmarked at ~5-10k/second), with the previous slowdown being on the step run status updates. 

The new internal queue is called `STEP_RUN_UPDATE_V2` and is only responsible for handling job run and workflow run updates. 

## Type of change


internal/services/controllers/jobs/controller.go
```diff
@@ -976,7 +976,7 @@ func (ec *JobsControllerImpl) handleStepRunFailed(ctx context.Context, task *msg
 }
 
 func (ec *JobsControllerImpl) failStepRun(ctx context.Context, tenantId, stepRunId, errorReason string, failedAt time.Time) error {
-	stepRun, err := ec.repo.StepRun().GetStepRunForEngine(ctx, tenantId, stepRunId)
+	oldStepRun, err := ec.repo.StepRun().GetStepRunForEngine(ctx, tenantId, stepRunId)
```
abelanger5, line 979: We were doing an unnecessary read later on to get the step run's worker before the update, so I removed that

internal/services/controllers/jobs/queue.go
```diff
@@ -356,6 +375,64 @@ func (q *queue) processStepRunUpdates(ctx context.Context, tenantId string) (boo
...
+			q.updateStepRunV2Operations.RunOrContinue(tenantId)
+		}
+	}
+}
+
+func (q *queue) processStepRunUpdatesV2(ctx context.Context, tenantId string) (bool, error) {
+	ctx, span := telemetry.NewSpan(ctx, "process-step-run-updates-v2")
+	defer span.End()
+
+	dbCtx, cancel := context.WithTimeout(ctx, 300*time.Second)
```
grutt, line 404: intentional 300 sec?

grutt (approved):

internal/services/controllers/jobs/queue.go
abelanger5: I think I left it the same as ProcessStepRunUpdates, I think I just left this intentionally high before.

reillyse (commented):
looks good!

reillyse (approved):

PR #943, opened 2024-10-04
fix: don't use xmin hack

abelanger5:
# Description

Seeing seq scans across the entire events table when using the `xmin` hack.

## Type of change

### DIFF
--- a/pkg/repository/prisma/dbsqlc/events.sql
+++ b/pkg/repository/prisma/dbsqlc/events.sql
@@ -104,11 +104,9 @@ INSERT INTO "Event" (
 
 
 -- name: GetInsertedEvents :many
-
 SELECT * FROM "Event"
-WHERE xmin::text = (txid_current() % (2^32)::bigint)::text
+WHERE "id" = ANY(@ids::uuid[])
 ORDER BY "insertOrder" ASC;
-
 
 -- name: ListEvents :many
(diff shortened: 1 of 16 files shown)


pkg/repository/prisma/dbsqlc/events.sql
```diff
@@ -104,12 +104,10 @@ INSERT INTO "Event" (
 
 
 -- name: GetInsertedEvents :many
-
 SELECT * FROM "Event"
-WHERE xmin::text = (txid_current() % (2^32)::bigint)::text
+WHERE "id" = ANY(@ids::uuid[])
```
grutt, line 108: possible this can get very large?

pkg/repository/prisma/event.go
```diff
@@ -333,11 +333,13 @@ func (r *eventEngineRepository) BulkCreateEvent(ctx context.Context, opts *repos
 			return nil, nil, err
 		}
 		params := make([]dbsqlc.CreateEventsParams, len(opts.Events))
+		ids := make([]pgtype.UUID, len(opts.Events))
 
 		for i, event := range opts.Events {
+			eventId := uuid.New().String()
```
grutt, line 339: nit sqlchelpers.UUIDFromStr here so we dont call it twice

grutt (approved):

PR #951, opened 2024-10-08
fix: event keys

grutt:
# Description

Inserts unique event keys into new table and fixes list query to not use distinct

## Type of change


reillyse (approved):
looks good

abelanger5 (approved):

PR #956, opened 2024-10-10
feat(throughput): single process per queue

abelanger5:
# Description

Refactors the queueing logic so that each queue is given its own process to run in with the goal of increasing throughput. The tricky part about this (and why we didn't do it earlier) is that slots on workers and rate limits can be shared between many possible queues, so there's a large amount of contention for slots if there are very many busy queues assigning to the same slots. We also risk losing transactional integrity by separating the process which handles queueing from the process which handles worker assignment. 

There are different orders of transactional integrity that we need to satisfy:
- Queue items and semaphore slots on workers require the highest order of correctness and resilience to engine restarts. To solve for this, each slot is pessimistically reserved and only released on the engine when we get an explicit ack after committing the transaction. 
- We require slightly less integrity on rate limits and use a deferred flush model to periodically update rate limits in the database. We still pessimistically reserve rate limits and wait for an ack on the transaction, but we only commit periodically on a 1 second interval. If we don't do this, scenarios where we have very many rate limits all under contention from multiple queues would slow down queueing significantly. 

Both workers and queues use the `Lease` model to reserve certain workers and queues for a period of time. Right now, we attempt to lease both workers and queues based on the tenants granted to the `ControllerPartition`, but a future optimization could reserve groups of workers which assign to distinct sets of queues to allow for horizontal scalibility. In other words, the only horizontal scaling for queueing is on the tenant at the moment.   

## Type of change


internal/services/controllers/jobs/controller.go
```diff
@@ -939,7 +953,7 @@ func (ec *JobsControllerImpl) handleStepRunFinished(ctx context.Context, task *m
 	}
 
 	// recheck the tenant queue
-	ec.checkTenantQueue(ctx, metadata.TenantId)
+	ec.checkTenantQueue(ctx, metadata.TenantId, "", false, true)
```
grutt, line 956: empty string?

internal/services/controllers/jobs/queue.go
```diff
@@ -290,6 +327,9 @@ func (q *queue) runTenantUpdateStepRuns(ctx context.Context) func() {
 			return
 		}
 
+		// TODO: MOVE ELSEWHERE!
```
grutt: todo

grutt: i think you already moved this?

pkg/repository/prisma/dbsqlc/lease.sql
```diff
@@ -0,0 +1,61 @@
...
+FROM (
+    SELECT
+        unnest(@resourceIds::text[]) AS "resourceId"
+    ) AS input
+-- On conflict, acquire the lease if the existing lease has expired.
+ON CONFLICT ("tenantId", "kind", "resourceId") DO UPDATE
+SET
+    "expiresAt" = EXCLUDED."expiresAt"
+WHERE
+    "Lease"."expiresAt" < now() OR
```
grutt, line 35: Intended OR?
`Returns the renewed leases.` will always update on conflict for existingLeaseIds even if not expired?

pkg/scheduling/v2/rate_limit.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	return rls
+}
+
+func (r *rateLimiter) subtractUnacked(candidateRls map[string]int32, currRls rateLimitSet) {
+	r.unackedMu.RLock()
+	defer r.unackedMu.RUnlock()
+
+	for _, set := range r.unacked {
```
grutt, line 263: what is happening here...?

pkg/scheduling/v2/rate_limit.go
grutt, line 1: note to self: re read one more time

pkg/scheduling/v2/scheduler.go
```diff
@@ -0,0 +1,602 @@
...
+		repo:         newSchedulerDbQueries(cf.queries, cf.pool, tenantId),
+		tenantId:     tenantId,
+		l:            cf.l,
+		actions:      make(map[string]*action),
+		unackedSlots: make(map[int]*slot),
+		rl:           rl,
+	}
+}
+
+// TODO: ACK IN BULK!
```
grutt: do now or future?

pkg/scheduling/v2/scheduler.go
```diff
@@ -0,0 +1,602 @@
...
+
+		storedAction := s.actions[actionId]
+
+		// determine if we match the conditions above
+		var replenish bool
+		activeCount := storedAction.activeCount()
+
+		if activeCount == 0 {
+			replenish = true
+		} else if activeCount <= (storedAction.lastReplenishedSlotCount / 2) {
```
grutt, line 197: make configurable?

pkg/scheduling/v2/scheduler.go
```diff
@@ -0,0 +1,602 @@
...
+	})
+
+	if err != nil {
+		return err
+	}
+
+	// FUNCTION 3: list unacked slots (so they're not counted towards the worker slot count)
+	workersToUnackedSlots := make(map[string][]*slot)
+
+	// TODO: CHECK CORRECTNESS OF UNACKED SLOTS, PERHAPS WE SHOULD LOCK THIS FOR THE ENTIRE REPLENISH?
```
grutt: it looks like it is locked for the entire replenish with the defer?

pkg/scheduling/v2/scheduler.go
```diff
@@ -0,0 +1,602 @@
...
+		if _, ok := workersToUnackedSlots[workerId]; !ok {
+			workersToUnackedSlots[workerId] = make([]*slot, 0)
+		}
+
+		workersToUnackedSlots[workerId] = append(workersToUnackedSlots[workerId], s)
+	}
+
+	// FUNCTION 4: write the new slots to the scheduler and clean up expired slots
+	now := time.Now()
+	expires := now.Add(5 * time.Second)
```
grutt: what is this arb expires?

pkg/scheduling/v2/scheduler.go
```diff
@@ -0,0 +1,602 @@
...
+		if _, ok := s.actions[actionId]; !ok {
+			s.actions[actionId] = &action{
+				slots:                      newSlots,
+				lastReplenishedSlotCount:   len(newSlots),
+				lastReplenishedWorkerCount: len(actionsToWorkerIds[actionId]),
+			}
+		} else {
+			// we overwrite the slots for the action
+
+			// TODO: order the slots by worker id to fairly distribute the slots
```
grutt: todo

pkg/scheduling/v2/scheduler.go
```diff
@@ -0,0 +1,602 @@
...
+
+	// third pass: remove any actions which have no slots
+	for actionId, storedAction := range s.actions {
+		if len(storedAction.slots) == 0 {
+			delete(s.actions, actionId)
+		}
+	}
+
+	// DEBUG PRINT
+	// for actionId, storedAction := range s.actions {
```
grutt: debug logg config?

pkg/scheduling/v2/scheduler.go
```diff
@@ -0,0 +1,602 @@
...
+				defer wg.Done()
+				assigned := make([]*AssignedQueueItem, 0, len(qis))
+				errored := make([]*erroredQueueItem, 0, len(qis))
+				unassigned := make([]*dbsqlc.QueueItem, 0, len(qis))
+				schedulingTimedOut := make([]*dbsqlc.QueueItem, 0, len(qis))
+				rateLimited := make([]*rateLimitResult, 0, len(qis))
+
+				startAssignment := time.Now()
+
+				// TODO: ADD A NOTE ABOUT SKIPPING SINCE THERE MIGHT BE CONTENTION ON THE CANDIDATE SLOTS
```
grutt: would be a nice note

internal/services/controllers/jobs/controller.go
abelanger5: yes, only matters for queueing, not for releasing a slot

pkg/repository/prisma/dbsqlc/lease.sql
abelanger5: Yes, I actually merged the renewal and acquire logic into a single query, so it's expected that it renews the lease as well. Deleting the old RenewLeases queries since they're unused.

pkg/scheduling/v2/scheduler.go
abelanger5: realized that the skip may be incorrect in certain cases, so turned this into a `ringOffset` which is basically a hint for where to start the iteration.

pkg/scheduling/v2/scheduler.go
abelanger5: removed the expires logic as I don't think it's doing anything

grutt (approved):

PR #962, opened 2024-10-14
fix: concurrency

grutt:
# Description

Fixes concurrency issues with cancel in progress queue strategy.

## Type of change

### DIFF
--- a/internal/services/controllers/workflows/queue.go
+++ b/internal/services/controllers/workflows/queue.go
@@ -551,7 +551,11 @@ func (wc *WorkflowsControllerImpl) queueByCancelInProgress(ctx context.Context, 
 	wc.l.Info().Msgf("handling queue with strategy CANCEL_IN_PROGRESS for %s", groupKey)
 
 	mutex := wc.getLock(fmt.Sprintf("%s:%s", tenantId, groupKey))
-	mutex.Lock()
+
+	if ok := mutex.TryLock(); !ok {
+		return nil
+	}
+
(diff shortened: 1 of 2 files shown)


internal/services/controllers/workflows/queue.go
```diff
@@ -538,90 +539,93 @@ func (ec *WorkflowsControllerImpl) runGetGroupKeyRunReassignTenant(ctx context.C
...
 func (wc *WorkflowsControllerImpl) queueByCancelInProgress(ctx context.Context, tenantId, groupKey string, workflowVersion *dbsqlc.GetWorkflowVersionForEngineRow) error {
 	ctx, span := telemetry.NewSpan(ctx, "queue-by-cancel-in-progress")
 	defer span.End()
 
 	wc.l.Info().Msgf("handling queue with strategy CANCEL_IN_PROGRESS for %s", groupKey)
 
-	// list all workflow runs that are running for this group key
-	running := db.WorkflowRunStatusRunning
+	mutex := wc.getLock(fmt.Sprintf("%s:%s", tenantId, groupKey))
+	mutex.Lock()
```
abelanger5:
```suggestion
	if ok := mutex.TryLock(); !ok {
	  return nil 
        }
```

Perhaps we should just exit if we can't get a lock, otherwise we'll be blocked on this method for a while if we get a lot of workflow runs in.

abelanger5 (changes requested):

abelanger5 (approved):

PR #968, opened 2024-10-16
fix: ack rate limits properly

abelanger5:
# Description

Fixes an issue with ack'ing rate limits so they can be flushed to the database. 

## Type of change


reillyse (approved):
looks good

PR #981, opened 2024-10-23
fix: more efficient step run events, reduce caching on queue

abelanger5:
# Description

Adds a query to make writing step run events more efficient by using `DISTINCT` on the stepRunId instead of a subquery for each row. Additionally reduces the caching on the `Queue` table as we want to frequently update the `lastActive` time on the queue. 

## Type of change


reillyse (approved):
looks good

PR #984, opened 2024-10-24
add multiple rate limiter in grpc using a token bucket

reillyse:
# Description

Added in a rate limit for GRPC requests.

We have separate rate limits for each of Workflow, Event and Dispatcher. 
Each api token has it's own rate limit.
The default is 1k / second

I've added a config grpcRateLimit, SERVER_GRPC_RATE_LIMIT

This config is used for each of the rate limits


internal/services/grpc/middleware/auth.go
```diff
@@ -41,6 +43,10 @@ func (a *GRPCAuthN) Middleware(ctx context.Context) (context.Context, error) {
 		return nil, forbidden
 	}
 
+	// we want to rate limit by API token
+
+	ctx = context.WithValue(ctx, "rate_limit_token", hashToken(token))
```
abelanger5: The token's ID is encoded in the token and should be relatively easy to return from the token package (similar to how we're returning the tenant id). I'd feel more comfortable from a security perspective using the UUID corresponding to the token than setting a hash of the actual token value in the context.

internal/services/grpc/middleware/rate_limit.go
```diff
@@ -0,0 +1,150 @@
...
+	fmt.Println("the service is", serviceName)
+
+	rateLimitToken := ctx.Value("rate_limit_token").(string)
+
+	if rateLimitToken == "" {
+		return status.Errorf(codes.Unauthenticated, "no rate limit token found")
+	}
+
+	switch matchServiceName(serviceName) {
+	case "dispatcher":
```
abelanger5, line 69: I don't think we should be rate-limiting the dispatcher, or at least the default rate limit should be higher for the dispatcher. We are not currently seeing issues on this service.

pkg/config/server/server.go
```diff
@@ -97,6 +97,9 @@ type ConfigFileRuntime struct {
 	// GRPCMaxMsgSize is the maximum message size that the grpc server will accept
 	GRPCMaxMsgSize int `mapstructure:"grpcMaxMsgSize" json:"grpcMaxMsgSize,omitempty" default:"4194304"`
 
+	// GRPCRateLimit is the rate limit for the grpc server
```
abelanger5: should clarify that this is a per-second, per-engine rate limit

abelanger5 (changes requested):

internal/services/grpc/middleware/auth.go
reillyse: I now use the ID of the token from the db.

internal/services/grpc/middleware/rate_limit.go
reillyse: ok, set it to 10X the provided rate for the dispatcher - probably best to have some rate limit and a 10X ratio seems safe enough ?

pkg/config/server/server.go
reillyse: done

internal/services/grpc/middleware/rate_limit.go
abelanger5: sounds good, thanks!

abelanger5 (approved):

internal/services/grpc/middleware/rate_limit.go
```diff
@@ -0,0 +1,151 @@
...
+		rate:         r,
+		burst:        b,
+		l:            l,
+	}
+}
+
+// Limit is called before each request is processed. It should return an error if rate-limited.
+func (r *HatchetRateLimiter) Limit(ctx context.Context) error {
+
+	fmt.Printf("The rate limit is being checked it is %d \n", int(r.rate))
```
abelanger5: Remove debug line

pkg/client/client.go
```diff
@@ -215,9 +217,17 @@ func newFromOpts(opts *ClientOpts) (Client, error) {
 		transportCreds = credentials.NewTLS(opts.tls)
 	}
 
+	retryOpts := []grpc_retry.CallOption{
```
abelanger5, line 225: Can we case this specifically on `rate limit exceeded` or resource exhausted errors? For a lot of calls, there is no point in retrying

abelanger5 (changes requested):

pkg/client/client.go
reillyse: I think these are the defaults but I'll make it explicit

reillyse: changed to
	grpc_retry.WithCodes(codes.ResourceExhausted, codes.Unavailable),

should help with random network errors but let me know if I should get rid of the Unavailable

abelanger5 (approved):

PR #990, opened 2024-10-24
add a serial write for step run events

reillyse:
Adds the ability to serialize step run event creation

### DIFF
--- a/pkg/config/server/server.go
+++ b/pkg/config/server/server.go
@@ -142,9 +142,6 @@ type ConfigFileRuntime struct {
 	// FlushItemsThreshold is the default number of items to hold in memory until flushing to the database
 	FlushItemsThreshold int `mapstructure:"flushItemsThreshold" json:"flushItemsThreshold,omitempty" default:"100"`
 
-	// SerialBuffer controls whether the event buffer is serial or runs concurrently
-	SerialBuffer bool `mapstructure:"serialBuffer" json:"serialBuffer,omitempty" default:"false"`
-
 	// WorkflowRunBuffer represents the buffer settings for workflow runs
 	WorkflowRunBuffer buffer.ConfigFileBuffer `mapstructure:"workflowRunBuffer" json:"workflowRunBuffer,omitempty"`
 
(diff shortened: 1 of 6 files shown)


pkg/config/server/server.go
```diff
@@ -477,6 +477,7 @@ func BindAllEnv(v *viper.Viper) {
 
 	_ = v.BindEnv("runtime.eventBuffer.flushPeriodMilliseconds", "SERVER_EVENTBUFFER_FLUSH_PERIOD_MILLISECONDS")
 	_ = v.BindEnv("runtime.eventBuffer.flushItemsThreshold", "SERVER_EVENTBUFFER_FLUSH_ITEMS_THRESHOLD")
+	_ = v.BindEnv("runtime.eventBuffer.serialBuffer", "SERVER_EVENTBUFFER_SERIAL_BUFFER")
```
abelanger5, line 480: This needs to be set in the config values above.

abelanger5 (dismissed):

pkg/config/server/server.go
reillyse: ok added above

reillyse: actually I'm confused now @abelanger5 I think I have it in the buffer -> config.go file so that does it I think? 
package buffer

// ConfigFileBuffer is the configuration for the buffer. We store it here to prevent circular dependencies
type ConfigFileBuffer struct {
	// FlushPeriodMilliseconds is the number of milliseconds before flush
	FlushPeriodMilliseconds int `mapstructure:"flushPeriodMilliseconds" json:"flushPeriodMilliseconds,omitempty" default:"10"`

	// FlushItemsThreshold is the number of items to hold in memory until flushing to the database
	FlushItemsThreshold int `mapstructure:"flushItemsThreshold" json:"flushItemsThreshold,omitempty" default:"100"`

	// SerialBuffer is a flag to determine if the buffer should be serial or bulk
	SerialBuffer bool `mapstructure:"serialBuffer" json:"serialBuffer,omitempty" default:"false"`
}

pkg/repository/prisma/dbsqlc/queue.sql
```diff
@@ -1,40 +1,52 @@
...
+    WHERE
+        EXISTS (
+            SELECT
+                1
+            FROM
+                queue_exists
+        )
+        AND "tenantId" = @tenantId::uuid
+        AND "name" = @name::text
+    FOR UPDATE SKIP LOCKED
```
grutt, line 24: @abelanger5 I feel like we've seen issues with this pattern for heart beating on partitions, i suppose we'll be calling this less frequently

abelanger5: The solve for that was a statement timeout on the heartbeats, in case we accidentally have a connection holding a lock when it shouldn't. Will do the same thing here.

grutt (approved):

PR #991, opened 2024-10-24
Feat concurrency limit for flush

reillyse:
# Description

I think we should have some upper bounds on concurrency in the bulk writing. This puts some guardrails around how many bulk writes can happen per buffer at any given moment. Items can still buffer and we have some log warnings if we get too far behind.

- [X ] New feature (non-breaking change which adds functionality)


pkg/repository/buffer/buffered.go
```diff
@@ -313,6 +405,14 @@ func (b *IngestBuf[T, U]) BuffItem(item T) (chan *FlushResponse[U], error) {
 		return nil, fmt.Errorf("buffer not ready, in state '%v'", b.state.String())
 	}
 
+	if b.safeCheckSizeOfBuffer() >= b.maxCapacity*50 {
+		return nil, status.Errorf(codes.ResourceExhausted, "buffer is out of space %v", b.safeCheckSizeOfBuffer())
```
abelanger5, line 409: Why are we using GRPC status errors here?

abelanger5 (approved):

pkg/repository/buffer/buffered.go
reillyse: trying to make it work with the retry when we push that.

PR #1001, opened 2024-10-30
fix: filter alert runs by failure only

grutt:
# Description

Alerting was noisy and alerting to cancellations as well as failures.

## Type of change

## What's Changed

- [x] Alert runs are filtered by workflow run status AND job run status of failed


pkg/repository/prisma/dbsqlc/workflow_runs.sql
```diff
@@ -1415,3 +1415,22 @@ WHERE
...
+FROM "WorkflowRun" wr
+JOIN
+	"JobRun" jr on jr."workflowRunId" = wr."id"
+JOIN
+	"StepRun" sr on sr."jobRunId" = jr."id"
+WHERE
+	wr."status" = 'FAILED' AND
+	wr."id" = @workflowRunId::uuid AND
+    wr."tenantId" = @tenantId::uuid
+LIMIT 1;
```
abelanger5: This is a LIMIT 1 but you are joining job run and step run, of which there could be many, so you might be selecting a succeeded step run in a workflow with many steps.

abelanger5 (changes requested):

abelanger5 (approved):

PR #1014, opened 2024-11-05
fix: unacked messages in the queuer

abelanger5:
# Description

Prevents an issue where an unacked queue item was never removed from the queuer, therefore never being processed again. This was caused by an early return statement out of the scheduler when a queue item had hit the `schedulingTimedOut` case. 

## Type of change

### DIFF
--- a/pkg/scheduling/v2/scheduler.go
+++ b/pkg/scheduling/v2/scheduler.go
@@ -518,6 +518,11 @@ func (s *Scheduler) tryAssignBatch(
 		// if the action is not in the map, then we have no slots to assign to
 		for i := range res {
 			res[i].noSlots = true
+
+			// nack the rate limit result, if it exists
+			if res[i].rateLimitResult != nil {
+				res[i].rateLimitResult.nack()
+			}
 		}
 
 		return res, newRingOffset, nil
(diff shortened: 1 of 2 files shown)


pkg/scheduling/v2/scheduler.go
```diff
@@ -518,6 +518,11 @@ func (s *Scheduler) tryAssignBatch(
 		// if the action is not in the map, then we have no slots to assign to
 		for i := range res {
 			res[i].noSlots = true
+
+			// nack the rate limit result, if it exists
+			if res[i].rateLimitResult != nil {
+				res[i].rateLimitResult.nack()
```
abelanger5, line 524: This also fixes an issue where a rate limit is used when there are no slots available.

pkg/scheduling/v2/scheduler.go
```diff
@@ -755,22 +767,16 @@ func (s *Scheduler) tryAssign(
...
+						rateLimited: batchRateLimited,
+						unassigned:  batchUnassigned,
 					}
 
 					return nil
 				})
 
 				if err != nil {
 					s.l.Error().Err(err).Msg("error assigning queue items")
-					return
```
abelanger5, line 766: This was the early return statement which prevented us from sending the rest of the `assignResults` below.

pkg/scheduling/v2/scheduler.go
```diff
@@ -735,9 +743,13 @@ func (s *Scheduler) tryAssign(
...
 						if !singleRes.succeeded {
 							if singleRes.rateLimitResult != nil {
-								rateLimited = append(rateLimited, singleRes.rateLimitResult)
-							} else if singleRes.noSlots {
-								unassigned = append(unassigned, singleRes.qi)
+								batchRateLimited = append(batchRateLimited, singleRes.rateLimitResult)
+							} else {
+								batchUnassigned = append(batchUnassigned, singleRes.qi)
+
+								if !singleRes.noSlots {
```
abelanger5, line 750: This case shouldn't happen, and as far as I can tell hasn't happened, but it's good just to double check.

grutt (approved):

PR #1050, opened 2024-11-15
fix: tenant race conditions, cleanup logic, old workers getting assigned

abelanger5:
# Description

Fixes an issue where tenant schedulers may leak, causing old workers to get assigned work. This adds a safeguard to the scheduler to ignore old workers (shouldn't add any query latency since we're already selecting the workers) and fixes the logic around cleaning up lease managers and assigning new tenant managers. 

Fixes # (issue)

## Type of change

### DIFF
--- a/internal/services/dispatcher/server.go
+++ b/internal/services/dispatcher/server.go
@@ -553,14 +553,7 @@ func (s *DispatcherImpl) Heartbeat(ctx context.Context, req *contracts.Heartbeat
 		return nil, err
 	}
 
-	// if we haven't seen the dispatcher for 6 seconds (one interval plus latency), reject the heartbeat as the client
-	// should reconnect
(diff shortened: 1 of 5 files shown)


internal/services/dispatcher/server.go
```diff
@@ -553,14 +553,7 @@ func (s *DispatcherImpl) Heartbeat(ctx context.Context, req *contracts.Heartbeat
 		return nil, err
 	}
 
-	// if we haven't seen the dispatcher for 6 seconds (one interval plus latency), reject the heartbeat as the client
-	// should reconnect
-	if worker.DispatcherLastHeartbeatAt.Time.Before(time.Now().Add(-6 * time.Second)) {
```
abelanger5, line 558: Removed -- I don't think this check is useful and it's causing problems when the DB is under high load.

grutt (approved):

PR #1055, opened 2024-11-18
Add a dynamic strategy for flushing as a function of currently flushing

reillyse:
# Description

Fixes # (issue)

## Type of change

## What's Changed


go.mod
```diff
@@ -147,6 +148,7 @@ require (
 	github.com/rabbitmq/amqp091-go v1.10.0
 	github.com/robfig/cron/v3 v3.0.1 // indirect
 	github.com/rs/zerolog v1.33.0
+	github.com/sasha-s/go-deadlock v0.3.5
```
abelanger5: Think you need to `go mod tidy` to remove this

pkg/config/server/server.go
```diff
@@ -154,6 +154,9 @@ type ConfigFileRuntime struct {
 	// FlushItemsThreshold is the default number of items to hold in memory until flushing to the database
 	FlushItemsThreshold int `mapstructure:"flushItemsThreshold" json:"flushItemsThreshold,omitempty" default:"100"`
 
+	// FlushStrategy is the strategy to use for flushing the buffer
+	FlushStrategy buffer.BuffStrategy `mapstructure:"flushStrategy" json:"flushStrategy" default:"DYNAMIC"`
```
abelanger5, line 158: Have we tested that setting `STATIC` as an env var properly marshals in the static buffer strategy?

abelanger5: The reason I ask is that I've had problems with viper marshaling config in weird ways before...

abelanger5 (approved):

PR #1075, opened 2024-11-27
Feat skip states for workflows

reillyse:
# Description
The majority of our workflows can skip the "pending" state and go straight to running. This change implements that short-circuit. 

Fixes # (issue)

## Type of change

## What's Changed


pkg/repository/prisma/workflow_run.go
```diff
@@ -1689,10 +1709,18 @@ func createNewWorkflowRuns(ctx context.Context, pool *pgxpool.Pool, queries *dbs
 				})
 			}
 
+			jrStatus := dbsqlc.JobRunStatusPENDING
+
+			// TODO is this the correct logic?
+			if opt.GetGroupKeyRun == nil && opt.DedupeValue == nil {
+				jrStatus = dbsqlc.JobRunStatusRUNNING
```
abelanger5: We don't want to set `on_failure` steps in a running state.

reillyse: Is there something in the opts that I can check to see if it is an on_failure ?

reillyse: TriggeredBy maybe ?

pkg/repository/prisma/workflow_run.go
```diff
@@ -2265,3 +2488,66 @@ func bulkWorkflowRunEvents(
...
+
+	if err != nil {
+		l.Err(err).Msg("could not add message to tenant partition queue")
+		return fmt.Errorf("could not get tenant: %w", err)
+	}
+
+	if tenant.ControllerPartitionId.Valid {
+		err = mq.AddMessage(
+			ctx,
+			msgqueue.QueueTypeFromPartitionIDAndController(tenant.ControllerPartitionId.String, msgqueue.WorkflowController),
```
abelanger5: If we're short-circuiting, this shouldn't be necessary -- this signal will bump the workflow run queue, which we know isn't necessary in the short circuit case since we're explicitly checking for the concurrency condition.

pkg/repository/prisma/workflow_run.go
```diff
@@ -2265,3 +2488,66 @@ func bulkWorkflowRunEvents(
...
+				tenantId,
+				workflowRunId,
+			),
+		)
+		if err != nil {
+			return fmt.Errorf("could not add workflow run queued task: %w", err)
+		}
+	} else if tenant.SchedulerPartitionId.Valid {
+
+		for _, queueName := range workflowRun.StepRunQueueNames {
```
abelanger5: Why are we signaling all step run queues? Shouldn't we only be signaling the queue which corresponds to the step runs which actually started?

pkg/repository/prisma/dbsqlc/step_runs.sql
```diff
@@ -240,6 +240,143 @@ WHERE
         sr."tenantId" = sqlc.narg('tenantId')::uuid
     );
 
+
+-- name: GetStartableStepRunsForWorkflowRuns :many
```
abelanger5, line 244: I think it would be significantly cleaner to return a list of step run IDs here, and then call `GetStepForEngine` in the same tx to do the join. I think how the data is populated isn't exactly correct here, and this will fix it (will also flag where it's incorrect for reference)

pkg/repository/prisma/dbsqlc/step_runs.sql
```diff
@@ -240,6 +240,143 @@ WHERE
...
+    FROM
+        "StepRun" AS child_run
+    LEFT JOIN
+        "_StepRunOrder" AS step_run_order ON step_run_order."B" = child_run."id"
+    WHERE
+        child_run."jobRunId" IN (SELECT "jobRunId" FROM JobRuns)
+        AND child_run."status" = 'PENDING'
+        AND step_run_order."A" IS NULL
+),
+ChildCount AS (
```
abelanger5: Child count is the number of child workflows that the step spawned, so we know it will be 0 since the step hasn't started yet. The way this is written, it looks like it will be 1 for all step runs, which is incorrect.

pkg/repository/prisma/step_run.go
```diff
@@ -1801,43 +1810,56 @@ func (s *stepRunEngineRepository) QueueStepRun(ctx context.Context, tenantId, st
...
 
 		if err != nil {
-			return nil, fmt.Errorf("could not release worker semaphore queue items: %w", err)
+			return nil, err
 		}
 
-		// retries get highest priority to ensure that they're run immediately
-		priority = 4
-	}
+		err = s.bulkSemaphoreReleaser.FireForget(tenantId, semaphoreReleaseOpts{
```
abelanger5: Note: make sure to remove this `FireForget`. This was a recent bug fix we just merged into main in #1175

pkg/repository/prisma/step_run.go
```diff
@@ -1801,43 +1810,56 @@ func (s *stepRunEngineRepository) QueueStepRun(ctx context.Context, tenantId, st
...
+
+			if err != nil {
+				return nil, fmt.Errorf("could not release worker semaphore queue items: %w", err)
+			}
+
+			// retries get highest priority to ensure that they're run immediately
+			priority = 4
+		}
+
+		_, err := s.bulkQueuer.FireAndWait(ctx, tenantId, bulkQueueStepRunOpts{
```
abelanger5, line 1831: The way this is written doesn't use the `tx` we're passing in. I think we should use the method which bulk queues step runs on the `sharedRepository`. Since workflow runs are written in bulk, we can hook into that to bulk queue step runs -- so we don't need to use the buffer, if that makes sense.

pkg/repository/prisma/workflow_run.go
```diff
@@ -1953,39 +1970,42 @@ func createNewWorkflowRuns(ctx context.Context, pool *pgxpool.Pool, queries *dbs
 				})
 			}
 
+			jrStatus := dbsqlc.JobRunStatusPENDING
```
abelanger5, line 1985: Are we actually setting this to `RUNNING` in the short-circuit case?

pkg/repository/prisma/workflow_run.go
```diff
@@ -2123,49 +2147,248 @@ func createNewWorkflowRuns(ctx context.Context, pool *pgxpool.Pool, queries *dbs
...
+			return nil, err
+		}
+
+		if len(createdWorkflowRuns) > 0 && len(queueNames) > 0 {
+			// NOTE: this is a hack to just set the first queue names for all the workflow runs
+			// We need to have an array of a single struct return from the buffer function.
+			// We do this because we bunch all the workflow runs in a single query to get the queue names and after we return we will just
+			// hit the mq with all the queue names.
+
+			createdWorkflowRuns[0].StepRunQueueNames = queueNames
```
abelanger5: Perhaps a more descriptive field name would be `InitialStepRunQueueNames`?

pkg/repository/prisma/workflow_run.go
```diff
@@ -2123,49 +2147,248 @@ func createNewWorkflowRuns(ctx context.Context, pool *pgxpool.Pool, queries *dbs
...
+
+	var workflowRunIds []pgtype.UUID
+
+	for _, wfr := range wfrs {
+		workflowRunIds = append(workflowRunIds, wfr.WorkflowRun.ID)
+	}
+
+	// update the workflow run status to running
+
+	err := s.queries.SetWorkflowRunRunning(ctx, tx, workflowRunIds)
```
abelanger5, line 2266: Why are we wasting a round trip on this when we can just insert the workflow run with a RUNNING status?

pkg/repository/prisma/workflow_run.go
```diff
@@ -2123,49 +2147,248 @@ func createNewWorkflowRuns(ctx context.Context, pool *pgxpool.Pool, queries *dbs
...
+		"step_run_id": data.SRID,
+	}
+
+	queueOpts := &repository.QueueStepRunOpts{
+		IsRetry: false,
+	}
+
+	inputDataBytes := data.Input
+
+	if in := data.Input; len(in) == 0 || string(in) == "{}" {
```
abelanger5: This is a duplicate of the code in the job controller. Is there a way to consolidate?

pkg/repository/prisma/workflow_run.go
```diff
@@ -2265,3 +2488,66 @@ func bulkWorkflowRunEvents(
 		l.Err(err).Msg("could not create bulk workflow run event")
 	}
 }
+
+func CanShortCircuit(workflowRunRow *dbsqlc.GetWorkflowRunsInsertedInThisTxnRow) bool {
+
+	return !(workflowRunRow.ConcurrencyLimitStrategy.Valid || workflowRunRow.ConcurrencyGroupExpression.Valid || workflowRunRow.GetGroupKeyRunId.Valid || workflowRunRow.WorkflowRun.ConcurrencyGroupId.Valid || workflowRunRow.DedupeValue.Valid || workflowRunRow.FailureJob)
+}
+
+// TODO this shouldn't be in the repo probably, should be at the controller layer but I'm not sure where.
```
abelanger5: I would just pull this into a helper package which returns a `NotifyQueue` callback that can be passed into the repository method.

abelanger5 (changes requested):

pkg/repository/prisma/step_run.go
reillyse: this is run in the post commit so is outside of the tx

pkg/repository/prisma/workflow_run.go
reillyse: iirc when we create the workflow run we don't have all of the info available to us. We don't have a workflow run obviously so we'd have to reimplement the logic and we don't have dedupe values and possibly others (concurrency key and expressions?).

pkg/repository/prisma/workflow_run.go
reillyse: Yes later on in the flow

pkg/repository/prisma/workflow_run.go
reillyse: As I understand it that is what we are doing? These are the initial step run queue names

pkg/repository/prisma/workflow_run.go
reillyse: extracted into a util

mrkaye97:
closing in favor of v1

PR #1076, opened 2024-11-27
fix: make cron migration more performant

abelanger5:
# Description

Adds `NOT VALID` to the migration to improve performance on large datasets. 

## Type of change


hatchet-temporary (approved):
LGTM! Any thoughts on manually validating the constraint in application code / fixing any breakages after running this?

abelanger5:
@hatchet-temporary we actually know this constraint is satisfied since we're dropping the previous constraint in the same tx -- this is the equivalent to adding an empty `name` column to the existing constraints.

PR #1080, opened 2024-12-03
fix: remove deprecated swagger-cli

abelanger5:
# Description

Removes the `swagger-cli` dependency that was causing issues during builds. 

## Type of change


api/v1/server/oas/transformers/workflow.go
```diff
@@ -148,7 +148,7 @@ func ToWorkflowVersionConcurrency(concurrency *WorkflowConcurrency) *gen.Workflo
 
 	res := &gen.WorkflowConcurrency{
 		MaxRuns:             concurrency.MaxRuns.Int32,
-		LimitStrategy:       gen.WorkflowConcurrencyLimitStrategy(concurrency.LimitStrategy.ConcurrencyLimitStrategy),
+		LimitStrategy:       gen.ConcurrencyLimitStrategy(concurrency.LimitStrategy.ConcurrencyLimitStrategy),
```
hatchet-temporary, line 151: is this change related?

hatchet-temporary (approved):
LGTM 👍

api/v1/server/oas/transformers/workflow.go
abelanger5: Yes, somehow changing the package for bundling changed our generated API as well. Seems to not be an issue but will double check on staging.

PR #1083, opened 2024-12-04
feat(go-sdk): cron and schedules API, minor fixes

abelanger5:
# Description

1. Implements the `Cron` and `Schedule` APIs for the Go SDK. 
2. Fixes some small issues in the cron and scheduling APIs:
- Creating a schedule was failing with a 500-level error due to the workflow populator searching by ID. 
- The migration for existing cron schedules was setting the name to `NULL` instead of an empty string, causing foreign key problems

## Type of change


Taskfile.yaml
```diff
@@ -167,7 +167,7 @@ tasks:
       - mkdir -p ./python-sdk/certs/ && cp ./hack/dev/certs/ca.cert ./python-sdk/certs/
   pre-commit-install:
     cmds:
-      - pip install pre-commit
+      - brew install pre-commit
```
abelanger5: Was running into problems with `pip` globally, figured everyone using this command is on OSX and using brew and this is slightly cleaner anyway, but if this breaks someone's setup can revert.

abelanger5: Update: it broke our actions. Reverting.

frontend/docs/components/code/CodeBlock.tsx
```diff
@@ -3,65 +3,125 @@ import { parseDocComments } from "./codeParser";
...
-                    setCopied(true);
-                    setTimeout(() => setCopied(false), 2000);
-                }}>
-                    {copied ? <><CheckIcon className="w-4 h-4 mr-2" />Copied</> : <><CopyIcon className="w-4 h-4 mr-2" />Copy</>}
-                </Button>
-            </div>
+  return (
+    <>
+      <div className="sticky top-0 z-10 bg-background flex flex-row gap-4 justify-between items-center pl-2 mb-2">
+        <a
```
hatchet-temporary, line 30: nit but a thing for the future - I think it's probably recommended to use `Link` from `next/link` here instead of an `a` tag, but this is fine 👍

frontend/docs/pages/home/features/triggering-runs/cron-trigger.mdx
```diff
@@ -27,7 +27,9 @@ export const CronTriggerPyProgrammaticAsync = {
 };
 
 export const CronTriggerGoProgrammatic = {
-  path: "examples/cron-trigger/programatic-crons.go",
+  path: "examples/cron-programmatic/main.go",
+  // TODO: REMOVE BEFORE MERGE
```
hatchet-temporary: flagging this icymi

sql/migrations/20241121142159_v0.52.0.sql
```diff
@@ -10,6 +10,9 @@ CREATE TYPE "WorkflowTriggerCronRefMethods" AS ENUM ('DEFAULT', 'API');
 CREATE TYPE "WorkflowTriggerScheduledRefMethods" AS ENUM ('DEFAULT', 'API');
 -- Modify "WorkflowTriggerCronRef" table
 ALTER TABLE "WorkflowTriggerCronRef" ADD COLUMN "name" text NULL, ADD COLUMN "id" uuid NOT NULL, ADD COLUMN "method" "WorkflowTriggerCronRefMethods" NOT NULL DEFAULT 'DEFAULT', ADD CONSTRAINT "WorkflowTriggerCronRef_parentId_cron_name_key" UNIQUE ("parentId", "cron", "name");
+
+UPDATE "WorkflowTriggerCronRef" SET "name" = '' WHERE "name" IS NULL;
```
hatchet-temporary, line 14: fixing a broken past migration?

hatchet-temporary (approved):
looks good on the non-Golang stuff 👍 

one request / thought: I think for non-docs frontend changes, it'd be great to post screenshots / recordings of UI that's changed since it's easier to review that way IMO! WDYT?

reillyse (approved):
LGTM

sql/migrations/20241121142159_v0.52.0.sql
abelanger5: since we haven't released this as latest, this is alright -- this just makes the migration more zero-downtime

frontend/docs/components/code/CodeBlock.tsx
abelanger5: actually just ran `prettier` here, so this is really a 1-line change on the docs

hatchet-temporary: ah nice, thanks! in any case, I still think we should do this at some point :P but not now

PR #1094, opened 2024-12-05
fix: v0.52.5 migrations

abelanger5:
# Description

Fixes migrations for `v0.52.5` when there are existing cron refs without an id. 

## Type of change


sql/migrations/20241121142159_v0.52.0.sql
```diff
@@ -8,8 +8,22 @@ CREATE INDEX "RetryQueueItem_isQueued_tenantId_retryAfter_idx" ON "RetryQueueIte
...
 -- Create enum type "WorkflowTriggerScheduledRefMethods"
 CREATE TYPE "WorkflowTriggerScheduledRefMethods" AS ENUM ('DEFAULT', 'API');
--- Modify "WorkflowTriggerCronRef" table
-ALTER TABLE "WorkflowTriggerCronRef" ADD COLUMN "name" text NULL, ADD COLUMN "id" uuid NOT NULL, ADD COLUMN "method" "WorkflowTriggerCronRefMethods" NOT NULL DEFAULT 'DEFAULT', ADD CONSTRAINT "WorkflowTriggerCronRef_parentId_cron_name_key" UNIQUE ("parentId", "cron", "name");
+
+-- Step 1: Add the new columns with "id" as nullable
+ALTER TABLE "WorkflowTriggerCronRef" 
+ADD COLUMN "name" text NULL, 
+ADD COLUMN "id" uuid NULL, 
+ADD COLUMN "method" "WorkflowTriggerCronRefMethods" NOT NULL DEFAULT 'DEFAULT',
```
reillyse, line 16: How big is WorkflowTriggerCronRef ?  we will probably need to create an index on the id after we add it because once we alter table to be not null I presume that will perform a seq scan on the table to check for NULL


Between the update and the alter it's possible for someone to add a row into the table that will break the next step. It would be best to update the rows, lock the table, update any new rows and then alter the table in that transaction.

reillyse (changes requested):

reillyse (approved):

PR #1095, opened 2024-12-05
fix: race condition if a step run has multiple parents

abelanger5:
# Description

Fixes a rare race condition with large payloads if a step run has multiple parents. We buffer and write the status update to an internal queue before we update the lookup data for the job run -- so if the write to update the lookup data is slow (with large payloads), we may process the status update before we write the job run data, which can lead to a child step starting without the correct input data set. 

I believe we weren't seeing this with any payloads on the order of 10s of KBs, but larger payloads will trigger this occasionally.  

## Type of change


reillyse (approved):
looks good

PR #1097, opened 2024-12-06
fix: frontend improvements

abelanger5:
# Description

A set of minor tweaks to the frontend:
- In the trigger workflow form, creates a default state if there are no workflows present and sets the name of the workflow correctly when selected
- Makes the default for code editors and highlighters to have a copy button present
- Prevents overflow of events modal
- Improves logic of slug generation in onboarding

## Type of change

### DIFF
--- a/frontend/app/src/pages/main/events/index.tsx
+++ b/frontend/app/src/pages/main/events/index.tsx
@@ -501,7 +501,7 @@ function EventsTable() {
 
 function ExpandedEventContent({ event }: { event: Event }) {
   return (
-    <DialogContent className="w-fit max-w-[80%] min-w-[500px]">
+    <DialogContent className="w-fit max-w-[700px] overflow-hidden">
       <DialogHeader>
         <DialogTitle>Event {event.key}</DialogTitle>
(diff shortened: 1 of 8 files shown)


frontend/app/src/pages/main/events/index.tsx
```diff
@@ -501,7 +501,7 @@ function EventsTable() {
 
 function ExpandedEventContent({ event }: { event: Event }) {
   return (
-    <DialogContent className="w-fit max-w-[80%] min-w-[500px]">
+    <DialogContent className="w-fit max-w-[700px] overflow-hidden">
```
hatchet-temporary, line 504: I figure it doesn't matter much, but do we think about responsiveness at all (e.g. mobile / tablet)?

hatchet-temporary (approved):

frontend/app/src/pages/main/events/index.tsx
abelanger5: That's a good callout -- we have tried to make login + some of the basic views responsive, but it's not been a priority

PR #1103, opened 2024-12-06
Speed up the delete worker query 

reillyse:
# Description

Fixes # (issue)

## Type of change

## What's Changed


abelanger5 (changes requested):
I believe we're still missing the switch from `ON DELETE CASCADE` to `ON DELETE SET NULL` for the `WorkerAssignEvent` -> `Worker` FK relation (and additionally making the relation optional)

reillyse:
> I believe we're still missing the switch from `ON DELETE CASCADE` to `ON DELETE SET NULL` for the `WorkerAssignEvent` -> `Worker` FK relation (and additionally making the relation optional)

added a migration to make those changes

sql/migrations/20241206231312_v52.6.sql
abelanger5, line 1: Merge these migration files, and the correct number would be `v0.52.12`, since the last patch release was `v0.52.11`

sql/schema/schema.sql
```diff
@@ -1370,6 +1370,10 @@ CREATE UNIQUE INDEX "Worker_id_key" ON "Worker" ("id" ASC);
 -- CreateIndex
 CREATE UNIQUE INDEX "Worker_webhookId_key" ON "Worker" ("webhookId" ASC);
 
+-- CreateIndex
+
+CREATE INDEX "Worker_lastHeartbeatAt_idx" ON "Worker" ("lastHeartbeatAt", "tenantId");
```
abelanger5: The name of this index should be "Worker_tenantId_lastHeartbeatAt_idx", and I think we should swap the index `CREATE INDEX "Worker_lastHeartbeatAt_idx" ON "Worker" ("tenantId", "lastHeartbeatAt");`

sql/migrations/20241209235857_v52.7.sql
```diff
@@ -0,0 +1,2 @@
+-- Modify "WorkerAssignEvent" table
+ALTER TABLE "WorkerAssignEvent" DROP CONSTRAINT "WorkerAssignEvent_workerId_fkey", ALTER COLUMN "workerId" DROP NOT NULL, ADD CONSTRAINT "WorkerAssignEvent_workerId_fkey" FOREIGN KEY ("workerId") REFERENCES "Worker" ("id") ON UPDATE CASCADE ON DELETE SET NULL NOT VALID;
```
abelanger5: Just to confirm, have we tested this migration against a database which has existing Worker and WorkerAssignEvents in the table?

abelanger5 (changes requested):

sql/migrations/20241209235857_v52.7.sql
reillyse: yes seems fine - tested on cloud copy with 
cloud=> select count(*) from "WorkerAssignEvent";
  count
---------
 5307060
(1 row)

select count(*) from "Worker";
 count
-------
 92610
(1 row)

abelanger5 (approved):

PR #1108, opened 2024-12-09
add a monitoring probe

reillyse:
# Description

Fixes # (issue)

## Type of change

## What's Changed

This change adds monitoring probe endpoint that  runs a worker to test out the availability of the entire system. 

 curl -v   -H "Authorization: Bearer $HATCHET_CLIENT_TOKEN"   -X POST https://app.dev.hatchet-tools.com/api/v1/monitoring/707d0855-80ab-4e1f-a156-f1c4546cbf52/probe


api-contracts/openapi/components/schemas/monitoring.yaml
```diff
@@ -0,0 +1,4 @@
+RunProbe:
```
abelanger5: It looks like this object is just returning the `tenantId` which is getting passed in anyway, is there a reason we're returning this object?

api-contracts/openapi/paths/monitoring/monitoring.yaml
```diff
@@ -0,0 +1,41 @@
...
+          type: string
+          format: uuid
+          minLength: 36
+          maxLength: 36
+    responses:
+      "200":
+        content:
+          application/json:
+            schema:
+              $ref: "../../components/schemas/_index.yaml#/MonitoringRunProbe"
```
abelanger5: See above, I don't think this endpoint needs to return an object

api/v1/server/handlers/monitoring/probe.go
```diff
@@ -0,0 +1,317 @@
...
+	defer cancel()
+	var cleanup func() error
+
+	cf := clientconfig.ClientConfigFile{
+		Token:     token,
+		TenantId:  tenant.ID,
+		Namespace: randomNamespace(),
+		TLS: clientconfig.ClientTLSConfigFile{
+			Base: shared.TLSConfigFile{
+				TLSRootCAFile: os.Getenv("SERVER_TLS_ROOT_CA_FILE"),
```
abelanger5: We shouldn't be using `os.Getenv` here because config variables are not guaranteed to be set via env. Our config loader supports loading from config files as well. It might be worth taking a look at how we initialize TLS on the server.

api/v1/server/handlers/monitoring/probe.go
```diff
@@ -0,0 +1,317 @@
...
+			},
+		},
+	}
+	workflowName := "probe-workflow"
+	var wfrId *string
+	go func() {
+		wfrId, cleanup, err = m.run(cancellableContext, cf, workflowName, events, errorChan)
```
abelanger5: `.run` is non-blocking, so I'm not seeing a benefit to putting this inside of a `go func`? It also seems like it could easily cause race conditions with the casing on `cleanup` below.

api/v1/server/handlers/monitoring/probe.go
```diff
@@ -0,0 +1,317 @@
...
+			m.l.Warn().Msg("cleanup function was never set for probe")
+		}
+		if cleanupErr != nil {
+			m.l.Error().Msgf("error cleaning up probe worker: %s", err)
+		}
+	}()
+
+	defer func() {
+
+		time.Sleep(250 * time.Millisecond)
```
abelanger5: What's the purpose of sleeping in a deferred func?

api/v1/server/handlers/monitoring/probe.go
```diff
@@ -0,0 +1,317 @@
...
+
+	token, err := getBearerTokenFromRequest(ctx.Request())
+
+	if err != nil {
+		return nil, err
+	}
+
+	events := make(chan string, 50)
+	errorChan := make(chan error, 50)
+	interrupt := cmdutils.InterruptChan()
```
abelanger5: This is a channel which gets interrupted on `ctrl + c`, which should be occurring during cleanup on the server anyway, where requests get gracefully terminated and then the server gets shut down, so I don't think this is necessary

api/v1/server/handlers/monitoring/probe.go
```diff
@@ -0,0 +1,317 @@
...
+		return gen.MonitoringPostRunProbe403JSONResponse{}, err
+	}
+
+	token, err := getBearerTokenFromRequest(ctx.Request())
+
+	if err != nil {
+		return nil, err
+	}
+
+	events := make(chan string, 50)
```
abelanger5, line 51: I'm not seeing `events` or `errorChan` getting closed, and just wanted to clarify that's expected?

api/v1/server/handlers/monitoring/probe.go
```diff
@@ -0,0 +1,317 @@
...
+	}()
+
+	messages := []string{"This is a stream event for step-one", "This is a stream event for step-two"}
+	messageIndex := 0
+
+	for {
+
+		select {
+		case e := <-events:
+			if e != messages[messageIndex] {
```
abelanger5: Stream events aren't guaranteed to be ordered (they mostly should be, but it's not incorrect if they're not) -- so we can likely discard the ordering check.

api/v1/server/handlers/monitoring/probe.go
```diff
@@ -0,0 +1,317 @@
...
+		} else {
+			m.l.Warn().Msg("workflow run id was never set")
+		}
+
+	}()
+
+	messages := []string{"This is a stream event for step-one", "This is a stream event for step-two"}
+	messageIndex := 0
+
+	for {
```
abelanger5, line 81: We seem to only be checking for stream events -- I think it would be useful to double-check that the steps are called in the correct order and that the steps have the correct output. It'd probably be fine to just read from the database after the workflow run is done.

api/v1/server/handlers/monitoring/probe.go
```diff
@@ -0,0 +1,317 @@
...
+			}
+
+		case <-cancellableContext.Done():
+
+			if cancellableContext.Err() == context.DeadlineExceeded {
+				m.l.Error().Msg("timed out waiting for probe to complete")
+				return nil, fmt.Errorf("timed out waiting for probe to complete")
+			} else {
+				m.l.Debug().Msg("probe completed successfully")
+				return gen.MonitoringPostRunProbe200JSONResponse{}, nil
```
abelanger5: This case seems be returning a 200 response when the context gets cancelled without us necessarily checking any correctness conditions -- what's the thinking here?

api/v1/server/handlers/monitoring/probe.go
```diff
@@ -0,0 +1,317 @@
...
+
+type stepOneOutput struct {
+	Message        string `json:"message"`
+	UniqueStreamId string
+}
+
+func (m *MonitoringService) run(ctx context.Context, cf clientconfig.ClientConfigFile, workflowName string, events chan<- string, errors chan<- error) (*string, func() error, error) {
+
+	c, err := client.NewFromConfigFile(
+		&cf, client.WithLogLevel("info"),
```
abelanger5: Before merging, we might want to change this log level to match the server, otherwise this will be very noisy

api/v1/server/handlers/monitoring/probe.go
```diff
@@ -0,0 +1,317 @@
...
+
+		err := fmt.Errorf("tenant is not a monitoring tenant for this instance")
+
+		if len(m.permittedTenants) > 0 {
+			m.l.Error().Err(err).Msgf("monitoring tenants are %v", m.permittedTenants)
+		} else {
+			m.l.Error().Err(err).Msg("no monitoring tenants are configured")
+		}
+
+		return gen.MonitoringPostRunProbe403JSONResponse{}, err
```
abelanger5: Have we confirmed that returning `err` in addition to `gen.MonitoringPostRunProbe403JSONResponse` will correctly throw a `403` error and not a `500` error to the client?

api/v1/server/handlers/monitoring/probe.go
```diff
@@ -0,0 +1,317 @@
...
+)
+
+func (m *MonitoringService) MonitoringPostRunProbe(ctx echo.Context, request gen.MonitoringPostRunProbeRequestObject) (gen.MonitoringPostRunProbeResponseObject, error) {
+	if !m.enabled {
+		return nil, fmt.Errorf("monitoring is not enabled")
+	}
+
+	tenant := ctx.Get("tenant").(*db.TenantModel)
+
+	if tenant == nil || tenant.ID == "" {
```
abelanger5: This check already exists in the `authn` middleware, which checks that the authorization token belongs to the tenant which is set in the context.

abelanger5 (changes requested):

api/v1/server/handlers/monitoring/probe.go
reillyse: we were waiting a bit to let the workflow finish but I've changed it to poll on the workflow status to make sure it's in a completed state before we delete it to prevent errors from the workflow run finished task

api-contracts/openapi/paths/monitoring/monitoring.yaml
reillyse: ok, didn't realize I could just return a nil from the handler and it would be fine. Removed this object

api-contracts/openapi/components/schemas/monitoring.yaml
reillyse: nope was just getting errors if I removed it, didn't realize I could just not return an object

api/v1/server/handlers/monitoring/probe.go
reillyse: The issue with the TLS is it gets processed in the loader and we don't have the raw ca cert path. Everything else just takes a config in the init so to keep that interface clean I've modified the Runtime.Monitoring config to allow a Cert file to be passed and then we check in the loader to see if it has a value and if not I use the value from the Server. 

`	if cf.Runtime.Monitoring.TLSRootCAFile == "" {
		cf.Runtime.Monitoring.TLSRootCAFile = cf.TLS.TLSRootCAFile
	}
`

api/v1/server/handlers/monitoring/probe.go
reillyse: went down a bit of a rabbit hole pulling them from the DB, I think they are not necessarily in order - went back to doing it via the steps.

api/v1/server/handlers/monitoring/probe.go
reillyse: you are right, it was returning a 500 feels weird but returning the response and a nil

api/v1/server/handlers/monitoring/probe.go
abelanger5: gotcha, that seems reasonable

api-contracts/openapi/components/schemas/monitoring.yaml
abelanger5: Still looks like this is here, we should be good to delete this file along with `MonitoringRunProbe` defined in `_index.yaml`

api/v1/server/handlers/monitoring/probe.go
```diff
@@ -0,0 +1,320 @@
...
+			if err != nil {
+				m.l.Error().Msgf("error getting workflow run: %s", err)
+
+			}
+
+			if wrfRow.Status != dbsqlc.WorkflowRunStatusRUNNING {
+
+				workflowId := sqlchelpers.UUIDToStr(wrfRow.Workflow.ID)
+
+				_, err = m.config.APIRepository.Workflow().DeleteWorkflow(cf.TenantId, workflowId)
```
abelanger5: Let's modify this method so that it takes a context param as the first argument which gets passed to the underlying DB call, and use a shared context with a timeout here?

api/v1/server/handlers/monitoring/probe.go
```diff
@@ -0,0 +1,320 @@
...
+	}()
+
+	defer func() {
+		if wfrId == nil || *wfrId == "" {
+			m.l.Warn().Msg("workflow run id was never set for probe")
+			return
+		}
+
+		for i := 0; i < 10; i++ {
+			wrfRow, err := m.config.APIRepository.WorkflowRun().GetWorkflowRunById(context.Background(), cf.TenantId, *wfrId)
```
abelanger5: Use a shared context with a timeout

api/v1/server/handlers/monitoring/probe.go
```diff
@@ -0,0 +1,320 @@
...
+		},
+	}
+
+	wfrId, cleanup, err := m.run(cancellableContext, cf, m.workflowName, events, stepChan, errorChan)
+	if err != nil {
+		m.l.Error().Msgf("error running probe: %s", err)
+		return nil, err
+	}
+
+	defer func() {
```
abelanger5: To make this more readable, I think we can write this as `defer cleanup()` and:

1. The callee (`m.run`) should ensure that if `err == nil`, then `cleanup` is never nil
2. Errors should be logged from within the cleanup method itself

api/v1/server/handlers/monitoring/probe.go
abelanger5: @reillyse just confirming that you saw this?

api/v1/server/handlers/monitoring/probe.go
```diff
@@ -0,0 +1,320 @@
...
+		}
+	}()
+
+	go func() {
+		testEvent := probeEvent{
+			UniqueStreamId: streamValue,
+		}
+
+		err := c.Event().Push(
+			context.Background(),
```
abelanger5: Is there a reason we're not using the shared method context here?

abelanger5 (changes requested):

api/v1/server/handlers/monitoring/probe.go
```diff
@@ -0,0 +1,313 @@
...
+
+					workflowId := sqlchelpers.UUIDToStr(wrfRow.Workflow.ID)
+
+					_, err = m.config.APIRepository.Workflow().DeleteWorkflow(ctx, cf.TenantId, workflowId)
+
+					m.l.Info().Msgf("deleted workflow %s", workflowId)
+					if err != nil {
+						m.l.Error().Msgf("error deleting workflow: %s", err)
+					}
+					return
```
abelanger5, line 283: It seems like we're returning here no matter what, even if we've failed to delete the workflow?

api/v1/server/handlers/monitoring/probe.go
```diff
@@ -0,0 +1,313 @@
...
+		if err != nil {
+			m.l.Error().Msgf("error cleaning up worker: %s", err)
+		}
+		defer func() {
+			if wfrId == "" {
+				m.l.Warn().Msg("workflow run id was never set for probe")
+				return
+			}
+
+			ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
```
abelanger5, line 262: This timeout seems quite low for a PG delete -- perhaps something like 10 seconds?

abelanger5 (changes requested):

api/v1/server/handlers/monitoring/probe.go
reillyse: What do you think the behavior should be? If we are erroring here i assume this is not recoverable so I was just returning  - I modified to retry 10 times.

abelanger5: Ah, I see -- the retry is just for the workflow run being in the final state.

abelanger5 (approved):

PR #1109, opened 2024-12-11
add pagination for workflows 

reillyse:
# Description

\
## What's Changed

Added pagination for workflows and extended workflow list to 200 workflows


frontend/app/src/pages/main/workflows/components/workflow-table.tsx
```diff
@@ -40,18 +41,51 @@ export function WorkflowTable() {
...
+  ]);
+
   const data = useMemo(() => {
+
     const data = listWorkflowQuery.data?.rows || [];
+    setPageIndex(listWorkflowQuery.data?.pagination?.num_pages || 0);
 
     return data;
-  }, [listWorkflowQuery.data?.rows]);
+  }, [listWorkflowQuery.data?.rows , searchParams]);
```
grutt: this wont be autofixed with lint:fix

pkg/repository/prisma/dbsqlc/workflows.sql
```diff
@@ -82,7 +82,8 @@ SELECT
 FROM
     "Workflow" as workflows
 WHERE
-    workflows."tenantId" = $1 AND
+    workflows."tenantId" = @tenantId::uuid AND
+    workflows."name" LIKE '%' || @name::text || '%' AND
```
grutt:
```suggestion
    ( @name::text  IS NULL OR workflows."name" LIKE '%' || @name::text || '%' AND)
```
We should null check the input first,
we might need to use sqlc.nargs here

grutt:
```
AND
        (
            sqlc.narg('search')::text IS NULL OR
            workflow.name like concat('%', sqlc.narg('search')::text, '%')
        )
```

Also, curious how SQLC interprets the `||` is it the same as concat?

grutt (changes requested):
lint required

frontend/app/src/pages/main/workflows/components/workflow-table.tsx
reillyse: linted

pkg/repository/prisma/dbsqlc/workflows.sql
reillyse: || is just string concatenation  - standardized it to your way

grutt (approved):
ship it!

PR #1085, opened 2024-12-05
Bugfix -  stalled worklowruns 

reillyse:
# Description

If workflows failed to start step runs (in this case because of suspected DB CPU issues) the step run can fail but the workflow run is now in a Running State with no running step runs and will not be retried. 
We wrap it all in a txn to get it to fail or complete atomically.  

Fixes # (issue)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## What's Changed

- [ ] Add a list of tasks or features here...

PR #1114, opened 2024-12-11
refactor(repository): improve usability of repository

abelanger5:
# Description

This PR consolidates the repository methods which had become fractured so they're all defined in a single package. It moves the buffers and the scheduler repositories into the same package as other repository methods. It also introduces 3 patterns for recent issues we've seen while developing complex features:

1. Transactions may need to be shared between methods that are defined on separate repositories. For example, to get transactional safety for workflow run versus step run state as part of #1085, we need to run queries against step runs which were previous in the `StepRunRepository`. The solution in that PR (and elsewhere in the repo) was to pass `StepRunRepository` as a field dependency of `WorkflowRunRepository`. 

    There is now a `sharedRepository` which can be used to share underlying methods and transactions. As an example, I've modified the `DeferredStepRunEvent` (previously called from WorkflowRunRepository) to:

    ```go 
      func (s *sharedRepository) deferredStepRunEvent(
	    tenantId string,
	    opts repository.CreateStepRunEventOpts,
    ) {
	    // fire-and-forget for events
	    err := s.bulkEventBuffer.FireForget(tenantId, &opts)
    
	    if err != nil {
		    s.l.Error().Err(err).Msg("could not buffer event")
	    }
    }
    ```
    Each repository should embed `*sharedRepository`. This also makes initializing repositories a bit cleaner, as we don't need to pass the same validator, logger, or pgxpool around everywhere. 

2. Similar to the above, buffers are now defined on the `sharedRepository`, which gives us a single place where we need to track buffers and ensure they're cleaned up. They can also be called across multiple separate repositories. I've made the additional change of changing the buffer signature from `BuffItem`, which previously returned a `chan` with a result or error, and modified the signature to the following:

    ```go
    func FireForget(item T) error 
    
    func FireAndWait(ctx context.Context, item T) (*U, error)
    ```

    The reasoning here is that these are the only 2 patterns that have been used in the repository for buffering items, and dealing with multiple channels and context passed back to the caller is incredibly error-prone (I found a few instances of results not being awaited when they should have been). This also makes it much easier for a buffered write to respect a context timeout (we were previously setting a timer and adding an additional case for timeouts when awaiting results). 

3. Callback patterns -- there are several instances where we've wanted to hook into a pub/sub message or other type of callback inside of repository methods ( this was a challenge with #1075 ), but we've been hesitant to do this because we don't want to pass a pub/sub dependency through to the repository. 

    There's now an easy way to define callbacks on repository methods via the following:
  
    ```go
    // In the core repository package, when defining the repository's interface:
    RepositoryMethod(ctx context.Context, p params, opts ...CallbackOptFunc[string]) error
    
    // In the method implementation
    func (w *repository) RepositoryMethod(ctx context.Context, p params, opts ...CallbackOptFunc[string]) error {
        // start tx
    
        repository.RunPreCommit(w.l, tenantId, workflowRunId, opts)
    
        // commit tx
    
        repository.RunPostCommit(w.l, tenantId, workflowRunId, opts)
    
        return nil
    }
    ```
  
    The callbacks are utilizing the same underlying `Do` method as previously, which is non-blocking and only logs errors. We may have to tweak signatures here if we want a failed pre-commit hook to fail the transaction. 

## Type of change

### DIFF
--- a/pkg/repository/prisma/shared.go
+++ b/pkg/repository/prisma/shared.go
@@ -1,0 +1,124 @@
+package prisma
+
+import (
+	"github.com/hashicorp/go-multierror"
+	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/rs/zerolog"
+
+	"github.com/hatchet-dev/hatchet/pkg/config/server"
+	"github.com/hatchet-dev/hatchet/pkg/repository"
+	"github.com/hatchet-dev/hatchet/pkg/repository/buffer"
+	"github.com/hatchet-dev/hatchet/pkg/repository/prisma/dbsqlc"
+	"github.com/hatchet-dev/hatchet/pkg/validator"
+)
+
+type sharedRepository struct {
+	pool    *pgxpool.Pool
+	v       validator.Validator
+	l       *zerolog.Logger
+	queries *dbsqlc.Queries
+
+	bulkStatusBuffer      *buffer.TenantBufferManager[*updateStepRunQueueData, pgtype.UUID]
+	bulkEventBuffer       *buffer.TenantBufferManager[*repository.CreateStepRunEventOpts, int]
+	bulkSemaphoreReleaser *buffer.TenantBufferManager[semaphoreReleaseOpts, pgtype.UUID]
+	bulkQueuer            *buffer.TenantBufferManager[bulkQueueStepRunOpts, pgtype.UUID]
+	bulkUserEventBuffer   *buffer.TenantBufferManager[*repository.CreateEventOpts, dbsqlc.Event]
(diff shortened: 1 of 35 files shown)


pkg/repository/prisma/shared.go
```diff
@@ -0,0 +1,124 @@
...
+
+	"github.com/hatchet-dev/hatchet/pkg/config/server"
+	"github.com/hatchet-dev/hatchet/pkg/repository"
+	"github.com/hatchet-dev/hatchet/pkg/repository/buffer"
+	"github.com/hatchet-dev/hatchet/pkg/repository/prisma/dbsqlc"
+	"github.com/hatchet-dev/hatchet/pkg/validator"
+)
+
+type sharedRepository struct {
+	pool    *pgxpool.Pool
```
reillyse, line 17: nice

reillyse (approved):
Hard saying how easy/hard this will be without using the structure a bit but I like the direction.

PR #1119, opened 2024-12-12
feat: postgres-backed message queue

### LINKED ISSUE #89: feat(engine): PG-backed task queue
Support a postgres-backed task queue to remove the RabbitMQ requirement.

abelanger5:
# Description

Adds a message queue backed by Postgres to make the RabbitMQ dependency optional. 

This is a WIP, and still needs the following:
- [x] Non-durable queues need to delete `MessageQueue` after the consumer exits
- [x] Some fanout queues are incorrectly registered with exclusive consumers
- [x] Load testing up to 100 events/second
- [x] Add e2e and load tests which use both postgres and rabbitmq

Fixes #89

## Type of change


internal/msgqueue/postgres/exchange.go
```diff
@@ -0,0 +1,16 @@
+package postgres
+
+import (
+	"context"
+)
+
+func (p *PostgresMessageQueue) addTenantExchangeMessage(ctx context.Context, tenantId string, msgBytes []byte) error {
+	// determine if the exchange message is greater than 8kb
```
grutt, line 8: why do we have this constraint?

internal/msgqueue/postgres/msgqueue.go
```diff
@@ -0,0 +1,333 @@
...
+		upsertedQueues: make(map[string]bool),
+	}
+}
+
+func (p *PostgresMessageQueue) cleanup() error {
+	return nil
+}
+
+func (p *PostgresMessageQueue) Clone() (func() error, msgqueue.MessageQueue) {
+	// TODO: switch out with a new instance of the message queue
```
grutt: address todo?

internal/msgqueue/postgres/msgqueue.go
```diff
@@ -0,0 +1,333 @@
...
+	pCp := NewPostgresMQ(p.repo)
+
+	return pCp.cleanup, pCp
+}
+
+// TODO: actually use QOS to set a limit on the number of messages that can be processed at once
```
grutt: address todo?

internal/msgqueue/postgres/msgqueue.go
```diff
@@ -0,0 +1,333 @@
...
+
+		if err != nil {
+			p.l.Error().Err(err).Msg("error processing messages")
+		}
+
+		return len(messages) == p.qos, nil
+	}))
+
+	// we poll once per second for new messages
+	ticker := time.NewTicker(time.Second)
```
grutt, line 225: why are we polling? is it possible we miss a message on the channel?

internal/msgqueue/postgres/msgqueue.go
```diff
@@ -0,0 +1,333 @@
...
+
+	// we use the listener to poll for new messages more quickly
+	newMsgCh := make(chan struct{})
+
+	// start the listener
+	go func() {
+		err = p.repo.Listen(subscribeCtx, queue.Name(), func(ctx context.Context, notification *repository.PubMessage) error {
+			// if this is an exchange queue, and the message starts with JSON '{', then we process the message directly
+			if queue.FanoutExchangeKey() != "" && notification.Payload[0] == '{' {
```
grutt: is it possible the payload can be nil?

pkg/repository/prisma/buffer_add_mq.go
```diff
@@ -0,0 +1,83 @@
...
+		i := index
+		res = append(res, &i)
+
+		p = append(p, dbsqlc.BulkAddMessageParams{
+			QueueId: pgtype.Text{
+				String: opt.queue,
+				Valid:  true,
+			},
+			Payload:   opt.payload,
+			ExpiresAt: sqlchelpers.TimestampFromTime(time.Now().UTC().Add(5 * time.Minute)),
```
grutt, line 57: nit: pass interval instead of timestamp so clock is based on db since we're using now() for delete

sql/migrations/20241212154910_v0.53.0.sql
```diff
@@ -0,0 +1,61 @@
...
+        "autoDeleted" boolean NOT NULL DEFAULT false,
+        "exclusive" boolean NOT NULL DEFAULT false,
+        "exclusiveConsumerId" uuid NULL,
+        PRIMARY KEY ("name")
+    );
+
+-- Create "MessageQueueItem" table
+CREATE TABLE
+    "MessageQueueItem" (
+        "id" bigint NOT NULL GENERATED ALWAYS AS IDENTITY,
```
grutt: assuming we're always using id for order?
i think we needed to increase timestamp resolution if we are using it for order for other events

grutt (commented):
a few questions

internal/msgqueue/postgres/exchange.go
abelanger5: 8kb is the max payload size of a NOTIFY message

internal/msgqueue/postgres/msgqueue.go
abelanger5: Yes, potentially: 

1. Not all setups will support listen/notify (pgbouncer)
2. Connections have an idle timeout of 15 minutes, so we might miss a message during a connection reset and miss a message

And I'm not sure about the reliability of listen/notify more generally. We can probably reduce this polling interval if we need to, or in some cases turn it off.

internal/msgqueue/postgres/msgqueue.go
abelanger5: Good catch 👍

pkg/repository/prisma/buffer_add_mq.go
abelanger5: The issue is `copyfrom`: `:copyfrom doesn't support non-parameter values`

sql/migrations/20241212154910_v0.53.0.sql
abelanger5: Ordering isn't strictly guaranteed, so I'm not sure why this would be necessary?

grutt (approved):

PR #1127, opened 2024-12-17
feat: `CANCEL_NEWEST` strategy and make cancel in progress more reliable 

abelanger5:
# Description

Adds a new `CANCEL_NEWEST` strategy for workflow run queues. If a workflow run's group key matches an existing value, we drop the newer workflow runs. This is essentially a deduplication strategy.  

Also improves the reliability of `CANCEL_IN_PROGRESS` and fixes an issue where the strategy did not consider currently queued runs as cancellable. 

Both strategies should have fewer race conditions as we're placing an explicit lock on the rows we'd like to update. If there are colliding rows, future enqueues will be blocked until the transaction commits.  

## Type of change


api-contracts/workflows/workflows.proto
```diff
@@ -48,9 +48,10 @@ message CreateWorkflowVersionOpts {
 
 enum ConcurrencyLimitStrategy {
     CANCEL_IN_PROGRESS = 0;
-    DROP_NEWEST = 1;
-    QUEUE_NEWEST = 2;
+    DROP_NEWEST = 1; // deprecated
+    QUEUE_NEWEST = 2; // deprecated
```
grutt, line 52: maybe not for this PR but do we want to support queue by implementing a convenience method for GRR with a key of "default"

internal/services/controllers/workflows/controller.go
```diff
@@ -536,19 +536,15 @@ func (wc *WorkflowsControllerImpl) runPollActiveQueuesTenant(ctx context.Context
...
 			return err
 		})
 	}
 
 	return false, errGroup.Wait()
 }
 
-func (wc *WorkflowsControllerImpl) bumpQueue(ctx context.Context, tenantId string, workflowVersionId string, groupKey *string) error {
+func (wc *WorkflowsControllerImpl) bumpQueue(ctx context.Context, tenantId string, workflowVersionId string) error {
 	workflowVersion, err := wc.repo.Workflow().GetWorkflowVersionById(ctx, tenantId, workflowVersionId)
```
grutt: ttl cache on version id?

internal/services/controllers/workflows/queue.go
```diff
@@ -746,9 +774,7 @@ func (wc *WorkflowsControllerImpl) queueByGroupRoundRobin(ctx context.Context, t
 
 func (wc *WorkflowsControllerImpl) cancelWorkflowRun(ctx context.Context, tenantId, workflowRunId string) error {
 	// cancel all running step runs
-	runningStatus := dbsqlc.StepRunStatusRUNNING
 	stepRuns, err := wc.repo.StepRun().ListStepRuns(ctx, tenantId, &repository.ListStepRunsOpts{
```
grutt, line 754: confirming: i suppose we're assuming downstream we're not canceling things in a finalized state?

pkg/repository/prisma/workflow_run.go
```diff
@@ -838,6 +839,221 @@ func (w *workflowRunEngineRepository) GetScheduledChildWorkflowRun(ctx context.C
...
+
+	tx, commit, rollback, err := sqlchelpers.PrepareTx(ctx, w.pool, w.l, 15000)
+
+	if err != nil {
+		return nil, nil, err
+	}
+
+	defer rollback()
+
+	// place a FOR UPDATE lock on queued and running workflow runs to prevent concurrent updates
```
grutt, line 855: ok just talking through this, so behavior here if a run finishes while this lock is acquired will force this cancelling state to be written but then the lock for success will be released putting it into a success state and the cancelled state will then be ignored?

pkg/repository/prisma/workflow_run.go
```diff
@@ -838,6 +839,221 @@ func (w *workflowRunEngineRepository) GetScheduledChildWorkflowRun(ctx context.C
 
 	return w.queries.GetScheduledChildWorkflowRun(ctx, w.pool, childParams)
 }
+
+func (w *workflowRunEngineRepository) PopWorkflowRunsCancelInProgress(ctx context.Context, tenantId, workflowId string, maxRuns int) ([]*dbsqlc.WorkflowRun, []*dbsqlc.WorkflowRun, error) {
+	ctx, span := telemetry.NewSpan(ctx, "queue-by-cancel-in-progress")
+	defer span.End()
+
+	tx, commit, rollback, err := sqlchelpers.PrepareTx(ctx, w.pool, w.l, 15000)
```
grutt, line 847: think thorugh: implications of FOR UPDATE + 15s timeout

pkg/repository/prisma/workflow_run.go
```diff
@@ -838,6 +839,221 @@ func (w *workflowRunEngineRepository) GetScheduledChildWorkflowRun(ctx context.C
...
+		for _, row := range toProcess {
+			if row.Status == dbsqlc.WorkflowRunStatusRUNNING {
+				runningWorkflowRuns = append(runningWorkflowRuns, row)
+			} else if row.Status == dbsqlc.WorkflowRunStatusQUEUED {
+				queuedWorkflowRuns = append(queuedWorkflowRuns, row)
+			}
+		}
+
+		// iterate over the running workflow runs and cancel them
+		toCancel := max(0, len(runningWorkflowRuns)+len(queuedWorkflowRuns)-maxRuns)
```
grutt, line 895: do we need to consider cancelling?

api-contracts/workflows/workflows.proto
abelanger5: yes

internal/services/controllers/workflows/queue.go
```diff
@@ -603,79 +603,107 @@ func (wc *WorkflowsControllerImpl) getLock(key string) *sync.Mutex {
...
+func (wc *WorkflowsControllerImpl) queueByCancelInProgress(ctx context.Context, tenantId string, workflowVersion *dbsqlc.GetWorkflowVersionForEngineRow) error {
 	ctx, span := telemetry.NewSpan(ctx, "queue-by-cancel-in-progress")
 	defer span.End()
 
-	wc.l.Info().Msgf("handling queue with strategy CANCEL_IN_PROGRESS for %s", groupKey)
+	workflowId := sqlchelpers.UUIDToStr(workflowVersion.WorkflowVersion.WorkflowId)
+	maxRuns := int(workflowVersion.ConcurrencyMaxRuns.Int32)
 
-	mutex := wc.getLock(fmt.Sprintf("%s:%s", tenantId, groupKey))
+	mutex := wc.getLock(fmt.Sprintf("%s:%s", tenantId, workflowId))
```
reillyse: Do we still need the mutex?

internal/services/controllers/workflows/queue.go
```diff
@@ -603,79 +603,107 @@ func (wc *WorkflowsControllerImpl) getLock(key string) *sync.Mutex {
...
+		workflowRunId := sqlchelpers.UUIDToStr(row.ID)
 
 		err = wc.cancelWorkflowRun(ctx, tenantId, workflowRunId)
 
 		if err != nil {
 			return fmt.Errorf("could not cancel workflow run: %w", err)
 		}
+	}
+
+	for i := range toStart {
```
reillyse, line 625: what happens if we cancel all of these and then we fail this step ? Do we lose the runs and not start any ? Should we do this in the Pop txn

internal/services/controllers/workflows/queue.go
```diff
@@ -603,79 +603,107 @@ func (wc *WorkflowsControllerImpl) getLock(key string) *sync.Mutex {
...
+}
+
+func (wc *WorkflowsControllerImpl) queueByCancelNewest(ctx context.Context, tenantId string, workflowVersion *dbsqlc.GetWorkflowVersionForEngineRow) error {
+	ctx, span := telemetry.NewSpan(ctx, "queue-by-cancel-newest")
+	defer span.End()
+
+	workflowId := sqlchelpers.UUIDToStr(workflowVersion.WorkflowVersion.WorkflowId)
+	maxRuns := int(workflowVersion.ConcurrencyMaxRuns.Int32)
+
+	mutex := wc.getLock(fmt.Sprintf("%s:%s", tenantId, workflowId))
```
reillyse: ditto with this mutex, are we protected by the locks in the DB?

internal/services/controllers/workflows/queue.go
abelanger5: Discussed offline, we may cancel things in a finalized state, but we're intentionally leaving the behavior undefined for now, as it can end up in either of the two states.

pkg/repository/prisma/workflow_run.go
abelanger5: Discussed offline

pkg/repository/prisma/workflow_run.go
abelanger5: Discussed offline

pkg/repository/prisma/workflow_run.go
abelanger5: I don't believe so, as those no longer contribute to the maxRuns count.

internal/services/controllers/workflows/controller.go
abelanger5: adding

internal/services/controllers/workflows/queue.go
abelanger5: I ended up removing the mutex, good catch

internal/services/controllers/workflows/queue.go
abelanger5: I did briefly look into moving all of this into the Pop txn, but it's much more trouble than it's worth at the moment -- this is also still a (minor) issue in the `PopWorkflowRunsByRoundRobin` logic, because we still require a message sent over the task queue. But I think we've reduced the probability of these cases significantly.

internal/services/controllers/workflows/queue.go
abelanger5: Removed

grutt (approved):

PR #1154, opened 2025-01-03
Feature  add version info

reillyse:
# Description
Add a version route to hit the API and server for version info.

Api route is 
/api/v1/version 
and it is displayed in the top right of the dashboard. (see screenshot).

The engine now has a route
/version
which is exposed on the same port as the health checks. Locally we hit

```bash
curl -v    127.0.0.1:8733/version
```

Fixes # (issue)

## Type of change

[
Screenshot 2025-01-03 at 1 42 46 PM
](url)


.github/workflows/test.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       - name: Install Protoc
         uses: arduino/setup-protoc@v3
         with:
-          version: "28.2"
+          version: "29.2"
```
grutt, line 15: will we need to update sdks?

frontend/app/src/pages/main/info/components/version-info.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { useEffect, useState } from 'react';
+// import api from '@/lib/api';
```
grutt: rm comment

frontend/app/src/pages/main/info/components/version-info.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { Spinner } from '@/components/ui/loading';
+import api from '@/lib/api';
+
+export const VersionInfo: React.FC = () => {
+  const [version, setVersion] = useState<string | null>(null);
+  const [isLoading, setIsLoading] = useState<boolean>(true);
+  const [error, setError] = useState<string | null>(null);
+
+  const fetchVersion = async () => {
```
grutt: let's use react query here
`useQuery`

frontend/app/src/pages/main/tenant-settings/overview/index.tsx
```diff
@@ -22,6 +22,7 @@ export default function TenantSettings() {
           {capitalize(tenant.name)} Overview
         </h2>
         <Separator className="my-4" />
+        <Separator className="my-4" />
```
grutt: intentional added separator?

grutt (changes requested):

grutt:
Can you please post a Screenshot of how the version renders?

frontend/app/src/pages/main/info/components/version-info.tsx
```diff
@@ -0,0 +1,46 @@
...
+    return (
+      <div className="flex items-center justify-center">
+        <Spinner />
+      </div>
+    );
+  }
+
+  if (error) {
+    return <div className="text-red-500 text-xs">{error}</div>;
+  }
```
mrkaye97: I'd think if there's an error the user can't do anything about, it'd be good to hide this component instead - WDYT?

frontend/app/src/pages/main/info/components/version-info.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  return (
+    <div className="text-sm">
+      <p>
+        <span className="text-xs"> {version}</span>
+      </p>
+    </div>
```
mrkaye97: nit: I think these styles will override each other. can probably simplify like this, unless we need the `div` and `span` for something I'm missing (untested):

```suggestion
    <p className="text-xs">{version}</p>
```

mrkaye97:
the component looks nice! nice work on this 🙌, left a couple small comments in addition to what Gabe suggested

frontend/app/src/pages/main/info/components/version-info.tsx
reillyse: If this was a consumer app I'd agree with you but I think surfacing errors and getting feedback that something is broken is a higher priority for us, especially as everyone who uses this will be technical and will be able to give us  feedback

.github/workflows/test.yml
reillyse: unsure, I'll try and build them now

reillyse: not 100% sure how to verify this, I slacked you

frontend/app/src/pages/main/info/components/version-info.tsx
mrkaye97: sounds good! I'd think we'd get notified by Sentry or some such, but maybe that's a conversation for another day 😅

mrkaye97 (approved):
Non-Golang bits LGTM!

.github/workflows/test.yml
grutt: i think we should be okay

grutt (approved):
looks good!

PR #1163, opened 2025-01-07
Fix  webhook stop healthcheck

grutt:
# Description

Fixes edge cases on #1162

### DIFF
--- a/internal/services/webhooks/webhooks.go
+++ b/internal/services/webhooks/webhooks.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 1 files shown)


internal/services/webhooks/webhooks.go
```diff
@@ -337,6 +342,11 @@ func (c *WebhooksController) run(tenantId string, webhookWorker *dbsqlc.WebhookW
 			case <-ctx.Done():
 				return
 			case <-ticker.C:
+
+				if !c.registeredWorkerIds[sqlchelpers.UUIDToStr(webhookWorker.ID)] {
```
grutt: i'm not sure if i need to cancel() here... i dont think so because we're returning

abelanger5 (approved):

PR #1165, opened 2025-01-08
fix- fix a race that happened locally

reillyse:
# Description

Got a panic about a concurrent access to a map, when I ran with trace this showed up so fixing it so it doesn't happen some other time.

Fixes # (issue)

## Type of change

## What's Changed


abelanger5:
Already resolved in #1163

PR #1170, opened 2025-01-09
add in worker logging using zerolog and logrus

reillyse:
# Description

Fixes # (issue)

## Type of change

## What's Changed

Added combined logging to the worker. Now users can bring their own logger and by calling 
```
logger := context.NewCombinedLogrusLogger(logrusLogger)
```
or 
```
logger  := context.NewCombinedZerologLogger(zerologLogger)
```

they can log to hatchet and to their local logging.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,6 +21,9 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version: "1.23"
+          cache: true
+          cache-dependency-path: go.sum
+
 
       - name: Setup pnpm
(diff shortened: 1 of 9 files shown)


grutt:
hm... i dont think i like this pattern and think what you were going for in the spec makes more sense. can we not expose a register logger or something on the worker or client to set the logger fn globally?

i don't think folks will want to init in every step

nikoksr:
@grutt I was coincidentally working on something similar. We're actively using and building on Hatchet and need a way to plug in our own logger. Submitted #1171 as a draft to quickly chip in an alternative to this here PR.

PR #1172, opened 2025-01-09
Feat: Fix switching tenants not working

mrkaye97:
Switching tenants right now in production doesn't work right away (works on refresh). I think it's because of the state handling in the `useTenantContext` relying on `useState` for derived data (from the search params) as opposed to getting and setting the search params directly

Also renamed `useTenantContext` -> `useTenant` since I think `Context` is a bit of a "reserved" thing in React (since this isn't really a Context, just a hook). And I also changed the return type of `useTenant` from `[Tenant | undefined, (string) -> void]` to an object so we can unpack that object as we need as opposed to needing e.g. indexing like `[1]` to get the setter out

### DIFF
--- a/frontend/app/src/components/molecules/analytics-provider.tsx
+++ b/frontend/app/src/components/molecules/analytics-provider.tsx
@@ -14,7 +14,7 @@ const AnalyticsProvider: React.FC<
 
   const [loaded, setLoaded] = React.useState(false);
 
-  const [tenant] = useTenantContext();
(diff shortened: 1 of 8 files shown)


mrkaye97:
@grutt I think I realized why we did the last tenant thing so I'll add that back - if you get redirected to a page without the query param set, I think this will reset you to the default. will fix that

abelanger5 (approved):

PR #1186, opened 2025-01-14
fix: hard sticky strategy with no desired worker id

abelanger5:
# Description

In the case of `stickyStrategy=HARD` and no desired worker id set, step runs can't be scheduled. This PR modifies the behavior to assign to any available worker in this case.  

We weren't seeing this behavior in most tests, because the workflows we were testing with `stickyStrategy=HARD` were child workflows spawned with `sticky=True`, where there's guaranteed to be a desired worker id. 

## Type of change

### DIFF
--- a/pkg/scheduling/v2/slot.go
+++ b/pkg/scheduling/v2/slot.go
@@ -186,10 +186,12 @@ func getRankedSlots(
 			continue
 		}
 
-		// if this is a HARD sticky strategy, it can only be assigned to the desired worker if the desired
-		// worker id is set. otherwise, it cannot be assigned.
+		// if this is a HARD sticky strategy, and there's a desired worker id, it can only be assigned to that
+		// worker. if there's no desired worker id, we assign to any worker.
 		if qi.Sticky.Valid && qi.Sticky.StickyStrategy == dbsqlc.StickyStrategyHARD {
 			if qi.DesiredWorkerId.Valid && workerId == sqlchelpers.UUIDToStr(qi.DesiredWorkerId) {
+				validSlots.addSlot(slot, 0)
+			} else if !qi.DesiredWorkerId.Valid {
 				validSlots.addSlot(slot, 0)
 			}
 


pkg/scheduling/v2/slot.go
```diff
@@ -186,11 +186,13 @@ func getRankedSlots(
...
 		}
 
-		// if this is a HARD sticky strategy, it can only be assigned to the desired worker if the desired
-		// worker id is set. otherwise, it cannot be assigned.
+		// if this is a HARD sticky strategy, and there's a desired worker id, it can only be assigned to that
+		// worker. if there's no desired worker id, we assign to any worker.
 		if qi.Sticky.Valid && qi.Sticky.StickyStrategy == dbsqlc.StickyStrategyHARD {
 			if qi.DesiredWorkerId.Valid && workerId == sqlchelpers.UUIDToStr(qi.DesiredWorkerId) {
 				validSlots.addSlot(slot, 0)
+			} else if !qi.DesiredWorkerId.Valid {
```
mrkaye97, line 194: nit (not blocking): can we use an or in the original condition? e.g.:

```suggestion
			if (qi.DesiredWorkerId.Valid && workerId == sqlchelpers.UUIDToStr(qi.DesiredWorkerId)) || !qi.DesiredWorkerId.Valid {
				validSlots.addSlot(slot, 0)
			}
```

mrkaye97 (approved):

pkg/scheduling/v2/slot.go
abelanger5: Yes, that's better, though I think it makes it more difficult to reason about -- probably something like this would work best?

```go
matchesDesired := qi.DesiredWorkerId.Valid && workerId == sqlchelpers.UUIDToStr(qi.DesiredWorkerId)
desiredNotSet := !qi.DesiredWorkerId.Valid

if matchesDesired || desiredNotSet {
  validSlots.addSlot(slot, 0)
}
```

mrkaye97: yeah, that looks good to me 👍

PR #1188, opened 2025-01-15
Feature - configurable logging backend

reillyse:
# Description

Fixes # (issue)

## Type of change

## What's Changed


cmd/hatchet-api/api/run.go
```diff
@@ -9,7 +9,7 @@ import (
 
 func Start(cf *loader.ConfigLoader, interruptCh <-chan interface{}, version string) error {
 	// init the repository
-	configCleanup, sc, err := cf.LoadServerConfig(version)
+	configCleanup, sc, err := cf.CreateServerFromConfig(version)
```
abelanger5:
```suggestion
	configCleanup, server, err := cf.CreateServerFromConfig(version)
```

nit + not important, but if we are doing a quick pass for consistency this would match the other files

internal/services/ingestor/server.go
```diff
@@ -228,7 +228,13 @@ func (i *IngestorImpl) PutLog(ctx context.Context, req *contracts.PutLogRequest)
 		return nil, status.Errorf(codes.InvalidArgument, "Invalid request: %s", apiErrors.String())
 	}
 
-	_, err := i.logRepository.PutLog(ctx, tenantId, opts)
+	// Make sure we are writing to a step run owned by this tenant
+	_, err := i.stepRunRepository.GetStepRunForEngine(ctx, tenantId, req.StepRunId)
```
abelanger5: This seems like it would add a significant read load to the database for tens of log lines per second. All we're trying to do is map a `stepRunId <> tenantId`, so perhaps we use an `*lru.Cache` like we're using in other places to get some load off the database?

pkg/repository/logs.go
```diff
@@ -55,9 +58,11 @@ type ListLogsResult struct {
 type LogsAPIRepository interface {
 	// ListLogLines returns a list of log lines for a given step run.
 	ListLogLines(tenantId string, opts *ListLogsOpts) (*ListLogsResult, error)
+	WithAdditionalConfig(validator.Validator, *zerolog.Logger) LogsAPIRepository
```
abelanger5, line 61: Why are these returning a `LogsAPIRepository`? Aren't validator and logger written directly to the instance of `LogsAPIRepository`?

pkg/repository/prisma/log.go
```diff
@@ -139,6 +139,17 @@ type logEngineRepository struct {
 	l       *zerolog.Logger
 }
 
+// Used as hook a hook to allow for additional configuration to be passed to the repository if it is instantiated a different way
+func (le *logAPIRepository) WithAdditionalConfig(v validator.Validator, l *zerolog.Logger) repository.LogsAPIRepository {
+	panic("not implemented in this repo")
```
abelanger5, line 144: Should we be panicking here or just returning `le`?

abelanger5 (changes requested):
A few questions but overall looks good. I like the naming changes, they make things more clear.

pkg/repository/prisma/log.go
reillyse: I think we panic because people are doing it wrong - worst thing we can do is have a situation where people expect the codebase to behave with one set of functionality and it behaves in a different way in effect we would be opening ourselves up to a whole other set of failure modes. This way if people are using the code wrong it errors out straight away at init time.

pkg/repository/logs.go
reillyse: It just makes it explicit when we assign and allows other patterns for when we want to do the actual overwriting. E.g. we can chain them and assign at the end. 

e.g. 

`logsAPIRepo = opts.logsAPIRepository.WithAdditionalConfig(opts.v, opts.l).WithSomethingElse(opts.v, opts.l)`

internal/services/ingestor/server.go
reillyse: I like this - makes sense to reduce the load on the DB as much as possible

cmd/hatchet-api/api/run.go
reillyse: agreed

abelanger5: @reillyse sorry for the confusion, this does lead to a compile error with just this change

internal/services/ingestor/server.go
```diff
@@ -229,12 +229,20 @@ func (i *IngestorImpl) PutLog(ctx context.Context, req *contracts.PutLogRequest)
 	}
 
 	// Make sure we are writing to a step run owned by this tenant
-	_, err := i.stepRunRepository.GetStepRunForEngine(ctx, tenantId, req.StepRunId)
-	if err != nil {
-		return nil, err
+	if t, ok := i.steprunTenantLookupCache.Get(opts.StepRunId); ok {
+		if t != tenantId {
+			return nil, status.Errorf(codes.PermissionDenied, "Permission denied: step run does not belong to tenant")
```
abelanger5, line 234: Minor nit, but shouldn't block merge -- I believe this will return a 500-level error to the user, which would be really confusing (though I'm not sure how they'd end up in this case)

abelanger5 (approved):

internal/services/ingestor/server.go
reillyse: yea it's really just a sanity check, getting into this case would be hard perhaps if the stepRun was "" and we pulled out the wrong tenantId then we know we have an error.

PR #1193, opened 2025-01-16
Feat: Propagate Error Through Context

mrkaye97:
Propagating errors in upstream step runs through the context by attaching a new field to the `input` that's `errors`, which is a dictionary of key-value pairs where the keys are the step names.

Really not sure if this is implemented in a reasonable way, but it does seem to work - would love some feedback here :) 

Example input from running the on failure example in Go:

```
{
  "input": {},
  "errors": {
    "step-one": "\"test on failure\""
  },
  "parents": {},
  "overrides": {},
  "user_data": null,
  "triggered_by": "manual"
}
```

### DIFF
--- a/examples/on-failure/main.go
+++ b/examples/on-failure/main.go
@@ -27,7 +27,7 @@ func StepOne(ctx worker.HatchetContext) (result *stepOneOutput, err error) {
 func OnFailure(ctx worker.HatchetContext) (result *stepOneOutput, err error) {
 	// run cleanup code or notifications here
 
-	// you can access the error from the failed step(s) like this
+	// 👀 you can access the error from the failed step(s) like this
 	fmt.Println(ctx.StepRunErrors())
 
 	return &stepOneOutput{
--- a/examples/on-failure/main.go
+++ b/examples/on-failure/main.go
@@ -26,6 +26,10 @@ func StepOne(ctx worker.HatchetContext) (result *stepOneOutput, err error) {
 
 func OnFailure(ctx worker.HatchetContext) (result *stepOneOutput, err error) {
 	// run cleanup code or notifications here
+
+	// you can access the error from the failed step(s) like this
+	fmt.Println(ctx.StepRunErrors())
+
 	return &stepOneOutput{
 		Message: "Failure!",
 	}, nil
--- a/internal/datautils/job_data.go
+++ b/internal/datautils/job_data.go
@@ -26,4 +26,7 @@ type StepRunData struct {
 
 	// overrides set from the playground
 	Overrides map[string]interface{} `json:"overrides"`
+
+	// errors in upstream steps (only used in on-failure step)
+	Errors map[string]string `json:"errors,omitempty"`
 }
--- a/internal/datautils/job_data.go
+++ b/internal/datautils/job_data.go
@@ -28,5 +28,5 @@ type StepRunData struct {
 	Overrides map[string]interface{} `json:"overrides"`
 
 	// errors in upstream steps (only used in on-failure step)
-	Errors map[string]string `json:"errors,omitempty"`
+	StepRunErrors map[string]string `json:"errors,omitempty"`
 }
--- a/internal/datautils/job_data.go
+++ b/internal/datautils/job_data.go
@@ -28,5 +28,5 @@ type StepRunData struct {
 	Overrides map[string]interface{} `json:"overrides"`
 
 	// errors in upstream steps (only used in on-failure step)
-	StepRunErrors map[string]string `json:"errors,omitempty"`
+	StepRunErrors map[string]string `json:"step_run_errors,omitempty"`
 }
--- a/internal/services/controllers/jobs/controller.go
+++ b/internal/services/controllers/jobs/controller.go
@@ -750,6 +750,18 @@ func (ec *JobsControllerImpl) queueStepRun(ctx context.Context, tenantId, stepId
 	// If the step run input is not set, then we should set it. This will be set upstream if we've rerun
 	// the step run manually with new inputs. It will not be set when the step is automatically queued.
 	if in := data.Input; len(in) == 0 || string(in) == "{}" {
+		upstreamErrors := make(map[string]string)
+
+		if stepRun.JobKind == dbsqlc.JobKindONFAILURE {
+			failedStepErrors, err := ec.repo.WorkflowRun().GetUpstreamErrorsForOnFailureStep(ctx, stepRunId)
+
+			if err == nil {
+				for _, failedStepError := range failedStepErrors {
+					upstreamErrors[failedStepError.StepReadableId.String] = failedStepError.Error.String
+				}
+			}
+		}
+
 		lookupDataBytes := data.JobRunLookupData
 
 		if lookupDataBytes != nil {
@@ -779,6 +791,7 @@ func (ec *JobsControllerImpl) queueStepRun(ctx context.Context, tenantId, stepId
 				Parents:     lookupData.Steps,
 				UserData:    userData,
 				Overrides:   map[string]interface{}{},
+				Errors:      upstreamErrors,
 			}
 
 			inputDataBytes, err = json.Marshal(inputData)
--- a/internal/services/controllers/jobs/controller.go
+++ b/internal/services/controllers/jobs/controller.go
@@ -786,12 +786,12 @@ func (ec *JobsControllerImpl) queueStepRun(ctx context.Context, tenantId, stepId
 
 			// input data is the triggering event data and any parent step data
 			inputData := datautils.StepRunData{
-				Input:       lookupData.Input,
-				TriggeredBy: lookupData.TriggeredBy,
-				Parents:     lookupData.Steps,
-				UserData:    userData,
-				Overrides:   map[string]interface{}{},
-				Errors:      upstreamErrors,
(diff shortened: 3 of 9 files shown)


internal/datautils/job_data.go
```diff
@@ -26,4 +26,7 @@ type StepRunData struct {
 
 	// overrides set from the playground
 	Overrides map[string]interface{} `json:"overrides"`
+
+	// errors in upstream steps (only used in on-failure step)
+	Errors map[string]string `json:"errors,omitempty"`
```
mrkaye97: Is this a reasonable way to go?

pkg/repository/prisma/dbsqlc/workflow_runs.sql
```diff
@@ -1551,3 +1551,26 @@ WHERE
...
+FROM "WorkflowRun" wr
+JOIN "JobRun" jr ON wr."id" = jr."workflowRunId"
+JOIN "StepRun" sr ON jr."id" = sr."jobRunId"
+JOIN "Step" s ON sr."stepId" = s."id"
+JOIN "Job" j ON jr."jobId" = j.id
+WHERE
+    wr."id" = (SELECT "id" FROM workflow_run)
+    -- Don't include the on-failure step in the step runs listed
+    AND j."kind" <> 'ON_FAILURE'
+    AND sr."error" IS NOT NULL;
```
mrkaye97: would love ideas about how to make this less redundant!

mrkaye97:
Is there a way to test this with the other SDKs? I was trying to run a Python worker against my local setup but having issues:

```
[ERROR] 🪓 -- 2025-01-16 14:21:25,080 - Could not put workflow: <_InactiveRpcError of RPC that terminated with:
        status = StatusCode.UNAVAILABLE
        details = "failed to connect to all addresses; last error: UNKNOWN: ipv4:127.0.0.1:443: Ssl handshake failed (TSI_PROTOCOL_FAILURE): SSL_ERROR_SSL: error:10000438:SSL routines:OPENSSL_internal:TLSV1_ALERT_INTERNAL_ERROR: Invalid certificate verification context"
        debug_error_string = "UNKNOWN:Error received from peer  {created_time:"2025-01-16T14:21:25.079756-05:00", grpc_status:14, grpc_message:"failed to connect to all addresses; last error: UNKNOWN: ipv4:127.0.0.1:443: Ssl handshake failed (TSI_PROTOCOL_FAILURE): SSL_ERROR_SSL: error:10000438:SSL routines:OPENSSL_internal:TLSV1_ALERT_INTERNAL_ERROR: Invalid certificate verification context"}"
>
```

reillyse:
I need a bit more context here to understand what we are trying to do.

mrkaye97:
> I need a bit more context here to understand what we are trying to do.

Ah yeah, sorry about that 😅 

A couple of people have been asking for if they can access the error that was thrown upstream in their on_failure steps, presumably to do different things depending on the error. So I was trying to hook up the Python SDK to my local dev setup to see if I could e.g. run the on_failure example there and see how the error propagated through / how it'd need to be parsed, but not sure how to do that (was having issues - error up there ^^)

pkg/repository/prisma/dbsqlc/workflow_runs.sql
abelanger5: You shouldn't need the join on `Step` and `Job` since an onFailureJob is guaranteed to only have a single step run, and only one onFailureJob is defined per WorkflowRun. So something like this would work:

```sql 
SELECT
    sr."id" AS "stepRunId",
    s."readableId" AS "stepReadableId",
    sr."error"
FROM "WorkflowRun" wr
JOIN "JobRun" jr ON wr."id" = jr."workflowRunId"
JOIN "StepRun" sr ON jr."id" = sr."jobRunId"
WHERE 
    sr."stepRunId" != @onFailureStepRunId::uuid
    AND sr."error" IS NOT NULL
```

abelanger5:
> Is there a way to test this with the other SDKs? I was trying to run a Python worker against my local setup but having issues:
> 
> ```
> [ERROR] 🪓 -- 2025-01-16 14:21:25,080 - Could not put workflow: <_InactiveRpcError of RPC that terminated with:
>         status = StatusCode.UNAVAILABLE
>         details = "failed to connect to all addresses; last error: UNKNOWN: ipv4:127.0.0.1:443: Ssl handshake failed (TSI_PROTOCOL_FAILURE): SSL_ERROR_SSL: error:10000438:SSL routines:OPENSSL_internal:TLSV1_ALERT_INTERNAL_ERROR: Invalid certificate verification context"
>         debug_error_string = "UNKNOWN:Error received from peer  {created_time:"2025-01-16T14:21:25.079756-05:00", grpc_status:14, grpc_message:"failed to connect to all addresses; last error: UNKNOWN: ipv4:127.0.0.1:443: Ssl handshake failed (TSI_PROTOCOL_FAILURE): SSL_ERROR_SSL: error:10000438:SSL routines:OPENSSL_internal:TLSV1_ALERT_INTERNAL_ERROR: Invalid certificate verification context"}"
> >
> ```

@mrkaye97 I usually have to trust the Caddy root cert:

```
HATCHET_CLIENT_TLS_ROOT_CA_FILE="/Users/belanger/Library/Application Support/Caddy/pki/authorities/local/root.crt"
```

Also, ensure that `HATCHET_CLIENT_TLS_STRATEGY` **is not set**

internal/datautils/job_data.go
abelanger5: I think so considering the comment -- `Errors` could be construed with internal errors from the engine, so perhaps `StepRunErrors` would make sense, but I don't have a strong opinion

internal/services/controllers/jobs/controller.go
```diff
@@ -750,6 +750,18 @@ func (ec *JobsControllerImpl) queueStepRun(ctx context.Context, tenantId, stepId
 	// If the step run input is not set, then we should set it. This will be set upstream if we've rerun
 	// the step run manually with new inputs. It will not be set when the step is automatically queued.
 	if in := data.Input; len(in) == 0 || string(in) == "{}" {
```
abelanger5, line 752: We should double check that a workflow run which gets replayed and throws new errors adds the new errors to the input data, not the old errors. I believe we clear all step run inputs on a replay, but would be good to double check!

internal/datautils/job_data.go
mrkaye97: This sounds like a better name to me, I'll go with this - think it'll be easier to understand on the client side 👍

pkg/repository/prisma/dbsqlc/workflow_runs.sql
mrkaye97: Ah nice, thanks! Do I still need to join to the `Step` to get the `readableId`? or is there a way around that?

abelanger5: Ah missed that, yeah there's no way around that

mrkaye97: Sounds good, will fix the other one 👍

internal/services/controllers/jobs/controller.go
mrkaye97: So, it looks like this maybe kinda works? Is it possible that on replay the error is updated in the db with the new one? It looks like what's happening is the first run throws an error and that's passed along to the `input` correctly, then when I replay the workflow, _all_ of the past run inputs get overwritten with the new error. Is that how it's supposed to happen?

Tested this by printing out a timestamp in the error message, and then on replay that timestamp would update in the original run (in addition to being updated in the replayed run)

abelanger5: Yes, that's the correct behavior, if you're saying both the step run and the on failure step run are getting the new timestamp.

mrkaye97: yep 👍

abelanger5 (approved):

examples/on-failure/main.go
```diff
@@ -26,6 +26,10 @@ func StepOne(ctx worker.HatchetContext) (result *stepOneOutput, err error) {
 
 func OnFailure(ctx worker.HatchetContext) (result *stepOneOutput, err error) {
 	// run cleanup code or notifications here
+
+	// you can access the error from the failed step(s) like this
```
abelanger5:
```suggestion
	// 👀 you can access the error from the failed step(s) like this
```

pkg/repository/prisma/dbsqlc/workflow_runs.sql
```diff
@@ -1551,3 +1551,23 @@ WHERE
...
+    FROM "WorkflowRun" wr
+    JOIN "JobRun" jr ON wr."id" = jr."workflowRunId"
+    JOIN "StepRun" sr ON jr."id" = sr."jobRunId"
+    WHERE sr."id" = @onFailureStepRunId::uuid
+)
+SELECT
+    sr."id" AS "stepRunId",
+    s."readableId" AS "stepReadableId",
+    sr."error" AS "stepRunError"
+FROM "WorkflowRun" wr
```
abelanger5: Very minor nit -- making this `FROM workflow_run wr` would prevent you from having to do `WHERE wr."id" = (SELECT "id" FROM workflow_run)`

abelanger5 (approved):

PR #1194, opened 2025-01-16
Fix: Change which exceptions retry

mrkaye97:
Fixing which types of exceptions retry in the Go SDK - basically what it says on the tin. `ResourceExhausted` shouldn't be retried, but these other ones probably should

Tested by running the `simple` example when the engine wasn't running on local. 

On main:

```
matt@Matts-MacBook-Pro hatchet % (cd examples/simple && go run main.go)
{"level":"debug","service":"client","time":"2025-01-16T15:09:33.072857-05:00","message":"connecting to grpc.dev.hatchet-tools.com:443 with TLS server name grpc.dev.hatchet-tools.com"}
panic: error registering workflow: could not create workflow simple: rpc error: code = Unavailable desc = unexpected HTTP status code received from server: 502 (Bad Gateway); malformed header: missing HTTP content-type

goroutine 1 [running]:
main.main()
        /Users/matt/Documents/GitHub/hatchet/examples/simple/main.go:36 +0xdc
exit status 2
```

On this branch:

```
matt@Matts-MacBook-Pro hatchet % (cd examples/simple && go run main.go)
{"level":"debug","service":"client","time":"2025-01-16T15:08:20.533618-05:00","message":"connecting to grpc.dev.hatchet-tools.com:443 with TLS server name grpc.dev.hatchet-tools.com"}
{"level":"debug","service":"client","time":"2025-01-16T15:08:25.954563-05:00","message":"grpc_retry attempt: 1, backoff for rpc error: code = Unavailable desc = unexpected HTTP status code received from server: 502 (Bad Gateway); malformed header: missing HTTP content-type"}
{"level":"debug","service":"client","time":"2025-01-16T15:08:35.414001-05:00","message":"grpc_retry attempt: 2, backoff for rpc error: code = Unavailable desc = unexpected HTTP status code received from server: 502 (Bad Gateway); malformed header: missing HTTP content-type"}
{"level":"debug","service":"client","time":"2025-01-16T15:08:54.188588-05:00","message":"grpc_retry attempt: 3, backoff for rpc error: code = Unavailable desc = unexpected HTTP status code received from server: 502 (Bad Gateway); malformed header: missing HTTP content-type"}
```

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -269,6 +269,7 @@ func newFromOpts(opts *ClientOpts) (Client, error) {
 
 	if !opts.noGrpcRetry {
 		retryOnCodes := []codes.Code{
+			codes.ResourceExhausted,
 			codes.DeadlineExceeded,
 			codes.FailedPrecondition,
 			codes.Internal,
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -269,7 +269,10 @@ func newFromOpts(opts *ClientOpts) (Client, error) {
 
 	if !opts.noGrpcRetry {
 		retryOnCodes := []codes.Code{
-			codes.ResourceExhausted,
+			codes.DeadlineExceeded,
+			codes.FailedPrecondition,
+			codes.Internal,
+			codes.Unavailable,
 		}
 
 		retryOpts := []grpc_retry.CallOption{


reillyse (commented):
If I recall the original idea for adding the retries was because of rate limiting issues - Resource Exhausted is what we return in that case. Do we have a retry mechanism for rate limit problems? 

To test against rate limit problems you can set a low rate limit locally and spawn a bunch of workflows in a go func and you'll quickly hit the rate limit.

see 
pkg/config/server/server.go

for the env vars - SERVER_GRPC_RATE_LIMIT is the relevant one.

mrkaye97:
> If I recall the original idea for adding the retries was because of rate limiting issues - Resource Exhausted is what we return in that case. Do we have a retry mechanism for rate limit problems?
> 
> To test against rate limit problems you can set a low rate limit locally and spawn a bunch of workflows in a go func and you'll quickly hit the rate limit.
> 
> see pkg/config/server/server.go
> 
> for the env vars - SERVER_GRPC_RATE_LIMIT is the relevant one.

Thanks! Maybe we could just add back the `ResourceExhausted` check then here? I'd seen that in the past for things like being over your daily limits, which we probably don't want to retry - WDYT? Add back that case to the list?

reillyse:
also , just thinking back to a previous PR
https://github.com/hatchet-dev/hatchet/pull/1046
I recall there was an issue with the timing of the retries - I think I made this change because the retry was longer than the heartbeat and so the heartbeats were coming in (from unavailable) after they were stale.

I think we should probably drop the default retry timeout so we are well within the heartbeat and maybe shorten the number of default retries. We want to catch network errors (for this Unavailable retry) so after that there probably isn't a lot of sense in retrying.

reillyse:
> > If I recall the original idea for adding the retries was because of rate limiting issues - Resource Exhausted is what we return in that case. Do we have a retry mechanism for rate limit problems?
> > To test against rate limit problems you can set a low rate limit locally and spawn a bunch of workflows in a go func and you'll quickly hit the rate limit.
> > see pkg/config/server/server.go
> > for the env vars - SERVER_GRPC_RATE_LIMIT is the relevant one.
> 
> Thanks! Maybe we could just add back the `ResourceExhausted` check then here? I'd seen that in the past for things like being over your daily limits, which we probably don't want to retry - WDYT? Add back that case to the list?

I think we probably need to handle this differently for each endpoint. GRPC ratelimits probably need to be retried, others not so much (like you said daily limits are probably just problematic to retry). 

Perhaps adding it back in and reducing the default retries makes sense (I think configuring a per gRPC function retry is a little bit of a heavier lift and not worth it). So I think add ResourceExhausted and maybe only retry one time. If we are retrying a rate limit or other limit more than that we are probably just stacking more load on top of ourself (like if the rate limit hasn't gotten better after a couple of retries it's probably gotten way worse).

abelanger5 (approved):

PR #1195, opened 2025-01-16
feat(go-sdk): adds preset labels on workers for autoscaling

abelanger5:
# Description

Adds an option to the Go SDK to parse the environment variable `HATCHET_AUTOSCALING_TARGET` and set it as a label on workers. This is required to get autoscaling to work properly, as it gives us a mechanism to identify a group of workers that belong to the same autoscaling group. 

There's a decent amount of annoying config-drilling down to the dispatcher, but fixing that would require a rewrite of the client that we're not going to tackle right now. 

## Type of change

### DIFF
--- a/pkg/client/dispatcher.go
+++ b/pkg/client/dispatcher.go
@@ -160,15 +160,18 @@ type dispatcherClientImpl struct {
 	v validator.Validator
 
 	ctx *contextLoader
+
+	presetWorkerLabels map[string]string
 }
 
-func newDispatcher(conn *grpc.ClientConn, opts *sharedClientOpts) DispatcherClient {
+func newDispatcher(conn *grpc.ClientConn, opts *sharedClientOpts, presetWorkerLabels map[string]string) DispatcherClient {
 	return &dispatcherClientImpl{
-		client:   dispatchercontracts.NewDispatcherClient(conn),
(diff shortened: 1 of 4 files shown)


pkg/client/dispatcher.go
```diff
@@ -160,15 +160,18 @@ type dispatcherClientImpl struct {
 	v validator.Validator
 
 	ctx *contextLoader
+
+	presetWorkerLabels map[string]string
 }
 
-func newDispatcher(conn *grpc.ClientConn, opts *sharedClientOpts) DispatcherClient {
+func newDispatcher(conn *grpc.ClientConn, opts *sharedClientOpts, presetWorkerLabels map[string]string) DispatcherClient {
```
reillyse, line 167: so I suppose sharedClientOpts is shared among other client services so we shouldn't add the labels directly in there. Should we create a dispatcher opts and add it there just to keep this interface clean.

reillyse (approved):
Looks good, one little comment.

pkg/client/dispatcher.go
abelanger5: I don't feel it makes a huge difference so unless you feel strongly, I'm going to leave it for now (lots of Go client improvements for the future though)

grutt (approved):

PR #1197, opened 2025-01-18
clean up rabbit mq session stuff, add a quick ack and error processin…

reillyse:
…g for AddMessage

# Description

Fixes # (issue)

## Type of change

## What's Changed

### DIFF
--- a/internal/msgqueue/rabbitmq/rabbitmq.go
+++ b/internal/msgqueue/rabbitmq/rabbitmq.go
@@ -59,7 +59,8 @@ type MessageQueueImpl struct {
 
 	l *zerolog.Logger
 
-	ready bool
+	readyMux sync.Mutex
+	ready    bool
 
 	disableTenantExchangePubs bool
 
@@ -67,8 +68,27 @@ type MessageQueueImpl struct {
 	tenantIdCache *lru.Cache[string, bool]
 }
 
-func (t *MessageQueueImpl) IsReady() bool {
+func (t *MessageQueueImpl) setNotReady() {
+	t.safeSetReady(false)
+}
+
+func (t *MessageQueueImpl) setReady() {
+	t.safeSetReady(true)
+}
+
+func (t *MessageQueueImpl) safeCheckReady() bool {
+	t.readyMux.Lock()
+	defer t.readyMux.Unlock()
 	return t.ready
+}
+func (t *MessageQueueImpl) safeSetReady(ready bool) {
+	t.readyMux.Lock()
+	defer t.readyMux.Unlock()
+	t.ready = ready
+}
+
+func (t *MessageQueueImpl) IsReady() bool {
+	return t.safeCheckReady()
 }
 
 type MessageQueueImplOpt func(*MessageQueueImplOpts)
@@ -676,11 +696,12 @@ func (t *MessageQueueImpl) redial(ctx context.Context, l *zerolog.Logger, pool *
 			newSession, err := t.establishSessionWithRetry(ctx, l, pool)
 			if err != nil {
 				l.Error().Msg("failed to establish session after retries")
-				t.ready = false
+				t.setNotReady()
+
 				return
 			}
 
-			t.ready = true
+			t.setReady()
 			t.monitorSession(ctx, l, newSession)
 
 			select {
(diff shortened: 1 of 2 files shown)


reillyse:
Quick first PR on dealing with the errors back from AddMessage. This is my first time dipping my toes into the RabbitMQ stuff so I'm not 100% sure yet why we need to do such complicated session channel wizardry - but Chesterton's fence I assume someone reasonably talented spent a lot of time creating it so it's probably there for a reason.  I've tested this by killing and restarting rabbitmq a bunch of times, let me know if you think there is a more in depth way to test - I figure this is going to break spectacularly or just work though.

reillyse:
Also, I've tried to make this change have as low a surface area as possible. I don't feel comfortable jumping in and rewriting the entire mq layer yet without getting an idea for how it's actually working, but I think this is going to fix a class of errors (basically where we were pretending everything was ok but we were just dropping the message)

abelanger5:
> Quick first PR on dealing with the errors back from AddMessage. This is my first time dipping my toes into the RabbitMQ stuff so I'm not 100% sure yet why we need to do such complicated session channel wizardry - but Chesterton's fence I assume someone reasonably talented spent a lot of time creating it so it's probably there for a reason. I've tested this by killing and restarting rabbitmq a bunch of times, let me know if you think there is a more in depth way to test - I figure this is going to break spectacularly or just work though.

Hey @reillyse to give some context -- the core logic/session stuff is based on the example here: https://github.com/rabbitmq/amqp091-go/blob/main/_examples/pubsub/pubsub.go 

At some point later we added `puddle.Pool` on top to reuse connections. 

There's definitely a cleaner way to write it, but it hasn't been a priority.

internal/msgqueue/rabbitmq/rabbitmq.go
```diff
@@ -215,12 +221,47 @@ func (t *MessageQueueImpl) AddMessage(ctx context.Context, q msgqueue.Queue, msg
...
 		Message: msg,
 		q:       q,
+		ackChan: ackC,
 	}
 
-	return nil
+	select {
+	case t.msgs <- &msgWithQueue:
+	case <-time.After(5 * time.Second):
+		return fmt.Errorf("timeout sending message to %s queue", q.Name())
```
abelanger5, line 263: Are we worried about a potential memory leak on `ackC`? I see the channel doesn't get closed, and just want to make sure we've covered all paths.

internal/msgqueue/rabbitmq/rabbitmq.go
```diff
@@ -370,6 +411,7 @@ func (t *MessageQueueImpl) startPublishing() func() error {
 
 						// retry failed delivery on the next session
 						if err != nil {
+							msg.ackChan <- ack{e: &err}
```
abelanger5: Considering this is a blocking write, is there any scenario where the channel is full (perhaps if we've retried a bunch of times?)

If this does block, this would leak a goroutine.

internal/msgqueue/rabbitmq/rabbitmq.go
```diff
@@ -646,6 +672,39 @@ func (t *MessageQueueImpl) redial(ctx context.Context, l *zerolog.Logger, pool *
...
+
+func (t *MessageQueueImpl) monitorSession(ctx context.Context, l *zerolog.Logger, newSession session) {
+	closeCh := newSession.Connection.NotifyClose(make(chan *amqp.Error, 1))
+
+	go func() {
+		select {
+		case <-ctx.Done():
+		case <-closeCh:
+			l.Warn().Msg("session closed, marking as not ready")
+			t.ready = false
```
abelanger5: Re-reading this -- while this is probably in line with our previous logic, it doesn't look correct. First, I think there's a potential data race on `t.ready` since it can be written to from multiple goroutines, so if we were going to go this route we should write a mutex-protect `setReady` method. 

But it also seems that we're setting `t.ready = false` potentially directly after setting `t.ready = true`, which means that even a recovered instance could still fail a health check and get restarted. 

I think the logic we're going for here is: is there at least one connection to the RabbitMQ instance which is healthy? But I don't think this gets us there.

abelanger5 (changes requested):

internal/msgqueue/rabbitmq/rabbitmq.go
reillyse: I've wrapped the ready to make it race resistant - tbh though I don't think the ready is that important unless I've missed something.

internal/msgqueue/rabbitmq/rabbitmq.go
reillyse: made it non-blocking

abelanger5 (approved):

PR #1199, opened 2025-01-19
fix: correct the number of rows selected in Workflow Runs table, afte…

joaopjoaquim:
…r refresh of web page

# Description

Fixes # After changing the number of rows that can be seen in the Workflow Runs page and then refreshing the page, the URL argument "pageSize" and the "Rows per page" value remain the same, but the number of rows actually displaying goes back to the standard 50 rows. This fix corrects this issue, keeping the variation made before the refresh

## Type of change

## What's Changed

### DIFF
--- a/frontend/app/src/pages/main/workflow-runs/components/workflow-runs-table.tsx
+++ b/frontend/app/src/pages/main/workflow-runs/components/workflow-runs-table.tsx
@@ -11,11 +11,11 @@ import {
 import { useMutation, useQuery } from '@tanstack/react-query';
 import invariant from 'tiny-invariant';
 import api, {
+  queries,
   ReplayWorkflowRunsRequest,
   WorkflowRunOrderByDirection,
   WorkflowRunOrderByField,
   WorkflowRunStatus,
-  queries,
 } from '@/lib/api';
 import { TenantContextType } from '@/lib/outlet';
 import { useOutletContext, useSearchParams } from 'react-router-dom';
@@ -181,6 +181,13 @@ export function WorkflowRunsTable({
     return { pageIndex, pageSize };
   });
 
+  const setPageSize = (newPageSize: number) => {
+    setPagination((prev) => ({
+      ...prev,
+      pageSize: newPageSize,
+    }));
+  };
+
   useEffect(() => {
     const newSearchParams = new URLSearchParams(searchParams);
     if (sorting.length) {
@@ -216,8 +223,6 @@ export function WorkflowRunsTable({
     setSearchParams,
     searchParams,
   ]);
-
-  const [pageSize, setPageSize] = useState<number>(50);
 
   const offset = useMemo(() => {
     if (!pagination) {
@@ -295,7 +300,7 @@ export function WorkflowRunsTable({
   const listWorkflowRunsQuery = useQuery({
     ...queries.workflowRuns.list(tenant.metadata.id, {
       offset,
-      limit: pageSize,
+      limit: pagination.pageSize,
       statuses,
       workflowId: workflow,
       parentWorkflowRunId,


abelanger5:
Hey @joaopjoaquim, really appreciate the PR! Would you mind running a quick `pnpm run lint:fix` inside of the `./frontend/app` folder?

joaopjoaquim:
hi @abelanger5 , sure! do you want me to push the modifications that were created after the command?

abelanger5:
> hi @abelanger5 , sure! do you want me to push the modifications that were created after the command?

yep!

joaopjoaquim:
> > hi @abelanger5 , sure! do you want me to push the modifications that were created after the command?
> 
> yep!

its done and repushed

frontend/app/src/pages/main/workflow-runs/components/workflow-runs-table.tsx
```diff
@@ -217,7 +217,9 @@ export function WorkflowRunsTable({
     searchParams,
   ]);
 
-  const [pageSize, setPageSize] = useState<number>(50);
+  const [pageSize, setPageSize] = useState<number>(() => {
+    return Number(searchParams.get('pageSize') || pagination);
```
abelanger5: Hey @joaopjoaquim this doesn't look quite correct. `pagination` is an object with a `.pageSize` parameter. 

I think it would be better if we removed the `pageSize` state parameter and instead modified the `limit:` call and modified how we're calling `setPageSize` to only use the `pagination` parameter as the source of truth. What do you think?

abelanger5 (changes requested):

frontend/app/src/pages/main/workflow-runs/components/workflow-runs-table.tsx
joaopjoaquim: you're right! I just repushed the code with the modifications that you suggested. I got rid of the pageSize state and created a new function  called setPageSize that will change directly the attribute inside pagination

abelanger5: thanks @joaopjoaquim! we'll take a look and hopefully get this merged today 💯

frontend/app/src/pages/main/workflow-runs/components/workflow-runs-table.tsx
```diff
@@ -684,6 +696,11 @@ export function WorkflowRunsTable({
   );
 }
 
+/**
+ * This component fetches workflow run metrics,
+ * displays a loading skeleton while data is being fetched,
+ * and finally renders a zoomable bar chart with the metrics data once it's available.
+ */
```
mrkaye97: nit: thanks for adding this! but I'm not sure it's in the scope of this PR, and would prefer we removed it - we definitely have documentation work to do, so we'll take an action item here to improve this in the future :)

frontend/app/src/pages/main/workflow-runs/components/workflow-runs-table.tsx
```diff
@@ -72,6 +75,10 @@ export interface WorkflowRunsTableProps {
   showMetrics?: boolean;
 }
 
+/**
+ * define a range of time, based on the current date and time and the given range
+ * @param timeRange range given
+ */
```
mrkaye97: same here re: docs

frontend/app/src/pages/main/workflow-runs/components/workflow-runs-table.tsx
```diff
@@ -60,6 +60,9 @@ import { DateTimePicker } from '@/components/molecules/time-picker/date-time-pic
 import useCloudApiMeta from '@/pages/auth/hooks/use-cloud-api-meta';
 import { AdditionalMetadataClick } from '../../events/components/additional-metadata';
 
+/**
+ * this interface structures the configuration settings for displaying a table of workflow runs
+ */
```
mrkaye97: and here

mrkaye97 (approved):
This LGTM! Left a couple of comments on documentation - we'll work on that more later! Thanks for the contribution :) 

Feel free to dismiss Alexander's review request

frontend/app/src/pages/main/workflow-runs/components/workflow-runs-table.tsx
joaopjoaquim: I took off the comments, then

PR #1208, opened 2025-01-22
Fix: Empty string workflow run status

mrkaye97:
Bug fix for https://discord.com/channels/1088927970518909068/1331450119861829664

TL;DR is that the workflow run status seems to get serialized as an empty string instead of `nil` since it's initialized as a string, which then makes Pydantic fail downstream since it wants an enum or None


grutt (approved):
Good find!

reillyse (approved):
looks good

PR #1209, opened 2025-01-22
Feat: Sample Sentries in the Engine

mrkaye97:
Setting up sampling for Sentries in the engine so we don't flood our quota. Configurable via an env var which should default to 1.0

### DIFF
--- a/pkg/config/server/server.go
+++ b/pkg/config/server/server.go
@@ -218,6 +218,9 @@ type SentryConfigFile struct {
 
 	// Environment is the environment that the instance is running in
 	Environment string `mapstructure:"environment" json:"environment,omitempty" default:"development"`
+
+	// Sample rate is the rate at which to sample events. Default is 1.0 to sample all events.
+	SampleRate float64 `mapstructure:"sampleRate" json:"sampleRate,omitempty" default:"1.0"`
 }
 
 type AnalyticsConfigFile struct {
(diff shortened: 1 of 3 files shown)


pkg/config/server/server.go
```diff
@@ -218,6 +218,9 @@ type SentryConfigFile struct {
 
 	// Environment is the environment that the instance is running in
 	Environment string `mapstructure:"environment" json:"environment,omitempty" default:"development"`
+
+	// Sample rate is the rate at which to sample events. Default is 1.0 to sample all events.
+	SampleRate float64 `mapstructure:"sampleRate" json:"sampleRate,omitempty" default:"1.0"`
```
grutt, line 223: should we default to lower? i dont have a strong opinion but cc @abelanger5 since we'll need to set this on all instances if we want to lower sampling

grutt (approved):

pkg/config/server/server.go
mrkaye97: my thought was that if people are self-hosting and using sentry now and we default to lower, they'll see our sampling logic kick in without them knowing to set this env var. but not 100% sure this is the right path 🤷

grutt: good call. https://docs.hatchet.run/self-hosting/configuration-options#alerting-configuration

we are publishing sentry, i thought we were not.

let's remember to set this in the next release

mrkaye97: sounds good! do I need to update this on the docs site? I didn't think to look there

grutt: nice to have

mrkaye97: ticket: https://linear.app/hatchet-dev/issue/HAT-323/documentation-on-sentry-sampling

PR #1215, opened 2025-01-24
simple OOM Monitor - we gracefully quit before we hit our OOM limit

reillyse:
# Description

## What's Changed

- [ ] Added an OOM monitor that we can use to gracefully exit when we are close to our real memory limit.

### DIFF
--- a/internal/services/oom/monitor.go
+++ b/internal/services/oom/monitor.go
@@ -1,0 +1,125 @@
+package oommonitor
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"runtime"
+	"syscall"
+	"time"
+
+	"github.com/rs/zerolog"
+)
+
+type MonitorOpts struct {
+	ThresholdBytes uint64
+	Signal         string
+	CheckInterval  time.Duration
+	L              *zerolog.Logger
+}
+
+type Monitor struct {
+	ThresholdBytes uint64
+	Signal         os.Signal
+	CheckInterval  time.Duration
+	l              *zerolog.Logger
+}
+
+func NewMonitor(opts MonitorOpts) (*Monitor, error) {
+	if opts.ThresholdBytes <= 0 {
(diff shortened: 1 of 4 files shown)


reillyse:
SERVER_OOM_ENABLED=true SERVER_OOM_THRESHOLD_BYTES=27574104 SERVER_OOM_SIGNAL="SIGTERM" SERVER_OOM_CHECK_INTERVAL="20s"  task start-dev

internal/services/oom/monitor.go
```diff
@@ -0,0 +1,125 @@
...
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				m.l.Info().Msg("OOM monitor stopped")
+				return
+			default:
+				runtime.GC()
+				var memStats runtime.MemStats
+				runtime.ReadMemStats(&memStats)
```
abelanger5, line 85: This concerns me -- I don't think we can necessarily get a true estimate of memory usage from this. More more information: https://www.datadoghq.com/blog/go-memory-metrics/

Has there been any investigation into whether the memory reported by `ReadMemStats` matches the system memory (divided by two, according to the DataDog article) when we're running in a Docker image?

From what I understand, not only may the count not be accurate, but we also have to multiply threshold bytes by 2 to get the actual threshold.

reillyse: this is 100% a tuning thing, we need to experiment and find out what the correct limit is, memory is always very difficult to calculate on linux boxes, even top and the various memory tools do not give an accurate picture, shared libraries are shared etc. The strategy here is to set a conservative limit and increase it. What we have now is OOMs which are a disaster and very difficult to mitigate - all of our code can just stop running right in the middle of executing which is the worst possible failure. We are running redundant engines right? So restarting an engine should be a normal part of our operations and not lead to any downtime.

PR #1228, opened 2025-01-28
Feat  canceled state

grutt:
# Description
Adds a Cancelled state for workflow runs and UI affordances to filter by Cancelled state

## Type of change

## What's Changed

- [x] Cancelling a workflow run, job, or step will mark the workflow run as cancelled
- [x] The cancelled state is queryable in list and metrics end points
- [x] Previously these runs would be marked as failed


docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: "3.8"
 services:
   postgres:
     image: postgres:15.6
-    command: postgres -c 'max_connections=200'
+    command: postgres -c 'max_connections=500'
```
grutt, line 5: this may be a symptom of a different problem, but i need to keep bumping this value

frontend/app/src/components/ui/badge.tsx
```diff
@@ -14,6 +14,8 @@ const badgeVariants = cva(
         destructive:
           'border-transparent bg-destructive text-destructive-foreground shadow hover:bg-destructive/80',
         outline: 'text-foreground',
+        outlineDestructive:
+          'border border-destructive rounded-sm px-1 font-normal text-red-800 dark:text-red-300 bg-transparent',
```
grutt, line 18: image
image

mrkaye97: ui nit: do you think we can make the status buttons grow so they're all the same width?

grutt:
@abelanger5 going to open a ticket for similar changes in cloud for timescale db for now

frontend/app/src/components/ui/badge.tsx
```diff
@@ -14,6 +14,8 @@ const badgeVariants = cva(
         destructive:
           'border-transparent bg-destructive text-destructive-foreground shadow hover:bg-destructive/80',
         outline: 'text-foreground',
+        outlineDestructive:
```
mrkaye97, line 17: nit: I think it's more tailwind-thonic to use `kebab-case` e.g. `outline-destructive` for things like this

api-contracts/openapi/components/schemas/event.yaml
```diff
@@ -115,6 +115,10 @@ EventWorkflowRunSummary:
       type: integer
       format: int64
       description: The number of failed runs.
+    cancelled:
+      type: integer
+      format: int64
+      description: The number of cancelled runs.
```
mrkaye97, line 121: nit but something I'm noticing in the code as I'm working on it now - this will never be null, right? since we have the `COALESCE`, so I imagine that it makes sense to mark this as `required` so we don't need to handle the null case in the TS

mrkaye97 (approved):
LGTM, left some small comments on things but I think we should think about those things more in the v2 work - lots to do 😄

api-contracts/openapi/components/schemas/event.yaml
grutt: yes, but this would be a breaking change for sdks so leaving for now. something we should implement in v2

mrkaye97: good call 👍

frontend/app/src/components/ui/badge.tsx
grutt: personal pref to not have kebab in ts keys... not strongly opinionated here

grutt: >ui nit: do you think we can make the status buttons grow so they're all the same width? 
github is broken since there are two commends on the same line i think.

no, this is just the badge... lets tweak the general layout in a separate pr!

mrkaye97: works for me!

PR #1230, opened 2025-01-28
Fix: Tenant switching, part II

mrkaye97:
# Description

Fixing two more small issues in switching tenants:

1. Adding a `useEffect` hook to the `useTenant` hook to always set the tenant as a query param (mostly helps with debugging)
2. Ensuring that we correctly switch to the newly created tenant when a new tenant is created by calling `setTenant` explicitly

### DIFF
--- a/frontend/app/src/lib/atoms.ts
+++ b/frontend/app/src/lib/atoms.ts
@@ -1,7 +1,7 @@
 import { atom, useAtom } from 'jotai';
 import { Tenant, queries } from './api';
 import { useSearchParams } from 'react-router-dom';
-import { useCallback, useMemo } from 'react';
+import { useCallback, useEffect, useMemo } from 'react';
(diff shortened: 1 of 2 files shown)


frontend/app/src/lib/atoms.ts
```diff
@@ -88,16 +100,38 @@ export function useTenant(): TenantContext {
...
+      tenant: undefined,
+      tenantId: undefined,
+      setTenant,
+    };
+  }
+
   return {
-    tenant: currTenant || computedCurrTenant,
+    tenant,
+    tenantId: tenant.metadata.id,
```
mrkaye97, line 134: this is just a helper to shorten needing to use `tenant.metadata.id` every time we want to grab the tenant

grutt (approved):

abelanger5 (approved):

Issue #264, opened 2024-03-13
bug(engine): grpc_broadcast_address hatchet-engine


Jak2k:
This gets logged when starting the worker script:

```
🪓 [INFO/HatchetClient] Initialized HatchetClient
(node:21382) [DEP0128] DeprecationWarning: Invalid 'main' field in '/home/jak2k/Projects/news-analysis/downloader/node_modules/@hatchet-dev/typescript-sdk/package.json' of 'dist/index.js'. Please either fix that or report it to the module author
(Use `node --trace-deprecation ...` to show where the warning was created)
🪓 [ERROR/Admin] Error: /WorkflowService/PutWorkflow UNAVAILABLE: Name resolution failed for target dns:hatchet-engine:7070
🪓 [ERROR/Admin] Error: /WorkflowService/PutWorkflow UNAVAILABLE: Name resolution failed for target dns:hatchet-engine:7070
🪓 [ERROR/Admin] Error: /WorkflowService/PutWorkflow UNAVAILABLE: Name resolution failed for target dns:hatchet-engine:7070
🪓 [ERROR/Admin] Error: /WorkflowService/PutWorkflow UNAVAILABLE: Name resolution failed for target dns:hatchet-engine:7070
🪓 [ERROR/Admin] Error: /WorkflowService/PutWorkflow UNAVAILABLE: Name resolution failed for target dns:hatchet-engine:7070
/home/jak2k/Projects/news-analysis/downloader/node_modules/.pnpm/@hatchet-dev+typescript-sdk@0.1.27/node_modules/@hatchet-dev/typescript-sdk/clients/worker/worker.js:86
                throw new hatchet_error_1.default(`Could not register workflow: ${e.message}`);
                      ^

HatchetError: Could not register workflow: /WorkflowService/PutWorkflow UNAVAILABLE: Name resolution failed for target dns:hatchet-engine:7070
    at Worker.<anonymous> (/home/jak2k/Projects/news-analysis/downloader/node_modules/.pnpm/@hatchet-dev+typescript-sdk@0.1.27/node_modules/@hatchet-dev/typescript-sdk/clients/worker/worker.js:86:23)
    at Generator.throw (<anonymous>)
    at rejected (/home/jak2k/Projects/news-analysis/downloader/node_modules/.pnpm/@hatchet-dev+typescript-sdk@0.1.27/node_modules/@hatchet-dev/typescript-sdk/clients/worker/worker.js:6:65)

Node.js v20.10.0
```

[...]

PR #1262, opened 2025-02-11
Update docker-compose.mdx

KonradAdamczyk:
Change host.docker.internal to hatchet-engine. Suggest adding depends_on to worker.  host.docker.internal is not production ready. Without depends_on worker will fail randomly depending of starting order.

# Description

Probably Fixes # ([issue](https://github.com/hatchet-dev/hatchet/issues/264))

## Type of change


mrkaye97 (approved):
thanks for this fix!

PR #1287, opened 2025-02-25
fix: scheduled workflows query

abelanger5:
# Description

(bug introduced in `v0.54.8`)

The scheduled workflows query was incorrectly sorting by `"version" DESC`, which is essentially a random ordering. `version` is a user-friendly display name for the version, but `order` is the source of truth for workflow versioning. 

This was causing scheduled workflows to essentially execute a random version of the workflow. 

## Type of change

### DIFF
--- a/pkg/repository/prisma/dbsqlc/tickers.sql
+++ b/pkg/repository/prisma/dbsqlc/tickers.sql
@@ -104,6 +104,8 @@ WITH latest_workflow_versions AS (
         MAX("order") as max_order
     FROM
         "WorkflowVersion"
+    WHERE
+        "deletedAt" IS NULL
     GROUP BY "workflowId"
 ),
 active_cron_schedules AS (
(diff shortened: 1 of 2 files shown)


pkg/repository/prisma/dbsqlc/tickers.sql
```diff
@@ -158,14 +165,7 @@ WITH not_run_scheduled_workflows AS (
     JOIN
         "Workflow" AS workflow ON workflow."id" = versions."workflowId"
     JOIN
-        (
-            -- Subquery to get the latest version per workflow
-            SELECT DISTINCT ON ("workflowId")
-                "id", "workflowId", "version"
-            FROM "WorkflowVersion"
-            WHERE "deletedAt" IS NULL
```
mrkaye97, line 166: do we want to keep this `deletedAt` filter?

mrkaye97 (approved):

pkg/repository/prisma/dbsqlc/tickers.sql
abelanger5: nice catch, just pushed

PR #1293, opened 2025-02-26
fix: migration issues on `pgx.ErrNoRows`

abelanger5:
# Description

Unlike the go prisma client, a `pgx.ErrNoRows` error from `sqlc` doesn't necessarily mean that the return value of the queried row is `nil`, instead it's a pointer to a zero-value of that struct. 

This fixes instances where the casing assumed that the queried row value was nil. 

## Type of change


abelanger5:
Wrong fork.

PR #1312, opened 2025-03-09
Feat: Python V1

mrkaye97:
[Easier to review without whitespace](https://github.com/hatchet-dev/hatchet/pull/1312/files?diff=unified&w=1) - lots of indentation changes in the examples

Finalizing the external-facing API for the new SDK here. Basically, you'll have the workflow declaration via`hatchet.workflow` (which is a `Workflow` factory), and then that workflow object has task and on_failure_task methods that let you transform functions into tasks and register them to the workflow.

I also refactored the giant top-level `workflows.py` into `runnables` with `task`, `workflow`, and `types`

Other misc changes / highlights:

* Added a workflows kwarg to hatchet.worker so you can pass a list of workflows to register instead of needing to register each individually (see examples)
* Also added worker.register_workflows (plural) to register many workflows at once
* Updated all the examples
* Got rid of the new_client, new_client_raw, from_environment, etc. stuff from before - it should be easy enough to instantiate ClientConfig directly and pass whatever variables you want, especially now that it's using pydantic-settings
* Set up `contextvars` to propagate `workflow_run_id`, `worker_id`, and `step_run_id` when calling `run_workflow` variants and removed `spawn_*`
* Changed the signature of a Hatchet task from `Workflow, Context -> Any` to `TWorkflowInput, Context -> JSONSerializable | BaseModel` and removed the need to call `get_workflow_input` everywhere to access the input (since now it's passed directly, and parsed correctly)
* `parents` of a task are now a list of `Task` objects, not a list of strings, which allows for better type checking / linting of those arrays
* Similar to ^, we can now have a `task_output` method on the `Context` which supersedes the old `step_output`, and is type-safe - it accepts a task and returns a typed object from the return value of the task (see DAG example for how this works)

TODO:
- [x] Try to get `spawn_index` counters working correctly
- [x] Try to clean up the `.pop()` stuff that we have buried in the internals (maybe use a separate model that we validate?)
- [x] Contextvars for `run_workflows` and other variants
- [x] Look more into step outputs that are non-objects
- [ ] Fix all the docs


sdks/python/examples/bulk_fanout/bulk_trigger.py
```diff
@@ -1,33 +1,30 @@
...
-                }
-            ),
-        )
-        for i in range(20)
-    ]
-
-    workflowRunRefs = hatchet.admin.run_workflows(
-        workflow_runs,
+async def main() -> None:
+    workflow_run_refs = bulk_parent_wf.run_many(
```
grutt, line 11: this is a bit confusing to me

sdks/python/examples/fanout/stream.py
```diff
@@ -21,9 +22,8 @@ async def main() -> None:
     # This key gets propagated to all child workflows
     # and can have an arbitrary property name.
 
-    hatchet.admin.run_workflow(
-        "Parent",
-        {"n": 2},
+    parent_wf.run(
```
grutt, line 25: https://github.com/hatchet-dev/hatchet/pull/1316/files sdks/typescript/src/v1/workflow.ts

I'm playing with the idea of simplifying this further by exposing 2 endpoints

1. enqueue which behaves likes the current run -- returns the ref where you can get the id or await a result

2. run which awaits the result directly

what are your thoughts here?

sdks/python/examples/cancellation/test_cancellation.py
```diff
@@ -1,13 +1,15 @@
...
+from examples.cancellation.worker import wf
 from hatchet_sdk import Hatchet, Worker
 
 
 # requires scope module or higher for shared event loop
 @pytest.mark.asyncio(scope="session")
 @pytest.mark.parametrize("worker", ["cancellation"], indirect=True)
 async def test_run(hatchet: Hatchet, worker: Worker) -> None:
-    run = hatchet.admin.run_workflow("CancelWorkflow", {})
+    run = wf.run()
```
grutt: nit: do we want to encourage naming all the wokflows wf>

sdks/python/examples/cancellation/worker.py
```diff
@@ -1,30 +1,26 @@
 import asyncio
 
-from hatchet_sdk import BaseWorkflow, Context, Hatchet
+from hatchet_sdk import Context, EmptyModel, Hatchet
 
 hatchet = Hatchet(debug=True)
 
-wf = hatchet.declare_workflow(on_events=["user:create"])
+wf = hatchet.workflow(name="CancelWorkflow", on_events=["user:create"])
```
grutt: nit: maybe its time to drop the default on_events from all the examples

grutt: we'll need to swap tests to .run

sdks/python/examples/dag/worker.py
```diff
@@ -1,61 +1,60 @@
...
-        return {
-            "step4": "step4",
-        }
+wf = hatchet.workflow(
+    name="DAGWorkflow", on_events=["dag:create"], schedule_timeout="10m"
+)
+
+
+@wf.task(timeout="5s")
+def step1(input: EmptyModel, context: Context) -> StepOutput:
```
grutt: nit: context -> ctx for sdk consistency?

sdks/python/examples/delayed/worker.py
```diff
@@ -11,41 +11,37 @@ class PrinterInput(BaseModel):
...
+    print(f"printed at \t {now.strftime('%H:%M:%S')}")
+    print(f"message \t {input.message}")
 
 
 def main() -> None:
-    worker = hatchet.worker("delayed-worker", max_runs=4)
-    worker.register_workflow(PrintSchedule())
-    worker.register_workflow(PrintPrinter())
+    worker = hatchet.worker(
+        "delayed-worker", max_runs=4, workflows=[print_schedule_wf, print_printer_wf]
```
grutt: id like to rename max_runs -> slots, thoughts?

sdks/python/examples/manual_trigger/worker.py
```diff
@@ -2,53 +2,50 @@
...
 
-wf = hatchet.declare_workflow(on_events=["man:create"])
+wf = hatchet.workflow(name="ManualTriggerWorkflow", on_events=["man:create"])
 
 
-class ManualTriggerWorkflow(BaseWorkflow):
-    config = wf.config
+@wf.task()
+def step1(input: EmptyModel, context: Context) -> dict[str, str]:
+    res = context.playground("res", "HELLO")
```
grutt: we're deprecating playground

grutt (commented):
This is looking great! Getting into minor cleanup territory now ✌️

sdks/python/examples/bulk_fanout/bulk_trigger.py
mrkaye97: the `run_many`? is it the naming? I wasn't sure what to call this 😅 open to ideas here

sdks/python/examples/fanout/stream.py
mrkaye97: I think this is pretty reasonable - seems like a common pattern to want to get the results of child runs before returning, and this feels similar to `FireAndForget` / `FireAndWait`. WDYT?

sdks/python/examples/cancellation/test_cancellation.py
mrkaye97: good call, I'll change these to be better named

sdks/python/examples/cancellation/worker.py
mrkaye97: yeah will do, was thinking about this too 👍 it should make running different examples simpler too b/c there won't be a bunch of phantom runs queued for workflows that aren't being registered anymore which has been annoying in the current setup

sdks/python/examples/delayed/worker.py
mrkaye97: sounds good to me!

sdks/python/examples/manual_trigger/worker.py
mrkaye97: sounds good, would gladly remove this 😅

frontend/docs/pages/sdks/python-sdk/asyncio.mdx
```diff
@@ -1,110 +1,60 @@
...
-        "data": data,
-    }
-...
+    return 42
 ```
 
-### More Resources for AsyncIO Development
+### More Resources for working with `asyncio`
+
+## TODO: FastAPI links here
```
mrkaye97: note to self: fill this in

frontend/docs/pages/sdks/python-sdk/asyncio.mdx
```diff
@@ -1,110 +1,60 @@
...
-Hatchet's Python SDK makes heavy use of AsyncIO under the hood to manage persistent connections,
-listeners, the runners for each workflow, and many other things! AsyncIO is a form of concurrency
-that makes use of [cooperative multitasking](https://en.wikipedia.org/wiki/Cooperative_multitasking).
-This means that when blocking non-async-safe functions are added to the mix, it can cause issues and
-degrade the performance of the SDK. While using async methods is optional with Hatchet, using async
-methods incorrectly can cause issues and degrade the performance of the SDK.
+Hatchet's Python SDK, similarly to other popular libraries like FastAPI, Langchain, etc., makes heavy use of `asyncio`, and recommends that you do as well!
 
-### Using the `sync_to_async` Decorator
+> To learn the basics of `asyncio`, check out [this introduction](LINK HERE)
```
mrkaye97: note to self: fill this in

sdks/python/hatchet_sdk/utils/aio_utils.py
```diff
@@ -1,83 +1,4 @@
 import asyncio
-import inspect
-from concurrent.futures import Executor
-from functools import partial, wraps
-from typing import Any
-
-
-## TODO: Stricter typing here
-def sync_to_async(func: Any) -> Any:
```
mrkaye97, line 9: for discussion - I don't think we should encourage this. IMO, people should use `asyncio.to_thread` directly and we should have guidance on that (see asyncio docs), but I don't think we should expose an automagic method that will "just make sync things run async"

one other option is to whittle this down to basically just being a `to_thread` wrapper, but at that point, might as well just use `to_thread` I think

Taskfile.yaml
```diff
@@ -179,3 +180,7 @@ tasks:
   pre-commit-run:
     cmds:
       - pre-commit run --all-files || pre-commit run --all-files
+  docs:
+    cmds:
+      - npx serve -l 4001 &
+      - cd frontend/docs/ && pnpm run dev
```
mrkaye97, line 186: important: this should make it a lot easier to run the docs and do the github emoji thingy going forward

grutt: nice, this is better than the go fileserver

sdks/python/examples/cron/programatic-async.py
```diff
@@ -6,7 +6,7 @@
 async def create_cron() -> None:
     # ❓ Create
     cron_trigger = await hatchet.cron.aio_create(
-        workflow_name="simple-cron-workflow",
+        workflow_name="CronWorkflow",
```
grutt: should we import and define this from the workflow dfn?

sdks/python/examples/cron/programatic-sync.py
```diff
@@ -4,7 +4,7 @@
 
 # ❓ Create
 cron_trigger = hatchet.cron.create(
-    workflow_name="simple-cron-workflow",
+    workflow_name="CronWorkflow",
```
grutt: same as previous

sdks/python/examples/worker_existing_loop/worker.py
```diff
@@ -1,27 +1,27 @@
...
     try:
-        workflow = MyWorkflow()
-        worker = hatchet.worker("test-worker", max_runs=1)
-        worker.register_workflow(workflow)
+        worker = hatchet.worker(
+            "test-worker", slots=1, workflows=[existing_loop_worker]
+        )
         worker.start()
 
         ref = hatchet.admin.run_workflow("MyWorkflow", input={})
```
grutt: change to .run?

grutt (approved):
Looks great! Minor nits that we can continue to address over time, nothing should be breaking 🥳

sdks/python/examples/worker_existing_loop/worker.py
mrkaye97: good catch, thanks

sdks/python/examples/cron/programatic-async.py
mrkaye97: yep good call, will fix this

PR #1315, opened 2025-03-10
Feat  UI tweaks

grutt:
# Description

UI Tweaks for upgrade routes

### DIFF
--- a/frontend/app/src/components/molecules/nav-bar/banner.tsx
+++ b/frontend/app/src/components/molecules/nav-bar/banner.tsx
@@ -1,0 +1,57 @@
+import React from 'react';
+import { Button } from '@/components/ui/button';
+
+export interface BannerProps {
+  message: React.ReactNode;
+  type?: 'info' | 'warning' | 'success' | 'error';
+  actionText?: string;
+  onAction?: () => void;
+  showBanner?: boolean;
+}
+
+export const Banner: React.FC<BannerProps> = ({
+  message,
+  type = 'info',
+  actionText,
+  onAction,
+  showBanner = true,
+}) => {
+  if (!showBanner) {
+    return null;
+  }
+
+  const getBgColor = () => {
+    switch (type) {
+      case 'warning':
+        return 'bg-amber-50 dark:bg-amber-900/20 text-amber-800 dark:text-amber-200';
+      case 'success':
+        return 'bg-green-50 dark:bg-green-900/20 text-green-800 dark:text-green-200';
+      case 'error':
+        return 'bg-red-50 dark:bg-red-900/20 text-red-800 dark:text-red-200';
+      default:
+        return 'bg-blue-50 dark:bg-blue-900/20 text-blue-800 dark:text-blue-200';
+    }
+  };
+
+  return (
+    <div className={`w-full py-2 px-4 h-12 ${getBgColor()}`}>
+      <div className="flex items-center justify-between">
+        <div className="flex flex-1 items-center">
+          <p className="text-sm font-medium">{message}</p>
+        </div>
+        <div className="flex items-center gap-2">
+          {actionText && onAction && (
+            <Button
+              variant="outline"
+              size="sm"
+              onClick={onAction}
+              className="text-xs font-medium hover:bg-opacity-20"
+            >
+              {actionText}
+            </Button>
+          )}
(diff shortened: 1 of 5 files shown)


grutt:
I think we should consider a more aggressive revision for some of these things at some point... but just want to get these working changes in

frontend/app/src/components/molecules/nav-bar/banner.tsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import { Button } from '@/components/ui/button';
+
+export interface BannerProps {
+  message: React.ReactNode;
```
mrkaye97: qq: can we make this a string?

frontend/app/src/components/molecules/nav-bar/banner.tsx
```diff
@@ -0,0 +1,57 @@
...
+  type?: 'info' | 'warning' | 'success' | 'error';
+  actionText?: string;
+  onAction?: () => void;
+  showBanner?: boolean;
+}
+
+export const Banner: React.FC<BannerProps> = ({
```
mrkaye97, line 11: stylistic nit: I've been defining components like this:

```typescript
export const Banner = ({foo, bar}: BannerProps) => {}
```

I don't care which we end up going for, but I think we should try to be consistent. WDYT? I kinda like the way I've been doing it because it's a little more terse, but I'm fine either way

frontend/app/src/components/molecules/nav-bar/banner.tsx
```diff
@@ -0,0 +1,57 @@
...
+  message,
+  type = 'info',
+  actionText,
+  onAction,
+  showBanner = true,
+}) => {
+  if (!showBanner) {
+    return null;
+  }
```
mrkaye97: I think this should happen at the level of the parent component - e.g. if we're going to pass `showBanner={someBooleanState}` we should instead do `!someBooleanState && <Banner ...>`

frontend/app/src/components/molecules/nav-bar/nav-bar.tsx
```diff
@@ -105,7 +106,7 @@ function HelpDropdown() {
   );
 }
 
-function AccountDropdown({ user }: MainNavProps) {
+function AccountDropdown({ user, tenant }: MainNavProps) {
```
mrkaye97: I've been preferring to use `useTenant()` inside the components instead of drilling as a prop

frontend/app/src/components/molecules/nav-bar/nav-bar.tsx
```diff
@@ -215,29 +222,89 @@ function VersionUpgradeButton() {
 
 interface MainNavProps {
   user: User;
+  tenant?: Tenant;
+  bannerProps?: BannerProps;
```
mrkaye97: unused props I think?

mrkaye97 (commented):
left some comments a couple things - thanks for cleaning this up!

frontend/app/src/components/molecules/nav-bar/banner.tsx
grutt: i think we may want to include an icon or additional links here (i.e. link to migration docs)

frontend/app/src/components/molecules/nav-bar/banner.tsx
grutt: agreed, was playing with making the banner hideable but removed this

frontend/app/src/components/molecules/nav-bar/banner.tsx
grutt: personally prefer React.FC<BannerProps> since it adds some additional type safety if needed later

frontend/app/src/components/molecules/nav-bar/banner.tsx
mrkaye97: cool, sounds good

frontend/app/src/components/molecules/nav-bar/banner.tsx
mrkaye97: ah nice, vs. just `JSX.Element`? sounds good to me

mrkaye97 (approved):

PR #1316, opened 2025-03-10
wip: typescript v1

grutt:
# Description

Improves Typescript SDK Dev Ex with a core focus on:
1. Simplicity and clarity of interface
2. Input and output type safety
3. Consistent naming x-ui and sibling sdks

## Type of change

## What's Changed

- [x] Improves type safety through factory-based workflow definition 
- [x] Exposes clear Run, Enqueue, Scheduling, and CRON methods
- [x] Exposes clear REST API


sdks/typescript/src/util/workflow-run-ref.ts
```diff
@@ -71,6 +71,15 @@ export default class WorkflowRunRef<T> {
     return this.client.stream(workflowRunId);
   }
 
+  // TODO test this..
```
grutt: need to javadoc this file and test this method

sdks/python/examples/concurrency_limit_rr_load/worker.py
```diff
@@ -0,0 +1,67 @@
+import time
+import random
```
mrkaye97: this looks like it snuck in unintentionally

sdks/python/examples/concurrency_limit_rr_load/event.py
```diff
@@ -0,0 +1,12 @@
+from hatchet_sdk import Hatchet
```
mrkaye97: this one too. Ithink but not 100% sure

sdks/typescript/src/examples/webhooks.e2e.ts
```diff
@@ -1,14 +1,14 @@
...
+import { Workflow, V0Worker } from '..';
 import sleep from '../util/sleep';
-import Hatchet from '../sdk';
 
 const port = 8369;
 
 describe('webhooks', () => {
-  let hatchet: Hatchet;
-  let worker: Worker;
+  let hatchet: Hatchet; // TODO upgrade to v1
```
mrkaye97: do we need to do this one here?

sdks/typescript/src/step.ts
```diff
@@ -245,15 +257,56 @@ export class Context<T, K = {}> {
...
+  }
+
   /**
    * Spawns multiple workflows.
    *
    * @param workflows an array of objects containing the workflow name, input data, and options for each workflow
    * @returns a list of references to the spawned workflow runs
+   * @deprecated use bulkEnqueueChildren or bulkRunChildren instead
    */
-  spawnWorkflows<Q = JsonValue, P = JsonValue>(
```
mrkaye97, line 254: nit: on the Python side I've been calling this `Record<string, any>` `JSONSerializableMapping` to make it clear what we're expecting - I kinda like that type of naming (i.e. `JsonValue`)

sdks/typescript/src/util/workflow-run-ref.ts
mrkaye97: bumping this icymi!

sdks/typescript/src/v1/client/client.ts
```diff
@@ -0,0 +1,151 @@
...
+  /**
+   * Triggers a workflow run without waiting for completion.
+   * @template T - The input type for the workflow
+   * @template K - The return type of the workflow
+   * @param workflow - The workflow to run, either as a Workflow instance or workflow name
+   * @param input - The input data for the workflow
+   * @param options - Configuration options for the workflow run
+   * @returns A WorkflowRunRef containing the run ID and methods to interact with the run
+   */
+  enqueue<T extends Record<string, any> = any, K = any>(
```
mrkaye97, line 78: I think we should agree on naming for these across SDKs - want to call them `enqueue` and `run` in both cases? could there be any confusion there? I kinda like `run_and_wait` and `run_and_forget`, but open to ideas 🤷

sdks/typescript/src/v1/client/worker.ts
```diff
@@ -0,0 +1,123 @@
...
+  /** Maximum number of concurrent runs on this worker */
+  slots?: number;
+  /** Array of workflows to register */
+  workflows?: Workflow<any, any>[] | V0Workflow[];
+  /** Worker labels for affinity-based assignment */
+  labels?: WorkerLabels;
+  /** Whether to handle kill signals */
+  handleKill?: boolean;
+  /** @deprecated Use slots instead */
+  maxRuns?: number;
```
mrkaye97, line 22: I've been hard-removing these params instead of deprecating - WDYT of that? I figure that since this is a major version, we're within our bounds to do that, and it might be a while before we have another opportunity

sdks/typescript/src/v1/examples/child_workflows/workflow.ts
```diff
@@ -0,0 +1,57 @@
...
+
+type ChildInput = {
+  N: number;
+};
+
+type ChildOutput = {
+  value: {
+    Value: number;
+  };
+};
```
mrkaye97, line 11: nice, this looks really similar to what I have on the python side 👍

sdks/typescript/src/v1/task.ts
```diff
@@ -0,0 +1,59 @@
...
+   */
+  name: string;
+
+  /**
+   * The function to execute when the task runs.
+   * @param input The input data for the workflow invocation.
+   * @param ctx The execution context for the task.
+   * @returns The result of the task execution.
+   */
+  fn: (input: T, ctx: Context<T>) => K | Promise<K>;
```
mrkaye97, line 20: 🙏 awesome

sdks/typescript/src/v1/workflow.ts
```diff
@@ -0,0 +1,252 @@
...
+   */
+  sticky?: WorkflowV0['sticky'];
+  /**
+   * Optional schedule timeout for the workflow.
+   */
+  scheduleTimeout?: WorkflowV0['scheduleTimeout'];
+  /**
+   * Optional on config for the workflow.
+   */
+  on?: WorkflowV0['on'];
```
mrkaye97, line 60: another small one - I can change the `on_crons`, `on_events`, etc. to this - WDYT? seems like another easy way to have the APIs be closer together. that'd be good anyways since we could pull some more validation into Pydantic

mrkaye97 (commented):
left some comments / questions - looks great! nice work on this 🥳

sdks/python/examples/concurrency_limit_rr_load/worker.py
mrkaye97: bumping this one

sdks/typescript/src/examples/webhooks.e2e.ts
grutt: good catch, had to expose webhook stuff on v1

sdks/typescript/src/v1/client/client.ts
grutt: i personally like run and enqueue since they're less verbose. do you think people will be confused?

i dont like _and_wait here because both requests are async and will technically need to be awaited

sdks/typescript/src/v1/client/worker.ts
grutt: typescript respects the @deprecated tag in most IDEs much better than python in my experience... personally I'm a bigger fan of backwards compatibility and eventually removing features.

by having the v0 client in ts my thoughts were we can eventually rip and replace the v0 client with a long term v1 client and start printing warnings before we fully remove the feature

sdks/typescript/src/v1/workflow.ts
grutt: can we register both crons and events on the same workflow? part of this single "on" design gave us type saftey to only allow one or the other

mrkaye97: ah I'm not sure! I assumed we could

grutt: lets test this but leave it for now

sdks/typescript/src/v1/client/client.ts
mrkaye97: I think it's a little confusing but I don't have really strong feelings on naming 🤷 agreed on the `_and_wait` being confusing since `await` is its own thing. could use celery terminology with `run` and `delay` 😂 

I think `enqueue` and `run` is good - I'll do this on the Python side too so we can be consistent

mrkaye97 (approved):

PR #1366, opened 2025-03-18
fix(ts):  context cleanup

grutt:
# Description
Fixes potential leak for context and future objects not being deleted


.github/workflows/app.yml
```diff
@@ -1,12 +1,26 @@
...
-      - 'sdks/**'
 
 jobs:
+  check_path:
+    runs-on: ubuntu-latest
+    outputs:
+      sdk_only: ${{ steps.filter.outputs.sdk_only }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dorny/paths-filter@v2
```
abelanger5: Is there no way we can just use a simple bash script here instead? Given recent security incidents with third-party GHA scripts ([example](https://orca.security/resources/blog/github-action-tj-actions-changed-files-compromised)) I'd prefer to avoid relying on this

abelanger5 (changes requested):

.github/workflows/app.yml
grutt: good call

abelanger5 (approved):

PR #1371, opened 2025-03-19
Feat  ts v1 tweaks

grutt:
# Description
Minor improvements


mrkaye97:
I think this needs a merge - I tried changing the base but I think we're still behind

grutt:
Looks like its up to date for me?

mrkaye97:
> Looks like its up to date for me?

Do we want the SQL changes and such in here?

PR #1372, opened 2025-03-19
feat: add back tenant alerting to v1

abelanger5:
# Description

Adds tenant alerting back to v1. We poll for failed runs from the OLAP controller, which seems appropriate since it tracks statuses of each workflow run. 

Also fixes an issue from #1368 where replays will cause workflows to get stuck. 

## Type of change


cmd/hatchet-migrate/migrate/migrations/20250319192344_v1_0_5.sql
```diff
@@ -0,0 +1,307 @@
...
+$$
+BEGIN
+    WITH new_retry_rows AS (
+        SELECT
+            nt.id,
+            nt.inserted_at,
+            nt.retry_count,
+            nt.tenant_id,
+            -- Convert the retry_after based on min(retry_backoff_factor ^ retry_count, retry_max_backoff)
+            NOW() + (LEAST(nt.retry_max_backoff, POWER(nt.retry_backoff_factor, nt.app_retry_count)) * interval '1 second') AS retry_after
```
grutt, line 14: should this be 1 second or 1 ms?

grutt (approved):

cmd/hatchet-migrate/migrate/migrations/20250319192344_v1_0_5.sql
abelanger5: 1 second

PR #1383, opened 2025-03-21
Fix: Missing fields in API

mrkaye97:
# Description

Fixing a couple missing fields causing Pydantic errors on the Python side. I guess nothing in the FE uses these 🤷

## Type of change

### DIFF
--- a/api/v1/server/oas/transformers/v1/tasks.go
+++ b/api/v1/server/oas/transformers/v1/tasks.go
@@ -53,6 +53,9 @@ func ToTaskSummary(task *sqlcv1.PopulateTaskRunDataRow) gen.V1TaskSummary {
 			CreatedAt: task.InsertedAt.Time,
 			UpdatedAt: task.InsertedAt.Time,
 		},
+		Input:              jsonToMap(task.Input),
+		Output:             jsonToMap(task.Output),
+		Type:               gen.V1WorkflowTypeTASK,
 		DisplayName:        task.DisplayName,
 		Duration:           durationPtr,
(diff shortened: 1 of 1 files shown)


api/v1/server/oas/transformers/v1/tasks.go
```diff
@@ -53,6 +53,9 @@ func ToTaskSummary(task *sqlcv1.PopulateTaskRunDataRow) gen.V1TaskSummary {
 			CreatedAt: task.InsertedAt.Time,
 			UpdatedAt: task.InsertedAt.Time,
 		},
+		Input:              jsonToMap(task.Input),
```
abelanger5, line 56: Input and output really shouldn't be sent in the `TaskSummary` -- can we just remove them from this field and the underlying query? There's a pretty noticeable slowdown if a user has large inputs and outputs.

abelanger5 (changes requested):

api/v1/server/oas/transformers/v1/tasks.go
mrkaye97: circling back here - I'm pretty sure we need these to render the inputs and outputs on the FE (also: there are a bunch of places where we create an instance of this struct and this is just one of them). are you thinking that we'll get the output dynamically on the FE from the task events? or have a separate query for it? or something else I'm missing?

if you don't think it's dire, I'd prefer to merge this in as-is and then circle back to refactoring this after since this will unbreak some python tests for V1.

abelanger5 (approved):

PR #1384, opened 2025-03-22
Feat: Python V1 Durable Features

mrkaye97:
Adds durable features to the Python SDK.

Basically what's happened here is we run two action listeners, two action runners, and have separate registries inside of the worker depending on what's registered. The `durable` ones only run if there's at least one durable workflow registered, and the same is true for the "normal" ones, and they both run if at least one of each is registered.

From the user's PoV, it looks like two workers are running instead of one, with totally separate action registries: as if they'd assigned the durable workflows to the durable worker manually. 

The other thing is that now you can define a task as being durable with `@workflow.durable()`, which is exactly like a regular task, except it marks the task as durable and gives it the `DurableContext` instead of a normal `Context` (importantly, this type checks correctly)

**Note: This is a pretty hacky implementation -  I think it'll be fine for now, but I'm sure there's a better way**


sdks/python/hatchet_sdk/clients/dispatcher/action_listener.py
```diff
@@ -152,7 +152,6 @@ def __init__(self, config: ClientConfig, worker_id: str) -> None:
         self.config = config
         self.worker_id = worker_id
 
-        self.client = DispatcherStub(new_conn(self.config, False))  # type: ignore[no-untyped-call]
```
mrkaye97, line 155: unused

sdks/python/hatchet_sdk/clients/durable_event_listener.py
mrkaye97, line 1: this is the guts of it - a lot of copy paste from the `PooledWorkflowRunListener`

sdks/python/hatchet_sdk/runnables/task.py
```diff
@@ -57,8 +70,11 @@ def __init__(
         skip_if: list[Condition | OrGroup] = [],
         cancel_if: list[Condition | OrGroup] = [],
     ) -> None:
-        self.fn = fn
-        self.is_async_function = is_async_fn(fn)
+        self.is_durable = is_durable
+
+        self.fn = _fn
+        self.is_async_function = is_async_fn(self.fn)  # type: ignore
```
mrkaye97, line 76: typing here was a huge PITA so I just added some ignores for now - I can fix this later 🤷

sdks/python/hatchet_sdk/clients/durable_event_listener.py
```diff
@@ -0,0 +1,318 @@
...
+        ) = None
+        self.listener_task: asyncio.Task[None] | None = None
+
+        self.curr_requester: int = 0
+
+        self.events: dict[int, _Subscription] = {}
+
+        self.interrupter: asyncio.Task[None] | None = None
+
+    async def _interrupter(self) -> None:
```
abelanger5, line 121: Is interrupter actually called? I think this is a relic of the action listener (which needs an external interrupt signal) but I don't think this does?

sdks/python/hatchet_sdk/runnables/workflow.py
```diff
@@ -455,6 +466,91 @@ def inner(
 
         return inner
 
+    def durable(
```
grutt: nit: i think implicit durable_task would make this more clear

grutt (approved):

sdks/python/hatchet_sdk/runnables/workflow.py
mrkaye97: good call, I'll change that 👍

sdks/python/hatchet_sdk/clients/durable_event_listener.py
mrkaye97: heh I'm not sure, I can poke around a little bit

mrkaye97: okay yeah I don't think we're using the interrupt anywhere, but I'm not going to mess with this now - it's referenced in a handful of places

PR #1387, opened 2025-03-23
feat: go sdk v1

grutt:
# Description

V1 SDK changes


pkg/v1/workflow/declaration.go
```diff
@@ -0,0 +1,357 @@
...
+		}
+	}
+
+	return req, fns, onFailureFn
+}
+
+// NewTaskDeclaration creates a new task declaration with the specified options.
+// @params opts - The options for creating a task
+// @params fn - The function to execute when the task runs (must be a function that takes an input and a Hatchet context and returns an output and an error)
+func NewTaskDeclaration[I any](opts task.CreateOpts[I], fn func(input I, ctx worker.HatchetContext) (*any, error)) *task.TaskDeclaration[I] {
```
abelanger5: Why does the TaskDeclaration not take an output parameter instead of returning `*any`?

pkg/v1/task/task.go
```diff
@@ -0,0 +1,234 @@
...
+	"fmt"
+	"time"
+
+	contracts "github.com/hatchet-dev/hatchet/internal/services/shared/proto/v1"
+	"github.com/hatchet-dev/hatchet/pkg/client/types"
+)
+
+// TaskDefaults defines default configuration values for tasks within a workflow.
+type TaskDefaults struct {
+	// (optional) ExecutionTimeout specifies the maximum duration a task can run after starting before being terminated
```
abelanger5, line 13: All optionals should document defaults

internal/services/admin/v1/server.go
grutt, line 1: idk what's up here but something was nil on a normal request..

examples/v1/workflows/dag.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+type DagResult struct {
+	Step1 SimpleOutput
+	Step2 SimpleOutput
+}
+
+func DagWorkflow(hatchet *v1.HatchetClient) workflow.WorkflowDeclaration[any, DagResult] {
+
+	simple := v1.WorkflowFactory[any, DagResult](
```
mrkaye97, line 23: qq: do we have to use `any` here? this is the input type right?

examples/v1/workflows/dag.go
```diff
@@ -0,0 +1,65 @@
...
+		workflow.CreateOpts[any]{
```
mrkaye97, line 24: same here re: any, but also: is it possible to do this without requiring the generic to be specified? coming from not knowing much about Go's type system

examples/v1/workflows/dag.go
```diff
@@ -0,0 +1,65 @@
...
+			Name: "simple-dag",
+		},
+		hatchet,
+	)
+
+	step1 := simple.Task(
+		task.CreateOpts[any]{
```
mrkaye97, line 31: same here I think - ideally we'd not need to provide this? idk if it's possible 🤷

examples/v1/workflows/dag.go
```diff
@@ -0,0 +1,65 @@
...
+			Name: "Step1",
+			Fn: func(_ any, ctx worker.HatchetContext) (*SimpleOutput, error) {
```
mrkaye97, line 33: does Go have something like `never`?

examples/v1/workflows/durable-sleep.go
```diff
@@ -0,0 +1,53 @@
...
+		},
+		hatchet,
+	)
+
+	simple.DurableTask(
+		task.CreateOpts[DurableSleepInput]{
+			Name: "Sleep",
+			Fn: func(input DurableSleepInput, ctx worker.DurableHatchetContext) (*SleepOutput, error) {
+
+				_, err := ctx.SleepFor(time.Minute)
```
mrkaye97, line 39: nit: in Python I'm calling this `wait_for` - do you think we should be consistent here? or is this just a wrapper around `wait_for`? if it is, maybe I should implement that too

pkg/v1/client.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// Workflow creates a new workflow declaration with the provided options.
+func (c *v1HatchetClientImpl) Workflow(opts workflow.CreateOpts[any]) workflow.WorkflowDeclaration[any, any] {
+	var v0 v0Client.Client
+	if c.v0 != nil {
+		v0 = *c.v0
+	}
+
+	return workflow.NewWorkflowDeclaration[any, any](opts, &v0)
```
mrkaye97, line 100: odd use of `any` here I think too but I'm not 100% sure

pkg/v1/features/crons.go
```diff
@@ -0,0 +1,161 @@
...
+	return &cronsClientImpl{
+		api:      api,
+		tenantId: tenantIdUUID,
+	}
+}
+
+// ValidateCronExpression validates that a string is a valid cron expression.
+func ValidateCronExpression(expression string) bool {
+	// Basic cron validation regex matching the TypeScript implementation
+	cronRegex := regexp.MustCompile(`^(\*|([0-9]|1[0-9]|2[0-9]|3[0-9]|4[0-9]|5[0-9])|\*\/([0-9]|1[0-9]|2[0-3])) (\*|([0-9]|1[0-9]|2[0-3])|\*\/([0-9]|1[0-9]|2[0-3])) (\*|([1-9]|1[0-9]|2[0-9]|3[0-1])|\*\/([1-9]|1[0-9]|2[0-9]|3[0-1])) (\*|([1-9]|1[0-2])|\*\/([1-9]|1[0-2])) (\*|([0-6])|\*\/([0-6]))$`)
```
mrkaye97, line 64: 🤮

mrkaye97 (commented):
left a couple comments but broadly makes sense - mostly am unsure about how much we're leaning on `any` in a lot of places but I might just be misunderstanding the Go compiler / patterns

examples/v1/workflows/durable-sleep.go
grutt: its a wrapper in go and typescript

abelanger5 (approved):

PR #1388, opened 2025-03-23
feat(v1): tenant limits

abelanger5:
# Description

Adds (optional) tenant limits to v1 for workers, worker slots, workflow runs, task runs, and events.

## Type of change


api/v1/server/handlers/v1/workflow-runs/trigger.go
```diff
@@ -64,6 +64,10 @@ func (t *V1WorkflowRunsService) V1WorkflowRunCreate(ctx echo.Context, request ge
 				return gen.V1WorkflowRunCreate400JSONResponse(
 					apierrors.NewAPIErrors(e.Message()),
 				), nil
+			case codes.ResourceExhausted:
+				return gen.V1WorkflowRunCreate400JSONResponse(
+					apierrors.NewAPIErrors(e.Message()),
+				), nil
```
mrkaye97, line 70: this is the case where someone tries to trigger a workflow and is at their limit?

pkg/repository/metered/metered.go
```diff
@@ -30,6 +30,69 @@ func NewMetered(entitlements repository.EntitlementsRepository, l *zerolog.Logge
...
+			return false, 0, fmt.Errorf("could not check tenant limit: %w", err)
+		}
+
+		canCreate = &c
+		percent = p
+
+		if percent <= 50 || percent >= 100 {
+			m.c.Set(key, c)
+		}
+	}
```
mrkaye97, line 57: not sure I'm following this - what are the percents here? usage?

mrkaye97 (approved):
left a couple questions, but looks good to me 👍

api/v1/server/handlers/v1/workflow-runs/trigger.go
abelanger5: yes, in this case from the REST API

pkg/repository/metered/metered.go
abelanger5: yeah we don't cache as we get close to the limit -- this is old code that was copy-pasted, we might want to change this...

mrkaye97: gotcha makes sense - yeah it might be good for readability at least to factor this out into a function so it's easier to understand what it's trying to do at least 🤷

PR #1396, opened 2025-03-24
fix(v1): bump logging line limit

abelanger5:
# Description

We're passing in a limit of 5k in some places which is breaking with the current validation.

## Type of change


mrkaye97 (approved):
is there any risk to this? storage issues? performance?

PR #1402, opened 2025-03-25
feat: standalone tasks

grutt:
Standalone tasks, on success


sdks/typescript/src/v1/declaration.ts
```diff
@@ -439,3 +557,41 @@ export function CreateWorkflow<T extends JsonObject = any, K extends JsonObject
...
+): TaskWorkflowDeclaration<I, O> {
+  // Note: We're using TaskWorkflowDeclaration here since task and durableTask
+  // share the same declaration structure but with different task types
+  const taskWorkflow = new TaskWorkflowDeclaration<I, O>(options as any, client);
+
+  // Move the task from tasks to durableTasks
+  if (taskWorkflow.definition._tasks.length > 0) {
+    const task = taskWorkflow.definition._tasks[0];
+    taskWorkflow.definition._tasks = [];
+    taskWorkflow.definition._durableTasks.push(task as any);
```
mrkaye97: I'm not totally clear on why we need this cast to any (also a bunch of similar type casts in other places) - is `_tasks` not an array of `Task` and compatible with `_durableTasks`?

sdks/typescript/src/v1/examples/infered-typing/run.ts
mrkaye97, line 1: nit: typo in file name (infered -> inferred)

sdks/typescript/src/v1/examples/infered-typing/workflow.ts
```diff
@@ -0,0 +1,50 @@
...
+      TransformedMessage: input.Message.toUpperCase(),
+    };
+  },
+});
+
+export const crazyWorkflow = hatchet.workflow({
+  name: 'crazy-workflow',
+});
+
+const step1 = crazyWorkflow.task(declaredType);
```
mrkaye97: nit: I'm calling this `add_task` but not a big deal

sdks/typescript/src/v1/examples/infered-typing/workflow.ts
```diff
@@ -0,0 +1,50 @@
...
+// crazyWorkflow.task(inferredTypeDurable);
+
+crazyWorkflow.task({
+  parents: [step1],
+  ...inferredType.taskDef,
+});
```
mrkaye97: IMO, we shouldn't advertise this (or the above thing with `.task()` - I think in this case, people _really_ should be using the DAG toolkit

sdks/typescript/src/v1/examples/on_success/workflow.ts
```diff
@@ -0,0 +1,50 @@
...
+// ❓ On Success Task
+export const onSuccess = hatchet.task({
+  name: 'run-on-success',
+  // 👀 onSuccess handler will run if the fn task succeeds
+  onSuccess: (_, ctx) => {
+    console.log('onSuccess for run:', ctx.workflowRunId());
+    return {
+      'on-success': 'success',
+    };
+  },
```
mrkaye97, line 13: I wasn't exposing `on_success` or `on_failure` for the single task thing either. Do you think we should do this? I sorta think no, for a similar reason to the DAG stuff

mrkaye97 (commented):
Left some comments but mostly looks good! looks pretty similar to mine

sdks/typescript/src/v1/declaration.ts
grutt: good catch!

sdks/typescript/src/v1/examples/on_success/workflow.ts
grutt: i dont think there's much harm in exposing it for now, i think its a good goal to make task <> workflow as close as possible

PR #1413, opened 2025-03-26
[Python] Feat: Replace `REST` Client

mrkaye97:
A bunch going on here:

1. Totally ripping out the `RestClient` in favor of a bunch of smaller, individual clients wrapping different parts of the REST API, like `hatchet.workflows`, `hatchet.workers`, and so on
2. Cleans up a bunch of places where we weren't closing event loops and globally setting them, causing problems in Pytest and other weird errors
3. Adds bulk cancellations and bulk replays to the `runs` API
4. Fixes the need for datetime format patching logic by setting the datetime format on the API client
5. Got rid of a bunch of dead + duped code!
6. Removed some unused instance variables from a couple places (context, runner, ...)
7. Changed `Workflow.schedule` from taking a list to taking a single time to schedule at


sdks/python/hatchet_sdk/clients/v1/api_client.py
```diff
@@ -0,0 +1,89 @@
...
+        if loop and loop.is_running():
+            return loop.run_until_complete(async_func(*args, **kwargs))
+        else:
+            with ThreadPoolExecutor() as executor:
+                future = executor.submit(
+                    lambda: self._run_async_function_do_not_use_directly(
+                        async_func, *args, **kwargs
+                    )
+                )
+                return future.result()
```
mrkaye97, line 81: important change: before, we have some global `asyncio.set_event_loop` in a couple of places, which was causing side effects and unpredictability in other places in the codebase and also causing Pytest issues. This should fix that by creating a new loop in a thread, which has nothing to do with the main thread

sdks/python/hatchet_sdk/clients/v1/api_client.py
```diff
@@ -0,0 +1,89 @@
...
+        self.client_config = config
+        self.api_config = Configuration(
+            host=config.server_url,
+            access_token=config.token,
+        )
+
+        self.api_config.datetime_format = "%Y-%m-%dT%H:%M:%S.%fZ"
+
+    def client(self) -> AsyncContextManager[ApiClient]:
+        return ApiClient(self.api_config)
```
mrkaye97, line 46: There were also some issues with aiohttp clients not being closed, which I think was happening because we're supposed to be using this as an async context manager, which is what we'll do now

sdks/python/hatchet_sdk/clients/v1/api_client.py
```diff
@@ -0,0 +1,89 @@
...
+    def __init__(self, config: ClientConfig) -> None:
+        self.tenant_id = config.tenant_id
+
+        self.client_config = config
+        self.api_config = Configuration(
+            host=config.server_url,
+            access_token=config.token,
+        )
+
+        self.api_config.datetime_format = "%Y-%m-%dT%H:%M:%S.%fZ"
```
mrkaye97, line 43: This removes the need for that datetime format patch

grutt (approved):

abelanger5 (approved):

abelanger5 (approved):

PR #1426, opened 2025-03-27
docs: eject v0

grutt:
# Description

Having V0 and V1 docs in one place made search and UX difficult -- moves v0 docs to a new project that will be hosted at https://v0-docs.hatchet.run


mrkaye97 (approved):
⏏️

PR #1428, opened 2025-03-27
fix(v1): dag statuses should wait for all tasks to be created

abelanger5:
# Description

When a task is in a waiting state (to wait for a trigger condition), the UI displays the parent DAG in a final state instead of displaying it in a running state. 
 
## Type of change

### DIFF
--- a/pkg/repository/v1/sqlcv1/olap.sql
+++ b/pkg/repository/v1/sqlcv1/olap.sql
@@ -91,7 +91,8 @@ INSERT INTO v1_dags_olap (
     workflow_version_id,
     input,
     additional_metadata,
-    parent_task_external_id
+    parent_task_external_id,
+    total_tasks
 ) VALUES (
     $1,
     $2,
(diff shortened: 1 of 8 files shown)


pkg/repository/v1/sqlcv1/olap.sql
```diff
@@ -777,14 +781,16 @@ WITH locked_events AS (
...
+        d.id, d.inserted_at, d.total_tasks
 ), updated_dags AS (
     UPDATE
         v1_dags_olap d
     SET
         readable_status = CASE
             -- If we only have queued events, we should keep the status as is
             WHEN dtc.queued_count = dtc.task_count THEN d.readable_status
+            -- If the task count is not equal to the total tasks, we should set the status to running
+            WHEN dtc.task_count != dtc.total_tasks THEN 'RUNNING'
```
mrkaye97, line 793: any footguns here?

mrkaye97 (approved):

pkg/repository/v1/sqlcv1/olap.sql
abelanger5: I don't think so?

mrkaye97: will the task count include skipped tasks / cancelled tasks? that's the only thing that comes to mind

PR #1432, opened 2025-03-27
Proposal: Python Output Validator

mrkaye97:
Proposal for doing output validation similarly to input validation with Pydantic

Just a proposal


mrkaye97:
Closing b/c I don't think this is a good idea

PR #1443, opened 2025-03-28
Fix: More Path Filters

mrkaye97:
Adding more path filters in CI and removing that custom check path action now that required checks seem to be working right


.github/workflows/sdk-typescript.yml
```diff
@@ -11,7 +11,7 @@ on:
     branches:
       - main
     paths:
-      - 'sdks/typescript/**'
+      - 'sdks/typescript/package.json'
```
mrkaye97, line 14: Important: I think this is a reasonable way to only run in the correct cases on main, since you need to change this file (to bump the version) to publish anyways, and if we're not going to publish then no reason to leave it unchanged.

it's not perfect b/c you could e.g. change a dependency and not the package version, but it should be close at least

.github/workflows/sdk-python.yml
```diff
@@ -11,7 +11,7 @@ on:
     branches:
       - main
     paths:
-      - 'sdks/python/**'
+      - 'sdks/python/pyproject.toml'
```
mrkaye97, line 14: Important: I think this is a reasonable way to only run in the correct cases on main, since you need to change this file (to bump the version) to publish anyways, and if we're not going to publish then no reason to leave it unchanged.

it's not perfect b/c you could e.g. change a dependency and not the package version, but it should be close at least

.pre-commit-config.yaml
```diff
@@ -9,7 +9,7 @@ repos:
       - id: trailing-whitespace
       - id: check-yaml
   - repo: https://github.com/golangci/golangci-lint
-    rev: v1.62.0
+    rev: v2.0.2
```
mrkaye97: I assume this is fine? the old version was breaking the lint CI step

grutt (approved):

PR #1444, opened 2025-03-28
fix: default type and export duration

grutt:
# Description

Default type from any -> unknown, I could be convinced this is a bad idea since it forces types but if you don't want types you can set your input type to any like an animal 🐮


mrkaye97 (approved):
nice 👍

mrkaye97:
I'm very much in favor of `unknown` over `any` fwiw - I'd prefer we had that rule to not allow explicit or implicit `any` on :P

PR #1452, opened 2025-03-30
[Python] Fix: Remove global event loop setters everywhere

mrkaye97:
Finally getting around to removing all of our global event loop setters everywhere, which I think were largely the cause of all of our async headaches (pytest, etc.). TL;DR here is that now, everything async (except for starting a worker basically) assumes there's a loop running. IMO, this is super important since we don't want our library to be trying to set a global event loop if e.g. someone tries to run a workflow, because that could mess with their existing code (attached to different loop errors, etc.)

I also added a hack to patch some typing for generated stubs so we'd get a little better mypy support there


sdks/python/hatchet_sdk/clients/dispatcher/dispatcher.py
```diff
@@ -95,6 +93,9 @@ async def send_step_action_event(
     async def _try_send_step_action_event(
         self, action: Action, event_type: StepActionEventType, payload: str
     ) -> grpc.aio.UnaryUnaryCall[StepActionEvent, ActionEventResponse]:
+        aio_conn = new_conn(self.config, True)
+        aio_client = DispatcherStub(aio_conn)  # type: ignore[no-untyped-call]
```
mrkaye97, line 97: IMPORTANT: This is one place (I think the only one) where I'm nervous about performance. The issue here is that we can't create an async client like this at the top level without requiring there to be an event loop, so our options are:

1. Throw if there's no event loop (I think this is a bad idea)
2. Create the event loop ourselves (the point of this PR is to _stop_ doing this
3. Do this here (we might be able to cache, I'm not sure though)

Any thoughts on this? I can try timing it and see how it goes

grutt (approved):

PR #1453, opened 2025-03-30
Fix  typescript types

grutt:
# Description

Improves type consistency and adds test coverage to typing


Scalahansolo:
Minor thing Gabe. It looks like this log messages isn't flowing through the passed custom logger. Figured it would be easy enough to sneak into this to not require another PR?

CleanShot 2025-03-30 at 18 06 58@2x

sdks/typescript/src/step.ts
```diff
@@ -296,6 +289,14 @@ export class Context<T, K = {}> {
     return this.action.workflowRunId;
   }
 
+  /**
+   * Gets the ID of the current workflow run.
+   * @returns The workflow run ID.
+   */
+  taskRunId(): string {
```
mrkaye97: nit: potentially confusing mix of `workflow run id` and `task run id` here

mrkaye97 (approved):
nice!

sdks/typescript/src/step.ts
grutt: good catch

msalafia:
@grutt with the new types for workflow declaration i cannot pass my input type:
```ts
export interface TimeoutsToCloudWatchData {
  timeout: number;
}

let _declaration: WorkflowDeclaration<
    TimeoutsToCloudWatchData,          //<--- error here: Index signature of type `string` is missing for type `TimeoutsToCloudWatchData`
    TaskResponse>;
```

The only way i have at the moment to bypass this is declaring a mapped type
```ts
export type Expanse<T extends Record<string, any>> = T & {
  [x: string]: string;
};
```

and use it every time i use an input type

```ts
let _declaration: WorkflowDeclaration<
    Expanse<TimeoutsToCloudWatchData>,          //<--- error here: Now it bypass the error
    TaskResponse>;
```

Is this something you really want to accomplish?

grutt:
Hi @msalafia, yes this is now the intended behavior. 

To clarify, a workflow must return Record<string, Record<string, JSONType>> where JSONType is a json serializable object.

The first dictionary prop is the name of the workflow task where the result will be posted. If it is not included, we can't type the return object for a specific task from the workflow object.

Does that make sense?

msalafia:
@grutt Yes for the return type.

I don't like the idea to wrap my input types with `Expanse<>` just to add the part `[x: string]: JsonValues` to my type because it is cumbersome from my point of view. My type `TimeoutsToCloudWatchData` it is already a json serializable object imho, and i would not add that additional index signatures (which in fact does not exist at all in my application logic for that type) just because the sdk is forcing me to do so for unknown reasons.

A fun fact: if i declare `TimeoutsToCloudWatchData` as `type` rather than an `interface` the compiler does not complain and i have no idea why this happen as for my typescript knowledge. But i cannot declare every input type as `type` because most of them already exist in my codebase as interfaces or classes and i cannot change them.

From my point of view, the hatchet sdk should allow something like
```ts
let _declaration: WorkflowDeclaration<
    TimeoutsToCloudWatchData,       
    TaskResponse>;
```
becase in my opinion is clearer.

But maybe i am missing something :D

grutt:
Right, I think this is a feature since it enforces the type on the correct task within the workflow. 

The recommended way to write your type without Expanse would actually be:
```
export interface TimeoutsToCloudWatchData {
 taskName: {
    timeout: number;
  }
}
```
This will then allow the type checker to enforce that the `taskName` task fn returns the correct type. This would further type the results of `.run` to include a fully typed object `res.taskName.timeout`.


Related, but we added a helper for a single task workflow where the `Output` type is directly bound without the need for the task-result mapping:

```javascript
export type Input = {
  Message: string;
};

export type Output = {
  TransformedMessage: string
}

export const simple = hatchet.task<Input, Output>({
  name: 'simple',
  fn: (input) => {
    return {
      TransformedMessage: input.Message.toLowerCase(),
    };
  },
});
```

msalafia:
I am getting confused. In the documentation [here](https://docs.hatchet.run/home/dags#defining-a-workflow) the yellow bow says to add the task name only in the output: "Declaring input and output types are optional, but recommended. When declaring output types, the keys of the object will be the names of the tasks in the workflow where the output will be returned."

Furthermore your example is completely different from 
```ts
type DagInput = {
  Message: string;
};
 
type DagOutput = {
  reverse: {
    Original: string;
    Transformed: string;
  };
};
 
export const simple = hatchet.workflow<DagInput, DagOutput>({
  name: "simple",
});
```
which is the example in the doc. Maybe you are referring to task, whilst i am referring to workflow, i don't know. I haven't tested, but i am quite sure that if i declare DagInput as an interface, the compiler returns an error.

grutt:
Wow, my fault -- I didn't realize we were talking about Inputs... yes that might be a bug

msalafia:
Could you confirm this is a bug, please? Because we are migrating our project to the new sdk and we want to be sure to make things using proper types for our workflow definitions.

grutt:
@msalafia opening a PR #1464 that should address using interfaces

PR #1456, opened 2025-03-31
Feat: Non-Retryable Exceptions

mrkaye97:
Adding non-retrying exceptions, which the client can raise to have a task not retry.


pkg/repository/v1/task.go
```diff
@@ -1003,8 +1010,9 @@ func (r *TaskRepositoryImpl) ProcessTaskTimeouts(ctx context.Context, tenantId s
 				InsertedAt: task.InsertedAt,
 				RetryCount: task.RetryCount,
 			},
-			IsAppError:   true,
-			ErrorMessage: fmt.Sprintf("Task exceeded timeout of %s", task.StepTimeout.String),
+			IsAppError:     true,
+			ErrorMessage:   fmt.Sprintf("Task exceeded timeout of %s", task.StepTimeout.String),
+			IsNonRetryable: false,
```
abelanger5, line 1015: Minor nit: Go default value for a boolean is `false`, so this is unnecessary though ok to be explicit

abelanger5 (approved):

PR #1458, opened 2025-03-31
feat: add cookie check and tooling

grutt:
# Description

Adds tooling for docs observability and cookie check


frontend/docs/components/ui/cookie-banner.tsx
```diff
@@ -0,0 +1,138 @@
...
+    const [isOpen, setIsOpen] = useState(false);
+    const [hide, setHide] = useState(false);
+    const [consentGiven, setConsentGiven] = useState('');
+
+    const accept = () => {
+        setIsOpen(false);
+        document.cookie = "cookieConsent=true; expires=Fri, 31 Dec 9999 23:59:59 GMT";
+
+        localStorage.setItem('cookie_consent', 'yes');
+        setConsentGiven('yes');
```
mrkaye97, line 30: qq: can we use only a cookie here instead of both a cookie and local storage?

frontend/docs/components/ui/cookie-banner.tsx
```diff
@@ -0,0 +1,138 @@
...
+    return localStorage.getItem('cookie_consent');
+}
+
+
+export default function CookieConsent({ variant = "default", demo = false, onAcceptCallback = () => { }, onDeclineCallback = () => { } }) {
+    const [isOpen, setIsOpen] = useState(false);
+    const [hide, setHide] = useState(false);
+    const [consentGiven, setConsentGiven] = useState('');
+
+    const accept = () => {
```
mrkaye97, line 25: I think this should probably be wrapped in a `useCallback` (same with the `decline` below) to avoid rerenders if we can

frontend/docs/components/ui/cookie-banner.tsx
```diff
@@ -0,0 +1,138 @@
...
+                    }, 700);
+                }
+            }
+        }
+        catch (e) {
+            // console.log("Error: ", e);
+        }
+    }, []);
+
+  const [reo, setReo] = useState<boolean>(false);
```
mrkaye97, line 70: what's `reo`?

also nit: I don't think we need the generic, and this also should probably go up above with the other state management unless it needs to be here 🤷

mrkaye97 (commented):
Left a couple small questions!

mrkaye97:
heh I was too late, not a big deal though

PR #1464, opened 2025-04-01
fix: typescript v1 bug bash

grutt:
# Description
Bug fixes for V1 SDK

## Type of change

## What's Changed

- [x] Exports base interface types to enforce strict type checking:
```
interface DagInput extends WorkflowInputType {
  Message: string;
}

interface DagOutput extends WorkflowOutputType {
  reverse: {
    Original: string;
    Transformed: string;
  };
}
```
- [x] Fix Standalone Task bulk context operation return types

### DIFF
--- a/sdks/python/README.md
+++ b/sdks/python/README.md
@@ -29,31 +29,22 @@ poetry add hatchet-sdk
 Here's a simple example of how to use the Hatchet Python SDK:
 
 ```python
-from hatchet_sdk import Hatchet, WorkflowContext
+from hatchet_sdk import Context, EmptyModel, Hatchet
 
-# Initialize the Hatchet client
-client = Hatchet(
-    api_key="your-api-key",  # For Hatchet Cloud
-    # For self-hosted:
-    # url="your-hatchet-instance-url",
-    # api_key="your-api-key"
-)
+hatchet = Hatchet(debug=True)
 
-# Define your workflow
-@client.workflow("example-workflow")
-def example_workflow(ctx: WorkflowContext, input_data: dict):
-    # Your workflow logic here
-    result = ctx.step(
-        "process-data",
-        lambda: process_data(input_data)
-    )
-    return result
 
-# Start your workflow
-workflow_run = client.start_workflow(
-    "example-workflow",
-    input_data={"key": "value"}
-)
+@hatchet.task(name="SimpleWorkflow")
+def step1(input: EmptyModel, ctx: Context) -> None:
+    print("executed step1")
+
+
+def main() -> None:
+    worker = hatchet.worker("test-worker", slots=1, workflows=[step1])
+    worker.start()
+
+if __name__ == "__main__":
+    main()
 ```
 
 ## Features
@@ -68,7 +59,6 @@ workflow_run = client.start_workflow(
 
 For detailed documentation, examples, and best practices, visit:
 - [Hatchet Documentation](https://docs.hatchet.run)
-- [Python SDK Reference](https://docs.hatchet.run/sdks/python)
 - [Examples](https://docs.hatchet.run/examples)
 
 ## Contributing
--- a/sdks/python/README.md
+++ b/sdks/python/README.md
@@ -1,3 +1,80 @@
 # Hatchet Python SDK
 
-This is the [Hatchet](https://hatchet.run) Python SDK. For usage, see the [docs](https://docs.hatchet.run).
+<div align="center">
+
+[![PyPI version](https://badge.fury.io/py/hatchet-sdk.svg)](https://badge.fury.io/py/hatchet-sdk)
+[![Documentation](https://img.shields.io/badge/docs-hatchet.run-blue)](https://docs.hatchet.run)
+[![License: MIT](https://img.shields.io/badge/License-MIT-purple.svg)](https://opensource.org/licenses/MIT)
+
+</div>
+
+This is the official Python SDK for [Hatchet](https://hatchet.run), a distributed, fault-tolerant task queue. The SDK allows you to easily integrate Hatchet's task scheduling and workflow orchestration capabilities into your Python applications.
+
+## Installation
+
+Install the SDK using pip:
+
+```bash
+pip install hatchet-sdk
+```
+
+Or using poetry:
+
+```bash
+poetry add hatchet-sdk
+```
+
+## Quick Start
+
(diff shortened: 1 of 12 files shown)


sdks/python/README.md
```diff
@@ -1,3 +1,80 @@
 # Hatchet Python SDK
 
-This is the [Hatchet](https://hatchet.run) Python SDK. For usage, see the [docs](https://docs.hatchet.run).
+<div align="center">
```
mrkaye97, line 3: thanks!

sdks/python/README.md
```diff
@@ -1,3 +1,80 @@
...
+## Quick Start
+
+Here's a simple example of how to use the Hatchet Python SDK:
+
+```python
+from hatchet_sdk import Hatchet, WorkflowContext
+
+# Initialize the Hatchet client
+client = Hatchet(
+    api_key="your-api-key",  # For Hatchet Cloud
```
abelanger5: Hallucination? I'm pretty sure this isn't a thing, but perhaps @mrkaye97 can confirm

sdks/python/README.md
```diff
@@ -1,3 +1,80 @@
...
+def example_workflow(ctx: WorkflowContext, input_data: dict):
+    # Your workflow logic here
+    result = ctx.step(
+        "process-data",
+        lambda: process_data(input_data)
+    )
+    return result
+
+# Start your workflow
+workflow_run = client.start_workflow(
```
abelanger5: This all looks hallucinated

abelanger5 (changes requested):

sdks/python/README.md
grutt: yes it is

sdks/python/README.md
```diff
@@ -1,3 +1,70 @@
...
+- 🔁 **Automatic Retries**: Configure retry policies for handling transient failures
+- 📊 **Observability**: Track workflow progress and monitor execution metrics
+- ⏰ **Scheduling**: Schedule workflows to run at specific times or on a recurring basis
+- 🔄 **Event-Driven**: Trigger workflows based on events in your system
+
+## Documentation
+
+For detailed documentation, examples, and best practices, visit:
+- [Hatchet Documentation](https://docs.hatchet.run)
+- [Examples](https://docs.hatchet.run/examples)
```
mrkaye97: this one 404s too - maybe link to the examples in the repo here?

sdks/typescript/README.md
```diff
@@ -1,9 +1,109 @@
...
+- 🔁 **Automatic Retries**: Configure retry policies for handling transient failures
+- 📊 **Observability**: Track workflow progress and monitor execution metrics
+- ⏰ **Scheduling**: Schedule workflows to run at specific times or on a recurring basis
+- 🔄 **Event-Driven**: Trigger workflows based on events in your system
+
+## Documentation
+
+For detailed documentation, examples, and best practices, visit:
+- [Hatchet Documentation](https://docs.hatchet.run)
+- [Examples](https://docs.hatchet.run/examples)
```
mrkaye97: same here

grutt: i hate ai

mrkaye97 (approved):

abelanger5 (approved):

msalafia:
I still think this is not a solution. Forcing the sdk users to extend this new interface `WorkflowInputType` just to add the missing index signature (as discussed in #1453 ) is not something elegant in my opinion. 

Let us imagine the following scenario:

```ts
//This is an existing interface in my codebase that i could not or should not want to change
// and represents the input of a workflow
interface MyInputData {
  a: string;
  b: number;
}

//Accoding to the new solution you provide i must create a new interface that both MyInputData
// and WorkflowInputType

interface MyWorkflowInput extends WorkflowInputType, MyInputData {}

let x: MyWorkflowInput = {a: 'hello', b: 1};

//As a result x now is not a MyInputData as i would expect but something different that allows me to do something like
// this, which is obviously something that should not be allowed
x.notExisting = 3; //<---- MyInputData has no property "notExisting" but no error is returned here
```

And just to be clear this is absolutely not needed because i could use a type instead which does not suffer from index signature requirements contrariwise to interfaces (see [here](https://github.com/microsoft/TypeScript/issues/15300) ).

```ts
interface MyInputData {
  a: string;
  b: number;
}

type MyWorkflowInput = Pick<MyInputData, keyof MyInputData>;

let workflow: WorkflowDeclaration<MyWorkflowInput>;   <--- no error in a workflow declaration generic type

let x: MyWorkflowInput = {a: 'hello', b: 1};
x.notExisting = 3; // <----- this gives me an error back now and it is still not allowed and developers will be happy
```

But this is just a naive solution, because `Pick` should be applied recursively to all thus instead of Pick a better solution is adopt a brand new Mapped Type. Currently im using this new type `TypeOf<T>` while migrating my projects to hatchet v1:

```ts
/**
 * TypeOf<T> is a utility type that recursively maps over the properties of T.
 *
 * NOTE: This is an utility type to bypass wrong type definition introduced on
 * Hatchet sdk v1.0.4. You can remove this and all its usages in the codebase
 * once hatchet devs fix the type definition.
 */
export type TypeOf<T> =
  T extends Array<infer U>
    ? Array<TypeOf<U>>
    : T extends Record<string, any>
      ? {
          [K in keyof T]: TypeOf<T[K]>;
        }
      : T;
``` 

This is just a workaround that can be improved for sure. This allows me to do something like `hatchet.workflow<TypeOf<MyInputData>, TaskResponse>(...)` and still using my existing data types. I strongly believe that improvements should be made at the SDK level to enhance type safety, ensuring that developers are not forced to modify their own types just to make things work.

I haven't dived in the hatchet sdk in depth so i don't know which particular requirements you have at type level. As far as i understand you require that the type of the input must be a valid JSON serializable object, therefore i would define a MappedType like JSON<T> which check exactly this, but it would be nice if this type is used internally by the sdk, so that it can return type errors when i am using a wrong type in the generic type of `WorkflowDeclaration`. In this way if i get an error from the type checker is not due to the lack of an index signature i have never heard before, but to to the fact that my type is not a valid JSON serializable type, which is in line with the statement from the Hatchet documentation.

@grutt what do you think?

grutt:
@msalafia moving this conversation to #1466

msalafia:
@grutt ok i will follow it. Thank you.

PR #1467, opened 2025-04-01
fix: add readme to dist

grutt:
# Description

Adds the Readme to the TS dist package


mrkaye97 (approved):
looks good, but I'm kinda surprised this doesn't work automatically?

PR #1469, opened 2025-04-01
Fix: Doc Light Mode

mrkaye97:
Fixing doc light mode. Had to change a bunch of stuff in the `globals.css` and remove the `forcedTheme` thing. Also removed the rogue footer

#### Dark

Screenshot 2025-04-01 at 3 23 47 PM

#### Light

Screenshot 2025-04-01 at 3 23 55 PM

### DIFF
--- a/frontend/docs/pages/_meta.js
+++ b/frontend/docs/pages/_meta.js
@@ -3,6 +3,9 @@ export default {
     "title": "User Guide",
     "type": "page"
   },
+  "_setup": {
+    "display": "hidden"
+  },
   "compute": {
     "title": "Managed Compute",
     "type": "page",
--- a/frontend/docs/styles/global.css
+++ b/frontend/docs/styles/global.css
@@ -151,6 +151,15 @@
 /* ========================= */
 /* Dark Mode Fixes (Nextra)  */
 /* ========================= */
+html {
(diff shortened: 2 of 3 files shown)


abelanger5 (changes requested):
@mrkaye97 nice! I'm a little confused though, the colors aren't quite right on the new dark mode? I'd prefer to keep our background color which is relatively consistent between landing/app/docs: 

image

mrkaye97:
> @mrkaye97 nice! I'm a little confused though, the colors aren't quite right on the new dark mode? I'd prefer to keep our background color which is relatively consistent between landing/app/docs:
> 
> image

Good catch, yeah lemme see if I can figure out what went wrong there

grutt (approved):

abelanger5 (commented):
This still isn't quite the right color? 

image

mrkaye97:
> This still isn't quite the right color?
> 
> image

I think I might be losing my mind (or vision) - these look the same to me. let me check the styles in the browser...

Edit: You're right, it's not picking up the right color, one sec

mrkaye97:
alright I think this is actually fixed now - had pasted some things in the wrong place

abelanger5 (approved):
nice, thanks!

PR #1470, opened 2025-04-01
[Python] Single listener base class + bug fixes + refactors

mrkaye97:
Things going on here:

* I changed the openapi generation from using `aiohttp` to not being async, and then inverted all of the APIs to go from wrapping async methods as sync to wrapping sync methods as async (much easier), which got rid of a lot of needing to manage threads for running async work.
* Created an ABC for the pooled listener and used that for both the durable event listener and the workflow run listener
* Added a hack fix to the sync `result` method on the workflow run ref that polls the API for results instead of trying to use the listener in a sync context
* Explicitly added `loop_scope="session"` to all the tests and did a bit of fixture cleanup
* Removed a bunch of dead code
* Some misc refactors, like renaming all the protobuf converters to `to_proto`, moving the task-specific ones under the `Task` class, and removing params from functions that are just getting shared config passed in (e.g. `namespace`)
* Fixes the `task_output` method on the context to type correctly for async tasks
* Removes the ability to pass in your own event loop to the worker, which wasn't working at all (it'd just exit immediately without actually starting the worker). Now, we'll throw an error if you try to run the worker in a process with an already-existing event loop, e.g. with `asyncio.run` so you can't get into a weird situation with unexpected behavior

TODO:

- [x] Remove print statement cruft


sdks/python/hatchet_sdk/clients/rest/api/api_token_api.py
```diff
@@ -109,17 +109,17 @@ async def api_token_create(
...
-        await response_data.read()
+        response_data.read()
         return self.api_client.response_deserialize(
             response_data=response_data,
             response_types_map=_response_types_map,
         ).data
 
     @validate_call
-    async def api_token_create_with_http_info(
+    def api_token_create_with_http_info(
```
grutt, line 122: are these changes breaking?

sdks/python/examples/dag/worker.py
```diff
@@ -26,26 +26,26 @@ def step1(input: EmptyModel, ctx: Context) -> StepOutput:
...
 
 @dag_workflow.task(execution_timeout=timedelta(seconds=5))
-def step2(input: EmptyModel, ctx: Context) -> StepOutput:
+async def step2(input: EmptyModel, ctx: Context) -> StepOutput:
     return StepOutput(random_number=random.randint(1, 100))
 
 
 @dag_workflow.task(parents=[step1, step2])
-def step3(input: EmptyModel, ctx: Context) -> RandomSum:
+async def step3(input: EmptyModel, ctx: Context) -> RandomSum:
```
grutt, line 34: do we have any tested sync examples left?

sdks/python/hatchet_sdk/clients/rest/rest.py
```diff
@@ -16,16 +16,23 @@
...
-from typing import Optional, Union
 
-import aiohttp
-import aiohttp_retry
+import urllib3
 
 from hatchet_sdk.clients.rest.exceptions import ApiException, ApiValueError
 
-RESTResponseType = aiohttp.ClientResponse
+SUPPORTED_SOCKS_PROXIES = {"socks5", "socks5h", "socks4", "socks4a"}
```
grutt, line 24: why are we adding this?

grutt: oh is this generated?

grutt (approved):

sdks/python/hatchet_sdk/clients/rest/api/api_token_api.py
grutt: yes, but we're exposing aio versions in the feature clients?

mrkaye97: yeah these are breaking but I can't imagine anyone is using any of these directly, it'd be a big pain - the corresponding changes in the feature client wrappers are non-breaking

sdks/python/examples/dag/worker.py
mrkaye97: yeah lots - this one is partially sync and partially async, just wanted to improve the coverage of `Context.task_output`

sdks/python/hatchet_sdk/clients/rest/rest.py
mrkaye97: yeah

PR #1472, opened 2025-04-01
Docs:  run and wait, child run in run and wait, run no wait, bulk

grutt:
# Description

Docs pages and BULK sdk updates


frontend/docs/pages/home/bulk-run.mdx
```diff
@@ -0,0 +1,83 @@
...
+import UniversalTabs from "@/components/UniversalTabs";
+import { GithubSnippet, getSnippets } from "@/components/code";
+
+export const SimpleTs = {
+  path: "src/v1/examples/simple/bulk.ts",
+};
+
+export const SimplePy = {
+  path: "examples/simple/bulk.py",
+};
```
mrkaye97: I think this path is going to be broken now

mrkaye97:
```suggestion
export const SimplePy = {
  path: "examples/child/bulk.py",
};
```

frontend/docs/pages/home/bulk-run.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+You can use the `aio_run_many` method to bulk run a task. This will return a list of results.
+
+<GithubSnippet src={SimplePy} target="Bulk Run a Task" />
+
+<Callout type="info">
+  `step1.create_bulk_run_item` is helper to create the inputs for each task.
+</Callout>
+
+There are additional bulk methods available on the `Task` object.
```
mrkaye97:
```suggestion
There are additional bulk methods available on the `Workflow` object.
```

mrkaye97: (also true on the `Standalone` but maybe not worth mentioning)

frontend/docs/pages/home/run-with-results.mdx
```diff
@@ -6,42 +6,118 @@ export const SimpleTs = {
...
 
-{/* TODO V1 DOCS: Fanout, Loops, and Conditional Logic */}
+## Running Tasks in Parallel
+
+Sometimes you may want to run multiple tasks concurrently. Here's how to do that in each language:
+
+<UniversalTabs items={["Python", "Typescript", "Go"]}>
+<Tabs.Tab title="Python">
+
+Since the `aio_run` method returns a coroutine, you can spawn multiple tasks in parallel and await using `asyncio.gather`.
```
mrkaye97, line 98: I don't think we should recommend this - IMO, we should always recommend using the bulk methods which do the `gather` for you behind the scenes over individually running these. WDYT?

sdks/python/examples/child/bulk.py
```diff
@@ -0,0 +1,38 @@
...
+    # this will await all results and return a list of results
+    print(results)
+    # !!
+
+    # ❓ Running Multiple Tasks
+    result1 = child_task.aio_run(SimpleInput(message="Hello, World!"))
+    result2 = child_task.aio_run(SimpleInput(message="Hello, Moon!"))
+
+    #  gather the results of the two tasks
+    results = list(await asyncio.gather(result1, result2))
```
mrkaye97: nit: I don't think we need the `list` here - the code below should work fine with a `tuple`
```suggestion
    results = await asyncio.gather(result1, result2)
```

mrkaye97 (commented):
Left some small comments but otherwise looks good!

sdks/python/examples/child/bulk.py
grutt: linting was complaining..

mrkaye97: Fixed [here](https://github.com/hatchet-dev/hatchet/pull/1472/commits/6e73ab58d271876cb423cdc673b6a5babcec7b8f#diff-e52d071bf1d78f66ba72401685a2269263fc331fc7a6022d78c9a0e74a8b1a83R33) - the problem is the mutation of `results` to something of a different type. MyPy doesn't know which type the thing is once you do that (this is dumb but 🤷 )

mrkaye97 (approved):

PR #1476, opened 2025-04-02
fix: `LockSignalCreatedEvents` performance

abelanger5:
# Description

Attempts to improve performance of `LockSignalCreatedEvents` with more explicit locking and ordering targeting the unique index on the events table. 

## Type of change


sdks/typescript/src/v1/examples/child_workflows/workflow.ts
grutt, line 1: Let's revert this

grutt (changes requested):

grutt (approved):

PR #1488, opened 2025-04-03
Fix: Workflow state refetch

mrkaye97:
Adding a refetch interval to the workflow state query

### DIFF
--- a/frontend/app/src/pages/main/v1/workflow-runs-v1/hooks/workflow-details.tsx
+++ b/frontend/app/src/pages/main/v1/workflow-runs-v1/hooks/workflow-details.tsx
@@ -1,8 +1,20 @@
-import { queries } from '@/lib/api';
+import { queries, V1TaskStatus } from '@/lib/api';
 import { useQuery } from '@tanstack/react-query';
 import { AxiosError, isAxiosError } from 'axios';
 import { useParams } from 'react-router-dom';
 import invariant from 'tiny-invariant';
+
+function isTerminalState(status: V1TaskStatus | undefined) {
+  if (!status) {
+    return false;
+  }
+
+  return [
+    V1TaskStatus.COMPLETED,
+    V1TaskStatus.FAILED,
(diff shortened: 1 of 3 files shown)


frontend/app/src/pages/main/v1/workflow-runs-v1/hooks/workflow-details.tsx
```diff
@@ -17,6 +29,15 @@ export const useWorkflowDetails = () => {
 
       return true;
     },
+    refetchInterval: (query) => {
+      const data = query.state.data;
+
+      if (isTerminalState(data?.run?.status)) {
```
grutt, line 35: is this the only place? what about the task page?

mrkaye97: yeah can change that hook too, one sec

mrkaye97: well actually, so that one fetches every 5 seconds - do we want it to fetch more often than that? it'd be a ton of API calls

mrkaye97: that's [here](https://github.com/hatchet-dev/hatchet/blob/fix--workflow-state-refetch/frontend/app/src/pages/main/v1/workflow-runs-v1/hooks/task-runs.tsx#L48-L54)

grutt: no i meant for the detail pages --we have task-run and workflow-run pages, right?

mrkaye97: ah yeah good call, just fixed that - I'll factor that query out into a hook later, it's used a lot of places

grutt (approved):

PR #1497, opened 2025-04-07
fix: v1 edge cases on concurrency, go SDK, parent outputs

abelanger5:
# Description

Fixes the following edge cases reported with v1 migrations:

- [X] Issues with replays and workflow concurrency slots getting blocked. Annotating this inline but gist of it is that there's a simpler and more performant query to check whether workflow concurrency slots should be deleted that requires an additional pg_advisory_lock
- [X] Fixes overflow on queue metrics modal
- [X] Fixes an issue in the Go SDK where args seem to be passed in an incorrect order
- [X] Fixes an issue on DAGs where parent outputs aren't always queried correctly
- [X] Removes the recursive query for parent outputs in favor of a hopefully more performant query (will need some validation against load tests to confirm)
- [X] Fixes a bug in the OLAP repository when replayed children don't get monitoring events sent

## Type of change

### DIFF
--- a/cmd/hatchet-migrate/migrate/migrations/20250407012633_v1_0_11.sql
+++ b/cmd/hatchet-migrate/migrate/migrations/20250407012633_v1_0_11.sql
@@ -371,6 +371,9 @@ END;
 $$
 LANGUAGE plpgsql;
 
+DROP TRIGGER IF EXISTS after_v1_workflow_concurrency_slot_update ON v1_workflow_concurrency_slot;
+DROP FUNCTION IF EXISTS after_v1_workflow_concurrency_slot_update_function();
+
 -- +goose StatementEnd
 
 -- +goose Down
@@ -745,5 +748,43 @@ DROP FUNCTION IF EXISTS cleanup_workflow_concurrency_slots(
     p_workflow_version_id UUID,
     p_workflow_run_id UUID
 );
+
+-- After we update the v1_workflow_concurrency_slot, we'd like to check whether all child_strategy_ids are
+-- in the completed_child_strategy_ids. If so, we should delete the v1_workflow_concurrency_slot.
+CREATE OR REPLACE FUNCTION after_v1_workflow_concurrency_slot_update_function()
+RETURNS trigger AS $$
+BEGIN
+    -- place a lock on new_table
+    WITH slots_to_delete AS (
+        SELECT
+            wcs.strategy_id, wcs.workflow_version_id, wcs.workflow_run_id
+        FROM
+            new_table wcs
+        WHERE
+            CARDINALITY(wcs.child_strategy_ids) = CARDINALITY(wcs.completed_child_strategy_ids)
+        ORDER BY
+            wcs.strategy_id, wcs.workflow_version_id, wcs.workflow_run_id
+        FOR UPDATE
+    )
+    DELETE FROM
+        v1_workflow_concurrency_slot wcs
+    WHERE
+        (strategy_id, workflow_version_id, workflow_run_id) IN (
+            SELECT
+                strategy_id, workflow_version_id, workflow_run_id
+            FROM
+                slots_to_delete
+        );
+
+    RETURN NULL;
+END;
+
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER after_v1_workflow_concurrency_slot_update
+AFTER UPDATE ON v1_workflow_concurrency_slot
+REFERENCING NEW TABLE AS new_table
+FOR EACH STATEMENT
+EXECUTE FUNCTION after_v1_workflow_concurrency_slot_update_function();
 
 -- +goose StatementEnd
--- a/cmd/hatchet-migrate/migrate/migrations/20250407012633_v1_0_11.sql
+++ b/cmd/hatchet-migrate/migrate/migrations/20250407012633_v1_0_11.sql
@@ -1,0 +1,749 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION after_v1_concurrency_slot_insert_function()
+RETURNS trigger AS $$
+BEGIN
+    WITH parent_slot AS (
+        SELECT
+            *
+        FROM
+            new_table cs
+        WHERE
+            cs.parent_strategy_id IS NOT NULL
+    ), parent_to_child_strategy_ids AS (
+        SELECT
+            wc.id AS parent_strategy_id,
+            wc.tenant_id,
+            ps.workflow_id,
+            ps.workflow_version_id,
+            ps.workflow_run_id,
+            MAX(ps.sort_id) AS sort_id,
+            MAX(ps.priority) AS priority,
+            MAX(ps.key) AS key,
+            ARRAY_AGG(DISTINCT wc.child_strategy_ids) AS child_strategy_ids
+        FROM
+            parent_slot ps
+        JOIN v1_workflow_concurrency wc ON wc.workflow_id = ps.workflow_id AND wc.workflow_version_id = ps.workflow_version_id AND wc.id = ps.parent_strategy_id
+        GROUP BY
+            wc.id,
+            wc.tenant_id,
+            ps.workflow_id,
+            ps.workflow_version_id,
+            ps.workflow_run_id
+    )
+    INSERT INTO v1_workflow_concurrency_slot (
+        sort_id,
+        tenant_id,
+        workflow_id,
+        workflow_version_id,
+        workflow_run_id,
+        strategy_id,
+        child_strategy_ids,
+        priority,
+        key
+    )
+    SELECT
+        pcs.sort_id,
+        pcs.tenant_id,
+        pcs.workflow_id,
+        pcs.workflow_version_id,
+        pcs.workflow_run_id,
+        pcs.parent_strategy_id,
+        pcs.child_strategy_ids,
+        pcs.priority,
+        pcs.key
+    FROM
+        parent_to_child_strategy_ids pcs
+    ON CONFLICT (strategy_id, workflow_version_id, workflow_run_id) DO NOTHING;
+
+    -- If the v1_step_concurrency strategy is not active, we set it to active.
+    WITH inactive_strategies AS (
+        SELECT
+            strategy.*
+        FROM
+            new_table cs
+        JOIN
+            v1_step_concurrency strategy ON strategy.workflow_id = cs.workflow_id AND strategy.workflow_version_id = cs.workflow_version_id AND strategy.id = cs.strategy_id
+        WHERE
+            strategy.is_active = FALSE
+        ORDER BY
+            strategy.id
+        FOR UPDATE
+    )
+    UPDATE v1_step_concurrency strategy
+    SET is_active = TRUE
+    FROM inactive_strategies
+    WHERE
+        strategy.workflow_id = inactive_strategies.workflow_id AND
+        strategy.workflow_version_id = inactive_strategies.workflow_version_id AND
+        strategy.step_id = inactive_strategies.step_id AND
+        strategy.id = inactive_strategies.id;
+
+    RETURN NULL;
+END;
+
+$$ LANGUAGE plpgsql;
+
+CREATE OR REPLACE FUNCTION cleanup_workflow_concurrency_slots(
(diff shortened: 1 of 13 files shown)


pkg/v1/workflow/declaration.go
```diff
@@ -237,7 +237,7 @@ func (w *workflowDeclarationImpl[I, O]) Task(opts create.WorkflowTask[I, O], fn
 	genericFn := func(ctx worker.HatchetContext, input I) (*any, error) {
 		// Use reflection to call the specific function
 		fnValue := reflect.ValueOf(fn)
-		inputs := []reflect.Value{reflect.ValueOf(input), reflect.ValueOf(ctx)}
+		inputs := []reflect.Value{reflect.ValueOf(ctx), reflect.ValueOf(input)}
```
abelanger5, line 240: @grutt would be very helpful to double check that these are all correct. I first noticed a panic on an on-failure step, but I suspect all of these are in the wrong order?

pkg/v1/task/task.go
```diff
@@ -159,6 +154,19 @@ func makeContractTaskOpts(t *TaskShared, taskDefaults *create.TaskDefaults) *con
 		taskOpts.ScheduleTimeout = &scheduleTimeout
 	}
 
+	// Only set Retries if it's not nil
+	if t.Retries != nil {
+		taskOpts.Retries = *t.Retries
+	}
+
+	if t.RetryBackoffFactor != nil {
```
abelanger5, line 162: Discovered that `RetryBackoffFactor` and `RetryMaxBackoffSeconds` weren't getting passed through on the Go SDK -- moved the `t.Retries` logic down here to keep it consolidated

pkg/repository/v1/sqlcv1/tasks.sql
```diff
@@ -671,76 +671,47 @@ LEFT JOIN
...
+    t.step_id,
+    t.workflow_id,
     e.data AS output
 FROM
-    augmented_tasks at
+    v1_task t1
+JOIN
+    v1_dag_to_task dt ON dt.dag_id = t1.dag_id AND dt.dag_inserted_at = t1.dag_inserted_at
 JOIN
-    v1_task_event e ON e.task_id = at.id AND e.task_inserted_at = at.inserted_at AND e.retry_count = at.retry_count
```
abelanger5, line 741: I believe this is the cause of the bug that was reported with parent outputs not making it to a child step in a DAG. There's a race condition where parent tasks get a reassignment even though they're been completed -- this doesn't fix that, but allows us to read the previous retry count's output to pass to the child step while we dig deeper into the root cause.

pkg/repository/v1/sqlcv1/concurrency.sql
```diff
@@ -305,18 +306,13 @@ WITH filled_parent_slots AS (
         cs.*
     FROM
         v1_concurrency_slot cs
+    JOIN
+        eligible_slots_per_group es ON cs.task_id = es.task_id
     WHERE
-        (task_inserted_at, task_id, task_retry_count, tenant_id, strategy_id) IN (
```
abelanger5, line 309: Noticed very high query latency with this high cardinality `IN` statement and it seems like this is a pattern we should avoid when the right side of the join has a lot of rows. Anecdotally this improved performance a lot.

pkg/repository/v1/sqlcv1/concurrency.sql
```diff
@@ -12,17 +12,18 @@ WHERE
 -- name: GetWorkflowConcurrencyQueueCounts :many
 SELECT
     w."name" AS "workflowName",
-    cs.key,
+    wcs.key,
     COUNT(*) AS "count"
 FROM
-    v1_concurrency_slot cs
+    v1_workflow_concurrency_slot wcs
```
abelanger5, line 18: This is for rendering queue counts in the UI -- slightly more intuitive to show the workflow concurrency slots instead of the task concurrency slot counts

internal/services/controllers/v1/task/controller.go
```diff
@@ -1021,6 +1029,79 @@ func (tc *TasksControllerImpl) signalTasksCreated(ctx context.Context, tenantId
 	return eg.Wait()
 }
 
+func (tc *TasksControllerImpl) signalTasksReplayedFromMatch(ctx context.Context, tenantId string, tasks []*sqlcv1.V1Task) error {
```
abelanger5, line 1032: This fixes the issue where a task gets replayed and never seems to make progress (even when cancelled under the hood)

sql/schema/v1-core.sql
```diff
@@ -571,20 +571,7 @@ BEGIN
         pcs.key
     FROM
         parent_to_child_strategy_ids pcs
-    ON CONFLICT (strategy_id, workflow_version_id, workflow_run_id) DO UPDATE
-        -- When concurrency slots are INSERTED:
-        -- We need to REMOVE their strategy_ids from the parent's completed_child_strategy_ids
-        -- because these child slots are now active again, not completed
-        -- This correctly handles bulk inserts by removing ALL strategy_ids in the current batch
-        SET completed_child_strategy_ids = ARRAY(
```
abelanger5, line 579: We are no longer using `completed_child_strategy_ids` to case on whether to remove the workflow concurrency slot

sql/schema/v1-core.sql
```diff
@@ -620,77 +607,133 @@ REFERENCING NEW TABLE AS new_table
 FOR EACH STATEMENT
 EXECUTE FUNCTION after_v1_concurrency_slot_insert_function();
 
-CREATE OR REPLACE FUNCTION after_v1_concurrency_slot_delete_function()
-RETURNS trigger AS $$
+CREATE OR REPLACE FUNCTION cleanup_workflow_concurrency_slots(
```
abelanger5, line 611: A future optimization could be to allow this function to take a set of records instead of just 1 -- but given the issues we've been seeing with `IN` queries, and given there's not a clean way to write this without that type of query, I figured it's slightly safer to just use a singleton.

sql/schema/v1-core.sql
```diff
@@ -620,77 +607,133 @@ REFERENCING NEW TABLE AS new_table
...
+    -- Get the sort_id for the specific workflow concurrency slot
+    SELECT sort_id INTO v_sort_id
+    FROM v1_workflow_concurrency_slot
+    WHERE strategy_id = p_strategy_id
+      AND workflow_version_id = p_workflow_version_id
+      AND workflow_run_id = p_workflow_run_id;
+
+    -- Acquire an advisory lock for the strategy ID
+    -- There is a small chance of collisions but it's extremely unlikely
+    PERFORM pg_advisory_xact_lock(1000000 * p_strategy_id + v_sort_id);
```
abelanger5, line 628: This gives us 1 million slots before a collision and still gives us trillions of strategy IDs in the DB. I think this is alright for now. There's no `(bigint, bigint)` signature for advisory locks.

examples/v1/workflows/dag.go
```diff
@@ -23,19 +25,45 @@ type DagResult struct {
...
+	)
+
 	// ❓ Defining a Task
-	simple.Task(
+	step1 := simple.Task(
 		create.WorkflowTask[DagInput, DagResult]{
-			Name: "step",
+			Name:    "step1",
+			Retries: 3,
+			// RetryBackoffFactor:     2.0,
```
grutt: rm?

sql/schema/v1-core.sql
```diff
@@ -620,77 +607,133 @@ REFERENCING NEW TABLE AS new_table
...
-            v1_retry_queue_item rqi ON rqi.task_id = t.id AND rqi.task_inserted_at = t.inserted_at
+            v1_lookup_table lt ON wcs.workflow_run_id = lt.external_id AND lt.dag_id IS NOT NULL
         WHERE
-            cs.parent_strategy_id IS NOT NULL
-            AND rqi.task_id IS NULL
-    ), parent_slots_grouped AS (
+            wcs.strategy_id = p_strategy_id
+            AND wcs.workflow_version_id = p_workflow_version_id
+            AND wcs.workflow_run_id = p_workflow_run_id
+            AND NOT EXISTS (
```
abelanger5, line 645: Given that we're processing singletons, I'm not too concerned about `NOT EXISTS` but will need to test

sql/schema/v1-core.sql
```diff
@@ -620,77 +607,133 @@ REFERENCING NEW TABLE AS new_table
...
+    ), final_concurrency_slots_for_tasks AS (
+        -- If the workflow run id corresponds to a single task, we can safely delete the workflow concurrency slot
+        SELECT
+            wcs.strategy_id,
+            wcs.workflow_version_id,
+            wcs.workflow_run_id
+        FROM
+            v1_workflow_concurrency_slot wcs
+        JOIN
+            v1_lookup_table lt ON wcs.workflow_run_id = lt.external_id AND lt.task_id IS NOT NULL
```
abelanger5, line 675: Note to self: test single tasks with workflow concurrency slots

grutt (approved):

pkg/v1/workflow/declaration.go
grutt: image
good catch, i think the rest are ok -- will look closer tomorrow

examples/v1/workflows/dag.go
abelanger5: reverted those changes

sql/schema/v1-core.sql
```diff
@@ -620,77 +607,133 @@ REFERENCING NEW TABLE AS new_table
...
+            AND NOT EXISTS (
+                -- Check if any task in this DAG has a v1_concurrency_slot
+                SELECT 1
+                FROM v1_dag_to_task dt
+                JOIN v1_task t ON dt.task_id = t.id AND dt.task_inserted_at = t.inserted_at
+                JOIN v1_concurrency_slot cs2 ON cs2.task_id = t.id AND cs2.task_inserted_at = t.inserted_at AND cs2.task_retry_count = t.retry_count
+                WHERE dt.dag_id = lt.dag_id
+                AND dt.dag_inserted_at = lt.inserted_at
+            )
+            AND CARDINALITY(wcs.child_strategy_ids) <= (
```
mrkaye97, line 654: not sure how much you tested here & how big the number of records here is but I think `CARDINALITY` could be expensive?

cmd/hatchet-migrate/migrate/migrations/20250407012633_v1_0_11.sql
```diff
@@ -0,0 +1,749 @@
...
+        pcs.workflow_id,
+        pcs.workflow_version_id,
+        pcs.workflow_run_id,
+        pcs.parent_strategy_id,
+        pcs.child_strategy_ids,
+        pcs.priority,
+        pcs.key
+    FROM
+        parent_to_child_strategy_ids pcs
+    ON CONFLICT (strategy_id, workflow_version_id, workflow_run_id) DO NOTHING;
```
mrkaye97, line 57: what's the reason for not updating here?

mrkaye97 (approved):
left two small clarifying questions

sql/schema/v1-core.sql
abelanger5: `CARDINALITY` here will be very cheap since this is a column on the `v1_workflow_concurrency_slot` table, which is at most the number of steps in the DAG.

cmd/hatchet-migrate/migrate/migrations/20250407012633_v1_0_11.sql
abelanger5: The only field that we were previously updating was the `completed_child_strategy_ids` which is now deprecated (will add a comment). 

I'm wondering if there's an extremely rare edge case where we end up hitting the `ON CONFLICT` case concurrently with another query that's decided to delete the `v1_workflow_concurrency_slot` -- even though that query takes out a `FOR UPDATE` lock. I think `ON CONFLICT` acquires a `RowExclusiveLock` on the the primary key (https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/backend/executor/execIndexing.c;h=1f1181b56044a599398433a0f7feb45206a28abd;hb=HEAD) but I'm not 100% sure. 

In the worst case, however, we end up deleting the concurrency slot early, which is better than the alternative where the concurrency slot is never deleted.

PR #1503, opened 2025-04-07
fix: filter scope for bulk ops

grutt:
# Description
fixes scopes similar to #1502


sdks/typescript/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hatchet-dev/typescript-sdk",
-  "version": "1.1.3",
+  "version": "1.1.5",
```
mrkaye97, line 3: two version bump?

mrkaye97 (approved):

PR #1504, opened 2025-04-07
[Python]: Batch Bulk Run Calls

mrkaye97:
Internally batching bulk run calls to align with engine max size of 1000

[Better to review with whitespace hidden](https://github.com/hatchet-dev/hatchet/pull/1504/files?diff=unified&w=1)


abelanger5 (approved):
Nice! Assuming the Go SDK will the a different PR?

mrkaye97:
> Nice! Assuming the Go SDK will the a different PR?

Yeah, can do that separately

PR #1507, opened 2025-04-08
fix: match condition writes and retry counts on failure

abelanger5:
# Description

Fixes an issue related to replaying a workflow which has the following conditions:
- Multi-step DAG
- Workflow has CEL-based concurrency expression
- Parent step in the DAG failed and the child was cancelled
- The workflow is a child workflow
- Replaying the parent step in the DAG which failed would cause an `input is nil` error while parsing the CEL expression

This was due to an ordering issue when inserting matches into the database which unintentionally scrambled the `SIGNAL` conditions with the `TRIGGER` conditions (this is why the workflow needed to be both a child workflow and a DAG). 

Also improves handling of the failure events: the previous logic would ignore `v1_task_runtimes` which weren't associated with the most recent retry count on the task, which could cause the `v1_task_runtime` to never get removed from the DB in rare cases. This moves the check on retry count to the query which determines whether the failure event should trigger a retry, which ensures that we call `releaseTasks` on the runtimes which are still in the DB but shouldn't be. 

There are some additional OLAP improvements to make here (most importantly, not sending reassigned/timed out events which corresponded to a no-op in the database).  

## Type of change

### DIFF
--- a/internal/services/admin/v1/server.go
+++ b/internal/services/admin/v1/server.go
@@ -711,6 +711,36 @@ func getCreateTaskOpts(tasks []*contracts.CreateTaskOpts, kind string) ([]v1.Cre
 					})
 				}
 			}
+
+		}
+
+		if stepCp.Concurrency != nil {
+			for _, concurrency := range stepCp.Concurrency {
+				// Skip nil concurrency
+				if concurrency == nil {
+					continue
+				}
+
+				if concurrency.Expression == "" {
+					return nil, status.Error(
+						codes.InvalidArgument,
+						fmt.Sprintf("CEL expression is required for concurrency (step %s)", stepCp.ReadableId),
+					)
+				}
+
+				var limitStrategy *string
+
+				if concurrency.LimitStrategy != nil && concurrency.LimitStrategy.String() != "" {
+					s := concurrency.LimitStrategy.String()
+					limitStrategy = &s
+				}
+
+				steps[j].Concurrency = append(steps[j].Concurrency, v1.CreateConcurrencyOpts{
+					Expression:    concurrency.Expression,
+					MaxRuns:       concurrency.MaxRuns,
+					LimitStrategy: limitStrategy,
+				})
+			}
 		}
 	}
 
--- a/nodemon.api.json
+++ b/nodemon.api.json
@@ -1,13 +1,6 @@
 {
-    "restartable": "rs",
-    "verbose": true,
-    "ext": "go,mod,sum",
-    "watch": [
-        "cmd/hatchet-api",
-        "api",
-        "internal",
-        "go.mod",
-        "go.sum"
-    ]
-
+  "restartable": "rs",
+  "verbose": true,
+  "ext": "go,mod,sum",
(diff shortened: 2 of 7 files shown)


internal/services/admin/v1/server.go
```diff
@@ -711,6 +711,36 @@ func getCreateTaskOpts(tasks []*contracts.CreateTaskOpts, kind string) ([]v1.Cre
 					})
 				}
 			}
+
+		}
+
+		if stepCp.Concurrency != nil {
```
abelanger5, line 717: Drive-by to add in concurrency at the task level, which we weren't doing before

nodemon.api.json
```diff
@@ -1,13 +1,6 @@
...
-        "api",
-        "internal",
-        "go.mod",
-        "go.sum"
-    ]
-
+  "restartable": "rs",
+  "verbose": true,
+  "ext": "go,mod,sum",
+  "watch": ["cmd/hatchet-api", "api", "internal", "pkg", "go.mod", "go.sum"]
```
abelanger5, line 5: Adds `pkg` to the nodemon scripts so we get hot reload when those files are changed (not sure why this was working until now, we probably upgraded nodemon)

pkg/repository/v1/sqlcv1/tasks.sql
```diff
@@ -250,14 +251,10 @@ WITH input AS (
         t.step_id
     FROM
         v1_task t
-    -- only fail tasks which have a v1_task_runtime equivalent to the current retry count. otherwise,
-    -- a cancellation which deletes the v1_task_runtime might lead to a future failure event, which triggers
-    -- a retry.
     JOIN
-        v1_task_runtime rt ON rt.task_id = t.id AND rt.task_inserted_at = t.inserted_at AND rt.retry_count = t.retry_count
+        input i ON i.task_id = t.id AND i.task_inserted_at = t.inserted_at AND i.task_retry_count = t.retry_count
```
abelanger5, line 255: This only performs an increment of the retry count if the failure event corresponds to the latest retry count of the task, which should be correct (otherwise old retries could cause more retries)

pkg/repository/v1/match.go
```diff
@@ -863,6 +904,75 @@ func (m *sharedRepository) createEventMatches(ctx context.Context, tx sqlcv1.DBT
...
+	if existingTaskId != nil {
+		existingTaskIdStr = fmt.Sprintf("%d", *existingTaskId)
+	}
+
+	parentTaskIdStr := ""
+	if parentTaskId.Valid {
+		parentTaskIdStr = fmt.Sprintf("%d", parentTaskId.Int64)
+	}
+
+	return fmt.Sprintf("dag:%s:%d:%s:%s:%s:%s",
```
mrkaye97, line 939: this seems like it might be worth writing a test for

mrkaye97 (approved):
🤞

pkg/repository/v1/match.go
abelanger5: Will address in a separate PR, I'd like to set up a test runner for PG anyway

PR #1513, opened 2025-04-09
Feat: Priority

mrkaye97:
Implements priorities for runs - wiring up the APIs and adding the for Python SDK implementation for priority (mostly for testing purposes)


pkg/repository/postgres/workflow.go
```diff
@@ -1087,6 +1095,8 @@ func (r *workflowEngineRepository) createWorkflowVersionTxs(ctx context.Context,
 					String: "",
 					Valid:  true,
 				},
+				// TODO: Check this is right
```
mrkaye97, line 1105: @abelanger5 checking on this!

sdks/python/examples/priority/test_priority.py
```diff
@@ -0,0 +1,291 @@
...
+    test_run_id = str(uuid4())
+    choices: list[Priority] = ["low", "medium", "high", "default"]
+
+    run_refs = await priority_workflow.aio_run_many_no_wait(
+        [
+            priority_workflow.create_bulk_run_item(
+                options=TriggerWorkflowOptions(
+                    priority=(priority_to_int(priority := choice(choices))),
+                    additional_metadata={
+                        "priority": priority,
```
grutt, line 74: this confused me for a second because i thought we set priority through the additional meta prop... which maybe is a good idea...

mrkaye97: yeah I just wanted it to show up in the UI 🤷

mrkaye97: made it a lot easier to eyeball if this was working

abelanger5 (approved):

PR #1522, opened 2025-04-10
Feat  ts multiple wf concurrency

grutt:
# Description

Adds support for multiple workflow-level concurrency keys

## Type of change

## What's Changed

- [ ] Previously, workflows only supported 1 key

### DIFF
--- a/sdks/typescript/src/clients/worker/worker.ts
+++ b/sdks/typescript/src/clients/worker/worker.ts
@@ -288,6 +288,12 @@ export class V0Worker {
         ...(workflow.on?.cron ? [workflow.on.cron] : []),
       ];
 
+      const concurrencyArr = concurrency
+        ? Array.isArray(concurrency)
+          ? concurrency
+          : [concurrency]
+        : [];
+
       const registeredWorkflow = this.client.admin.putWorkflowV1({
         name: workflow.name,
         description: workflow.description || '',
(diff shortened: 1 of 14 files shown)


api-contracts/v1/workflows.proto
```diff
@@ -74,10 +74,13 @@ message CreateWorkflowVersionRequest {
     repeated string event_triggers = 4; // (optional) event triggers for the workflow
     repeated string cron_triggers = 5; // (optional) cron triggers for the workflow
     repeated CreateTaskOpts tasks = 6; // (required) the workflow jobs
+
+    // Deprecated: use concurrency_arr instead
     Concurrency concurrency = 7; // (optional) the workflow concurrency options
```
grutt: this is still required

sdks/typescript/src/clients/worker/worker.ts
```diff
@@ -327,6 +333,7 @@ export class V0Worker {
                 : [workflow.taskDefaults.concurrency]
               : [],
         })),
+        concurrency: undefined,
```
mrkaye97, line 336: ah good call - I didn't do this. is any idea if specifying both `concurrency` and `concurrency_arr` is broken? I'll mimic this in python

grutt: not sure

grutt: this a type requirement bc the proto wasnt actually optional

abelanger5 (approved):

PR #1541, opened 2025-04-11
Feat: Lifespans!

mrkaye97:
# Description

Lifespans (inspired by [FastAPI's](https://fastapi.tiangolo.com/advanced/events/)) allow you to share state across all tasks running on a worker, which is useful for creating connection pools, loading models into memory, and so on. If you don't supply a `lifespan` to the worker, it's a no-op (so the behavior is the same as now).

Other drive-by fixes here:

* Added a license
* Added a couple of errors for if you try to register an empty workflow or a worker with no workflows at all
* Added `durable_slots` to the worker to allow that to be customizable, with the same default of 1000


sdks/python/pyproject.toml
```diff
@@ -1,10 +1,11 @@
 [tool.poetry]
 name = "hatchet-sdk"
-version = "1.3.2"
+version = "1.4.0"
 description = ""
 authors = ["Alexander Belanger <alexander@hatchet.run>"]
 readme = "README.md"
 include = ["hatchet_sdk/py.typed"]
+license = "MIT"
```
mrkaye97, line 8: want to make sure this is fine - do we have some custom license we want to add?

sdks/python/examples/lifespans/worker.py
```diff
@@ -0,0 +1,85 @@
...
+
+        return TaskOutput(
+            num_rows=len(rows),
+            external_ids=[cast(UUID, row[0]) for row in rows],
+        )
+
+
+@lifespan_workflow.task()
+async def async_lifespan_task(input: EmptyModel, ctx: Context) -> TaskOutput:
+    pool = cast(ConnectionPool, ctx.lifespan["pool"])
```
grutt: i'm assuming this not typed?

sdks/python/hatchet_sdk/context/context.py
```diff
@@ -116,6 +119,10 @@ def was_triggered_by_event(self) -> bool:
     def workflow_input(self) -> JSONSerializableMapping:
         return self.input
 
+    @property
+    def lifespan(self) -> Any:
```
grutt, line 123: python ❤️

sdks/python/examples/lifespans/worker.py
mrkaye97: yeah, I don't think it's possible unfortunately unless we want to require you to pass the generic to every workflow or something. FastAPI's aren't typed either 🤷

sdks/python/hatchet_sdk/context/context.py
mrkaye97: just don't want people to overwrite it lol

frontend/docs/pages/home/lifespans.mdx
```diff
@@ -0,0 +1,36 @@
...
+import UniversalTabs from "@/components/UniversalTabs";
+import { GithubSnippet, getSnippets } from "@/components/code";
+
+export const Lifespans = {
+  path: "examples/lifespans/worker.py",
+};
+
+export const getStaticProps = ({}) => getSnippets([Lifespans]);
+
+# Lifespans
```
grutt, line 11: Let's add a note that this feature is experimental and signatures may change

mrkaye97: Good call

grutt (approved):

PR #1547, opened 2025-04-14
Fe overhaul  run form

grutt:
# Description

Adds a trigger modal

### DIFF
--- a/frontend/app/src/next/hooks/use-crons.ts
+++ b/frontend/app/src/next/hooks/use-crons.ts
@@ -7,6 +7,7 @@ import {
   useMutation,
   UseMutationResult,
   useQuery,
+  useQueryClient,
 } from '@tanstack/react-query';
 import useTenant from './use-tenant';
 import {
@@ -58,6 +59,7 @@ export default function useCrons({
   paginationManager = PaginationManagerNoOp,
 }: UseCronsOptions = {}): CronsState {
   const { tenant } = useTenant();
+  const queryClient = useQueryClient();
 
   const listCronsQuery = useQuery({
     queryKey: ['cron:list', tenant, filters, paginationManager],
@@ -102,7 +104,7 @@ export default function useCrons({
       return res.data;
     },
     onSuccess: () => {
-      listCronsQuery.refetch();
+      queryClient.invalidateQueries({ queryKey: ['cron:list'] });
     },
   });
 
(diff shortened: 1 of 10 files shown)


frontend/app/src/next/hooks/use-crons.ts
```diff
@@ -102,7 +104,7 @@ export default function useCrons({
       return res.data;
     },
     onSuccess: () => {
-      listCronsQuery.refetch();
+      queryClient.invalidateQueries({ queryKey: ['cron:list'] });
```
mrkaye97, line 107: can we grab this query from `queries.cronList` or whatever it's called? maybe `queries.cronList.queryKey` or something?

frontend/app/src/next/hooks/use-runs.tsx
```diff
@@ -181,6 +181,30 @@ export function useRuns({
     },
   });
 
+  const triggerNowMutation = useMutation({
+    mutationKey: ['workflow-run:create', tenant?.metadata.id],
+    mutationFn: async (data: {
+      workflowName: string;
+      input: object;
+      additionalMetadata: object;
```
mrkaye97, line 189: as of now we'll want priority here too I think :P

frontend/app/src/next/hooks/use-runs.tsx
```diff
@@ -181,6 +181,30 @@ export function useRuns({
...
+      const res = await api.v1WorkflowRunCreate(tenant.metadata.id, {
+        workflowName: data.workflowName,
+        input: data.input,
+        additionalMetadata: data.additionalMetadata,
+      });
+
+      return res.data;
+    },
+    onSuccess: () => {
+      listRunsQuery.refetch();
```
mrkaye97, line 204: maybe use `invalidateQueries` here too?

frontend/app/src/next/hooks/use-crons.ts
grutt: good idea, making a note to do this for all

frontend/app/src/next/hooks/use-runs.tsx
grutt: will need to generate and add this throughout

frontend/app/src/next/hooks/use-runs.tsx
grutt: adding a note to do this throughout

PR #1555, opened 2025-04-15
hotfix: priority nil pointer

abelanger5:
# Description

Patches a nil pointer exception on priority w/ cron triggers. 

## Type of change


mrkaye97 (approved):
thanks!

PR #1566, opened 2025-04-16
Fix  v1 admin

grutt:
# Description

Fixes # (issue)

## Type of change

## What's Changed

- [x] runNoWait methods are now async -- this is required so that we don't overwhelm the GRPC channel.


sdks/typescript/src/v1/client/admin.ts
```diff
@@ -0,0 +1,229 @@
...
+
+  /**
+   * Run a new instance of a workflow with the given input. This will create a new workflow run and return the ID of the
+   * new run.
+   * @param workflowName the name of the workflow to run
+   * @param input an object containing the input to the workflow
+   * @param options an object containing the options to run the workflow
+   * @returns the ID of the new workflow run
+   */
+  async runWorkflow<Q = object, P = object>(
```
mrkaye97, line 54: qq: are `Q` and `P` named that for some reason?

sdks/typescript/src/v1/client/admin.ts
```diff
@@ -0,0 +1,229 @@
...
+      };
+    }>,
+    batchSize: number = 500
+  ): Promise<WorkflowRunRef<P>[]> {
+    // Prepare workflows to be triggered in bulk
+    const workflowRequests = workflowRuns.map(({ workflowName, input, options }) => {
+      let computedName = workflowName;
+
+      if (this.config.namespace && !workflowName.startsWith(this.config.namespace)) {
+        computedName = this.config.namespace + workflowName;
```
mrkaye97, line 123: qq: do we need an underscore here or is that magicked away somehow?

sdks/typescript/src/v1/client/admin.ts
```diff
@@ -0,0 +1,229 @@
...
+
+    this.logger.debug(`batching ${batches.length} batches`);
+
+    try {
+      const allResults: WorkflowRunRef<P>[] = [];
+
+      // Process each batch
+      for (let batchIndex = 0; batchIndex < batches.length; batchIndex += 1) {
+        const batch = batches[batchIndex];
+        const { requests, originalIndices } = batch;
```
mrkaye97: qq: can we use `forEach` to loop through the batches directly? seems like we only use the index to get the batch again

sdks/typescript/src/v1/client/admin.ts
```diff
@@ -0,0 +1,229 @@
...
+
+        // Call the bulk trigger workflow method for this batch
+        const bulkTriggerWorkflowResponse = await this.grpc.bulkTriggerWorkflow(
+          BulkTriggerWorkflowRequest.create({
+            workflows: requests,
+          })
+        );
+
+        this.logger.debug(`batch ${batchIndex + 1} of ${batches.length}`);
```
mrkaye97, line 156: qq: do we want to log this? I think it'll be really noisy

sdks/typescript/src/v1/client/admin.ts
```diff
@@ -0,0 +1,229 @@
...
+
+        // Map the results back to their original indices
+        const batchResults = bulkTriggerWorkflowResponse.workflowRunIds.map((resp, index) => {
+          const originalIndex = originalIndices[index];
```
mrkaye97, line 160: this makes me nervous - can we do this without needing to rely on index slices here by returning the parentId from somewhere? it looks like that's the only thing we use from here

sdks/typescript/src/v1/client/admin.ts
```diff
@@ -0,0 +1,229 @@
...
+      return allResults;
+    } catch (e: any) {
+      throw new HatchetError(e.message);
+    }
+  }
+
+  private batchWorkflows(
+    workflowRequests: any[],
+    batchSize: number
+  ): Array<{ requests: any[]; originalIndices: number[] }> {
```
mrkaye97: I think this is worth writing unit tests for

sdks/typescript/src/v1/client/admin.ts
```diff
@@ -0,0 +1,229 @@
...
+    let currentBatchSize = 0;
+
+    for (let i = 0; i < workflowRequests.length; i += 1) {
+      const request = workflowRequests[i];
+      const requestSize = Buffer.byteLength(JSON.stringify(request), 'utf8');
+
+      // Check if adding this request would exceed either the payload limit or batch size
+      if (currentBatchSize + requestSize > payloadLimit || currentBatch.length >= batchSize) {
+        // If we have a batch, add it to batches
+        if (currentBatch.length > 0) {
```
mrkaye97: qq: do we need to check this condition? how could either of the conditions above be true and this be false?

sdks/typescript/src/v1/client/admin.ts
```diff
@@ -0,0 +1,229 @@
...
+      }
+
+      return allResults;
+    } catch (e: any) {
+      throw new HatchetError(e.message);
+    }
+  }
+
+  private batchWorkflows(
+    workflowRequests: any[],
```
mrkaye97: qq: possible to use a generic here instead of `any`?

mrkaye97 (commented):
Left some questions about things - looks like lots of breaking changes here? Context, worker types, run workflow, ...?

grutt:
> Left some questions about things - looks like lots of breaking changes here? Context, worker types, run workflow, ...?

The only breaking change should be the run no waits are now async -- the rest are deliberately internal from the v1 release -- i may need to confirm I'm exporting the correct Context type on root

sdks/typescript/src/v1/client/admin.ts
grutt: i set the underscore on config load so we dont have to worry about it anywhere else

mrkaye97: awesome, sounds great 🙏

sdks/typescript/src/v1/client/admin.ts
grutt: if yes, the reason is no longer in my cache

sdks/typescript/src/v1/client/admin.ts
grutt: good catch, that was my intent writing this separately

sdks/typescript/src/v1/client/admin.ts
grutt: rewrote this but yes first payloads can be too large and we're not failing on that case

sdks/typescript/src/v1/client/admin.ts
grutt: its not guaranteed all parents are the same... for reasons that are ambitious and future scope i think this is fine for now

sdks/typescript/src/v1/client/admin.ts
grutt: its debug only and _very_ helpful, i think we leave it for now

sdks/typescript/src/v1/client/admin.ts
grutt: tl;dr no... that was part of the problem with the old implementation -- this needs to be guaranteed serial and for is the best way to do that

mrkaye97 (approved):

PR #1573, opened 2025-04-18
[Python]: Fixing some more issues

mrkaye97:
# Description

* Default priority wasn't being sent from `hatchet.task` or `hatchet.durable_task`
* Returning a class from the `read_with_interrupt` thing for a little bit safer type checking

## Type of change

### DIFF
--- a/sdks/python/hatchet_sdk/clients/dispatcher/action_listener.py
+++ b/sdks/python/hatchet_sdk/clients/dispatcher/action_listener.py
@@ -9,7 +9,11 @@ import grpc
 import grpc.aio
 from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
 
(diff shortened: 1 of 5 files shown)


sdks/python/hatchet_sdk/clients/dispatcher/action_listener.py
```diff
@@ -275,15 +279,17 @@ async def _generator(self) -> AsyncGenerator[Action | None, None]:
 
                         break
 
-                    assigned_action, _, is_eof = t.result()
+                    result = t.result()
 
-                    if is_eof:
+                    if isinstance(result, UnexpectedEOF):
```
abelanger5, line 284: Do we have a reliable repro/has this been tested?

mrkaye97: worked fine locally but happy to test against staging if you want!

abelanger5 (approved):

PR #1576, opened 2025-04-18
fix: strict typing on populator

grutt:
# Description

Fixes 404 if cron is not found

## Type of change


api/v1/server/oas/transformers/workflow_run.go
```diff
@@ -549,6 +549,43 @@ func ToScheduledWorkflowsFromSQLC(scheduled *dbsqlc.ListScheduledWorkflowsRow) *
...
+
+	if cron.Input != nil {
+		err := json.Unmarshal(cron.Input, &input)
+		if err != nil {
+			return nil
+		}
+	}
+
+	res := &gen.CronWorkflows{
+		Metadata:           *toAPIMetadata(sqlchelpers.UUIDToStr(cron.ID_2), cron.CreatedAt_2.Time, cron.UpdatedAt_2.Time),
```
mrkaye97, line 572: nit: I think we should modify the query to explicitly select which created at we want here, since this seems very bug prone to me

grutt: i think i did already

PR #1582, opened 2025-04-20
wip: nestjs - hatchet fullstack example

### LINKED ISSUE #1581: [QUESTION] When queuing tasks, needing object references to the actual tasks/workflow makes it very difficult to use in DI frameworks
Given a simple task:

```ts
export type SimpleInput = {
  Message: string;
};

export const simple = hatchet.task({
  name: 'simple',
  fn: (input: SimpleInput) => {
    return {
      TransformedMessage: input.Message.toLowerCase(),
    };
  },
});
```

Providing this task in a Dependency Injection framework is easy.

But when using multiple processes it gets tricky. Lets say:

```
Process A: Node web framework that uses the tasks call `.run` or `.runNoWait` etc.
Process B: Node worker process that creates the workers and runs the tasks
```

If you are using a Dependency Injection framework (NestJS in my case), it becomes very frustrating to deal with the fact that the worker process needs object references to the Hatchet tasks. If these Hatchet Tasks have been modified to be provided by a DI container, the `fn` of these tasks is not available statically somewhere. It gets created when the DI container is initializing.

In the worker process, I would be forced to initialize the same DI dependencies and tasks in order to receive the entire Hatchet Task object with the correct `fn`. This in turn would mean that Process A initializes the tasks, and Process B initializes them as well.

The part I'm having trouble with is having to also then initialize all DI dependencies in Process A and B just to: queue the tasks, and to run the tasks.

Any pointers on how I should deal with this?

grutt:
# Description

Simple example to share how to use Hatchet within Nestjs

Fixes #1581 

## Type of change

### DIFF
--- a/fullstack-examples/nest-hatchet/.gitignore
+++ b/fullstack-examples/nest-hatchet/.gitignore
@@ -1,0 +1,56 @@
+# compiled output
+/dist
+/node_modules
+/build
+
+# Logs
+logs
+*.log
+npm-debug.log*
+pnpm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+lerna-debug.log*
+
+# OS
+.DS_Store
+
+# Tests
+/coverage
+/.nyc_output
+
+# IDEs and editors
+/.idea
+.project
+.classpath
+.c9/
+*.launch
+.settings/
+*.sublime-workspace
+
+# IDE - VSCode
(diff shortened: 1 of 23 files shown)


mrkaye97:
closing since this is pretty stale, can always reopen if we want to ship this later

PR #1592, opened 2025-04-22
feat: sampling

abelanger5:
# Description

Adds sampling for the monitoring controller. Can be enabled via env vars:

```
SERVER_SAMPLING_ENABLED=t
SERVER_SAMPLING_RATE=0.1
```

Known issues:
- Parent/child workflows don't maintain sampling consistency
- Should have a way to retain all failed
- Workflow runs can't be accessed directly in the UI if they're not sampled (which we can eventually fix with proxying to the gRPC service)

## Type of change


frontend/docs/pages/self-hosting/sampling.mdx
```diff
@@ -0,0 +1,13 @@
+# Sampling
```
grutt: nit: Trace or UI Sampling

grutt (approved):

PR #1593, opened 2025-04-22
Fix: Wire up outputs in REST API

mrkaye97:
# Description

Sending back outputs with the `V1WorkflowRunDetails` correctly over the API

### DIFF
--- a/cmd/hatchet-migrate/migrate/migrations/20250422170746_v1_0_16.sql
+++ b/cmd/hatchet-migrate/migrate/migrations/20250422170746_v1_0_16.sql
@@ -1,0 +1,11 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE v1_tasks_olap
+ADD COLUMN step_readable_id TEXT NOT NULL DEFAULT '';
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+ALTER TABLE v1_tasks_olap
+DROP COLUMN step_readable_id;
+-- +goose StatementEnd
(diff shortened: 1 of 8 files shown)


cmd/hatchet-migrate/migrate/migrations/20250422170746_v1_0_16.sql
```diff
@@ -0,0 +1,11 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE v1_tasks_olap
+ADD COLUMN step_readable_id TEXT NOT NULL DEFAULT '';
```
mrkaye97, line 4: important: I was thinking making this non-null would give us improved behavior in the code + it should be non-null going forward, but happy to make this nullable if that doesn't make sense!

I don't think we need to backfill in either case

pkg/repository/v1/sqlcv1/olap.sql
```diff
@@ -1073,9 +1075,18 @@ WITH runs AS (
...
+            t.step_readable_id,
+            e.output
+        ) FILTER (
+            WHERE e.readable_status = 'COMPLETED'
+            AND e.output IS NOT NULL
+            AND t.step_readable_id <> ''
+        )::JSONB AS output
     FROM
         relevant_events e
+    JOIN v1_tasks_olap t ON (e.task_id, e.task_inserted_at) = (t.id, t.inserted_at)
```
mrkaye97, line 1089: I think this should be fine but I'm not 100% sure - it's using both PKs 🤷

PR #1594, opened 2025-04-22
Fix: Add Retries to Log Pushes

mrkaye97:
# Description

Seeing some gRPC errors on log pushes, so this will naively add retries there which should help a bit

## Type of change

### DIFF
--- a/sdks/python/hatchet_sdk/clients/events.py
+++ b/sdks/python/hatchet_sdk/clients/events.py
@@ -149,6 +149,7 @@ class EventClient:
             ).events
         )
 
+    @tenacity_retry
     def log(self, message: str, step_run_id: str) -> None:
         request = PutLogRequest(
             stepRunId=step_run_id,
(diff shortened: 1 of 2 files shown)


sdks/python/hatchet_sdk/clients/events.py
```diff
@@ -149,6 +149,7 @@ def bulk_push(
             ).events
         )
 
+    @tenacity_retry
```
grutt, line 152: right now we reject logs in some cases (i.e. a lot of logs.. rejecting might add a bunch of extra pressure) 

do we need to be more specific for what we're retrying?

mrkaye97: good call, I didn't consider that - I'm not sure actually. it seems like right now the main problem is just the pushes flaking out so this fixes that, but do you know what status we'd get back if we intentionally rejected? (or is there some other way to know?)

grutt:
No I'm not sure.. probably 429

sdks/python/hatchet_sdk/clients/events.py
mrkaye97: > No I'm not sure.. probably 429

In this case, I do think we should retry

PR #1617, opened 2025-04-24
feat: read replica support and docs

abelanger5:
# Description

Adds support for configuring a read replica for the OLAP repository, along with some basic docs explaining how they can be configured.

## Type of change

### DIFF
--- a/pkg/repository/v1/olap.go
+++ b/pkg/repository/v1/olap.go
@@ -181,6 +181,8 @@ type UpdateDAGStatusRow struct {
 
 type OLAPRepository interface {
 	UpdateTablePartitions(ctx context.Context) error
+	SetReadReplicaPool(pool *pgxpool.Pool)
+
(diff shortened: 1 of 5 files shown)


pkg/repository/v1/olap.go
```diff
@@ -303,6 +308,10 @@ func (o *OLAPRepositoryImpl) UpdateTablePartitions(ctx context.Context) error {
 	return nil
 }
 
+func (o *OLAPRepositoryImpl) SetReadReplicaPool(pool *pgxpool.Pool) {
+	o.readPool = pool
```
mrkaye97, line 312: so the idea here is by default it'll use the leader unless a replica is provided?

mrkaye97 (approved):

pkg/repository/v1/olap.go
abelanger5: yep, exactly

PR #1620, opened 2025-04-25
Fix: Retry + Cancel Bugs

mrkaye97:
# Description

Fixing a couple of bugs that were popping up around retries + cancels:

* The retry count of the cancel event being sent from the engine was of the latest task, causing a race where a more recent task (a retry) could start, and then immediately be cancelled, leaving the old one running still
* Fixing a bug in the Python SDK where a cancellation signal would cause the `cancel` method to be called on the API and cancel the retry too (basically multiple cancellations)
* Fixing how the SDK keys actions to identify them by pairs of step run id + retry count

## Type of change

TODO: Still need to fix the Go SDK here

### DIFF
--- a/internal/services/dispatcher/dispatcher_v1.go
+++ b/internal/services/dispatcher/dispatcher_v1.go
@@ -314,10 +314,15 @@ func (d *DispatcherImpl) handleTaskCancelled(ctx context.Context, msg *msgqueuev
 
 	msgs := msgqueuev1.JSONConvert[tasktypesv1.SignalTaskCancelledPayload](msg.Payloads)
 
-	taskIds := make([]int64, 0)
+	taskIdsToRetryCounts := make(map[int64]int32)
 
 	for _, innerMsg := range msgs {
-		taskIds = append(taskIds, innerMsg.TaskId)
+		taskIdsToRetryCounts[innerMsg.TaskId] = innerMsg.RetryCount
+	}
+
+	taskIds := make([]int64, 0)
+	for taskId := range taskIdsToRetryCounts {
+		taskIds = append(taskIds, taskId)
 	}
 
 	tasks, err := d.repov1.Tasks().ListTasks(ctx, msg.TenantID, taskIds)
@@ -346,6 +351,18 @@ func (d *DispatcherImpl) handleTaskCancelled(ctx context.Context, msg *msgqueuev
 			d.l.Warn().Msgf("task %d not found", msg.TaskId)
 			continue
 		}
+
+		retryCount, ok := taskIdsToRetryCounts[msg.TaskId]
+		if !ok {
+			d.l.Warn().Msgf("task %d not found in retry counts", msg.TaskId)
+			continue
+		}
+
+		// Since we're handling cancellations, we need to use
+		// the _incoming_ retry count, not the one in the database,
+		// since we want to cancel the retry count of the task that was
+		// passed in.
+		task.RetryCount = retryCount
 
 		workerIdToTasks[msg.WorkerId] = append(workerIdToTasks[msg.WorkerId], task)
 	}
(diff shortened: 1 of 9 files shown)


internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -314,10 +314,15 @@ func (d *DispatcherImpl) handleTaskCancelled(ctx context.Context, msg *msgqueuev
 
 	msgs := msgqueuev1.JSONConvert[tasktypesv1.SignalTaskCancelledPayload](msg.Payloads)
 
-	taskIds := make([]int64, 0)
+	taskIdsToRetryCounts := make(map[int64]int32)
```
abelanger5: In theory (with buffering and all that) we could send cancellations from multiple retry counts. So perhaps this needs to be `make(map[int64][]int32)`?

internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -347,6 +352,18 @@ func (d *DispatcherImpl) handleTaskCancelled(ctx context.Context, msg *msgqueuev
...
+		if !ok {
+			d.l.Warn().Msgf("task %d not found in retry counts", msg.TaskId)
+			continue
+		}
+
+		// Since we're handling cancellations, we need to use
+		// the _incoming_ retry count, not the one in the database,
+		// since we want to cancel the retry count of the task that was
+		// passed in.
+		task.RetryCount = retryCount
```
abelanger5: Instead of overwriting the retry count, perhaps we pass it as an additional arg to the `CancelTask` method to be more explicit? I think it's potentially more confusing to overwrite this field.

sdks/python/hatchet_sdk/clients/dispatcher/action_listener.py
```diff
@@ -141,6 +144,18 @@ def otel_attributes(self) -> dict[str, str | int]:
...
 
+    @property
+    def key(self) -> ActionKey:
+        """
+        This key is used to uniquely identify a single step run by its id + retry count.
+        It's used when storing references to a task, a context, etc. in a dictionary so that
+        we can look up those items in the dictionary by a unique key.
+        """
+        if self.action_type == ActionType.START_GET_GROUP_KEY:
+            return f"{self.get_group_key_run_id}/{self.retry_count}"
```
abelanger5, line 155: The `retry_count` definitely isn't sent in v0 for group key runs -- does this handle the case where the retry count is null?

abelanger5 (changes requested):

sdks/python/hatchet_sdk/clients/dispatcher/action_listener.py
mrkaye97: oh, interesting 👀  - I'm a bit confused by how this was ever working, because we had `retry_count: int` in the v0 SDK. maybe since it wasn't pydantic it wasn't doing any validation.

in any case, I don't think we need to handle v0 here since you can't use the v1 SDK with v0 anyways?

internal/services/dispatcher/dispatcher_v1.go
mrkaye97: good idea, I'll update that

mrkaye97: are you thinking something more like this? https://github.com/hatchet-dev/hatchet/pull/1620/commits/75ba13eedf839fcd87eba16407e5897651a6b018

sdks/python/hatchet_sdk/clients/dispatcher/action_listener.py
abelanger5: gotcha, sounds good

abelanger5 (approved):

PR #1621, opened 2025-04-25
fix: critical deadlock bug in scheduler

abelanger5:
# Description

Fixes a deadlock when cleaning up a tenant in the scheduler, which can happen when a tenant is moved to a different partition during restarts or redeploys.  

The deadlock happens when the `Cleanup()` method is called on the `LeaseManager`. The lease manager is responsible for acquiring leases on workers, queues, and concurrency strategies. Leases are refreshed very often, and if we're not processing the lease operations, the queues, workers, and strategies that we're processing will all remain constant.  

The deadlocking was due to mutexes being acquired out of order:
1. The `Cleanup` method acquires a lock the `cleanupMu` and then acquires a lock on the worker, queue, and concurrency strategy mutexes. 
2. In the `acquire<Resource>Leases` methods, we acquire a lock on the worker, queue or concurrency strategy mutexes, and then acquire a lock on the `cleanupMu`. 

So we could end up in a state where one of the `acquire` methods are waiting for the `cleanupMu` lock and the `Cleanup` method is waiting for a resource lock. 

This led to some interesting downstream impact:
1. Because the lease manager wasn't updating its queues, each of the existing queues would continue to process, sometimes with or without workers. So this would lead to random `scheduling timed out` or `no worker available` errors because the lease's were completely out of sync (for example, a queue lease still being held with all worker leases expired). 

3. While the `queuer` is somewhat concurrency-safe since we use `for update skip locked`, the queue items would be read properly but there would be additional locks on other tables, like rate limits. This led to observed deadlocking on the rate limit tables (which we will get to later). 

The fix here simplifies lock management of the `LeaseManager` to a single process-level lock, which guards any lease-modifying operation from running concurrently. The resource-level locks aren't necessary after we added the cleanup mutex, which was effectively trying to operate as a process mutex. 

Based on the behavior observed downstream, there's an additional fix for guarding the rate limiter with an advisory lock in the database based on an FNV hash of the tenant id. The chance of collisions here is extremely unlikely (something like 1 in the 10^12). 

## Type of change

### DIFF
--- a/internal/services/scheduler/v1/scheduler.go
+++ b/internal/services/scheduler/v1/scheduler.go
@@ -193,8 +193,9 @@ func (s *Scheduler) Start() (func() error, error) {
 	_, err := s.s.NewJob(
 		gocron.DurationJob(time.Second*1),
 		gocron.NewTask(
-			s.runTenantSetQueues(ctx),
+			s.runSetTenants(ctx),
 		),
+		gocron.WithSingletonMode(gocron.LimitModeReschedule),
 	)
 
 	if err != nil {
@@ -332,7 +333,7 @@ func (s *Scheduler) handleCheckQueue(ctx context.Context, msg *msgqueue.Message)
 	return nil
 }
 
-func (s *Scheduler) runTenantSetQueues(ctx context.Context) func() {
+func (s *Scheduler) runSetTenants(ctx context.Context) func() {
 	return func() {
 		s.l.Debug().Msgf("partition: checking step run requeue")
 
--- a/pkg/repository/v1/scheduler_concurrency.go
+++ b/pkg/repository/v1/scheduler_concurrency.go
@@ -68,7 +68,7 @@ func (c *ConcurrencyRepositoryImpl) UpdateConcurrencyStrategyIsActive(
 
 	defer rollback()
 
-	err = c.queries.ConcurrencyAdvisoryLock(ctx, tx, strategy.ID)
+	err = c.queries.AdvisoryLock(ctx, tx, strategy.ID)
 
 	if err != nil {
 		return err
@@ -131,7 +131,7 @@ func (c *ConcurrencyRepositoryImpl) runGroupRoundRobin(
 
 	defer rollback()
 
-	err = c.queries.ConcurrencyAdvisoryLock(ctx, tx, strategy.ID)
+	err = c.queries.AdvisoryLock(ctx, tx, strategy.ID)
 
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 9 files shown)


internal/services/scheduler/v1/scheduler.go
```diff
@@ -332,7 +333,7 @@ func (s *Scheduler) handleCheckQueue(ctx context.Context, msg *msgqueue.Message)
 	return nil
 }
 
-func (s *Scheduler) runTenantSetQueues(ctx context.Context) func() {
+func (s *Scheduler) runSetTenants(ctx context.Context) func() {
```
abelanger5, line 336: Renamed for clarity

pkg/repository/v1/scheduler_concurrency.go
```diff
@@ -68,7 +68,7 @@ func (c *ConcurrencyRepositoryImpl) UpdateConcurrencyStrategyIsActive(
 
 	defer rollback()
 
-	err = c.queries.ConcurrencyAdvisoryLock(ctx, tx, strategy.ID)
+	err = c.queries.AdvisoryLock(ctx, tx, strategy.ID)
```
abelanger5, line 71: Renamed `ConcurrencyAdvisoryLock` to `AdvisoryLock` since it can be used wherever.

pkg/repository/v1/scheduler_rate_limit.go
```diff
@@ -95,3 +85,10 @@ func (d *rateLimitRepository) UpdateRateLimits(ctx context.Context, tenantId pgt
 
 	return res, &nextRefillAt, err
 }
+
+func tenantAdvisoryInt(tenantID string) int64 {
+	hasher := fnv.New64a()
```
abelanger5, line 90: Uses an FNV hash which we're using in other places for hash buckets.

pkg/scheduling/v1/lease_manager.go
```diff
@@ -325,78 +295,82 @@ func (l *LeaseManager) loopForLeases(ctx context.Context) {
...
-
 		return l.lr.ReleaseLeases(ctx, l.tenantId, l.concurrencyLeases)
 	})
 
 	if err := eg.Wait(); err != nil {
 		return err
 	}
 
+	// close channels: this is safe to do because each channel is guarded by l.cleanedUp + the process lock
+	close(l.workersCh)
```
abelanger5, line 370: Removed the `defer` when closing the channels since it's more difficult to reason about Go's LIFO stacking of defers.

pkg/scheduling/v1/queuer.go
```diff
@@ -512,59 +511,3 @@ func (q *Queuer) flushToDatabase(ctx context.Context, r *assignResults) int {
 
 	return len(succeeded) + len(r.schedulingTimedOut)
 }
-
-func getLargerDuration(s1, s2 string) (string, error) {
```
abelanger5, line 516: Removed unused code

pkg/scheduling/v1/tenant_manager.go
```diff
@@ -203,28 +201,6 @@ func (t *tenantManager) setConcurrencyStrategies(strategies []*sqlcv1.V1StepConc
 	t.concurrencyStrategies = newArr
 }
 
-func (t *tenantManager) refreshAll(ctx context.Context) {
```
abelanger5, line 206: Also unused

pkg/repository/v1/scheduler_rate_limit.go
```diff
@@ -21,25 +22,6 @@ func newRateLimitRepository(shared *sharedRepository) *rateLimitRepository {
 	}
 }
 
-func (d *rateLimitRepository) ListCandidateRateLimits(ctx context.Context, tenantId pgtype.UUID) ([]string, error) {
```
abelanger5, line 24: Removed unused code.

mrkaye97 (approved):

PR #1622, opened 2025-04-25
Fix: Queue blocking on many concurrency keys + failures

mrkaye97:
# Description

Fixes an issue where work can grind to a halt if cleanup isn't performed correctly on concurrency slots after they should be processed / cleaned up

## Type of change


abelanger5 (approved):

abelanger5 (changes requested):
We need a migration file for updating the trigger

abelanger5 (approved):

PR #1625, opened 2025-04-26
fix: cancel if behavior

abelanger5:
# Description

Fixes the behavior of `cancel_if` so that if a user overrides the behavior, we still register the cancellations on failure and upstream cancellation. The `cancel_if` behavior can only be overwritten based on a completed upstream task. 

## Type of change


mrkaye97 (approved):
🥳

PR #1631, opened 2025-04-28
test: improve Go testing harness

abelanger5:
# Description

Significantly improves our testing harness:
- [X] Uses `testcontainers` to create Postgres and RabbitMQ, which removes the requirement of having these components running through `docker-compose`
- [X] Adds a testing matrix for testing against:
  - Postgres 15, 16 and 17
  - Using RabbitMQ as the message broker vs using Postgres
  - Running tests again the "penultimate" vs "latest" migration files, to ensure that a new migration won't impact existing queues
- [X] Improves the usage of `goleak` so we can verify there are no unexpected goroutines whenever we use the engine

I'd like to use this pattern for all of our e2e and load tests on the Go side moving forward. 

## Type of change


mrkaye97 (approved):
This is really nice 🥳

PR #1633, opened 2025-04-28
Feat: Events in the OLAP Repo

mrkaye97:
# Description

Adding two tables `v1_events_olap` and `v1_event_to_run_olap` to the db to store events + do lookups between runs + event triggers (if there are any)

Also includes changes from #1634 and #1676 

## Type of change


cmd/hatchet-migrate/migrate/migrations/20250428175026_v1_0_17.sql
```diff
@@ -0,0 +1,34 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE v1_events_olap (
+    tenant_id UUID NOT NULL,
+    id BIGINT GENERATED ALWAYS AS IDENTITY,
+    inserted_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    generated_at TIMESTAMPTZ NOT NULL,
```
abelanger5: Is this referring to the original event timestamp?

abelanger5 (commented):
Looks really solid!

cmd/hatchet-migrate/migrate/migrations/20250428175026_v1_0_17.sql
mrkaye97: yeah I was thinking we'd want a timestamp corresponding to the time the event was pushed, not sure if we should consider that the "insert time" though

sql/schema/v1-olap.sql
```diff
@@ -373,6 +373,28 @@ BEGIN
...
+CREATE TABLE v1_event_to_run_olap (
+    run_id BIGINT NOT NULL,
+    run_inserted_at TIMESTAMPTZ NOT NULL,
+    event_id UUID NOT NULL,
+    event_seen_at TIMESTAMPTZ NOT NULL,
+
+    PRIMARY KEY (event_id, event_seen_at, run_id, run_inserted_at)
+) PARTITION BY RANGE(event_seen_at);
+
+
```
mrkaye97, line 399: was considering adding indexes here for filtering by key (at least) - curious for thoughts there

abelanger5 (approved):

pkg/repository/v1/sqlcv1/olap.sql
```diff
@@ -4,7 +4,9 @@ SELECT
     create_v1_hash_partitions('v1_task_status_updates_tmp'::text, @partitions::int),
     create_v1_olap_partition_with_date_and_status('v1_tasks_olap'::text, @date::date),
     create_v1_olap_partition_with_date_and_status('v1_runs_olap'::text, @date::date),
-    create_v1_olap_partition_with_date_and_status('v1_dags_olap'::text, @date::date);
+    create_v1_olap_partition_with_date_and_status('v1_dags_olap'::text, @date::date),
+    create_v1_range_partition('v1_events_olap'::text, @date::date),
+    create_v1_range_partition('v1_event_to_run_olap'::text, @date::date);
 
 -- name: ListOLAPPartitionsBeforeDate :many
```
abelanger5, line 11: We'll need to return event partitions before this date and make sure we detach old partitions from the events tables

abelanger5 (changes requested):

pkg/repository/v1/sqlcv1/olap.sql
mrkaye97: done in [this commit](https://github.com/hatchet-dev/hatchet/pull/1633/commits/da579ab280c120d5b3eb95dda89eda8a1ad3dd47) I think - is there anything else that'll need to be added for this? I grepped for `DETACH PARTITION` and this was what I found

abelanger5: This looks correct but the best way to test it to create partitions older than the retention window and make sure that they get cleaned up on server startup

mrkaye97: copy that, will give it a shot

mrkaye97: okay yep, just confirmed that works!

mrkaye97:
FYI @abelanger5, I bundled in the request for priority on event push in these two commits:
* https://github.com/hatchet-dev/hatchet/pull/1633/commits/9181f32176e41be1c82edc77f7d1d8dc19646306
* https://github.com/hatchet-dev/hatchet/pull/1633/commits/dc50db64783ebd4871d455db0b90089ceea01c71

abelanger5 (approved):

PR #1634, opened 2025-04-28
Feat: Write events into OLAP tables

mrkaye97:
# Description

Wiring up event writes into the OLAP tables by sending event + task metadata over the message queue to the OLAP repo and then writing the events and tying them to their corresponding task runs

## Type of change

- [ ] Figure out correct value to use for the `generated_at` timestamp


internal/services/ingestor/ingestor_v1.go
```diff
@@ -75,6 +76,26 @@ func (i *IngestorImpl) ingestSingleton(tenantId, key string, data []byte, metada
...
+				AdditionalMetadata: metadata,
+			},
+		)
+
+		if err != nil || olapEvent == nil {
+			return fmt.Errorf("could not create event in OLAP: %v", err)
+		}
+
+		return nil
+	}()
```
mrkaye97, line 97: I'm a Golang noob, but was thinking that doing this in a Goroutine so it's non-blocking would make sense? not sure if this is a good idea or not

internal/services/controllers/v1/olap/controller.go
```diff
@@ -332,6 +334,59 @@ func (tc *OLAPControllerImpl) handleCreatedDAG(ctx context.Context, tenantId str
...
+		for _, payload := range msg.Payloads {
+			opts := sqlcv1.CreateEventParams{
+				Tenantid:           sqlchelpers.UUIDFromStr(tenantId),
+				Generatedat:        sqlchelpers.TimestamptzFromTime(payload.EventGeneratedAt),
+				Key:                payload.EventKey,
+				Payload:            payload.EventPayload,
+				AdditionalMetadata: payload.EventAdditionalMetadata,
+			}
+
+			event, err := tc.repo.OLAP().CreateEvent(ctx, tenantId, opts)
```
mrkaye97: unsure about this - I think we might need to do it to get the id back, but it's a bit unfortunate to not be able to do this whole insert (both the events and the trigger bridge table) in one shot

pkg/repository/v1/trigger.go
```diff
@@ -153,27 +161,39 @@ func (r *TriggerRepositoryImpl) TriggerFromEvents(ctx context.Context, tenantId
...
 				input:              opt.Data,
 				additionalMetadata: additionalMetadata,
 			})
+
+			eventKeysTaskExternalIds = append(eventKeysTaskExternalIds, &EventDataForOlap{
+				EventKey:                workflow.EventKey,
+				TaskExternalId:          externalId,
+				EventAdditionalMetadata: additionalMetadata,
+				EventPayload:            opt.Data,
+			})
```
mrkaye97: wasn't sure about this but it seemed like a reasonable way to get the event data back to the message queue

internal/services/controllers/v1/olap/controller.go
abelanger5: I'm not quite understanding, why can't this all be in the same TX?

pkg/repository/v1/olap.go
```diff
@@ -211,6 +211,8 @@ type OLAPRepository interface {
 	ListTasksByExternalIds(ctx context.Context, tenantId string, externalIds []string) ([]*sqlcv1.FlattenTasksByExternalIdsRow, error)
 
 	GetTaskTimings(ctx context.Context, tenantId string, workflowRunId pgtype.UUID, depth int32) ([]*sqlcv1.PopulateTaskRunDataRow, map[string]int32, error)
+	CreateEvent(ctx context.Context, tenantId string, event sqlcv1.CreateEventParams) (*sqlcv1.V1EventsOlap, error)
```
abelanger5: This should be a bulk insert no?

pkg/repository/v1/trigger.go
```diff
@@ -103,11 +103,19 @@ func newTriggerRepository(s *sharedRepository) TriggerRepository {
...
-func (r *TriggerRepositoryImpl) TriggerFromEvents(ctx context.Context, tenantId string, opts []EventTriggerOpts) ([]*sqlcv1.V1Task, []*DAGWithData, error) {
+type EventDataForOlap struct {
+	EventKey                string
+	EventId                 string
+	TaskExternalId          string
+	EventAdditionalMetadata []byte
+	EventPayload            []byte
+}
+
+func (r *TriggerRepositoryImpl) TriggerFromEvents(ctx context.Context, tenantId string, opts []EventTriggerOpts) ([]*sqlcv1.V1Task, []*DAGWithData, []*EventDataForOlap, error) {
```
abelanger5: Once we get to this many return types I think it's worth just returning a struct containing the arrays -- especially if we only ever return `nil, nil, nil, err` or all 3 arrays.

abelanger5: I'm also curious if we just want to return a `map[*EventData][]string` which maps the event to a set of task external ids.

internal/services/controllers/v1/olap/controller.go
mrkaye97: think it can now, I think this was an issue with the autogenerated int ids causing ordering problems but using the uuids should be fine

pkg/repository/v1/trigger.go
```diff
@@ -103,11 +103,25 @@ func newTriggerRepository(s *sharedRepository) TriggerRepository {
...
+	EventId                 string
+	TaskExternalId          string
+	EventAdditionalMetadata []byte
+	EventPayload            []byte
+}
+
+type TriggerFromEventsResult struct {
+	Tasks  []*sqlcv1.V1Task
+	Dags   []*DAGWithData
+	Events []*EventDataForOlap
```
abelanger5: I think making this something like `map[string][]*Run` will make life much easier here (will need to create a type for `Run`). You don't need to worry about the `n^2` problem above or removing duplicate events in the OLAP controller.

internal/services/controllers/v1/task/controller.go
```diff
@@ -819,7 +819,60 @@ func (tc *TasksControllerImpl) handleProcessUserEventTrigger(ctx context.Context
...
+	result, err := tc.repov1.Triggers().TriggerFromEvents(ctx, tenantId, opts)
+
+	if err != nil {
+		return fmt.Errorf("could not trigger tasks from events: %w", err)
+	}
+
+	eventTriggerOpts := make([]tasktypes.CreatedEventTriggerPayloadSingleton, 0)
+	eventSeenAt := time.Now()
+
+	for _, event := range result.Events {
```
abelanger5: (see note below first) I think this blows up much faster than we think -- these buffers can easily saturate to 1000 events + tasks at peak, so 1 million iterations?

pkg/repository/v1/trigger.go
```diff
@@ -103,11 +103,25 @@ func newTriggerRepository(s *sharedRepository) TriggerRepository {
 	}
 }
 
-func (r *TriggerRepositoryImpl) TriggerFromEvents(ctx context.Context, tenantId string, opts []EventTriggerOpts) ([]*sqlcv1.V1Task, []*DAGWithData, error) {
+type EventDataForOlap struct {
+	EventKey                string
+	EventId                 string
+	TaskExternalId          string
```
abelanger5: If we decide to keep this, `TaskExternalId` isn't as accurate since it could be a DAG external id as well

pkg/repository/v1/olap.go
```diff
@@ -1333,3 +1334,32 @@ func (r *OLAPRepositoryImpl) GetTaskTimings(ctx context.Context, tenantId string
 
 	return tasksWithData, idsToDepth, nil
 }
+
+func (r *OLAPRepositoryImpl) BulkCreateEventsAndTriggers(ctx context.Context, events []sqlcv1.BulkCreateEventsParams, triggers []sqlcv1.BulkCreateEventTriggersParams) error {
+	tx, err := r.pool.Begin(ctx)
```
abelanger5: If you trace up the context, this looks like it's just using `context.Background` and there's no timeout on this insert, which can be an issue (we need timeouts on all database calls). I think it's also useful to call `sqlchelpers.PrepareTx` with a statement timeout for any writes.

internal/services/controllers/v1/task/controller.go
```diff
@@ -819,7 +819,60 @@ func (tc *TasksControllerImpl) handleProcessUserEventTrigger(ctx context.Context
...
+				eventTriggerOpts = append(eventTriggerOpts, tasktypes.CreatedEventTriggerPayloadSingleton{
+					RunId:         task.ID,
+					RunInsertedAt: task.InsertedAt.Time,
+					// FIXME: Should `SeenAt` be set on the SDK when the event is created?
+					EventSeenAt:             eventSeenAt,
+					EventKey:                event.EventKey,
+					EventId:                 event.EventId,
+					EventPayload:            event.EventPayload,
+					EventAdditionalMetadata: event.EventAdditionalMetadata,
+				})
```
abelanger5: Probably missing a `break` statement here if we get a match, should we decide to keep this logic

abelanger5 (changes requested):

pkg/repository/v1/trigger.go
mrkaye97: There's a bunch of downstream stuff that uses the `Tasks` and `Dags` from here - do you think we should return the `map[string][]*Run` and then refactor the downstream things to use `Run` instead of e.g. `V1Task` and `DagWithData`?

internal/services/controllers/v1/task/controller.go
mrkaye97: I think we need to complete the loop here but I might be misunderstanding, since I was thinking an event can trigger many different tasks or dags, so it wouldn't be safe to break out early b/c we'd lose a triggered run that way

pkg/repository/v1/trigger.go
mrkaye97: good call, will remove / replace this

pkg/repository/v1/trigger.go
mrkaye97: ahhh nvm I think I'm understanding, one sec, was confused by the highlighted block vs. just this line

mrkaye97: took another stab at this but also just returning a list, but still prevents the N^2 thing afaict - lmk if you think that's fine!

internal/services/controllers/v1/task/controller.go
```diff
@@ -819,7 +819,45 @@ func (tc *TasksControllerImpl) handleProcessUserEventTrigger(ctx context.Context
...
+		},
+	)
+
+	if err != nil {
+		return fmt.Errorf("could not create event trigger message: %w", err)
+	}
+
+	tc.pubBuffer.Pub(ctx, msgqueue.OLAP_QUEUE, msg, false)
 
 	if err != nil {
```
abelanger5, line 862: This error now looks unused? This might be a sign that Go linters aren't installed in your VSCode env, since it should warn about an error, something like `nil != nil impossible condition`

abelanger5 (changes requested):

internal/services/controllers/v1/task/controller.go
mrkaye97: Nice catch, thanks! Shouldn't pre commit catch this?

abelanger5: It might not be a strict rule in the golangci-lint config -- there are a lot of rules that render as warnings in my editor without failing pre commit

PR #1537, opened 2025-04-11
feat--frontend-overhaul

grutt:
# Description

A complete rework of the Hatchet frontend UI/UX to support a more consistent, scalable, and navigable developer experience.

Fixes # (issue)

## Type of change

- [x]  🚨 Refactor (major structural/UI overhaul)

## Goals

This PR addresses core UX and structural issues that have grown with the platform:

- Establish consistent layout and component structure across pages
- Improve clarity and traceability of parent-child workflow runs
- Simplify navigation across services and settings
- Enhance maintainability for contributors by reducing ad-hoc patterns

---

## Structural Changes

- Feature hooks
  - [x] `use-alerter.ts` notification component (top right) -- right now only invites
  - [x] `use-api-meta.ts` api metadata for capabilities -- bundles cloud and oss calls
  - [x] `use-api-tokens.ts` api token crud
  - [x] `use-billing.ts` cloud only billing state
  - [x] `use-breadcrumbs.tsx` breadcrumb state for navigation
  - [x] `use-can.ts` permission based on user and tenant state
  - [x] `use-cloud-feature-flags.ts` cloud feature flags
  - [x] `use-crons.ts` cron run crud
  - [x] `use-definitions.ts` definition crud
  - [x] `use-docs-sheet.ts` doc sidebar state
  - [x] `use-filters.tsx` sharable filter state
  - [x] `use-managed-compute.ts` cloud managed compute state
  - [x] `use-members.tsx` tenant member state and invite crud
  - [x] `use-mobile.tsx` is mobile
  - [x] `use-pagination.tsx` sharable page state
  - [x] `use-ratelimits.ts` ratelimit crud
  - [x] `use-run-detail.tsx` detailed view for workflow and task runs
  - [x] `use-runs.tsx` run list and actions for runs
  - [x] `use-schedules.ts` schedules crud
  - [x] `use-support-chat.ts` cloud chat helpers
  - [x] `use-tenant.ts` tenant crud
  - [x] `use-timeline-context.tsx` state for the waterfall
  - [x] `use-user.ts` user crud
  - [x] `use-workers.ts` worker crud
- [x] Initial sharable filter and pagination hooks and components
- [x] Replace all routes and links with route builder utility
- [x] Replace doc links with exported doc link generation
- [x] Add `can` package for consistent authorization checks
- [x] Introduce `breadcrumbs` across relevant pages
- [x] Consistent layout structure and ShadCN UI usage
- [x] Add `/next` routing namespace for new UI shell
[...]

PR #1639, opened 2025-04-29
feat: fe overhaul api changes

grutt:
# Description
API changes for #1537


api/v1/server/handlers/v1/workflow-runs/get-timings.go
```diff
@@ -0,0 +1,126 @@
...
+func DFSOrder(tasks []gen.V1TaskTiming) []gen.V1TaskTiming {
+	// build set of task IDs
+	idSet := make(map[openapi_types.UUID]struct{}, len(tasks))
+	for _, t := range tasks {
+		idSet[t.TaskExternalId] = struct{}{}
+	}
+
+	// group children and collect roots
+	children := make(map[openapi_types.UUID][]gen.V1TaskTiming)
+	var roots []gen.V1TaskTiming
```
mrkaye97, line 68: not sure what's idiomatic but I've been making a slice of `V1TaskTiming` of the same length as `tasks` and filling it in in each iter of the loop (or in this case maybe just appending)

api/v1/server/run/run.go
```diff
@@ -301,10 +301,6 @@ func (t *APIServer) registerSpec(g *echo.Group, spec *openapi3.T) (*populator.Po
 			return nil, "", err
 		}
 
-		if scheduled == nil {
-			return nil, "", fmt.Errorf("cron workflow not found")
-		}
-
```
mrkaye97: merge issue? I think we just added this

cmd/hatchet-admin/cli/seed.go
```diff
@@ -41,5 +46,146 @@ func runSeed(cf *loader.ConfigLoader) error {
...
+		// seed an example user
+		hashedPw, err := repository.HashPassword(dc.Seed.AdminPassword)
+
+		if err != nil {
+			return err
+		}
+
+		user, err := dc.APIRepository.User().GetUserByEmail(context.Background(), dc.Seed.AdminEmail)
+
+		if err != nil {
```
mrkaye97: this looks like a merge issue from Alexander's testcontainers change

cmd/hatchet-migrate/migrate/run.go
```diff
@@ -234,32 +215,9 @@ func RunMigrations(ctx context.Context, opts ...RunMigrationsOpt) {
 	}
 	goose.SetBaseFS(fsys)
 
-	switch {
-	case options.upToPenultimate:
```
mrkaye97: same here, think this needs a merge

mrkaye97 (changes requested):
Think we need a merge - looks like we're removing / re-adding some things we've changed recently (testcontainers / seeding stuff, error handling for missing crons, etc)

api/v1/server/run/run.go
grutt: ooph looks like i backmerged just at the wrong time

pkg/repository/postgres/mq.go
```diff
@@ -55,17 +55,7 @@ func (m *messageQueueRepository) Listen(ctx context.Context, name string, f func
...
-	if err != nil {
-		if errors.Is(err, context.Canceled) {
-			return nil
-		}
-
-		return err
-	}
-
-	return nil
+	return l.Listen(ctx)
```
mrkaye97: this was just changed too I think: https://github.com/hatchet-dev/hatchet/pull/1631/files#diff-7370cadfe1fad4544664f60ea1b590865daa2c10b6225467b3af38bd5b6717dcR58

pkg/v1/task/task.go
```diff
@@ -280,7 +279,3 @@ func (t *DurableTaskDeclaration[I]) GetName() string {
 func (t *NamedTaskImpl) GetName() string {
 	return t.Name
 }
-
-func getActionID(workflowName, taskName string) string {
-	return strings.ToLower(fmt.Sprintf("%s:%s", workflowName, taskName))
-}
```
mrkaye97: this too: https://github.com/hatchet-dev/hatchet/pull/1637/files#diff-adbea65fca87f25c1415af8d20c07d3891d0f7ecf39887b0ff07e56b1dc3b5f7R284-R286

mrkaye97:
not sure if you meant to re-request or not - flagged a couple of other things I think are merge issues

grutt:
_i think_ i got them all now, @mrkaye97 lmk if i missed anything

pkg/repository/v1/sqlcv1/olap.sql
```diff
@@ -1170,3 +1194,66 @@ WHERE
...
+    SELECT
+        t.id,
+        t.inserted_at,
+        t.tenant_id,
+        t.external_id,
+        t.parent_task_external_id,
+        ar.depth + 1 AS depth
+    FROM
+        v1_runs_olap r
+    JOIN all_runs ar ON ar.external_id = r.parent_task_external_id
```
mrkaye97, line 1228: I think this join might end up being really slow because the PK for `v1_runs_olap` is `(inserted_at, id, readable_status, kind)` and there's no index on the external id

I'm not sure what the best strategy is here though, I _think_ you can join on `(r.id, r.inserted_at) = (ar.id, ar.inserted_at)` but I'm not 100% sure (cc @abelanger5 to check on this)

pkg/worker/worker.go
```diff
@@ -354,12 +354,13 @@ func (w *Worker) RegisterAction(actionId string, method any) error {
 }
 
 func (w *Worker) registerAction(service, verb string, method any, compute *compute.Compute) error {
-	actionID := strings.ToLower(fmt.Sprintf("%s:%s", service, verb))
+
+	actionId := fmt.Sprintf("%s:%s", service, verb)
```
mrkaye97: I think we still want to use that helper function here

mrkaye97:
> _i think_ i got them all now, @mrkaye97 lmk if i missed anything

left a couple more comments 😅 - I think that query is problematic and I think we accidentally reverted one or two more of the `actionId` spots again

pkg/repository/v1/sqlcv1/olap.sql
abelanger5: So the thinking here is that we have an index on `v1_runs_olap.parent_task_external_id` and `all_runs` is a very small table. Granted it's not great to scan every partition but the other join condition won't work. To fix this we'll need to propagate the inserted_at/id of the parent to the v1_runs_olap table which is doable but can probably wait.

mrkaye97: sounds good to me!

grutt:
> > _i think_ i got them all now, @mrkaye97 lmk if i missed anything
> 
> left a couple more comments 😅 - I think that query is problematic and I think we accidentally reverted one or two more of the `actionId` spots again

at least you know the correct state 🫠

mrkaye97 (approved):

PR #1641, opened 2025-04-29
Feat: Add cron filters to the API

mrkaye97:
# Description

Adding two more filters to the crons API, `workflowName` and `cronName`, so you can optionally filter by those over the API + in the SDKs

Also bundled in a handful of misc Python changes / requests from #1636 

## Type of change

### DIFF
--- a/frontend/app/src/lib/api/generated/Api.ts
+++ b/frontend/app/src/lib/api/generated/Api.ts
@@ -1732,6 +1732,10 @@ export class Api<SecurityDataType = unknown> extends HttpClient<SecurityDataType
        * @maxLength 36
        */
       workflowId?: string;
+      /** The workflow name to get runs for. */
+      workflowName?: string;
+      /** The cron name to get runs for. */
+      cronName?: string;
       /**
        * A list of metadata key value pairs to filter by
        * @example ["key1:value1","key2:value2"]
--- a/sdks/typescript/src/clients/rest/generated/Api.ts
+++ b/sdks/typescript/src/clients/rest/generated/Api.ts
@@ -120,12 +120,10 @@ import {
   WorkflowUpdateRequest,
   WorkflowVersion,
   WorkflowWorkersCount,
-} from "./data-contracts";
-import { ContentType, HttpClient, RequestParams } from "./http-client";
+} from './data-contracts';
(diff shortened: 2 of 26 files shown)


frontend/app/src/lib/api/generated/Api.ts
```diff
@@ -1732,6 +1732,10 @@ export class Api<SecurityDataType = unknown> extends HttpClient<SecurityDataType
        * @maxLength 36
        */
       workflowId?: string;
+      /** The workflow name to get runs for. */
+      workflowName?: string;
```
grutt, line 1736: what happens if both name and id are set?

sdks/typescript/src/clients/rest/generated/Api.ts
```diff
@@ -1718,6 +1718,10 @@ export class Api<SecurityDataType = unknown> extends HttpClient<SecurityDataType
        * @maxLength 36
        */
       workflowId?: string;
+      /** The workflow name to get runs for. */
+      workflowName?: string;
+      /** The cron name to get runs for. */
```
grutt, line 1723: have you tested the sdk?

mrkaye97: nice catch, I need to wire this up evidently - I tested Python but will test here too to make sure 👍

frontend/app/src/lib/api/generated/Api.ts
mrkaye97: I was thinking about that. it's harmless I think, but there's an argument to be made that maybe only one of the two should be set? idk, it seems fine to me since they're both optional and all of these params are `AND` clauses so I think it should be pretty clear. WDYT?

grutt: agreed, its fine

sdks/typescript/src/clients/rest/generated/Api.ts
mrkaye97: ah turns out I forgot I did this - I forgot that the TS API wrapper in the crons client just passes the params along properly so no changes needed there. tested it and it works fine

mrkaye97:
@grutt re-requested you, this should be good I think 👍

grutt (approved):

PR #1649, opened 2025-04-30
config: allow buffer settings to be configurable

abelanger5:
# Description

Adds three environment variables for setting default buffer configuration. For example:

```
SERVER_DEFAULT_BUFFER_FLUSH_INTERVAL=20ms
SERVER_DEFAULT_BUFFER_SIZE=500
SERVER_DEFAULT_BUFFER_CONCURRENCY=5
```

## Type of change

### DIFF
--- a/internal/msgqueue/v1/mq_pub_buffer.go
+++ b/internal/msgqueue/v1/mq_pub_buffer.go
@@ -3,14 +3,47 @@ package v1
 import (
 	"context"
 	"encoding/json"
+	"os"
+	"strconv"
 	"sync"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


mrkaye97 (approved):
Is there any value in separate settings for pub and sub so they're not forced to be the same?

abelanger5:
> Is there any value in separate settings for pub and sub so they're not forced to be the same?

I'm not sure -- we can always add support for this, but I'm not sure we'd like to expose that level of granularity (I don't think we have a good understanding of how setting the pub/sub flush intervals to be different would impact the system)

PR #1661, opened 2025-05-01
Fix: Action key handling for Go SDK

mrkaye97:
# Description

Mirroring the impls in Python and TS - created a `getActionKey` helper + a corresponding type that returns a string with both the step run id _and_ the retry count of the step to prevent cancellations race conditions

## Type of change


mrkaye97:
@abelanger5 I'm not 100% sure this is right but I grepped for all the `sync.Map`s I could find and this is what turned up. lmk if you think this needs to be fixed other places too

mrkaye97:
closing in favor of V1 Go SDK

PR #1667, opened 2025-05-02
[Python] Fix: Task defaults, Patching fixes, Datetime conversions

mrkaye97:
# Description

Couple things here:
* Fixing task default propagation + expanding it to include a few more settings (retries, backoff) and wrote some tests to make sure that the defaults are being set properly on the task, regardless of how it's created, and that they don't change without breaking tests.
* Fixed how our patching works to use a python script to atomically apply patches
* Via ^, fixed a bug in POST body datetimes not being handled correctly when they're tz-naive - just patched the openapi generated stuff to print a warning and then replace a null `tzinfo` with `utc`
* Attempting to fix The Hanging Test™️  by running it on its own worker with one slot

## Type of change

- [x] New feature


sdks/python/hatchet_sdk/hatchet.py
```diff
@@ -294,8 +293,8 @@ def task(
         sticky: StickyStrategy | None = None,
         default_priority: int = 1,
         concurrency: ConcurrencyExpression | list[ConcurrencyExpression] | None = None,
-        schedule_timeout: Duration = DEFAULT_SCHEDULE_TIMEOUT,
-        execution_timeout: Duration = DEFAULT_EXECUTION_TIMEOUT,
+        schedule_timeout: Duration = timedelta(minutes=5),
```
grutt, line 296: are we sure we dont want a constant here?

mrkaye97: we do :( but it messes up the docs - they end up showing `DEFAULT_XYZ_TIMEOUT` instead of the actual value. I wrote the tests for this to try to enforce what the constant would give us 🤷

mrkaye97: Screenshot 2025-05-06 at 5 20 06 PM

grutt: gotcha, ok good enough

sdks/python/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "hatchet-sdk"
-version = "1.7.0"
+version = "1.8.0"
```
grutt, line 3: minor or patch?

grutt (approved):

sdks/python/pyproject.toml
mrkaye97: 🤷 I was thinking minor b/c it adds a couple new features (expanded task defaults) but either way is fine by me

grutt: your call

PR #1669, opened 2025-05-03
fix: validate cron expression without regex

Flo4604:
# Description

Fixes 

```
could not create workflow cron-workflow: rpc error: code = InvalidArgument desc = Validation failed with the following errors:
0: validation failed on field 'CreateWorkflowVersionOpts.CronTriggers[0]': Invalid cron expression
```

when the cron expression looks like `15 */6 * * *` 

## Type of change

## What's Changed

- [] Cron is no longer validated by regex but using the `github.com/robfig/cron/v3` package


abelanger5 (approved):
Thanks @Flo4604!

PR #1675, opened 2025-05-05
fix: improve tasks olap query

abelanger5:
# Description

Improves OLAP query for getting a single task run to ensure that the query always returns 1 row. 

## Type of change


abelanger5:
not sure this is still an issue since we made some changes to this query since, can revisit if we get another report

PR #1676, opened 2025-05-05
Feat: CEL-based filtering of events

mrkaye97:
# Description

Adding CEL-based filtering of events on trigger. Ended up being pretty simple with a few changes:

1. Added an `expression` to the `WorkflowTriggerEventRef`
2. Right before we go to trigger a workflow from an event, we evaluate the workflow's CEL expression for event filtering (if it exists) and if it returns `true`, we skip triggering the workflow.

Tested it with Python and it worked great! I see the runs being skipped correctly + not showing up in the OLAP table (also correctly)
 
## Type of change


internal/cel/cel.go
```diff
@@ -294,3 +294,31 @@ func (p *CELParser) CheckStepRunOutAgainstKnownV1(out *StepRunOut, knownType sql
 
 	return nil
 }
+
+func (p *CELParser) EvaluateBooleanExpression(expr string, inputData map[string]interface{}) (bool, error) {
```
mrkaye97: Wasn't sure if this was the right kind of thing to do or not, would love some ideas here!

pkg/repository/v1/workflow.go
```diff
@@ -485,12 +488,22 @@ func (r *workflowRepository) createWorkflowVersionTxs(ctx context.Context, tx sq
 	}
 
 	for _, eventTrigger := range opts.EventTriggers {
+		expression := pgtype.Text{
+			String: "",
+			Valid:  false,
+		}
```
mrkaye97: is this the right way to do this?

pkg/repository/v1/trigger.go
```diff
@@ -154,6 +187,22 @@ func (r *TriggerRepositoryImpl) TriggerFromEvents(ctx context.Context, tenantId
...
+			if workflow.EventExpression.Valid && workflow.EventExpression.String != "" {
+				shouldFilter, err := r.processWorkflowExpression(ctx, workflow, opt)
+
+				if err != nil {
+					r.l.Error().
+						Err(err).
+						Str("workflow_id", workflow.WorkflowId.String()).
+						Str("expression", workflow.EventExpression.String).
+						Msg("Failed to evaluate workflow expression")
+				}
```
mrkaye97: curious for thoughts on this - my thinking was that if the CEL errors, we should _not_ skip as the default behavior because running the workflow might be lower risk than _not_ running it, but could be convinced otherwise here

pkg/repository/v1/workflow.go
abelanger5: What you've initialized is the zero value of `pgtype.Text` (importantly because this _isn't_ a pointer, and the zero value of bools is `false` and strings is the empty string), so it's not necessary to even instantiate anything here since the zero value is the default value

pkg/repository/v1/trigger.go
abelanger5: Since we don't have any way to communicate to the user that the CEL expression errored, I suppose to reason this out: 
- If it does run, it most likely looks like a Hatchet bug
- If it doesn't run, it'll be shown in the events view as not having triggered a workflow (assuming they can track it down) which will give the user something to repro

I also think there are cases where running a workflow which fails a CEL is fairly high risk since  CELs can check for existence of certain input fields and we don't want to blow up a worker somehow.

On average it's probably lower risk to run these workflows but in the worst case it might actually be worse. 

I'd lean towards not running it and figuring out a mechanism where we can log failed CEL expressions in the UI somewhere (this is also more consistent with everywhere else that we're parsing CELs)

pkg/repository/v1/trigger.go
```diff
@@ -117,6 +117,39 @@ type TriggerFromEventsResult struct {
...
+		if err != nil {
+			return false, fmt.Errorf("failed to unmarshal input data: %w", err)
+		}
+	} else {
+		inputData = make(map[string]interface{})
+	}
+
+	match, err := r.celParser.EvaluateBooleanExpression(
+		workflow.EventExpression.String,
+		inputData,
```
abelanger5: In other places we structure our input data to have an `input` and `metadata` field (or something like that), did we double check that this is consistent with that?

pkg/repository/v1/trigger.go
mrkaye97: sounds good!

pkg/repository/v1/trigger.go
mrkaye97: do you mean in the cel parser or elsewhere in the code? I don't think I did this though, but just checking to make sure

abelanger5: So `EvaluateBooleanExpression` uses the `stepRunEnv` in the CEL environment has the following declarations:

```go
cel.Declarations(
			decls.NewVar("input", decls.NewMapType(decls.String, decls.Dyn)),
			decls.NewVar("additional_metadata", decls.NewMapType(decls.String, decls.Dyn)),
			decls.NewVar("parents", decls.NewMapType(decls.String, decls.NewMapType(decls.String, decls.Dyn))),
			decls.NewVar("workflow_run_id", decls.String),
			checksumDecl,
		)
```

Which is perhaps an indication that we _shouldn't_ be using the `stepRunEnv here and we should instead create a new `eventEnv` with different bindings (probably `input`, `additional_metadata`, and `event_id`)

(to clarify, each CEL environment needs to have explicit declarations for its inputs so that expressions can compile)

Does that make sense?

abelanger5 (changes requested):

pkg/repository/v1/trigger.go
mrkaye97: Got it! I think I fixed that [here](https://github.com/hatchet-dev/hatchet/pull/1676/commits/3bf14f81c827db46f478c13734105c658d26183e) - lmk if that's closer to what you're thinking

internal/cel/cel.go
```diff
@@ -67,9 +68,19 @@ func NewCELParser() *CELParser {
 		checksum,
 	)
 
+	eventEnv, _ := cel.NewEnv(
+		cel.Declarations(
+			decls.NewVar("input", decls.NewMapType(decls.String, decls.Dyn)),
+			decls.NewVar("additional_metadata", decls.NewMapType(decls.String, decls.Dyn)),
+			decls.NewVar("event_id", decls.NewMapType(decls.String, decls.Dyn)),
```
abelanger5: Shouldn't this be a string not a map? (might have to fix the `workflow_run_id` as well if this was copied from that)

internal/cel/cel.go
```diff
@@ -295,20 +306,22 @@ func (p *CELParser) CheckStepRunOutAgainstKnownV1(out *StepRunOut, knownType sql
 	return nil
 }
 
-func (p *CELParser) EvaluateBooleanExpression(expr string, inputData map[string]interface{}) (bool, error) {
-	ast, issues := p.stepRunEnv.Compile(expr)
+func (p *CELParser) EvaluateBooleanExpression(expr, eventId string, input, additionalMetadata map[string]interface{}) (bool, error) {
```
abelanger5: nit: naming -- should probably be `EvaluateEventExpression` since this can't evaluate any boolean expr

internal/cel/cel.go
mrkaye97: whoops yeah, thanks

api-contracts/workflows/workflows.proto
```diff
@@ -44,6 +44,7 @@ message CreateWorkflowVersionOpts {
     optional StickyStrategy sticky = 12; // (optional) the sticky strategy for assigning steps to workers
     optional WorkflowKind kind = 13; // (optional) the kind of workflow
     optional int32 default_priority = 14; // (optional) the priority of the workflow
+    optional string event_filter_expression = 15; // (optional) the event expression for the workflow
```
abelanger5: This should be a v1 workflow-only feature right?

internal/cel/cel.go
```diff
@@ -294,3 +305,33 @@ func (p *CELParser) CheckStepRunOutAgainstKnownV1(out *StepRunOut, knownType sql
 
 	return nil
 }
+
+func (p *CELParser) EvaluateEventExpression(expr, eventId string, input, additionalMetadata map[string]interface{}) (bool, error) {
```
abelanger5: Should be a `cel_test` file, we should add a unit test here

pkg/repository/v1/trigger.go
abelanger5: yep that looks right!

api-contracts/workflows/workflows.proto
abelanger5: curious what the thinking is for adding to the v0 workflow contracts

mrkaye97: ah whoops I meant to delete this, put it in the wrong place originally and then moved it after - thanks!

internal/cel/cel.go
mrkaye97: sounds good! will give it a shot, haven't written any Go tests yet 🥳

PR #1677, opened 2025-05-05
Feat: SDK changes for event filtering

mrkaye97:
# Description

SDK changes for event filtering + some other misc things in Python. I added a Python changelog since other people have been asking for it, so just going to paste the contents of this entry at the bottom here

## Type of change

### Added

- The main `Hatchet` client now has a `filters` attribute (a `Filters` client) which wraps basic CRUD operations for managing filters.
- Events can now be pushed with a `priority` attribute, which sets the priority of the runs triggered by the event.
- There are new `list` and `aio_list` methods for the `Events` client, which allow listing events.
- Workflow runs can now be filtered by `triggering_event_external_id`, to allow for seeing runs triggered by a specific event.
- There is now an `id` property on all `Workflow` objects (`Workflow` created by `hatchet.workflow` and `Standalone` created by `hatchet.task`) that returns the ID (UUID) of the workflow.
- Events can now be pushed with a `scope` parameter, which is required for using filters to narrow down the filters to consider applying when triggering workflows from the event.

### Changed

- The `name` parameter to `hatchet.task` and `hatchet.durable_task` is now optional. If not provided, the task name will be the same as the function name.


api/v1/server/handlers/v1/filters/list.go
```diff
@@ -32,16 +32,16 @@ func (t *V1FiltersService) V1FilterList(ctx echo.Context, request gen.V1FilterLi
 	workflowIdParams := make([]pgtype.UUID, numScopesOrIds)
 
 	if workflowIds != nil {
-		for _, id := range *workflowIds {
-			workflowIdParams = append(workflowIdParams, sqlchelpers.UUIDFromStr(id.String()))
+		for ix, id := range *workflowIds {
+			workflowIdParams[ix] = sqlchelpers.UUIDFromStr(id.String())
```
mrkaye97, line 36: this needs to be in the other PR, but I'll merge them together before merging so it should be fine

pkg/client/event.go
```diff
@@ -15,14 +15,16 @@ import (
...
+	scope              *string
 }
 
 type PushOpFunc func(*pushOpt) error
 
 type BulkPushOpFunc func(*eventcontracts.BulkPushEventRequest) error
 
 type EventClient interface {
-	Push(ctx context.Context, eventKey string, payload interface{}, options ...PushOpFunc) error
+	Push(ctx context.Context, eventKey string, payload interface{}, priority *int32, scope *string, options ...PushOpFunc) error
```
mrkaye97: this is a breaking change, but I don't know what the convention here is - can these go after `options`?

sdks/python/hatchet_sdk/features/filters.py
```diff
@@ -0,0 +1,172 @@
...
+
+        :return: A list of filters matching the specified criteria.
+        """
+        return await asyncio.to_thread(self.list, limit, offset, workflow_ids, scopes)
+
+    def list(
+        self,
+        limit: int | None = None,
+        offset: int | None = None,
+        workflow_ids: list[str] | None = None,
```
grutt: this is something we should really standardize... these are ids or names? what are we using elsewhere?

sdks/typescript/src/v1/client/features/filters.ts
```diff
@@ -0,0 +1,33 @@
...
+export class FiltersClient {
+  tenantId: string;
+  api: HatchetClient['api'];
+
+  constructor(client: HatchetClient) {
+    this.tenantId = client.tenantId;
+    this.api = client.api;
+  }
+
+  async list(opts?: Parameters<typeof this.api.v1FilterList>[1]) {
```
grutt: i think i do this in a number of places, but this is bad practice as most ides wont render the props

sdks/python/hatchet_sdk/features/filters.py
mrkaye97: these are IDs, which I think is the correct thing to do. IMO, we should be using IDs everywhere that we can, and especially here since this is just an API wrapper (otherwise we'll need to list all workflows matching some set of names, loop through the results to get the IDs, and then pass the IDs back into here, or change the API to accept names instead of IDs and then do the lookup in Go). I think those are both undesirable - WDYT?

sdks/typescript/src/v1/examples/on_event/event.e2e.ts
```diff
@@ -0,0 +1,374 @@
+import sleep from '@hatchet-dev/typescript-sdk/util/sleep';
+import { randomUUID } from 'crypto';
+import { Event } from '@hatchet-dev/typescript-sdk/protoc/events';
+import { SIMPLE_EVENT, lower, Input } from './workflow';
+import { hatchet } from '../hatchet-client';
+import { Worker } from '../../client/worker/worker';
+
+xdescribe('events-e2e', () => {
```
grutt, line 8: intentional skip?

mrkaye97: yeah I have no idea why this isn't passing in CI and don't have time to debug it if we want to get this out today 😬

sdks/typescript/src/v1/client/features/filters.ts
mrkaye97: yeahh just so easy to do this 😅 but I agree, I think we should file this away for later as something to improve

pkg/client/event.go
abelanger5: We can add these as `PushOpFunc` options (`WithPriority`, `WithScope`) to prevent the breaking change

mrkaye97: lovely, I'll do that - one sec!

PR #1690, opened 2025-05-07
fix: UI bugs

grutt:
# Description

bashes a number of layout and state bugs on the overhaul ui


api-contracts/openapi/components/schemas/v1/task.yaml
```diff
@@ -393,7 +397,7 @@ V1TaskTiming:
     - taskId
     - taskInsertedAt
     - tenantId
-
+    - workflowRunId
```
mrkaye97: I think making this required will be breaking, since if we generate the Python SDK pydantic will treat this as `str` and not `str | None`, and then will fail on older APIs (I've made this mistake at least once 😭 )

frontend/app/src/next/components/runs/run-id.tsx
```diff
@@ -46,7 +54,7 @@ export function RunId({ wfRun, taskRun, onClick }: RunIdProps) {
       <Tooltip>
         <TooltipTrigger asChild>
           <span>
-            {url && !onClick ? (
+            {url && !onClick && !noLink ? (
```
mrkaye97: nit: I usually prefer "positives" as opposed to double-negatives if possible (not a big deal)

api-contracts/openapi/components/schemas/v1/task.yaml
grutt: good catch -- i didnt realize we're validating api RESULTS... can we not?

mrkaye97: I don't think so since the openapi generator generates Pydantic, which I think is actually a good thing at the end of the day but is annoying for things like this

frontend/app/src/next/components/runs/run-id.tsx
grutt: good catch -- meant to remove this anyway

mrkaye97 (approved):

PR #1693, opened 2025-05-07
fix: reset scheduling timeouts on retries

abelanger5:
# Description

In the case that a task was retried and has a concurrency key (but isn't a retry w/ backoff), we weren't properly resetting the schedule timeout value, which would lead to tasks hitting their scheduling timeout too early.

## Type of change

### DIFF
--- a/cmd/hatchet-migrate/migrate/migrations/20250507193114_v1_0_17.sql
+++ b/cmd/hatchet-migrate/migrate/migrations/20250507193114_v1_0_17.sql
@@ -1,0 +1,360 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION v1_task_update_function()
+RETURNS TRIGGER AS
+$$
+BEGIN
+    WITH new_retry_rows AS (
+        SELECT
+            nt.id,
+            nt.inserted_at,
+            nt.retry_count,
+            nt.tenant_id,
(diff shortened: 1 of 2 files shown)


cmd/hatchet-migrate/migrate/migrations/20250507193114_v1_0_17.sql
```diff
@@ -0,0 +1,360 @@
...
+            -- Concurrency strategy id should never be null
+            AND nt.concurrency_strategy_ids[1] IS NOT NULL
+            AND (nt.retry_backoff_factor IS NULL OR ot.app_retry_count IS NOT DISTINCT FROM nt.app_retry_count OR nt.app_retry_count = 0)
+            AND ot.retry_count IS DISTINCT FROM nt.retry_count
+    ), updated_slot AS (
+        UPDATE
+            v1_concurrency_slot cs
+        SET
+            task_retry_count = nt.retry_count,
+            schedule_timeout_at = nt.schedule_timeout_at,
```
mrkaye97, line 76: this is the change, right?

abelanger5: yep

mrkaye97 (approved):
re-creating these triggers every time we change one line is brutal 😬

PR #1698, opened 2025-05-08
fix: one more issue in dag populator

abelanger5:
# Description

Fixes one more issue in the DAG populator where the max retry count isn't selected (related: #1697)

## Type of change


mrkaye97 (approved):
copilot comments look potentially useful this time?

PR #1701, opened 2025-05-09
[Python]: Fix: Adding Observability + Failure Handling

mrkaye97:
# Description

Working on adding some improved observability into the Python SDK:

* Logging about the status of the thread pool
* Setting the `stop_signal` to stop accepting new work on shutdown more forcefully
* ~[Proposal]: Run a health check no-op workflows on the healthcheck endpoint~ Need to think more about how to implement this (basically need to be able to target the worker)
* Removing some invalid concurrency strategies
* Fixing flaky cron test
* Adding error + hint on obviously invalid token
* Removes some buggy concurrency validation

## Type of change

### DIFF
--- a/sdks/python/hatchet_sdk/config.py
+++ b/sdks/python/hatchet_sdk/config.py
@@ -64,8 +64,10 @@ class ClientConfig(BaseSettings):
     )
 
     worker_preset_labels: dict[str, str] = Field(default_factory=dict)
+
     enable_force_kill_sync_threads: bool = False
     enable_thread_pool_monitoring: bool = False
+    enable_no_op_heartbeat_workflow: bool = False
 
     @model_validator(mode="after")
     def validate_token_and_tenant(self) -> "ClientConfig":
--- a/sdks/python/hatchet_sdk/config.py
+++ b/sdks/python/hatchet_sdk/config.py
@@ -73,6 +73,11 @@ class ClientConfig(BaseSettings):
         if not self.token:
             raise ValueError("Token must be set")
 
+        if not self.token.startswith("ey"):
+            raise ValueError(
+                f"Token must be a valid JWT. Hint: These are the first few characters of the token provided: {self.token[:5]}"
+            )
+
         if not self.tenant_id:
             self.tenant_id = get_tenant_id_from_jwt(self.token)
 
--- a/sdks/python/hatchet_sdk/config.py
+++ b/sdks/python/hatchet_sdk/config.py
@@ -65,6 +65,7 @@ class ClientConfig(BaseSettings):
 
     worker_preset_labels: dict[str, str] = Field(default_factory=dict)
     enable_force_kill_sync_threads: bool = False
+    enable_thread_pool_monitoring: bool = False
 
     @model_validator(mode="after")
     def validate_token_and_tenant(self) -> "ClientConfig":
--- a/sdks/python/hatchet_sdk/config.py
+++ b/sdks/python/hatchet_sdk/config.py
@@ -67,7 +67,6 @@ class ClientConfig(BaseSettings):
 
     enable_force_kill_sync_threads: bool = False
     enable_thread_pool_monitoring: bool = False
-    enable_no_op_heartbeat_workflow: bool = False
 
     @model_validator(mode="after")
     def validate_token_and_tenant(self) -> "ClientConfig":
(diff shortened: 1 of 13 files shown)


sdks/python/hatchet_sdk/worker/worker.py
```diff
@@ -199,6 +233,18 @@ def _setup_loop(self) -> None:
         asyncio.set_event_loop(self.loop)
 
     async def _health_check_handler(self, request: Request) -> Response:
+        if self.config.enable_no_op_heartbeat_workflow:
+            ref = await self.admin_client.aio_run_workflow(
+                workflow_name="__internal__no_op_heartbeat_workflow",
+                input={},
+            )
```
mrkaye97: We really need to set the desired worker id here, but I'm not sure what a reasonable way to do that is since the worker doesn't know its own id?

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -419,23 +465,18 @@ async def handle_cancel_action(self, action: Action) -> None:
         try:
             # call cancel to signal the context to stop
             if key in self.contexts:
-                context = self.contexts.get(key)
-
-                if context:
-                    context._set_cancellation_flag()
+                self.contexts[key]._set_cancellation_flag()
```
mrkaye97, line 466: drive by cleanup - the `.get` and then checking for truthiness in these is redundant b/c if the key is in the set of keys, then we know the value is truthy already

sdks/python/hatchet_sdk/worker/action_listener_process.py
```diff
@@ -291,6 +291,9 @@ async def cleanup(self) -> None:
         self.event_queue.put(STOP_LOOP)
 
     async def exit_gracefully(self) -> None:
+        if self.listener:
+            self.listener.stop_signal = True
```
mrkaye97, line 295: @grutt would like a sanity check here - I was thinking this should forcibly stop assignment to the worker (even before the pause processes). lmk if this is wrong

sdks/python/hatchet_sdk/runnables/types.py
```diff
@@ -26,8 +26,6 @@ class StickyStrategy(str, Enum):
 
 class ConcurrencyLimitStrategy(str, Enum):
     CANCEL_IN_PROGRESS = "CANCEL_IN_PROGRESS"
-    DROP_NEWEST = "DROP_NEWEST"
-    QUEUE_NEWEST = "QUEUE_NEWEST"
```
mrkaye97, line 30: drive by - these shouldn't be here

sdks/python/hatchet_sdk/runnables/types.py
```diff
@@ -82,31 +80,6 @@ class WorkflowConfig(BaseModel):
...
 
-    def _raise_for_invalid_expression(self, expr: str) -> None:
-        if not expr.startswith("input."):
-            return None
-
-        _, field = expr.split(".", maxsplit=2)
-
-        if field not in self.input_validator.model_fields.keys():
-            raise ValueError(
-                f"The concurrency expression provided relies on the `{field}` field, which was not present in `{self.input_validator.__name__}`."
```
mrkaye97, line 93: this is broken, since you could have a key like `foo.bar + foo.baz` or `foo.bar.baz`

sdks/python/hatchet_sdk/config.py
```diff
@@ -64,13 +64,20 @@ class ClientConfig(BaseSettings):
     )
 
     worker_preset_labels: dict[str, str] = Field(default_factory=dict)
+
     enable_force_kill_sync_threads: bool = False
+    enable_thread_pool_monitoring: bool = False
```
abelanger5, line 69: just to confirm, this automatically gets parsed as an environment variable?

abelanger5 (approved):

sdks/python/hatchet_sdk/config.py
mrkaye97: yeah, pydantic settings does that for us 🥳

mrkaye97: so it'll be `HATCHET_CLIENT_ENABLE_THREAD_POOL_MONITORING`

PR #1652, opened 2025-04-30
feat(ts): improved custom logger

grutt:
# Description

Improves ctx.logger and exposes additional logging capabilities

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## What's Changed

- [ ] adds retry count
- [] fixes metadata
- [] exposes additional custom logger properties

PR #1703, opened 2025-05-12
Fix: more UI bugs -- focus on run detail view

grutt:
# Description

Fixes and improvements for the run detail view.

Depends on #1652


frontend/app/src/next/components/runs/run-dag/dag-run-visualizer.tsx
```diff
@@ -190,25 +212,79 @@ function WorkflowRunVisualizerContent({
...
-    layoutedNodes.length === 0 ||
-    layoutedEdges.length === 0
-  ) {
-    return null;
-  }
+  // Track the last centered task ID to avoid unnecessary recentering
+  const lastCenteredTaskId = useRef<string | undefined>(undefined);
+
+  const recenter = useCallback(() => {
+    setTimeout(() => {
```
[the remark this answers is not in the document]
grutt: i personally dont understand this enough. i know the setTimeout is a hack that probably bandaids a deeper problem but reading the docs on requestAnimationFrame, i do not think this is correct

api/v1/server/oas/transformers/v1/log.go
```diff
@@ -8,9 +8,27 @@ import (
 )
 
 func ToV1LogLine(log *sqlcv1.V1LogLine) *gen.V1LogLine {
+
+	retryCount := int(log.RetryCount)
+	attempt := retryCount + 1
```
mrkaye97: nit: might be good to factor this into a helper? I don't super duper care since it's so self-explanatory though 🤷

frontend/app/src/next/components/runs/run-dag/dag-run-visualizer.tsx
```diff
@@ -27,36 +29,40 @@ const edgeTypes = {
 
 interface WorkflowRunVisualizerProps {
   workflowRunId: string;
-  onTaskSelect?: (taskId: string) => void;
+  patchTask?: V1TaskSummary | null;
+  onTaskSelect?: (taskId: string, childWfrId?: string) => void;
```
mrkaye97: nit but also fine for now: in general I prefer to spell things out instead of using abbreviations if we can, for clarity

frontend/app/src/next/components/runs/run-dag/dag-run-visualizer.tsx
```diff
@@ -232,6 +311,26 @@ function WorkflowRunVisualizerContent({
         }
         snapToGrid={true}
       />
+      <button
```
mrkaye97: shad button here?

frontend/app/src/next/components/runs/run-dag/dag-run-visualizer.tsx
```diff
@@ -232,6 +311,26 @@ function WorkflowRunVisualizerContent({
...
+        onClick={toggleExpand}
+        className={cn(
+          'absolute inset-x-0 -bottom-2 z-20 h-4 transition-all ease-linear after:absolute after:inset-x-0 after:top-1/2 after:h-[2px] hover:after:bg-border',
+          isExpanded ? 'cursor-n-resize' : 'cursor-s-resize',
+        )}
+        title="Toggle height"
+      />
+      <button
+        onClick={toggleExpand}
```
mrkaye97, line 317: shad button

frontend/app/src/next/hooks/use-run-detail.tsx
```diff
@@ -85,10 +87,16 @@ function RunDetailProviderContent({
   const [lastRefetchTime, setLastRefetchTime] = useState(Date.now());
   const { toast } = useToast();
 
+  // Memoize query keys to prevent unnecessary refetches
+  const runDetailsQueryKey = useMemo(
+    () => ['workflow-run-details:get', runId],
+    [runId],
+  );
```
mrkaye97: this seems like premature optimization IMO - it's a single value?

frontend/app/src/next/hooks/use-run-detail.tsx
```diff
@@ -104,21 +112,26 @@ function RunDetailProviderContent({
...
     },
-    refetchInterval: refetchInterval,
+    refetchInterval,
   });
 
+  // Memoize parent task ID to prevent unnecessary fetches
+  const parentTaskId = useMemo(
+    () => runDetails.data?.run.parentTaskExternalId,
+    [runDetails.data?.run.parentTaskExternalId],
+  );
```
mrkaye97: this too - I can't imagine this really helps but could be proven wrong here 😅

frontend/app/src/next/hooks/use-run-detail.tsx
```diff
@@ -129,47 +142,54 @@ function RunDetailProviderContent({
...
-              })
-            ).data?.rows || [],
-        );
-
+        // Batch API calls for better performance
         const [logs, events] = await Promise.all([
-          Promise.all(logPromises),
-          Promise.all(eventPromises),
+          Promise.all(
+            tasks.map((task) =>
```
mrkaye97, line 161: any risk here? how many API calls could this make concurrently?

frontend/app/src/next/pages/authenticated/dashboard/runs/run-detail.page.tsx
```diff
@@ -455,16 +411,20 @@ function RunDetailPageContent({ workflowRunId, taskId }: RunDetailPageProps) {
...
-                  navigate(
-                    ROUTES.runs.taskDetail(workflowRunId!, taskId, options),
-                  );
-                }}
-              />
-            </FilterProvider>
+            <RunEventLog
+              workflow={workflow}
+              onTaskSelect={(event) => {
+                console.log(event);
```
mrkaye97: nit: log cruft?

pkg/repository/v1/sqlcv1/olap.sql
```diff
@@ -325,9 +325,12 @@ JOIN v1_tasks_olap tsk
 ORDER BY a.time_first_seen DESC, t.event_timestamp DESC;
 
 -- name: PopulateSingleTaskRunData :one
-WITH latest_retry_count AS (
+WITH selected_retry_count AS (
     SELECT
-        MAX(retry_count) AS retry_count
+        CASE
+            WHEN sqlc.narg('retry_count')::int IS NOT NULL THEN sqlc.narg('retry_count')::int
```
mrkaye97, line 331: couldn't this end up returning N rows if there are N that match the `WHERE` clause? I think we need a `LIMIT 1` here probably

api/v1/server/oas/transformers/v1/log.go
grutt: generally would agree, i think this is fine though. bigger concern that i dont have a solve for is that we miss setting this on future transformers

frontend/app/src/next/components/runs/run-dag/dag-run-visualizer.tsx
grutt: 👍 ez n-uff

frontend/app/src/next/components/runs/run-dag/dag-run-visualizer.tsx
grutt: good catch

frontend/app/src/next/hooks/use-run-detail.tsx
grutt: ya good catch, was debugging something else and this snuck in

frontend/app/src/next/hooks/use-run-detail.tsx
grutt: this should really be a single endpoint (ideally using some kind of stream or last event timestamp cursor)
added a fixme for now

pkg/repository/v1/sqlcv1/olap.sql
grutt: yeah good catch, looks like this was potentially an issue in the previous implementation too

mrkaye97 (approved):
🚢

mrkaye97 (approved):

PR #1704, opened 2025-05-12
Feat: Dynamic (Event) Filters

mrkaye97:
# Description

Introduces dynamic event filters, which are persisted into `v1_filter` and then queried when we go to trigger a workflow from an event.

Some decisions made here / open questions:

1. I was thinking it'd make sense to have the `(tenant_id, id)` PK with a unique index on `(tenant_id, workflow_id, resource_hint, expression)` so that: 1) it's fast to look up + delete filters by the id and 2) it's fast to look up which filters match a workflow.
2. I was thinking we want to "FK" these to the `workflow_id` and not the workflow version to give the expected behavior to users (basically they know nothing about workflow versions and would expect that e.g. if they add an event trigger to their workflow, the filters would still be there). If we FK to the workflow version, we'd need to migrate these over each time.
3. Should this API be on the REST API or the gRPC one? Went with gRPC since events are pushed via gRPC + we use it for e.g. putting workflows, but don't feel strongly

TODO:

- [x] Implement `delete` method
- [x] Write tests for various cases (key used in CEL doesn't exist in input, no input provided, no hint provided, triggering events with no filters works as expected, ...)
- [x] I think that if no hint is provided, we need to make sure that the query for filters returns an empty list
- [x] Change this to use the REST API instead of the gRPC one
- [x] Add the payload to the CEL env

## Type of change


api-contracts/openapi/components/schemas/v1/filter.yaml
```diff
@@ -0,0 +1,62 @@
...
+    tenantId:
+      type: string
+      description: The ID of the tenant associated with this filter.
+    workflowId:
+      type: string
+      format: uuid
+      minLength: 36
+      maxLength: 36
+      description: The workflow id associated with this filter.
+    resourceHint:
```
abelanger5: nit on naming: I know that we're trying to keep consistency with our internal naming, but I really think should be `scope` or `partition` (have a slight preference towards the former), because I don't think `resourceHint` is very intuitive

abelanger5 (commented):
Looking pretty good -- one question that's not immediately clear to me, if we trigger a workflow based on a filter, I assume we'd want both the triggering event payload and the filter's payload to be present in the input. Where is that being propagated at the moment?

mrkaye97:
> Looking pretty good -- one question that's not immediately clear to me, if we trigger a workflow based on a filter, I assume we'd want both the triggering event payload and the filter's payload to be present in the input. Where is that being propagated at the moment?

~Ah that's a good thought, let me take a stab at this~

Edit: I think there's not necessarily a guarantee that a single filter will apply to any event / workflow run pair, which is also true in the code right now (we loop over the set of filters that match the workflow id and the resource hint). I'm not sure it makes sense to potentially send _all_ of the matching payloads back.

Also, how valuable is the filter payload if a workflow _was_ triggered (meaning the filter did not apply)?

api-contracts/openapi/components/schemas/v1/filter.yaml
mrkaye97: heh I prefer `partition` but am happy either way, agree that `resourceHint` isn't very helpful - will fix this one

pkg/repository/v1/input.go
```diff
@@ -101,6 +112,8 @@ func (s *sharedRepository) ToV1StepRunData(t *TaskInput) *V1StepRunData {
 		}
 	}
 
+	triggers["filter_payload"] = t.FilterPayload
```
mrkaye97, line 115: @abelanger5 I hacked this a bit, but this seemed to be the simplest / lowest friction route to getting the filter payload to be sent through the API correctly. think this is fine? I can also figure out how to get it to go through the db separately, so whatever works

internal/services/ingestor/server.go
```diff
@@ -276,12 +277,20 @@ func toEvent(e *dbsqlc.Event) (*contracts.Event, error) {
 	tenantId := sqlchelpers.UUIDToStr(e.TenantId)
 	eventId := sqlchelpers.UUIDToStr(e.ID)
 
+	var additionalMeta *string
+
+	if e.AdditionalMetadata != nil {
+		additionalMetaStr := string(e.AdditionalMetadata)
+		additionalMeta = &additionalMetaStr
+	}
```
mrkaye97, line 285: drive by - this wasn't being sent back on bulk push

pkg/repository/v1/sqlcv1/filters-overwrite.sql
```diff
@@ -0,0 +1,15 @@
...
+    SELECT
+        UNNEST(@tenantIds::UUID[]) AS tenant_id,
+        UNNEST(@workflowIds::UUID[]) AS workflow_id,
+        -- NOTE: this is nullable, so sqlc doesn't support casting to a type
+        UNNEST(@scopes::TEXT[]) AS scope
+)
+
+SELECT f.*
+FROM v1_filter f
+JOIN inputs i ON (f.tenant_id, f.workflow_id, f.scope) = (i.tenant_id, i.workflow_id, i.scope)
```
mrkaye97: this is important - if we get a `null` scope when we run this query, that scope will never match a filter. I think this is the desired behavior since we don't want to (e.g.) do the opposite and end up iterating through a zillion possible filters looking for the one that evals to `true`. and this also means that if someone does the "default" thing (pushing an event with no filters and no scopes), they get the behavior they expect (the task runs)

pkg/repository/v1/sqlcv1/filters-overwrite.sql
```diff
@@ -0,0 +1,15 @@
...
+LIMIT COALESCE(sqlc.narg('filterLimit')::BIGINT, 20000)
```
mrkaye97, line 12: added a limit here just in case but not sure what a reasonable thing for this to be set to is

pkg/repository/v1/trigger.go
```diff
@@ -116,6 +119,61 @@ type TriggerFromEventsResult struct {
 	EventExternalIdToRuns map[string][]*Run
 }
 
+type TriggerDecision struct {
+	ShouldTrigger bool
+	FilterPayload []byte
+}
+
+func (r *TriggerRepositoryImpl) makeTriggerDecision(ctx context.Context, filters []*sqlcv1.V1Filter, opt EventTriggerOpts) TriggerDecision {
```
mrkaye97, line 127: this is the meat of the filtering logic

pkg/repository/v1/trigger.go
```diff
@@ -116,6 +119,61 @@ type TriggerFromEventsResult struct {
...
+			FilterPayload: nil,
+		}
+	}
+
+	for _, filterPtr := range filters {
+		if filterPtr == nil {
+			continue
+		}
+
+		filter := *filterPtr
```
mrkaye97, line 141: do I need to do this or can I just refer to the pointer directly?

api-contracts/openapi/paths/v1/filters/filter.yaml
```diff
@@ -0,0 +1,217 @@
+V1FilterGetDelete:
+  get:
+    x-resources: ["tenant"]
```
abelanger5: our convention is to add `v1-filter` the `x-resources` array and use the `filter_id` which is passed in the path as the source of truth for the tenant ID. this is an additional layer of defense for preventing cross-tenant authz issues, because if both the tenant id and filter id are passed in the path, a user could pass in a UUID for a filter which doesn't belong to that tenant (the other safeguard is that the tenant id is used in *all* queries, which is a good mechanism it's easy to introduce a bug where the tenant id isn't used). 

So to remediate:
1. Add `["tenant", "v1-filter"]` as the resource tree, and remove the `tenant` from the path param
2. Add a `populator` (to the `run.go` file) for the filter, which returns the tenant id as the `uniqueParentId` field
3. Grab the filter from the context instead of from the request params, which will panic if not correctly (this is a feature of the middleware system, because it's nearly impossible to ship a panicking handler if this is tested at all)

api-contracts/openapi/paths/v1/filters/filter.yaml
```diff
@@ -0,0 +1,217 @@
...
+        content:
+          application/json:
+            schema:
+              $ref: "../../../components/schemas/_index.yaml#/APIErrors"
+        description: Forbidden
+    summary: Get a filter
+    tags:
+      - Filter
+  delete:
+    x-resources: ["tenant"]
```
abelanger5: Same as above, just flagging

api/v1/server/handlers/v1/filters/create.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (t *V1FiltersService) V1FilterCreate(ctx echo.Context, request gen.V1FilterCreateRequestObject) (gen.V1FilterCreateResponseObject, error) {
+	tenant := ctx.Get("tenant").(*dbsqlc.Tenant)
+
+	var payload []byte
+	if request.Body.Payload != nil {
+		marshalledPayload, err := json.Marshal(request.Body.Payload)
+
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal payload: %w", err)
```
abelanger5: Should probably return a 400-level response here?

api/v1/server/handlers/v1/filters/list.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+func (t *V1FiltersService) V1FilterList(ctx echo.Context, request gen.V1FilterListRequestObject) (gen.V1FilterListResponseObject, error) {
+	tenant := ctx.Get("tenant").(*dbsqlc.Tenant)
+
+	scopes := request.Params.Scopes
+	workflowIds := request.Params.WorkflowIds
+
+	if scopes != nil && workflowIds != nil && len(*scopes) != len(*workflowIds) {
+		return gen.V1FilterList400JSONResponse(apierrors.NewAPIErrors("resource hints and workflow ids must be the same length")), nil
```
abelanger5:
```suggestion
		return gen.V1FilterList400JSONResponse(apierrors.NewAPIErrors("scopes and workflow ids must be the same length")), nil
```

pkg/repository/v1/sqlcv1/filters-overwrite.sql.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"context"
+
+	"github.com/jackc/pgx/v5/pgtype"
+)
+
+const listFilters = `-- name: ListFilters :many
+WITH inputs AS (
+    SELECT
+        UNNEST($1::UUID[]) AS tenant_id,
```
abelanger5: we shouldn't need to unnest tenant_id here, it can be passed as a positional param in the join

pkg/repository/v1/sqlcv1/filters.sql
```diff
@@ -0,0 +1,39 @@
...
+    @workflowId::UUID,
+    @scope::TEXT,
+    @expression::TEXT,
+    @payload::JSONB
+)
+ON CONFLICT (tenant_id, workflow_id, scope, expression) DO UPDATE
+SET
+    payload = EXCLUDED.payload,
+    scope = EXCLUDED.scope,
+    expression = EXCLUDED.expression,
```
abelanger5: Why are we updating `scope/expression` here?

api/v1/server/handlers/v1/filters/create.go
```diff
@@ -0,0 +1,49 @@
...
+		Payload:    payload,
+	}
+
+	filter, err := t.config.V1.Filters().CreateFilter(
+		ctx.Request().Context(),
+		params,
+	)
+
+	if err != nil {
+		return gen.V1FilterCreate400JSONResponse(apierrors.NewAPIErrors("failed to create filter")), nil
```
abelanger5: shouldn't this one be a 500-level response? we have nothing actionable for the user here

abelanger5: (matters bc we log 400s as warnings and 500s as errors)

api/v1/server/handlers/v1/filters/list.go
```diff
@@ -0,0 +1,74 @@
...
+	if scopes != nil {
+		for _, hint := range *scopes {
+			scopeParams = append(scopeParams, &hint)
+		}
+	}
+
+	filters, err := t.config.V1.Filters().ListFilters(
+		ctx.Request().Context(),
+		sqlcv1.ListFiltersParams{
+			Tenantids:    tenantIds,
```
abelanger5: See note in query below, but using unnest for tenant ids makes me nervous (and usually prefer this type of param construction gets handled at the repository layer for future callers)

abelanger5: and I prefer all tenant-scoped repository methods to use `ctx context.Context, tenantId string` as an almost-required signature, it will make it a million times easier to add some tenant-level routing middleware at the database level

pkg/repository/v1/filters.go
```diff
@@ -0,0 +1,40 @@
+package v1
+
+import (
+	"context"
+
+	"github.com/hatchet-dev/hatchet/pkg/repository/v1/sqlcv1"
+)
+
+type FilterRepository interface {
+	CreateFilter(ctx context.Context, params sqlcv1.CreateFilterParams) (*sqlcv1.V1Filter, error)
```
abelanger5: We haven't been doing this recently but really should be -- I prefer that we create structs with custom validation params at the repository layer and call validate on these methods at this layer as well. The most fully-featured example is the methods which upsert a workflow version I think, which you can use as a reference.

pkg/repository/v1/match.go
```diff
@@ -485,7 +485,7 @@ func (m *sharedRepository) processEventMatches(ctx context.Context, tx sqlcv1.DB
 
 					switch matchData.Action() {
 					case sqlcv1.V1MatchConditionActionQUEUE:
-						opt.Input = m.newTaskInput(input, matchData)
+						opt.Input = m.newTaskInput(input, matchData, nil)
```
abelanger5, line 488: I think we need to pass this to this level somehow? These are steps which have been triggered in the DAG, should they not have access to the triggering payload?

pkg/repository/v1/trigger.go
abelanger5: it's fine to make a copy of the pointer `filter := filterPtr`, but in new versions of Go this isn't necessary since each iteration of the loop gets `filterPtr` reassigned to a new value. This didn't used to be the case in < 1.22 or something like that, so if anyone ever decides to compile against an older version (cough cough PR bots) it's still ok to do this. hopefully don't need to worry about this in a few years

abelanger5 (changes requested):

pkg/repository/v1/sqlcv1/filters.sql
mrkaye97: oh yeah I guess that's redundant given the constraint, will remove that - thanks!

api/v1/server/handlers/v1/filters/list.go
mrkaye97: cool, makes sense to me - I can add the tenant as a param and then just leave it unused

abelanger5: why would it be unused?

mrkaye97: I guess one other question on this is do you think it's preferable to use the tenant id that's passed in, even if that means needing to construct the other non-tenant params in two places? like:

```go
	filters, err := t.config.V1.Filters().ListFilters(
		ctx.Request().Context(),
		tenant.ID.String(),
		sqlcv1.ListFiltersParams{
			Workflowids:  workflowIdParams,
			Scopes:       scopeParams,
			FilterLimit:  request.Params.Limit,
			FilterOffset: request.Params.Offset,
		},
	)
```

and then:

```go
func (r *filterRepository) ListFilters(ctx context.Context, tenantId string, params sqlcv1.ListFiltersParams) ([]*sqlcv1.V1Filter, error) {
	return r.queries.ListFilters(ctx, r.pool, sqlcv1.ListFiltersParams{
		Tenantid:     sqlchelpers.UUIDFromStr(tenantId),
		Workflowids:  params.Workflowids,
		Scopes:       params.Scopes,
		FilterLimit:  params.FilterLimit,
		FilterOffset: params.FilterOffset,
	})
}
```

IMO this is pretty ugly + brittle

abelanger5: I get why you're setting `sqlcv1` params as args to these methods, but we probably shouldn't be doing that anyway -- we're still missing validation. I know it seems like overkill, but there are 3 places we can add validation: (1) at the API handler layer, (2) at the repository layer, and (3) as constraints in the database. At a minimum, we should have 1 layer -- in this case we have none. 

So I would actually split out a separate struct called `ListFilterOpts` in the repository package with validators on the limit, offset, and lists of workflow ids (enforcing that the workflow ids are UUIDs) and scopes. 

Once we do this, it's not any uglier to construct params, and additionally it provides more safety for other callers, if we ever call `ListFilters` from some internal process.

And once we have this struct, we can call `ValidateAPI` to construct an `apierrors` automatically which is user-friendly.

It would be ideal to generate `sqlcv1.ListFiltersParams` with these validation tags, and while sqlc sort of supports it, it's not a pattern we've built out yet (and I think DB params are always going to need some transforming from the API layer regardless, so perhaps not a pattern we'd like to adopt in the long run).

mrkaye97: nice, thanks! makes sense to me

pkg/repository/v1/match.go
mrkaye97: ah yeah good call, I wasn't sure how to do this but I'll figure it out

abelanger5 (approved):
💯

PR #1710, opened 2025-05-13
fix: attempt to optimize match condition query

abelanger5:
# Description

Improves queries for match conditions which involve filtering for an `event_resource_hint`. This hint is used for filtering the number of potential match conditions for an event -- we use it internally to determine whether a task in a DAG should be triggered. 

This query was using `COALESCE` to select for both null and not null rows on the match condition, which basically skips the index. I then tried `IS NOT DISTINCT FROM`, but this is also not indexable in Postgres. 

I ended up splitting this query in two, one query which looks for match conditions using the `event_resource_hint`, and one query which doesn't. I analyzed some queries against large datasets (1.5M conditions) and saw sub-millisecond responses, where previously we would see >500ms response times. 

## Type of change

### DIFF
--- a/pkg/repository/v1/match.go
+++ b/pkg/repository/v1/match.go
@@ -312,37 +312,44 @@ func (m *sharedRepository) processEventMatches(ctx context.Context, tx sqlcv1.DB
 		}
 	}
 
-	// list all match conditions
-	matchConditions, err := m.queries.ListMatchConditionsForEventWithHint(
-		ctx,
-		tx,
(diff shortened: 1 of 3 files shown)


grutt (approved):

mrkaye97 (commented):
looks like maybe the SQL file and the `.sql.go` file are out of sync?

abelanger5:
> looks like maybe the SQL file and the .sql.go file are out of sync?

good catch, just going to delete the query in the SQL file to avoid confusion

PR #1714, opened 2025-05-14
Fix: Multiple Queries for Event Partitions

mrkaye97:
# Description

Splitting up the event partitions queries

## Type of change

### DIFF
--- a/pkg/repository/v1/olap.go
+++ b/pkg/repository/v1/olap.go
@@ -275,6 +275,17 @@ func (o *OLAPRepositoryImpl) UpdateTablePartitions(ctx context.Context) error {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


abelanger5 (changes requested):
@mrkaye97 I think one more thing is when we're cleaning up old partitions, we should probably explicitly skip these tables -- I'm not sure if our queries will read the Timescale partitions as part of the events table and detach them.

mrkaye97:
> @mrkaye97 I think one more thing is when we're cleaning up old partitions, we should probably explicitly skip these tables -- I'm not sure if our queries will read the Timescale partitions as part of the events table and detach them.

Sounds good!

abelanger5 (approved):

PR #1719, opened 2025-05-14
fix: retry count cast

grutt:
# Description
Retry count can possibly be null

### DIFF
--- a/pkg/repository/v1/sqlcv1/olap.sql
+++ b/pkg/repository/v1/sqlcv1/olap.sql
@@ -621,7 +621,7 @@ SELECT
     q.queued_at::timestamptz as queued_at,
     e.error_message as error_message,
     o.output::jsonb as output,
-    t.latest_retry_count as retry_count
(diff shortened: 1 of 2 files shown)


pkg/repository/v1/sqlcv1/olap.sql
```diff
@@ -621,7 +621,7 @@ SELECT
     q.queued_at::timestamptz as queued_at,
     e.error_message as error_message,
     o.output::jsonb as output,
-    t.latest_retry_count as retry_count
+    COALESCE(t.latest_retry_count, 0)::int as retry_count
```
grutt, line 624: i think the cast is required since t.latest_retry_count can be nullable and sqlc doesnt seem to be able to interpret it

abelanger5: Yep, cast + coalesce should be fine

abelanger5 (approved):

PR #1720, opened 2025-05-14
Feat: Remove inputs and outputs from workflow run list

mrkaye97:
# Description

Adding a boolean flag to not include the `input` and `output` payloads on some query results. It's a bit hacky since they'll just end up being empty but I think that's fine. I didn't want to change the types to allow the input and output to be nullable since that's a breaking change too, so I think this is okay for now.

Will follow this up with FE + SDK changes 

## Type of change


api-contracts/openapi/openapi.yaml
```diff
@@ -34,6 +34,8 @@ paths:
     $ref: "./paths/v1/tasks/tasks.yaml#/listTasksByDAGIds"
   /api/v1/stable/tenants/{tenant}/workflow-runs:
     $ref: "./paths/v1/workflow-runs/workflow_run.yaml#/listWorkflowRuns"
+  /api/v2/stable/tenants/{tenant}/workflow-runs:
+    $ref: "./paths/v1/workflow-runs/workflow_run.yaml#/v2ListWorkflowRuns"
```
mrkaye97: naming here sucks - I wasn't sure what to call this so I called it `/v2`, but obviously that's problematic. open to ideas here!

abelanger5 (approved):

PR #1727, opened 2025-05-15
Hotfix: Go SDK scheduled workflows issues

mrkaye97:
# Description

Fixing a couple scheduled workflows issues:

1. Namespaces weren't being applied properly on creating a scheduled workflow
2. Input wasn't being returned from the API

## Type of change

### DIFF
--- a/api/v1/server/oas/transformers/workflow_run.go
+++ b/api/v1/server/oas/transformers/workflow_run.go
@@ -532,10 +532,7 @@ func ToScheduledWorkflowsFromSQLC(scheduled *dbsqlc.ListScheduledWorkflowsRow) *
 
 	input := make(map[string]interface{})
 	if scheduled.Input != nil {
-		err := json.Unmarshal(scheduled.Input, &input)
-		if err != nil {
-			return nil
-		}
+		json.Unmarshal(scheduled.Input, &input)
 	}
 
 	res := &gen.ScheduledWorkflows{
--- a/api/v1/server/oas/transformers/workflow_run.go
+++ b/api/v1/server/oas/transformers/workflow_run.go
@@ -530,6 +530,14 @@ func ToScheduledWorkflowsFromSQLC(scheduled *dbsqlc.ListScheduledWorkflowsRow) *
 		workflowRunIdPtr = &workflowRunId
 	}
 
+	input := make(map[string]interface{})
+	if scheduled.Input != nil {
+		err := json.Unmarshal(scheduled.Input, &input)
+		if err != nil {
+			return nil
+		}
+	}
+
 	res := &gen.ScheduledWorkflows{
 		Metadata:             *toAPIMetadata(sqlchelpers.UUIDToStr(scheduled.ID), scheduled.CreatedAt.Time, scheduled.UpdatedAt.Time),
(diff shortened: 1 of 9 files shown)


pkg/client/admin.go
```diff
@@ -190,6 +190,10 @@ func (a *adminClientImpl) ScheduleWorkflow(workflowName string, fs ...ScheduleOp
 		return err
 	}
 
+	if a.namespace != "" && !strings.HasPrefix(workflowName, a.namespace) {
```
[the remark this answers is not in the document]
grutt: maybe good idea? we use the `_` char

pkg/v1/features/schedules.go
```diff
@@ -40,25 +42,31 @@ type CreateScheduledRunTrigger struct {
...
-		tenantId: tenantIdUUID,
+		api:       api,
+		tenantId:  tenantIdUUID,
+		namespace: namespace,
 	}
 }
 
 // Create creates a new scheduled workflow run.
 func (s *schedulesClientImpl) Create(ctx context.Context, workflowName string, trigger CreateScheduledRunTrigger) (*rest.ScheduledWorkflows, error) {
+	if s.namespace != nil && *s.namespace != "" && !strings.HasPrefix(workflowName, *s.namespace) {
```
grutt: helper for this?

api/v1/server/oas/transformers/workflow_run.go
```diff
@@ -544,6 +549,7 @@ func ToScheduledWorkflowsFromSQLC(scheduled *dbsqlc.ListScheduledWorkflowsRow) *
 		WorkflowRunName:      &scheduled.WorkflowRunName.String,
 		Method:               gen.ScheduledWorkflowsMethod(scheduled.Method),
 		Priority:             &scheduled.Priority,
+		Input:                &input,
```
grutt, line 552: we're returning the input on the list query here? we'll likely have the same issues as list runs at some point... your call on how we want to handle this

grutt (changes requested):
i think having a helper for the namespace prefix check will be very helpful soon

pkg/client/admin.go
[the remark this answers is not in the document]
mrkaye97: I think we shouldn't make this call here - this is copy pasted a bunch of other places and agree the underscore would make sense (we do this in python) but don't think we should change this in this change (more on namespacing soon...)

pkg/v1/features/schedules.go
mrkaye97: sure, sounds good to me

api/v1/server/oas/transformers/workflow_run.go
mrkaye97: I think this is probably fine to return the input for - IMO we should add that same query param to determine whether or not to return payloads, but I'll do that in the other PR and not here

pkg/client/admin.go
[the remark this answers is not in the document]
mrkaye97: FYI, looks like we do the same here as in Python where we concatenate the `_` to the namespace initially (at least in some cases, not sure how this plays with env vars): https://github.com/hatchet-dev/hatchet/blob/hotfix--go-sdk-namespaces/pkg/client/client.go#L179-L183

grutt (approved):

PR #1742, opened 2025-05-16
Feat  improved billing

grutt:
# Description

Improvements for usage based billing

### DIFF
--- a/frontend/app/src/components/cloud/billing/subscription.tsx
+++ b/frontend/app/src/components/cloud/billing/subscription.tsx
@@ -20,11 +20,13 @@ import {
 } from '@/lib/api/generated/cloud/data-contracts';
 import { useApiError } from '@/lib/hooks';
 import { TenantContextType } from '@/lib/outlet';
+import { ROUTES } from '@/next/lib/routes';
+import { applyCouponsToPrice } from '@/next/pages/authenticated/dashboard/settings/usage/components/coupon-utils';
 import queryClient from '@/query-client';
 import { ExclamationTriangleIcon } from '@radix-ui/react-icons';
 import { useMutation } from '@tanstack/react-query';
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
-import { useOutletContext } from 'react-router-dom';
+import { Link, useOutletContext } from 'react-router-dom';
 
 interface SubscriptionProps {
   active?: TenantSubscription;
@@ -50,6 +52,10 @@ export const Subscription: React.FC<SubscriptionProps> = ({
   const { tenant } = useOutletContext<TenantContextType>();
   const { handleApiError } = useApiError({});
   const [portalLoading, setPortalLoading] = useState(false);
+
+  useEffect(() => {
+    setShowAnnual(active?.period?.includes('yearly') || false);
+  }, [active]);
 
   const manageClicked = async () => {
     try {
@@ -92,14 +98,25 @@ export const Subscription: React.FC<SubscriptionProps> = ({
   const activePlanCode = useMemo(
     () =>
       active?.plan
-        ? [active.plan, active.period].filter((x) => !!x).join(':')
-        : 'free',
+        ? {
+            plan: active.plan,
+            period: active.period,
+          }
+        : {
+            plan: 'free',
+            period: 'monthly',
+          },
     [active],
   );
 
-  useEffect(() => {
-    return setShowAnnual(active?.period?.includes('yearly') || false);
-  }, [active]);
+  const isCustomPlan = useMemo(() => {
+    if (!active?.plan || !plans) {
+      return false;
+    }
+    return !plans.some(
+      (p) => p.plan_code === active.plan && p.period === active.period,
+    );
+  }, [active, plans]);
 
   const sortedPlans = useMemo(() => {
     return plans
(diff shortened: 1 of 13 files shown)


frontend/app/src/next/hooks/use-billing.ts
```diff
@@ -114,10 +114,7 @@ export default function useBilling({
     onSuccess: async () => {
       await Promise.all([
         queryClient.invalidateQueries({
-          queryKey: ['tenant:*'],
-        }),
-        queryClient.invalidateQueries({
-          queryKey: ['billing:*'],
+          queryKey: ['*'],
```
grutt, line 117: @mrkaye97 heads up this doesnt work 🤷

pkg/repository/postgres/tenant.go
```diff
@@ -85,6 +94,23 @@ func (r *tenantAPIRepository) CreateTenant(ctx context.Context, opts *repository
 		return nil, err
 	}
 
+	if user != nil {
+		// add the user as an owner of the tenant
+		_, err = r.CreateTenantMember(ctx, tenantId, &repository.CreateTenantMemberOpts{
```
grutt, line 99: it makes more sense to me to create a tenant owner on create, without this the hook is racy

frontend/app/src/components/cloud/billing/subscription.tsx
```diff
@@ -51,6 +53,10 @@ export const Subscription: React.FC<SubscriptionProps> = ({
   const { handleApiError } = useApiError({});
   const [portalLoading, setPortalLoading] = useState(false);
 
+  useEffect(() => {
+    setShowAnnual(active?.period?.includes('yearly') || false);
+  }, [active]);
```
mrkaye97, line 58: I know this was just moved up from below, but I'd think we could get rid of this effect hook and use derived state?

```typescript
const showAnnual = active?.period?.includes("yearly") || false
```

frontend/app/src/components/v1/cloud/billing/payment-methods.tsx
```diff
@@ -80,7 +80,7 @@ export function PaymentMethods({
                       <Icon size={24} />
                       {method.brand.toUpperCase()}
                       {method.last4 && ` *** *** ${method.last4} `}
-                      {method.expiration && `(Expires {method.expiration})`}
```
mrkaye97, line 83: 😬

frontend/app/src/components/v1/cloud/billing/subscription.tsx
```diff
@@ -51,6 +53,10 @@ export const Subscription: React.FC<SubscriptionProps> = ({
   const { handleApiError } = useApiError({});
   const [portalLoading, setPortalLoading] = useState(false);
 
+  useEffect(() => {
+    setShowAnnual(active?.period?.includes('yearly') || false);
+  }, [active]);
+
```
mrkaye97, line 59: same here re: derived state

frontend/app/src/next/hooks/use-billing.ts
mrkaye97: which part of it? the wildcard? if yes, I don't think I'd expect it to, there's nothing on this in the docs AFAICT

frontend/app/src/next/pages/authenticated/dashboard/settings/usage/components/coupon-utils.ts
```diff
@@ -0,0 +1,35 @@
...
+      adjusted = Math.round(adjusted * (1 - coupon.percent / 100));
+    }
+  }
+
+  // Then apply fixed-amount coupons
+  for (const coupon of coupons) {
+    if (coupon.amount_cents) {
+      adjusted = Math.max(0, adjusted - coupon.amount_cents);
+    }
+  }
```
mrkaye97, line 32: would be great to use `reduce` for these IMO

pkg/repository/postgres/tenant.go
```diff
@@ -35,7 +36,15 @@ func NewTenantAPIRepository(shared *sharedRepository, cache cache.Cacheable, def
...
 }
 
-func (r *tenantAPIRepository) CreateTenant(ctx context.Context, opts *repository.CreateTenantOpts) (*dbsqlc.Tenant, error) {
+func (r *tenantAPIRepository) RegisterCreateCallback(callback repository.UnscopedCallback[*dbsqlc.Tenant]) {
+	if r.createCallbacks == nil {
+		r.createCallbacks = make([]repository.UnscopedCallback[*dbsqlc.Tenant], 0)
+	}
+
+	r.createCallbacks = append(r.createCallbacks, callback)
+}
```
mrkaye97, line 45: O_O does Go allow this? Haven't seen this pattern before so I'm curious about it, since this feels a lot more like Java getter/setter patterns to me than Go's no-class approach to things

mrkaye97 (commented):
left some comments / questions!

frontend/app/src/next/hooks/use-billing.ts
grutt: interesting... ok then this is a gabe hallucination and not an llm one -- i could have sworn it supported wildcard query keys

mrkaye97: heh yeah I'm not sure, I thought I'd seen them before too but I was looking in the docs and couldn't find anything. I think maybe the right thing would be to split up the query keys into e.g. `['billing', 'foo']`, `['billing', 'bar']`, .. and then use `'billing'` as the key to invalidate? idk

PR #1750, opened 2025-05-19
Feat: UI Upgrade Flow

mrkaye97:
# Description

Adding a basic UI upgrade / downgrade flow to / from the V1 UI using a feature flag to gate rollout. Also added a `TenantMajorUIVersion` enum to the db and a `uiVersion` column to the `Tenant` table to store the version. The upgrade / downgrade buttons are gated behind a flag, so nobody will see them yet.

## Type of change


grutt:
@mrkaye97 can we put the generated stuff into a separate pr...

grutt (approved):

PR #1755, opened 2025-05-20
Fix: Quote replacement in generated doc snippets

mrkaye97:
# Description

Gets rid of quote replacement logic in the docs snippets

## Type of change


frontend/snips/src/processors/snippets/snippet.processor.ts
mrkaye97, line 1: this is the real change

abelanger5 (approved):

PR #1756, opened 2025-05-20
Feat: Tenant-in-path

mrkaye97:
# Description

Prefixing all authed paths with `/tenants/:tenantId` to simplify our handling of tenant switching, error handling, null checking, etc.

The main highlight here is this dramatically simplifies our handling of tenants, meaning:

1. We don't need to check if the `tenantId` is `undefined` anymore on any authed routes, since we know it's in the path
2. When switching tenants, we redirect but no longer need to update local storage / query params
3. We can remove a bunch of effect hooks for updating tenant id query params
4. Slight performance improvements for a bunch of queries since we no longer need to wait for the `tenant` query to complete before sending subsequent ones to fetch data

Some notes:

* The redirect from the old UI to the new one wasn't working (I think because of a race with an effect hook), so I added a button which is super ugly but fine for now, I think we can circle back here later
* Onboarding flow needs a lot of styling work
* I did a bunch of manual QA but we need to go through more carefully to make sure all the edges work correctly (i.e. redirects across versions, etc.)

TODO:

- [x] Figure out how to determine the tenant when logging in
- [x] Test login, signup, and logout paths (look for other edges where we go from authed -> not and vice versa too)
- [x] Figure out how to determine the tenant when landing
- [x] Set up redirect to tenanted route w/ default
- [x] Fix onboarding flows

## Type of change


frontend/app/src/next/components/errors/unauthorized.tsx
```diff
@@ -21,29 +19,3 @@ export function Unauthorized() {
     </div>
   );
 }
-
-export function WrongTenant({ desiredTenantId }: { desiredTenantId: string }) {
```
grutt, line 25: just to sanity check here the idea is we never have a url<>uuid pair that is incompatible

frontend/app/src/next/pages/authenticated/dashboard/components/sidebar/sidebar.tsx
grutt, line 1: these linter memo warnings are usually pretty important to address

grutt (approved):
lets lint before final merge for memo things

frontend/app/src/next/components/errors/unauthorized.tsx
mrkaye97: yeah exactly, we can just never get into a state where we'd show this now. Unauthorized is possible still, but this isn't

frontend/app/src/next/pages/authenticated/dashboard/components/sidebar/sidebar.tsx
mrkaye97: I think I got them all in #1765  - I'll just smush these two together

PR #1760, opened 2025-05-21
Bigger workflow visualizer

Ymirke:
# Description

We have somewhat big workflows, and looking at them with the small window of react-flow is quite annoying.

This is a suggestion to make the workflow view bigger. For us it would make sense that it _JUST_ is bigger, but it might be that this should be changed to be adjustable.

Fixes # (issue)
- Increased react-flow height from 300 to 800px. 

## Type of change

## What's Changed

- [x] Made react-flow view bigger for workflows


mrkaye97:
hey @Ymirke - thanks for putting this in! going to close it b/c we're planning to ship a new frontend soon that'll allow you to increase the size of the visualizer :)

PR #1772, opened 2025-05-23
Update rate limit example, await hatchet.ratelimits.upsert

tylerkahn:
The docs here https://docs.hatchet.run/home/rate-limits#static-rate-limits have this example referenced but there should be an `await`.

`hatchet.ratelimits.upsert()` returns a promise and should be awaited

#### Type of change

- Bug fix (non-breaking change which fixes an issue)
- Documentation change (pure documentation change)


grutt (approved):

grutt:
Thanks for fixing this!

PR #1788, opened 2025-05-27
Fix: UI Bug Burndown, Part II

mrkaye97:
# Description

More 🐛 

* Making the sidebar thingy draggable and fixing a bunch of rendering + layout issues there
* Doing the same ^ for the docs
* Fixing some dead links
* Fixing workflows empty state
* Fixing some filter issues
* Adding event runs tooltip
* Fixing redirect button issues with hard redirect to new UI

## Type of change


frontend/app/src/next/components/ui/docs-button.tsx
```diff
@@ -32,10 +33,12 @@ export function DocsButton({
   ...props
 }: DocsButtonProps) {
   const { open } = useDocs();
+  const { close: closeSideSheet } = useSideSheet();
 
   const handleClick = (e: React.MouseEvent) => {
     if (method === 'sheet') {
       e.preventDefault();
+      closeSideSheet();
```
grutt, line 44: 🤔 not sure where we should put this, but i think maybe it should be in open instead -- we're not directly using this button in the sidebar docs button for example

grutt (changes requested):
one minor tweak.

can also probably wait but the draggable width not being persistent is a bit annoying---especially for my larger monitor. if i set a larger size id want all sidebars to be that size

frontend/app/src/next/components/ui/docs-button.tsx
mrkaye97: Ah I was thinking this was how the docs were opened everywhere? The intention here is to close the "side sheet" (the non-docs side window with run details and stuff) when the docs open. AFAICT, this is a bug on main because they both open at the same time (sometimes overlapping, sometimes side by side) which seems incorrect to me. WDYT?

grutt: maybe we should just merge these into one state and have the docs component rendered as a sidebar state?

im cool with the "one sidebar" design concept but i think it's going to get hairy to have to open and close each thing

mrkaye97: Very much agree - I think it'd be much easier with one component

mrkaye97:
> one minor tweak.
> 
> can also probably wait but the draggable width not being persistent is a bit annoying---especially for my larger monitor. if i set a larger size id want all sidebars to be that size

I can give this a shot with local storage

grutt (approved):

PR #1791, opened 2025-05-28
Fix ts  feature client namespaces

### LINKED ISSUE #1780: [BUG] Namespaces are broken in V1 TS SDK
**Describe the issue**
Namespaces are broken in V1 TS SDK

**Environment**
- SDK: TypeScript v1.6.1
- Engines: Cloud v0 and latest self-hosted hatchet-lite

**Expected behavior**
When defining a namespace `foo` in my Hatchet client, via `const hatchet = Hatchet.init({namespace: 'foo'});`, I expect workflows and workers to consistently be prefixed with `foo_`, as was the case in V0. This is not the case with the V1 SDK. This prevents Hatchet from working, basically.

Suppose for example that:
- I have a workflow named `bar`, which I construct with `const wf = hatchet.workflow({name: 'bar', ... });`,
- I have a worker named `quux`, which I construct with `const wk = await hatchet.worker('quux');`.

In the hatchet dashboard, the worker will show up as `fooquux` instead of the expected `foo_quux`. Furthermore if I `await wk.registerWorkflows([wf])` then the workflow shows up in the dashboard as `foobar` instead of the expected `foo_bar`.

When submitting a workflow run request in another node app via, say, `await hatchet.admin.runWorkflow('bar', ...)` or `await wf.runNoWait(...)` then it gets the namespacing right and tries to run `foo_bar`, which it doesn't find. However if the worker registers the workflow with a horrible hack along the lines of `await wk.registerWorkflows([{...wf, name: "_" + wf.name}])`, then everything works as expected.

I had similar results when setting the namespace via the env var instead.

What I described above is reproducible using the hatchet-lite docker-compose engine and a very bare bones TS setup.

**Additional context**
This seems to be a bug in the TS SDK because I've been experiencing the same thing in both a local hatchet-lite deployment (via docker compose) and in the cloud v0 engine, which we'd like to upgrade.

grutt:
# Description
Fixes namespaces for scheduled and cron runs

Fixes: #1780

## Type of change


sdks/typescript/src/util/with-namespace.ts
```diff
@@ -0,0 +1,7 @@
+export function withNamespace(name: string, namespace?: string) {
```
mrkaye97: nit: I think we should call this the same thing (and maybe match the impl?) that it's called in [Python](https://github.com/hatchet-dev/hatchet/blob/fix--bug-burndown-part-ii/sdks/python/hatchet_sdk/config.py#L157-L171) and [Go](https://github.com/hatchet-dev/hatchet/blob/fix--bug-burndown-part-ii/pkg/config/client/client.go#L85)

mrkaye97: Although I guess those impls aren't really the same so 🤦

sdks/typescript/src/clients/admin/admin-client.ts
```diff
@@ -182,9 +183,7 @@ export class AdminClient {
     let computedName = workflowName;
```
mrkaye97: I'm kinda surprised the linter is fine with this but alas - I think this should be:

```typescript
const computedName = withNamespace(workflowName, this.config.namespace)
```

?

same in other places too if we can 🤷

sdks/typescript/src/clients/worker/worker.ts
```diff
@@ -121,7 +122,7 @@ export class V0Worker {
     for (const workflow of workflows) {
       const wf: Workflow = {
         ...workflow,
-        id: this.client.config.namespace + workflow.id,
+        id: withNamespace(workflow.id, this.client.config.namespace),
```
mrkaye97: I guess `id` here really means `name`?

sdks/typescript/src/clients/worker/worker.ts
```diff
@@ -144,7 +145,7 @@ export class V0Worker {
   async registerWorkflow(initWorkflow: Workflow) {
     const workflow: Workflow = {
       ...initWorkflow,
-      id: (this.client.config.namespace + initWorkflow.id).toLowerCase(),
+      id: withNamespace(initWorkflow.id, this.client.config.namespace).toLowerCase(),
```
mrkaye97: same here 😅  this is a bit confusing - I guess it'd be a breaking change to fix this maybe? this feels like the kind of thing we'll try to reread in a few months and not understand though

sdks/typescript/src/util/config-loader/config-loader.ts
```diff
@@ -94,7 +97,7 @@ export class ConfigLoader {
         (this.env('HATCHET_CLIENT_LOG_LEVEL') as LogLevel) ??
         'INFO',
       tenant_id: tenantId,
-      namespace: namespace ? `${namespace}_`.toLowerCase() : '',
```
mrkaye97, line 97: just confirming: was this related to that other bug a few months ago with the missing underscore or double underscore or something?

sdks/typescript/src/util/uuid.ts
```diff
@@ -0,0 +1,10 @@
+export function isValidUUID(uuid: string): boolean {
+  return /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i.test(uuid);
```
mrkaye97: there must be a way to do this with a library 😅

sdks/typescript/src/v1/client/client.ts
```diff
@@ -91,13 +93,6 @@ export class HatchetClient implements IHatchetClient {
         logger: logConstructor,
       };
 
-      // FIXME: Remove this once we have a proper namespace validation
-      if (clientConfig.namespace) {
-        clientConfig.namespace = clientConfig.namespace.endsWith('_')
-          ? clientConfig.namespace.slice(0, -1)
-          : clientConfig.namespace;
-      }
```
mrkaye97, line 99: 🥳

sdks/typescript/src/util/with-namespace.ts
grutt: matching name is fine

sdks/typescript/src/clients/admin/admin-client.ts
grutt: yeah weird, resolved regardless good catch

sdks/typescript/src/clients/worker/worker.ts
grutt:
```
  /**
   * @deprecated use definition.name instead
   */
  get id() {
    return this.definition.name;
  }
```

yeah... i have this getter for backwards compat. i'll go through and clean this up another day

mrkaye97 (approved):

sdks/typescript/src/clients/worker/worker.ts
mrkaye97: yeah sounds good, not a big deal

sdks/typescript/src/util/uuid.ts
grutt: good catch, we already have zod as a dep

sdks/typescript/src/util/config-loader/config-loader.ts
grutt: yes, added a safer way to handle this above

mrkaye97 (approved):
🥳

PR #1800, opened 2025-06-02
configurable task ops jitter

grutt:
# Description

Adds jitter to v1 task operations

## Type of change

### DIFF
--- a/internal/services/controllers/v1/olap/controller.go
+++ b/internal/services/controllers/v1/olap/controller.go
@@ -44,7 +44,7 @@ type OLAPControllerImpl struct {
 	updateDAGStatusOperations    *queueutils.OperationPool
 	processTenantAlertOperations *queueutils.OperationPool
 	samplingHashThreshold        *int64
-	olapConfig                   *server.ConfigFileOLAP
+	olapConfig                   *server.ConfigFileOperations
 }
 
 type OLAPControllerOpt func(*OLAPControllerOpts)
@@ -58,7 +58,7 @@ type OLAPControllerOpts struct {
 	p                     *partition.Partition
 	ta                    *alerting.TenantAlertManager
 	samplingHashThreshold *int64
-	olapConfig            *server.ConfigFileOLAP
+	olapConfig            *server.ConfigFileOperations
 }
 
 func defaultOLAPControllerOpts() *OLAPControllerOpts {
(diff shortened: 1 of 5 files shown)


pkg/config/server/server.go
```diff
@@ -578,6 +581,7 @@ func BindAllEnv(v *viper.Viper) {
 	_ = v.BindEnv("runtime.maxInternalRetryCount", "SERVER_MAX_INTERNAL_RETRY_COUNT")
 	_ = v.BindEnv("runtime.preventTenantVersionUpgrade", "SERVER_PREVENT_TENANT_VERSION_UPGRADE")
 	_ = v.BindEnv("runtime.defaultEngineVersion", "SERVER_DEFAULT_ENGINE_VERSION")
+	_ = v.BindEnv("runtime.operationsJitter", "SERVER_OPERATIONS_JITTER")
```
abelanger5: Do we need to remove what was added in this PR? https://github.com/hatchet-dev/hatchet/pull/1759

grutt: i'm okay with one config for this, depends on how granular we want things

abelanger5: I think one config makes sense since it's really a function of the number of tenants for both -- can always split it again if that's incorrect

internal/services/controllers/v1/olap/controller.go
```diff
@@ -180,33 +180,41 @@ func New(fs ...OLAPControllerOpt) (*OLAPControllerImpl, error) {
...
+
+	// Override with config value if available
+	if o.olapConfig != nil && o.olapConfig.PollInterval > 0 {
+		pollInterval = time.Second * time.Duration(o.olapConfig.PollInterval)
 	}
 
 	o.updateTaskStatusOperations = queueutils.NewOperationPool(
 		opts.l,
-		time.Second*15,
+		pollInterval,
```
grutt: this now defaults to 5s to match task ops

abelanger5: This isn't the polling interval, it's the context timeout. Polling is done with a scheduler on `Start` (I've been wanting to move this to the operation pool for a long time, but we're not there yet)

abelanger5 (changes requested):

internal/services/controllers/v1/olap/controller.go
grutt: Ooph, reverting

abelanger5 (approved):

PR #1806, opened 2025-06-03
Feat: Filter + Events Improvements

mrkaye97:
# Description

A couple of filters + events improvements:

* Allow for triggering multiple runs if there are multiple filter matches
* Handle triggering decisions correctly in the case that no filters are found but some do exist
* Adding event wildcard matching
* Adding a `scope` column to `v1_events_olap` and populating it correctly (corresponding cloud change [here](https://github.com/hatchet-dev/hatchet-cloud/pull/72))

## Type of change


pkg/repository/v1/sqlcv1/triggers.sql
```diff
@@ -26,8 +30,8 @@ JOIN
     "WorkflowTriggers" as triggers ON triggers."workflowVersionId" = latest_versions."workflowVersionId"
 JOIN
     "WorkflowTriggerEventRef" as eventRef ON eventRef."parentId" = triggers."id"
-WHERE
-    eventRef."eventKey" = ANY(@eventKeys::text[]);
+JOIN event_keys k ON k.event_key LIKE REPLACE(eventRef."eventKey", '*', '%')
```
mrkaye97, line 34: This will do a sequential scan over the `WorkflowTriggerEventRef`, but I think this is fine since there won't be very many rows here AFAICT

abelanger5 (approved):
looks good 💯

PR #1809, opened 2025-06-03
Feat: Tenant getter + corresponding SDK warnings

mrkaye97:
# Description

A desperate attempt to resolve SDK - Engine version mismatch issues by printing a loud warning

Fixes # (issue): So many of them

## Type of change


pkg/v1/features/tenant.go
```diff
@@ -0,0 +1,42 @@
+package features
```
mrkaye97, line 1: Added this here but I don't think we need to use it unless we make breaking SDK changes

sdks/typescript/src/v1/client/client.ts
```diff
@@ -105,6 +107,20 @@ export class HatchetClient implements IHatchetClient {
...
+          this.config
+            .logger('client-init', LogLevel.INFO)
+            .warn(
+              '🚨⚠️‼️ YOU ARE USING A V0 ENGINE WITH A V1 SDK, WHICH IS NOT SUPPORTED. PLEASE UPGRADE YOUR ENGINE TO V1.🚨⚠️‼️'
+            );
+        }
+      });
+    } catch (e) {
+      // Do nothing here
+    }
```
mrkaye97: @grutt think this is fine?

api/v1/server/handlers/tenants/get.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/hatchet-dev/hatchet/api/v1/server/oas/apierrors"
+	"github.com/hatchet-dev/hatchet/api/v1/server/oas/gen"
+	"github.com/hatchet-dev/hatchet/api/v1/server/oas/transformers"
+	"github.com/hatchet-dev/hatchet/pkg/repository/postgres/dbsqlc"
+)
+
+func (t *TenantService) TenantGet(ctx echo.Context, request gen.TenantGetRequestObject) (gen.TenantGetResponseObject, error) {
+	maybeTenant := ctx.Get("tenant").(*dbsqlc.Tenant)
+
+	if maybeTenant == nil {
```
abelanger5: This can never happen -- the populator should fail and we would panic on line 13, so this can be removed

abelanger5 (changes requested):

api/v1/server/handlers/tenants/get.go
mrkaye97: nice!

abelanger5 (approved):

PR #1811, opened 2025-06-03
fix: dont copy acks map

grutt:
# Description

### DIFF
--- a/internal/services/dispatcher/server_v1.go
+++ b/internal/services/dispatcher/server_v1.go
@@ -135,9 +135,7 @@ func (s *DispatcherImpl) subscribeToWorkflowRunsV1(server contracts.Dispatcher_S
 		wg.Add(1)
 		defer wg.Done()
 
-		workflowRunIds := acks.getNonAckdWorkflowRunsMap()
-
-		if matchedWorkflowRunIds, ok := s.isMatchingWorkflowRunV1(msg, workflowRunIds); ok {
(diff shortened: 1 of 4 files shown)


internal/services/dispatcher/server_v1.go
```diff
@@ -987,14 +985,14 @@ func (s *DispatcherImpl) msgsToWorkflowEvent(msgId string, payloads [][]byte, fi
...
-func (s *DispatcherImpl) isMatchingWorkflowRunV1(msg *msgqueue.Message, workflowRunIds map[string]bool) ([]string, bool) {
+func (s *DispatcherImpl) isMatchingWorkflowRunV1(msg *msgqueue.Message, acks *workflowRunAcks) ([]string, bool) {
 	switch msg.ID {
 	case "workflow-run-finished":
 		payloads := msgqueue.JSONConvert[tasktypes.NotifyFinalizedPayload](msg.Payloads)
 		res := make([]string, 0)
 
 		for _, payload := range payloads {
-			if _, ok := workflowRunIds[payload.ExternalId]; ok {
+			if _, ok := acks.acks[payload.ExternalId]; ok {
```
abelanger5: Need to protect this with an `RLock`, would recommend adding a method directly to `acks`

abelanger5 (changes requested):

internal/services/dispatcher/server_v1.go
grutt: fixed

abelanger5 (approved):

PR #1817, opened 2025-06-04
fix: drop go-cron for time after

grutt:
# Description

Simplifies scheduling to eliminate potential goroutine leaks

## Type of change

### DIFF
--- a/internal/services/ticker/schedule_workflow.go
+++ b/internal/services/ticker/schedule_workflow.go
@@ -88,8 +88,10 @@ func (t *TickerImpl) handleScheduleWorkflow(ctx context.Context, scheduledWorkfl
 		return nil
 	}
 
+	duration := time.Until(triggerAt)
+
 	// create a cancellation context for this specific scheduled workflow
-	cancelCtx, cancel := context.WithCancel(context.Background())
+	cancelCtx, cancel := context.WithTimeout(context.Background(), duration*2)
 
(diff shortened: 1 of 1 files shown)


internal/services/ticker/schedule_workflow.go
```diff
@@ -91,24 +88,28 @@ func (t *TickerImpl) handleScheduleWorkflow(ctx context.Context, scheduledWorkfl
...
-	_, err = s.NewJob(
-		gocron.OneTimeJob(
-			gocron.OneTimeJobStartDateTime(triggerAt),
-		),
-		gocron.NewTask(
-			t.runScheduledWorkflow(tenantId, workflowVersionId, scheduledWorkflowId, scheduledWorkflow),
-		),
-	)
+	// create a cancellation context for this specific scheduled workflow
+	cancelCtx, cancel := context.WithCancel(context.Background())
```
abelanger5: Why don't we set a timeout on this context? We know what the lookahead time should be; we can probably set the timeout as twice the lookup time pretty safely

abelanger5 (changes requested):

internal/services/ticker/schedule_workflow.go
grutt: good idea

grutt: pushed

abelanger5 (approved):

PR #1825, opened 2025-06-05
Feat: Declaratively create filters on worker start

mrkaye97:
# Description

Adding SDK support for creating filters when declaring a workflow, which then get created when the workflow is registered

Also did a bit of refactoring in Python, removing some duplicate code, and changed all the times we use `datetime.now()` to be UTC

TODO:

- [x] Go
- [x] Test TS
- [x] Changelog
- [x] Bump versions

## Type of change


sdks/python/hatchet_sdk/runnables/standalone.py
grutt, line 1: is this breaking for imports that were using it?

mrkaye97: don't think so unless someone is importing it directly, which they shouldn't be since it's exported at the top level 🤷 and I can't see many cases where you'd need to import this regardless

sdks/typescript/src/workflow.ts
```diff
@@ -51,6 +60,7 @@ export const CreateWorkflowSchema = z.object({
   on: OnConfigSchema,
   steps: StepsSchema,
   onFailure: CreateStepSchema?.optional(),
+  defaultFilters: DefaultFilterSchema.array().optional(),
```
grutt: these are v0 signatures, i dont think we need it here?

mrkaye97: ah thanks, I was looking at the wrong `registerWorkflow` impl - will fix this!

api-contracts/v1/workflows.proto
```diff
@@ -83,6 +83,14 @@ message CreateWorkflowVersionRequest {
...
     optional int32 default_priority = 11; // (optional) the default priority for the workflow
     repeated Concurrency concurrency_arr = 12; // (optional) the workflow concurrency options
+    repeated DefaultFilter default_filters = 13; // (optional) the default filters for the workflow
+}
+
+
+message DefaultFilter {
+    string expression = 1; // (required) the CEL expression for the filter
+    string scope = 2; // (required) the scope of the filter
+    optional string payload = 3; // (optional) the payload for the filter, if any. A JSON object as a string.
```
abelanger5: Not sure if this impacts too much, but recently I've been using `bytes` as the type instead of `string` which seems more appropriate

internal/services/admin/v1/server.go
```diff
@@ -540,6 +540,21 @@ func getCreateWorkflowOpts(req *contracts.CreateWorkflowVersionRequest) (*v1.Cre
 		cronInput = []byte(*req.CronInput)
 	}
 
+	defaultFilters := make([]v1.DefaultFilter, 0)
+	for _, f := range req.DefaultFilters {
+		var payload []byte
+
+		if f.Payload != nil {
+			payload = []byte(*f.Payload)
```
abelanger5: There should be a check that this is valid JSON and a user-friendly error if it's not

pkg/repository/v1/workflow.go
```diff
@@ -179,6 +181,17 @@ type CreateWorkflowStepRateLimitOpts struct {
...
 
+type DefaultFilter struct {
+	// (required) the filter expression
+	Expression string `json:"expression" validate:"required"`
+
+	// (required) the scope for the filter
+	Scope string `json:"scope" validate:"required"`
+
+	// (optional) the payload for the filter
+	Payload *[]byte `json:"payload,omitempty" validate:"omitempty"`
```
abelanger5: Don't need `*[]byte` since the zero-value of `[]byte` is `nil`

pkg/repository/v1/sqlcv1/filters.sql
```diff
@@ -4,24 +4,67 @@ INSERT INTO v1_filter (
...
     payload = EXCLUDED.payload,
+    is_declarative = EXCLUDED.is_declarative,
     updated_at = NOW()
 WHERE v1_filter.tenant_id = @tenantId::UUID
   AND v1_filter.workflow_id = @workflowId::UUID
   AND v1_filter.scope = @scope::TEXT
   AND v1_filter.expression = @expression::TEXT
 RETURNING *;
 
+-- name: DangerouslyBulkUpsertDeclarativeFilters :many
```
abelanger5: Not a huge fan of this naming, it seems to imply this is somehow unsafe/not fit for use, don't think that's the case?

pkg/repository/v1/sqlcv1/filters.sql
```diff
@@ -4,24 +4,67 @@ INSERT INTO v1_filter (
...
+)
+SELECT
+    @tenantId::UUID,
+    @workflowId::UUID,
+    scope,
+    expression,
+    payload,
+    is_declarative
+FROM inputs
+ON CONFLICT (tenant_id, workflow_id, scope, expression) DO UPDATE
```
abelanger5: Why would there be a conflict if we're doing a delete/re-insert?

pkg/repository/v1/sqlcv1/filters.sql
mrkaye97: Haha yeah, I wasn't sure what to call this - the point is to imply that this drops and updates everything and is really only intended to be called in this one place, but open to other naming ideas 😅. The intention was to make it so it's less likely to be (re)-used though

internal/services/admin/v1/server.go
mrkaye97: good call, will fix that

api-contracts/v1/workflows.proto
mrkaye97: nice! yeah I just copied what we had for the additional metadata, but this makes sense to me

pkg/repository/v1/sqlcv1/filters.sql
mrkaye97: ah good catch, yeah this was left over from before - will remove this line!

pkg/repository/v1/sqlcv1/filters.sql
mrkaye97: So I think maybe the right strategy is to have the queries handle the `is_declarative` and not pass that in as a param, so I did that [here](https://github.com/hatchet-dev/hatchet/pull/1825/commits/a7809c411b0d55f3fbf325d077070901ce2db24d)

internal/services/admin/v1/server.go
```diff
@@ -541,17 +542,16 @@ func getCreateWorkflowOpts(req *contracts.CreateWorkflowVersionRequest) (*v1.Cre
...
 
 	defaultFilters := make([]v1.DefaultFilter, 0)
-	for _, f := range req.DefaultFilters {
-		var payload []byte
 
-		if f.Payload != nil {
-			payload = []byte(*f.Payload)
+	for _, f := range req.DefaultFilters {
+		if f.Payload != nil && !json.Valid(f.Payload) {
+			return nil, fmt.Errorf("default filter payload is not valid JSON: %s", f.Payload)
```
abelanger5: We probably don't want to log the payload here?

abelanger5 (approved):

internal/services/admin/v1/server.go
mrkaye97: I was thinking it might be helpful for debugging - are you worried about size / security risk?

PR #1836, opened 2025-06-07
fix: checksum instability on triggers

abelanger5:
# Description

Fixes an issue with checksum instability caused by client-side UUIDs being included in the checksum input. 

## Type of change


pkg/repository/v1/workflow.go
```diff
@@ -899,19 +905,49 @@ func checksumV1(opts *CreateWorkflowVersionOpts) (string, error) {
 		return "", err
 	}
 
+	// Generate a unique index for each or group id in the workflow, and add this to the trigger condition.
+	// We would like to update the workflow version checksum only when the combination of or group ids changes.
+	orGroupIdsToIndex := make(map[string]int32)
+
+	for i, task := range opts.Tasks {
+		for j, condition := range task.TriggerConditions {
```
grutt, line 912: do we need to stable sort trigger conditions so or group id index is stable?

abelanger5: Eventually, but we order the steps and the trigger conditions should be stable when sent from the client, and I'm not sure what condition we would use for stable sorting (we can't use the UUIDs for obvious reasons). I'd prefer to merge this and circle back here.

grutt (approved):

PR #1852, opened 2025-06-11
fix(frontend-v1): get managed worker views compatible with current

abelanger5:
# Description

Basically gets views compatible with the existing:

image

image

image

image

## Type of change


frontend/app/src/next/pages/authenticated/dashboard/workers/managed/components/metrics-tab.tsx
```diff
@@ -0,0 +1,246 @@
...
+                setDefaultTimeRange(value);
+                setCustomTimeRange(undefined);
+              } else {
+                setCustomTimeRange([
+                  getCreatedAfterFromTimeRange(defaultTimeRange),
+                  new Date().toISOString(),
+                ]);
+              }
+            }}
+          >
```
mrkaye97: I think ideally we'd be able to use the existing `TimeFilter` and `TimeFilterGroup` for this

frontend/app/src/next/pages/authenticated/dashboard/workers/managed/components/managed-worker-instances-table.tsx
```diff
@@ -0,0 +1,129 @@
...
+          aria-label="Refresh instances list"
+        >
+          <ArrowPathIcon
+            className={`h-4 w-4 transition-transform ${rotate ? 'rotate-180' : ''}`}
+          />
+        </Button>
+      </div>
+      <DataTable columns={columns} data={data} emptyState={emptyState} />
+    </div>
+  );
```
mrkaye97, line 86: I think we probably do / should have a component for this

frontend/app/src/next/pages/authenticated/dashboard/workers/managed/components/managed-worker-build-logs.tsx
```diff
@@ -0,0 +1,28 @@
...
+export function ManagedWorkerBuildLogs({ buildId }: { buildId: string }) {
+  const getBuildLogsQuery = useQuery({
+    ...queries.cloud.getBuildLogs(buildId),
+    refetchInterval: 5000,
+  });
+
+  const logs = getBuildLogsQuery.data?.rows || [
+    {
+      line: 'Loading...',
+      timestamp: new Date().toISOString(),
```
mrkaye97: I think this `new Date` needs to be in a `useState` or `useRef` probably so it doesn't cause infinite renders

mrkaye97: (not 100% sure of this)

frontend/app/src/next/pages/authenticated/dashboard/workers/managed/components/managed-worker-instances-table.tsx
```diff
@@ -0,0 +1,129 @@
...
+}
+
+const INSTANCE_STATUSES: Record<
+  string,
```
mrkaye97: IMO, probably better to have a type here that's:

```typescript
type InstanceStatus = {
  text: string;
  variant: 'secondary' | 'destructive' | 'outline' | 'default';
}

type InstanceStatusMap = {
  started: InstanceStatus;
  suspended: InstanceStatus;
  destroyed: InstanceStatus;
  stopped: InstanceStatus;
}
```

and then have a switch over the cases so we get type checking for exhaustiveness

mrkaye97: (or just not use the `Record<string ...` type hint 🤷

frontend/app/src/next/pages/authenticated/dashboard/workers/managed/components/managed-worker-build-logs.tsx
abelanger5: not sure I'm following, why would this cause infinite renders?

frontend/app/src/next/pages/authenticated/dashboard/workers/managed/components/managed-worker-instances-table.tsx
abelanger5: going to skip for now

frontend/app/src/next/pages/authenticated/dashboard/workers/managed/components/managed-worker-instances-table.tsx
abelanger5: sounds good, added

frontend/app/src/next/pages/authenticated/dashboard/workers/managed/components/managed-worker-build-logs.tsx
mrkaye97: I think because `new Date()` is unstable and will get recalculated over and over, but I might be wrong on this! I saw something like this happen a couple weeks ago 🤷

frontend/app/src/next/pages/authenticated/dashboard/workers/managed/components/metrics-tab.tsx
abelanger5: used this; think it's working as long as we don't turn those components/hooks stateful w/ query params

mrkaye97 (approved):

frontend/app/src/next/pages/authenticated/dashboard/workers/managed/components/managed-worker-build-logs.tsx
abelanger5: ah gotcha, yeah that could be an issue

PR #1856, opened 2025-06-12
fix: listener v2 status race

grutt:
# Description
This addresses the race condition where multiple listeners can interfere with each other's active status

 Race condition scenario:
1. Worker A establishes ListenV2 connection, sets worker active=true with timestamp T1
2. Worker B establishes ListenV2 connection, sets worker active=true with timestamp T2 (T2 > T1)
3. Worker A disconnects, sets worker active=false with timestamp T1
4. Since T1 <= T2, the database update succeeds and worker is marked inactive
5. Worker B is still connected but worker appears inactive, causing heartbeat rejections

This polling goroutine ensures that as long as ANY ListenV2 connection is active,
the worker status remains active=true, preventing the race condition

## Type of change


internal/services/dispatcher/server.go
```diff
@@ -492,6 +492,54 @@ func (s *DispatcherImpl) ListenV2(request *contracts.WorkerListenRequest, stream
 		s.workers.DeleteForSession(request.WorkerId, sessionId)
 	}()
 
+	// HOTFIX: Start a background goroutine to poll and ensure worker stays active
+	// This addresses the race condition where multiple listeners can interfere with each other's active status
+	//
+	// Race condition scenario:
+	// 1. Worker A establishes ListenV2 connection, sets worker active=true with timestamp T1
```
abelanger5, line 325: I would update this comment to indicate that it is the same worker that's establishing the connection, this is incorrect as written

abelanger5 (dismissed):

PR #1857, opened 2025-06-12
Fix: Attempts at improving logging fatigue

mrkaye97:
# Description

Fixes a couple logging issues

## Type of change

### DIFF
--- a/internal/msgqueue/v1/rabbitmq/rabbitmq.go
+++ b/internal/msgqueue/v1/rabbitmq/rabbitmq.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hatchet-dev/hatchet/internal/telemetry"
 	"github.com/hatchet-dev/hatchet/pkg/logger"
 	"github.com/hatchet-dev/hatchet/pkg/random"
+	"github.com/hatchet-dev/hatchet/pkg/repository/metered"
 )
 
 const MAX_RETRY_COUNT = 15
@@ -609,6 +610,10 @@ func (t *MessageQueueImpl) subscribe(
 
 				if err := preAck(msg); err != nil {
 					t.l.Error().Msgf("error in pre-ack on msg %s: %v", msg.ID, err)
+
+					if err == metered.ErrResourceExhausted {
+						return
+					}
 
 					// nack the message
 					if err := rabbitMsg.Reject(false); err != nil {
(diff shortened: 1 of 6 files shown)


internal/msgqueue/v1/rabbitmq/rabbitmq.go
```diff
@@ -610,6 +611,10 @@ func (t *MessageQueueImpl) subscribe(
 				if err := preAck(msg); err != nil {
 					t.l.Error().Msgf("error in pre-ack on msg %s: %v", msg.ID, err)
 
+					if err == metered.ErrResourceExhausted {
```
abelanger5: This probably isn't the right place for this -- I would put it in the tasks controller. The MQ shouldn't concern itself with specific error states and we have two MQ implementations, so this is a bit confusing

internal/services/dispatcher/server.go
```diff
@@ -549,8 +549,7 @@ func (s *DispatcherImpl) Heartbeat(ctx context.Context, req *contracts.Heartbeat
 		span.RecordError(err)
 		span.SetStatus(telemetry_codes.Error, "could not get worker")
 		if errors.Is(err, pgx.ErrNoRows) {
-			s.l.Error().Msgf("worker %s not found", req.WorkerId)
-			return nil, err
+			return nil, fmt.Errorf("worker %s not found", req.WorkerId)
```
abelanger5: This is still returning a 500-level error to the user. We want to return a 404 or something similar

internal/services/ticker/cron.go
```diff
@@ -95,6 +95,7 @@ func (t *TickerImpl) handleScheduleCron(ctx context.Context, cron *dbsqlc.PollCr
 	)
 
 	if err != nil {
+		t.repo.Workflow().DeleteInvalidCron(ctx, cron.ID)
```
abelanger5: handle error and make sure it's logged?

abelanger5: also: should we be pretty sure that this is a cron validation error, not some other random error? perhaps validation errors get silently removed, othe errors we can still be noisier about

abelanger5 (changes requested):

internal/services/dispatcher/server.go
mrkaye97: ah whoops, missed that 🤦 thanks

abelanger5 (approved):

PR #1859, opened 2025-06-12
Feat: Events Frontend, Redux

mrkaye97:
# Description

Adding back the events frontend + V1-ifying it with some new queries for fetching + filtering events, keys, etc.

TODO:

- [x] Figure out cached state for filters (add a filter, old runs still show up)
- [x] Make sure success / fail counts are correct (some duplicates, seemingly, or not all runs being fetched properly)
- [ ] Should additional meta filters be able to filter by non-string types? <-- I don't think we can really do this, which is a bummer

## Type of change


cmd/hatchet-migrate/migrate/migrations/20250612173730_v1_0_25.sql
```diff
@@ -0,0 +1,9 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE INDEX v1_events_olap_scope_idx ON v1_events_olap (tenant_id, scope, seen_at DESC) WHERE scope IS NOT NULL;
```
mrkaye97: partial index to enable filtering by scope + seen at. one note here is we can't create this index concurrently unless we loop over all the partitions, but I figured this was probably fine

pkg/repository/v1/sqlcv1/olap-overwrite.sql
```diff
@@ -219,3 +219,134 @@ SELECT *
 FROM to_insert
 RETURNING *
 ;
+
+
+-- name: ListEvents :many
```
mrkaye97: needed to move these queries here to play nicely with the status enum

pkg/repository/v1/sqlcv1/olap-overwrite.sql
```diff
@@ -219,3 +219,134 @@ SELECT *
...
+WITH included_events AS (
+    SELECT DISTINCT e.*
```
mrkaye97: important qq: since runs are now one to many with events, we need to dedupe somehow after joining to the runs. This is the simplest way I could think of, but I'm unsure about performance here - any other ideas? We could e.g. find the first run for each event and then join to that or some such

pkg/repository/v1/sqlcv1/olap.sql
```diff
@@ -1352,85 +1352,11 @@ VALUES (
...
-    e.key AS event_key,
-    e.payload AS event_payload,
-    e.additional_metadata AS event_additional_metadata,
-    sc.queued_count,
-    sc.running_count,
-    sc.completed_count,
-    sc.cancelled_count,
-    sc.failed_count
+-- name: ListEventKeys :many
+SELECT DISTINCT key
```
mrkaye97, line 1358: ugly diff, but also important - I was thinking a distinct over the keys would be fine with the index + the `seen_at` filter so we only scan one or two partitions at most, but correct me on this if not!

pkg/repository/v1/sqlcv1/olap-overwrite.sql
abelanger5: what specifically? there's a trick for the enums with `cast(...)`, perhaps that's the case you hit

pkg/repository/v1/sqlcv1/olap-overwrite.sql
abelanger5: seems alright to try this for now

pkg/repository/v1/sqlcv1/olap.sql
abelanger5: Yeah it's probably alright. We used to have an event key table and use a cached update on it, but we can always do something fancier when we hit problems here

sql/schema/v1-olap.sql
```diff
@@ -388,6 +388,7 @@ CREATE TABLE v1_events_olap (
 ) PARTITION BY RANGE(seen_at);
 
 CREATE INDEX v1_events_olap_key_idx ON v1_events_olap (tenant_id, key);
+CREATE INDEX v1_events_olap_scope_idx ON v1_events_olap (tenant_id, scope, seen_at DESC) WHERE scope IS NOT NULL;
```
abelanger5: isn't `seen_at` redundant here, since it's the partition key?

abelanger5 (changes requested):

pkg/repository/v1/sqlcv1/olap-overwrite.sql
mrkaye97: oh nice! yeah, it was a typing issue on the Go side I think, what's the cast trick? I'll try to find that

sql/schema/v1-olap.sql
mrkaye97: ah yeah good call, I'll remove that

abelanger5 (approved):

PR #1860, opened 2025-06-13
Feat: Batched replays

mrkaye97:
# Description

Implementing batching for bulk replay so we send batches of ~100 tasks to replay, sleep for a little, and continue. A couple cases for batching:

1. We group by workflow run external id, so all the tasks under a single workflow run are in the same batch
2. If a workflow run has >100 tasks to replay, it's its own batch
3. Otherwise, greedily add tasks to the batch until its size would exceed 100

It's not really optimal in any way (lots of looping + batch sizes won't be consistent), but I think it should pretty much work

## Type of change


internal/services/admin/v1/server.go
```diff
@@ -234,49 +240,100 @@ func (a *AdminServiceImpl) ReplayTasks(ctx context.Context, req *contracts.Repla
...
+
+		if err != nil {
+			return nil, err
+		}
+
+		for _, task := range batch {
+			replayedIds = append(replayedIds, taskIdInsertedAtRetryCountToExternalId[task].String())
+		}
+
+		time.Sleep(200 * time.Millisecond)
```
mrkaye97, line 332: do we need to sleep here? and for how long if we do?

internal/services/admin/v1/server.go
```diff
@@ -150,6 +151,11 @@ func (a *AdminServiceImpl) CancelTasks(ctx context.Context, req *contracts.Cance
 	}, nil
 }
 
+type TaskIdInsertedAtRetryCountWithWorkflowRunId struct {
+	v1.TaskIdInsertedAtRetryCount
+	WorkflowRunId pgtype.UUID `json:"workflow_run_id" db:"workflow_run_id"`
```
abelanger5: what is this `db:` tag?

internal/services/admin/v1/server.go
abelanger5: this seems reasonable

abelanger5 (approved):

internal/services/admin/v1/server.go
mrkaye97: lol, copilot autocomplete - thanks, will remove that

PR #1870, opened 2025-06-17
fix: Remove explicit public schema reference in WorkflowTriggerEventRef queries

yang6n:
# Description

When running Hatchet with a non-public schema (e.g., `?search_path=hatchet`), the "get workflows" API was returning a 500 error. The error occurred because some SQL queries were explicitly referencing the `public` schema for the `WorkflowTriggerEventRef` table, even though the search path was set to `hatchet`.

## Impact
This change allows Hatchet to work correctly with non-public schemas, making it more flexible for different database configurations.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue): https://github.com/hatchet-dev/hatchet/issues/1871

## What's Changed

- Removed explicit `public` schema references from the `WorkflowTriggerEventRef` table in the `listWorkflowsLatestRuns` query
- The query will now respect the configured search path instead of hardcoding the schema


grutt:
Hi @yang6n thank you for your contribution!

You'll need to generate the corresponding go fixture with `task generate-sqlc` (or make these corresponding changes manually in`pkg/repository/postgres/dbsqlc/workflows.sql.go`)

yang6n:
@grutt - added the go fixture. Let me know if there is anything else need to be updated. Thanks!

fairchild (approved):

grutt (approved):

PR #1886, opened 2025-06-20
Feat: Streaming Improvements

mrkaye97:
# Description

Fixing our streaming implementation by adding a stream event buffer and sending indices from the SDKs (have only done Python so far). Then we buffer based on the index for a couple of seconds at most so we can do our best to guarantee ordering.

If no index gets send, we should just do what we were doing before without buffering

## Type of change


internal/services/dispatcher/server_v1.go
```diff
@@ -558,6 +747,38 @@ func (s *DispatcherImpl) subscribeToWorkflowEventsByWorkflowRunIdV1(workflowRunI
 	var mu sync.Mutex     // Mutex to protect activeRunIds
 	var sendMu sync.Mutex // Mutex to protect sending messages
 
+	streamBuffer := NewStreamEventBuffer(5 * time.Second)
+
+	go func() {
+		ticker := time.NewTicker(50 * time.Millisecond)
```
mrkaye97: not sure if this is too small of a value or not

internal/services/dispatcher/server_v1.go
```diff
@@ -27,6 +27,194 @@ import (
 	tasktypes "github.com/hatchet-dev/hatchet/internal/services/shared/tasktypes/v1"
 )
 
+type StreamEventBuffer struct {
+	stepRunIdToWorkflowEvents map[string][]*contracts.WorkflowEvent
+	stepRunIdToExpectedIndex  map[string]int64
+	stepRunIdToLastSeenTime   map[string]time.Time
+	stepRunIdToCompletionTime map[string]time.Time
+	mu                        sync.RWMutex
```
abelanger5: Looks like read locking in `RWMutex` isn't used here so more straightforward to just use `Mutex`

internal/services/dispatcher/server_v1.go
```diff
@@ -27,6 +27,194 @@ import (
...
+	}
+
+	b.stepRunIdToWorkflowEvents[stepRunId] = append(b.stepRunIdToWorkflowEvents[stepRunId], event)
+
+	slices.SortFunc(b.stepRunIdToWorkflowEvents[stepRunId], sortByEventIndex)
+
+	return b.getReadyEvents(stepRunId)
+}
+
+func (b *StreamEventBuffer) GetTimedOutEvents() []*contracts.WorkflowEvent {
```
abelanger5: I think there's a simpler way to handle timed out events. The stream buffer should return a channel called `timedOutEvents` which has a max length (something like 100 will be fine) and buffer value. 

Internally you can write to this channel from an internal FIFO queue (also a channel), let's call it `timedOutEventProducer` with a `nextTimeoutAt` value representing the front of the queue. Then you just produce writes on the `timedOutEvents` channel from a single goroutine by calling `time.After` on `nextTimeoutAt` before processing the next event our of `timedOutEventProducer`.

Benefits this gets you:
- No for loops, just single writes to a channel
- The caller isn't responsible for polling the stream event buffer, you're just producing writes on a channel
- I think (?) this will cut down quite a bit on complexity

Happy to pair on this.

abelanger5: Actually, I think more broadly, there should be a single channel that the stream event buffer writes events to, and that should be returned after initializing the buffer. That way `AddEvent` doesn't need to return the slice of events either. There's only one producer and one consumer.

abelanger5 (changes requested):

PR #1890, opened 2025-06-23
Fix: Store `CreateWorkflowVersionOpts` for debugging

mrkaye97:
# Description

Adding a column to the workflow version to store the unhashed opts for debugging

## Type of change


pkg/repository/v1/sqlcv1/workflows.sql.go
```diff
@@ -741,21 +742,23 @@ INSERT INTO "WorkflowVersion" (
...
+	CreatedAt                 pgtype.Timestamp   `json:"createdAt"`
+	UpdatedAt                 pgtype.Timestamp   `json:"updatedAt"`
+	Deletedat                 pgtype.Timestamp   `json:"deletedat"`
+	Checksum                  string             `json:"checksum"`
+	Version                   pgtype.Text        `json:"version"`
+	Workflowid                pgtype.UUID        `json:"workflowid"`
+	Sticky                    NullStickyStrategy `json:"sticky"`
+	Kind                      NullWorkflowKind   `json:"kind"`
+	DefaultPriority           pgtype.Int4        `json:"defaultPriority"`
+	CreateWorkflowVersionOpts []byte             `json:"createWorkflowVersionOpts"`
```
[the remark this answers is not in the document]
mrkaye97: 🤦

pkg/repository/v1/workflow.go
```diff
@@ -370,10 +370,17 @@ func (r *workflowRepository) createWorkflowVersionTxs(ctx context.Context, tx sq
 		return sqlchelpers.UUIDToStr(oldWorkflowVersion.WorkflowVersion.ID), nil
 	}
 
+	optsJson, err := json.Marshal(opts)
```
abelanger5: Instead of passing in these `opts`, we should return the modified `opts` out of `checksumV1` since it's what we're actually using in the checksum (these are slightly different, and I'm not sure if checksumV1 modifies in-place or not)

abelanger5 (changes requested):

frontend/app/src/pages/main/v1/workflows/$workflow/index.tsx
```diff
@@ -204,44 +200,46 @@ export default function ExpandedWorkflow() {
...
+                        onClick={() => {
+                          setDeleteWorkflow(true);
+                        }}
+                      >
+                        Delete Workflow
+                      </Button>
+                    </div>
+                  </div>
+                </div>
               </div>
```
mrkaye97, line 230: this appeared to be doing nothing? just removed it 🤷

abelanger5 (approved):

PR #1893, opened 2025-06-24
Feat  improve auth error handling

grutt:
# Description
Adds basic rate limiting and consistent error responses for auth endpoints


api/v1/server/handlers/users/update_login.go
```diff
@@ -31,35 +30,39 @@ func (u *UserService) UserUpdateLogin(ctx echo.Context, request gen.UserUpdateLo
...
 
 	if err != nil {
-		return nil, fmt.Errorf("could not get user password: %w", err)
+		u.config.Logger.Err(err).Msg("failed to get user password")
+		return gen.UserUpdateLogin400JSONResponse(apierrors.NewAPIErrors(ErrInvalidCredentials)), nil
 	}
 
 	if verified, err := repository.VerifyPassword(userPass.Hash, request.Body.Password); !verified || err != nil {
-		return gen.UserUpdateLogin400JSONResponse(apierrors.NewAPIErrors("invalid password")), nil
+		return gen.UserUpdateLogin400JSONResponse(apierrors.NewAPIErrors(ErrInvalidCredentials)), nil
```
mrkaye97, line 58: should all of these be 400 or 401/403?

api/v1/server/handlers/users/update_password.go
```diff
@@ -42,27 +40,30 @@ func (u *UserService) UserUpdatePassword(ctx echo.Context, request gen.UserUpdat
...
 
 	if err != nil {
-		return nil, fmt.Errorf("could not get user password: %w", err)
+		u.config.Logger.Err(err).Msg("failed to get user password")
+		return gen.UserUpdatePassword400JSONResponse(apierrors.NewAPIErrors(ErrInvalidCredentials)), nil
 	}
 
 	if verified, err := repository.VerifyPassword(userPass.Hash, request.Body.Password); !verified || err != nil {
-		return gen.UserUpdatePassword400JSONResponse(apierrors.NewAPIErrors("invalid password", "password")), nil
+		return gen.UserUpdatePassword400JSONResponse(apierrors.NewAPIErrors(ErrInvalidCredentials)), nil
```
mrkaye97, line 48: maybe 401/403 here too?

api/v1/server/middleware/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,72 @@
...
+				return true
+			}
+
+			enableRateLimiting := enableRateLimitingInt.(bool)
+
+			return !enableRateLimiting
+		},
+		Store: echoMiddleware.NewRateLimiterMemoryStoreWithConfig(
+			echoMiddleware.RateLimiterMemoryStoreConfig{Rate: rate.Limit(m.config.Runtime.APIRateLimit), ExpiresIn: m.config.Runtime.APIRateLimitWindow},
+		),
```
mrkaye97, line 52: FYI, I think I have either this or something very similar set up for webhooks - might be useful to check that out and either consolidate, get rid of the other implementation, etc.!

api/v1/server/handlers/users/update_login.go
mnafees: I think 400 makes sense here since it is a bad request

api/v1/server/handlers/users/update_password.go
mnafees: I think 400 works fine here too

api/v1/server/handlers/users/update_login.go
mrkaye97: oooh sounds good! I read "invalid password" the other way 😅

api/v1/server/handlers/users/update_password.go
mrkaye97: same here

mrkaye97 (approved):

PR #1899, opened 2025-06-24
Fix: Filter list improvements

mrkaye97:
# Description

Couple of improvements to filter listing:

* Fixes a broken uuid validator - just relying on Go's typing instead
* Cleans up the filter list query a bit (no need for override anymore) to allow listing everything and have a stable ordering - we had the optional thing because we were passing a nullable scope in earlier on event pushes, but it's better to just not include those rows in the inputs since we'd never match anything anyways in that case
* Fixes a Python bug with the cached workflow id

## Type of change


pkg/repository/v1/sqlcv1/filters.sql
```diff
@@ -88,3 +88,24 @@ WHERE
...
+), num_filter_inputs AS (
+    SELECT COUNT(*) AS ct
+    FROM inputs
+    WHERE workflow_id IS NOT NULL AND scope IS NOT NULL
+)
+
+SELECT f.*
+FROM v1_filter f
+CROSS JOIN num_filter_inputs n
+JOIN inputs i ON (n.ct = 0) OR (f.workflow_id, f.scope) = (i.workflow_id, i.scope)
```
mrkaye97: any thoughts on this? it felt pretty hacky but not sure if we have a different way that's preferred

PR #1913, opened 2025-06-26
Fix: Streaming Bugs

mrkaye97:
# Description

Fixing two bugs in streaming:

1. We'd sometimes (maybe like 1/4 of the time) get an error about an unexpected end to JSON input, which turned out to be because we were `Unmarshal`ling `nil` sometimes (an empty byte array). I think this is probably caused by a race or something but I'm not sure - in any case, seems like just short circuiting fixes it
2. We weren't hanging up on failure or cancelled events, which caused the listeners in the SDK to just hang
3. Publish stream events even if tenant pubs is disabled

## Type of change

### DIFF
--- a/internal/msgqueue/v1/rabbitmq/rabbitmq.go
+++ b/internal/msgqueue/v1/rabbitmq/rabbitmq.go
@@ -231,7 +231,7 @@ func (t *MessageQueueImpl) pubMessage(ctx context.Context, q msgqueue.Queue, msg
 	}
 
 	// if this is a tenant msg, publish to the tenant exchange
-	if (!t.disableTenantExchangePubs || msg.ID != "task-stream-event") && msg.TenantID != "" {
+	if (!t.disableTenantExchangePubs || msg.ID == "task-stream-event") && msg.TenantID != "" {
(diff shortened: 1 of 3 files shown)


abelanger5 (approved):

internal/msgqueue/v1/rabbitmq/rabbitmq.go
```diff
@@ -231,7 +231,7 @@ func (t *MessageQueueImpl) pubMessage(ctx context.Context, q msgqueue.Queue, msg
 	}
 
 	// if this is a tenant msg, publish to the tenant exchange
-	if !t.disableTenantExchangePubs && msg.TenantID != "" {
+	if !t.disableTenantExchangePubs && msg.TenantID != "" && msg.ID != "task-stream-event" {
```
abelanger5: should this be

```
if msg.TenantID != "" && (!t.disableTenantExchangePubs || msg.ID == "task-stream-event")
```

(we want to send stream events even if we've disabled all other pubs?)

abelanger5 (changes requested):

abelanger5 (approved):

PR #1917, opened 2025-06-27
Feat: Top-level stream consumer in the SDKs

mrkaye97:
# Description

Adding a top-level stream consumer to the SDKs in the runs client to make it easier to subscribe and iterate over a stream without needing to worry about internal stuff.

Also wrote up a NextJS example. Recorded this video, but just left the backend part checked in, since I don't think we want to introduce NextJS here. It'll be enough to be able to use for a snippet in the docs to show how to proxy with Next

https://github.com/user-attachments/assets/54a402c0-8511-44ed-a66f-579a1d43ecd4

## Type of change


sdks/typescript/src/clients/listeners/run-listener/child-listener-client.ts
```diff
@@ -136,8 +136,11 @@ export class RunEventListener {
           });
         }
       }
+
+      this.eventEmitter.emit('complete');
```
mrkaye97, line 140: @grutt would be good to have your 👀 on these changes - it seems like in some situations the stream just would never hang up without this `complete` event (or a similar one) to indicate we need a `break`. It'd work in a one off script, but not with Next when trying to use fetch / `ReadableStream` / `Response`.

abelanger5 (approved):

PR #1926, opened 2025-06-30
fix: concurrency timeout from 5s -> 30s

abelanger5:
# Description

Adds some more overhead for concurrency timeouts with large queues (usually due to a spike in workflows that hit a scheduling timeout)

## Type of change


mnafees (approved):
LGTM

PR #1928, opened 2025-07-01
Remove `nginx` and use custom static fileservers

mnafees:
# Description

Remove dependency on `nginx` to serve frontend files in `frontend.dockerfile`

## Type of change

## What's Changed

- [x] Uses the `hatchet-fileserver` binary to serve static files for in `frontend.dockerfile`


build/package/frontend.dockerfile
```diff
@@ -9,13 +10,25 @@ RUN corepack pnpm@9.15.4 install --frozen-lockfile && corepack pnpm@9.15.4 store
...
-ARG APP_TARGET=client
+WORKDIR /app
+
+COPY go.mod go.sum ./
+COPY ./cmd/hatchet-staticfileserver/ ./cmd/hatchet-staticfileserver/
+
+RUN go build -o hatchet-staticfileserver ./cmd/hatchet-staticfileserver/main.go
+RUN chmod +x ./hatchet-staticfileserver
+
+# Stage 3: Run the static fileserver
```
mrkaye97, line 24: IMO, we should remove these types of comments :P

mrkaye97 (approved):

abelanger5 (approved):

build/package/frontend.dockerfile
mnafees: It is a multi stage Dockerfile and I think comments just make it easier for the eye to self navigate

PR #1934, opened 2025-07-01
Fix: Remove `task-runs` route + lots of dead code

mrkaye97:
# Description

Getting rid of a ton of dead code + the `task-runs` route in favor of just dynamically figuring out if something is a DAG or a task

Removed the whole `/next` directory - figure we can easily bring back individual parts, but it was making linters, etc. take forever and made searching the repo dramatically harder

This file is the one that's actually modified here: `frontend/app/src/pages/main/v1/workflow-runs-v1/$run/index.tsx`

## Type of change


frontend/app/src/pages/main/v1/workflow-runs-v1/$run/index.tsx
mrkaye97, line 1: this is the actual change

abelanger5 (approved):

PR #1937, opened 2025-07-01
Feat: Improve overflow / scroll behavior on table pages

mrkaye97:
# Description

### [Review without whitespace](https://github.com/hatchet-dev/hatchet/pull/1937/files?diff=unified&w=1)

Improving overflow / scroll behavior on various table pages (runs table, workers table, tasks table, events table, ...) to not overflow, and instead scroll within their given size. Doing this because the full-page scrolling where if you scroll down you lose all the header / metrics context and if you scroll up you lose the pagination state has been bothering me.

Also removed the page headers like `Events` or `Runs` since they take up a ton of space and don't add much since the sidebar is right there

Also got rid of the card views for the workers + workflows and fixed the durations in the runs view (did that after taking screenshots)

examples:
Screenshot 2025-07-01 at 4 16 18 PM

Screenshot 2025-07-01 at 4 16 28 PM

## Type of change


frontend/app/src/components/v1/molecules/data-table/data-table-pagination.tsx
```diff
@@ -66,7 +66,7 @@ export function DataTablePagination<TData>({
               <SelectValue placeholder={pagination.pageSize} />
             </SelectTrigger>
             <SelectContent side="top">
-              {[50, 100, 200, 500].map((pageSize) => (
```
mrkaye97, line 69: allowing 500 was causing the UI to lag a ton

abelanger5 (changes requested):
This overall looks great! A couple UI things:
1. I actually disagree with removing the headers, I think they're important in providing some nav context because so many of the table views look so similar (also eventually makes sense to attach docs links and descriptions there). They can definitely take up less real estate but I think we should keep them. 
2. The additional metadata popover disappears after a second or two of rendering, which is very frustrating.

mrkaye97:
> This overall looks great! A couple UI things:
> 
> 1. I actually disagree with removing the headers, I think they're important in providing some nav context because so many of the table views look so similar (also eventually makes sense to attach docs links and descriptions there). They can definitely take up less real estate but I think we should keep them.
> 2. The additional metadata popover disappears after a second or two of rendering, which is very frustrating.

Sounds good! Maybe I'll try to add the breadcrumbs or something to help with nav context a bit.

The additional meta disappearing is an existing bug with rerenders being caused by RQ refetches. I can try to fix that here and see how it goes 🤷

abelanger5:
> Sounds good! Maybe I'll try to add the breadcrumbs or something to help with nav context a bit.

I think perhaps you're looking for a Linear-like heading/breadcrumb section? Otherwise most SaaS apps that I like (Posthog, Stripe) have pretty large headings.

mrkaye97:
> > Sounds good! Maybe I'll try to add the breadcrumbs or something to help with nav context a bit.
> 
> I think perhaps you're looking for a Linear-like heading/breadcrumb section? Otherwise most SaaS apps that I like (Posthog, Stripe) have pretty large headings.

heh yeah I always think of Notion, Linear's is similar to them

mrkaye97:
alright I think I got this all working - hacked around some typing issues with the router and put in a janky fix for the metadata thing but it should be good for now, hoping that switching to TanStack at some point and ripping out the v0 code resolves a bunch of these things

abelanger5 (approved):
looks good!

PR #1948, opened 2025-07-03
Fix: Skipping conditions with multiple parents

mrkaye97:
# Description

Fixing a bug with skipping conditions not being respected when multiple parents are specified.

Also extended the Python Context's `wait_for` to allow for or groups, and wrote tests for both of these. 

## Type of change


abelanger5 (approved):
good stuff!

PR #1955, opened 2025-07-05
Add new streaming support to Go SDK

mnafees:
# Description

Add support for the new streaming functionality in the Go SDK.

## Type of change

## What's Changed

- [x] Enrich `HatchetContext` to support new `PutStream()`
- [x] Add ability to assign stream index just like TypeScript
- [x] Add examples for Go SDK new stream


examples/go/streaming/client/main.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	log.Printf("Workflow started with run ID: %s", workflowRun.RunId())
+	log.Println("Subscribing to stream events...")
+
+	// Subscribe to the stream and log events
+	err = hatchet.V0().Subscribe().Stream(context.Background(), workflowRun.RunId(), func(event client.StreamEvent) error {
+		log.Printf("Received stream event: %q", string(event.Message))
+		return nil
+	})
```
mrkaye97: One note here - I added a method on the runs client to subscribe to stream events from a run, which might be useful here too (it just takes the run id and returns the iterator, not sure what the equivalent is in Go)

pkg/worker/context.go
```diff
@@ -137,6 +137,9 @@ type hatchetContext struct {
 	indexMu    sync.Mutex
 	listener   *client.WorkflowRunsListener
 	listenerMu sync.Mutex
+
+	streamEventIndex   int64
+	streamEventIndexMu sync.Mutex
```
mrkaye97, line 144: I didn't do this in Python / TS but maybe I should add a lock around the counter 🤔 - was thinking if you're streaming things back concurrently you wouldn't expect any ordering guarantees, but it probably makes sense to try our best anyways. nice idea 👍

mrkaye97 (approved):
Looks nice! Mind adding these examples to the streaming docs too? we're missing Go examples there

examples/go/streaming/client/main.go
mnafees: That makes sense - I shall add that!

pkg/worker/context.go
mnafees: Yeah, I think in such cases it just made most sense to use a lock to make sure we can ignore any and all race conditions.

mrkaye97: Turns out in TS and Python we don't need this! Since they both use event loop based concurrency, creating the index automatically blocks, so it's sorta accidentally concurrency-safe 😅

mrkaye97 (approved):
Looks great! Thanks for adding the docs 🙌

PR #1965, opened 2025-07-09
Fix: Persist tenant to local storage

mrkaye97:
# Description

Persisting tenant id to local storage (and switching it when we redirect) so that we can correctly load the previous tenant when someone comes back to the root, logs in, etc.

## Type of change

### DIFF
--- a/frontend/app/src/lib/atoms.ts
+++ b/frontend/app/src/lib/atoms.ts
@@ -62,11 +62,11 @@ export function useTenant(): TenantContext {
   const pathParams = useParams();
 
   const setTenantInLocalStorage = useCallback(async (tenantId: string) => {
-    localStorage.setItem('__hatchet.tenantId', tenantId);
+    localStorage.setItem('tenantId', tenantId);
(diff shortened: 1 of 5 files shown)


frontend/app/src/lib/atoms.ts
```diff
@@ -152,6 +154,7 @@ export function useTenant(): TenantContext {
   // NOTE: This is helpful mostly for debugging to easily grab
   // the tenant from the URL.
   useEffect(() => {
+    console.log('Calling useEffect to set tenant from search params');
```
abelanger5: Seems correct, remove debug?

abelanger5: removed it to get this merged in

abelanger5 (approved):

frontend/app/src/lib/atoms.ts
mrkaye97: thanks yeah, we need a linter for this

PR #1966, opened 2025-07-09
Fix popover contents to have word wrap, sanitized input and max width

mnafees:
# Description

The contents of the popover when clicked on `View Error` of a task would not wrap contents properly and would span the entire screen.

## Type of change

## What's Changed

- [x] Make sure to sanitize inputs for popover contents
- [x] Make sure to have max width of the same


frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/events-columns.tsx
```diff
@@ -259,11 +259,25 @@ function ErrorWithHoverCard({ event }: { event: V1TaskEvent }) {
...
         <PopoverContent
-          className="min-w-fit p-0 bg-background border-none z-[80]"
+          className="p-0 bg-background border-none z-[80]"
           align="start"
           container={containerRef.current}
+          style={{
+            width: '600px',
+            maxWidth: '80vw',
+            minWidth: '300px',
+          }}
```
abelanger5: Why are we switching to inline styles instead of tailwind here?

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/step-run-error.tsx
```diff
@@ -1,18 +1,44 @@
...
+});
 
 export default function StepRunError({ text }: { text: string }) {
+  const sanitizedHtml = DOMPurify.sanitize(convert.toHtml(text || ''), {
+    USE_PROFILES: { html: true },
+  });
+
   return (
-    <div className="p-0 w-full min-w-[500px] rounded-md h-[400px] overflow-y-auto">
-      <LoggingComponent
```
abelanger5, line 6: Why do we not fix this in the `LoggingComponent` instead of adding a lot of repeated code to various components?

abelanger5 (changes requested):

abelanger5 (approved):

frontend/app/src/components/cloud/logging/logs.tsx
```diff
@@ -143,6 +143,10 @@ const LoggingComponent: React.FC<LogProps> = ({
       className="w-full mx-auto overflow-y-auto p-6 text-indigo-300 font-mono text-xs rounded-md max-h-[25rem] min-h-[25rem] bg-muted"
       ref={containerRef}
       onScroll={handleScroll}
+      style={{
```
abelanger5: We're again using inline styles instead of tailwind, why?

abelanger5 (changes requested):

frontend/app/src/components/cloud/logging/logs.tsx
mnafees: uh oh that was a mishap and quick test, was supposed to be in a follow up PR

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/events-columns.tsx
```diff
@@ -259,11 +258,13 @@ function ErrorWithHoverCard({ event }: { event: V1TaskEvent }) {
           </Button>
         </PopoverTrigger>
         <PopoverContent
-          className="min-w-fit p-0 bg-background border-none z-[80]"
+          className="p-0 bg-popover border-border shadow-lg z-[80] w-[600px] max-w-[80vw] min-w-[300px]"
```
mrkaye97: do we want fixed widths here? might be preferable to use breakpoints? I haven't run the code locally so not sure how this ends up looking, maybe it's fine

mrkaye97 (approved):
thanks! one small comment but I think it's fine - I usually just test on my laptop screen and on the 13-14" screen size to make sure

mrkaye97 (approved):

abelanger5 (approved):

PR #1976, opened 2025-07-10
Feat: Improve metrics feature client

mrkaye97:
# Description

Improves the metrics feature client to correctly fetch task metrics + queue metrics for V1. Breaking changes here, but I can't imagine anyone is using these since they wouldn't have worked before. 

## Type of change

### DIFF
--- a/sdks/python/hatchet_sdk/utils/serde.py
+++ b/sdks/python/hatchet_sdk/utils/serde.py
@@ -5,23 +5,27 @@ K = TypeVar("K")
 
 
 @overload
-def remove_null_unicode_character(data: str) -> str: ...
+def remove_null_unicode_character(data: str, replacement: str = "") -> str: ...
 
 
 @overload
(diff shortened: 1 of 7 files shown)


sdks/python/hatchet_sdk/utils/serde.py
```diff
@@ -5,35 +5,41 @@
...
 
 @overload
-def remove_null_unicode_character(data: list[T]) -> list[T]: ...
+def remove_null_unicode_character(data: list[T], replacement: str = "") -> list[T]: ...
 
 
 @overload
-def remove_null_unicode_character(data: tuple[T, ...]) -> tuple[T, ...]: ...
+def remove_null_unicode_character(
+    data: tuple[T, ...], replacement: str = ""
```
mrkaye97, line 23: drive-by: adding `replacement` so you can replace with a custom value

abelanger5 (approved):

PR #1978, opened 2025-07-10
Feat: Webhooks

mrkaye97:
# Description

Adds support for webhook-as-event, where we allow you to set up a webhook (via UI or API) and we'll handle incoming webhooks and process them as events within Hatchet as if they were published from a Hatchet SDK directly.

Example request to create a webhook:

```json
{
    "sourceName": "STRIPE",
    "name": "stripe-test-webhook",
    "eventKeyExpression": "'stripe:' + input.type",
    "authType": "HMAC",
    "auth": {
        "algorithm": "SHA256",
        "encoding": "HEX",
        "signatureHeaderName": "Stripe-Signature",
        "signingSecret": "whsec_<snip>"
    }
}
```

To get events sending from Stripe, create a webhook pointing at: `/api/v1/stable/tenants/< id >/webhooks/stripe-test-webhook` 

Misc other things:

1. On the FE, handled the Stripe / Github cases basically as wrappers of the HMAC auth option, where we pre-populate the fields we know from the Stripe / GH docs with their values instead of making the user fill them in
2. Used a built-in Echo rate limiting middleware for handling rate limiting - set to 50 base, 100 burst by default 🤷 

## Type of change

TODO:

- [x] Wire up writes of validation failures into a table in the OLAP DB for monitoring
- [x] Add struct validation on creation
- [x] Documentation
- [x] Populator tests
- [x] sqlc make nullable cols nullable (pgtype.Text) on insert
- [x] order by tenant, inserted at on the list query
- [x] maybe rm index for source name
- [x] set env var for max webhooks per tenant
- [x] set up partitioning like we do for other tables
- [x] implement in-memory rate limiter

### FE Screenshots:

Screenshot 2025-07-18 at 5 08 03 PM

Screenshot 2025-07-18 at 5 08 23 PM
Screenshot 2025-07-18 at 5 08 39 PM


api/v1/server/handlers/v1/webhooks/receive.go
```diff
@@ -0,0 +1,388 @@
...
+}
+
+func removePrefixesFromSignature(signature string) string {
+	signature = strings.TrimPrefix(signature, "sha1=")
+	signature = strings.TrimPrefix(signature, "sha256=")
+	signature = strings.TrimPrefix(signature, "sha512=")
+	signature = strings.TrimPrefix(signature, "md5=")
+
+	return signature
+}
```
mrkaye97: would be great to get a spot check on this!

api/v1/server/handlers/v1/tasks/get.go
```diff
@@ -13,7 +13,13 @@ func (t *TasksService) V1TaskGet(ctx echo.Context, request gen.V1TaskGetRequestO
...
 
 	if taskInterface == nil {
-		return nil, echo.NewHTTPError(404, "Task not found")
+		return gen.V1TaskGet404JSONResponse{
+			Errors: []gen.APIError{
+				{
+					Description: "task not found",
+				},
+			},
+		}, nil
```
mrkaye97, line 22: drive by to fix task errors having wrong statuses

api/v1/server/handlers/v1/webhooks/receive.go
```diff
@@ -0,0 +1,388 @@
...
+				Errors: []gen.APIError{
+					{
+						Description: fmt.Sprintf("failed to unmarshal request body: %v", err),
+					},
+				},
+			}, nil
+		}
+
+		delete(payloadMap, "tenant")
+		delete(payloadMap, "v1-webhook")
```
mrkaye97, line 105: note: if the incoming webhook has these fields, I'm not sure what'll happen here - could result in unexpected behavior

api/v1/server/handlers/v1/webhooks/create.go
```diff
@@ -0,0 +1,184 @@
...
+
+	discriminator := unparsedDiscriminator.(string)
+
+	authType := sqlcv1.V1IncomingWebhookAuthType(discriminator)
+	if authType == "" {
+		return "", fmt.Errorf("invalid auth type: %s", unparsedDiscriminator)
+	}
+
+	return authType, nil
+}
```
mrkaye97, line 86: This is a hack - I couldn't figure out how else to extract the discriminator because the discriminated union doesn't translate well (AFAICT) to Go types. would love some ideas for this one 😂 but it seems to work fine at least, just not type safe

api/v1/server/middleware/webhook_rate_limit.go
```diff
@@ -0,0 +1,43 @@
+package middleware
```
mrkaye97, line 1: Curious for thoughts on this! Seemed reasonable enough to me but not super familiar with the tools here

api/v1/server/middleware/webhook_rate_limit.go
```diff
@@ -0,0 +1,43 @@
...
+			}
+
+			tenantId := c.Param("tenant")
+			webhookName := c.Param("v1-webhook")
+
+			if tenantId == "" || webhookName == "" {
+				return true
+			}
+
+			return c.Request().URL.Path != fmt.Sprintf("/api/v1/stable/tenants/%s/webhooks/%s", tenantId, webhookName)
```
mrkaye97, line 30: This seems pretty brittle, but will add a comment in the OpenAPI spec pointing at this

pkg/config/server/server.go
```diff
@@ -134,6 +134,12 @@ type ConfigFileRuntime struct {
 	// GRPCRateLimit is the rate limit for the grpc server. We count limits separately for the Workflow, Dispatcher and Events services. Workflow and Events service are set to this rate, Dispatcher is 10X this rate. The rate limit is per second, per engine, per api token.
 	GRPCRateLimit float64 `mapstructure:"grpcRateLimit" json:"grpcRateLimit,omitempty" default:"1000"`
 
+	// WebhookRateLimit is the rate limit for webhook endpoints per second, per webhook
+	WebhookRateLimit float64 `mapstructure:"webhookRateLimit" json:"webhookRateLimit,omitempty" default:"50"`
```
mrkaye97, line 138: wasn't sure what reasonable values here were, but 50 / 100 RPS seemed okay to start with

api-contracts/openapi/paths/v1/webhooks/webhook.yaml
```diff
@@ -0,0 +1,264 @@
...
+        description: Forbidden
+      "404":
+        content:
+          application/json:
+            schema:
+              $ref: "../../../components/schemas/_index.yaml#/APIErrors"
+        description: Not found
+    tags:
+      - Webhook
+  post:
```
abelanger5, line 90: Are there any examples of webhooks which aren't a POST request?

pkg/repository/v1/sqlcv1/olap-overwrite.sql.go
```diff
@@ -385,21 +388,23 @@ INSERT INTO v1_events_olap (
...
-	Additionalmetadatas [][]byte             `json:"additionalmetadatas"`
-	Scopes              []*string            `json:"scopes"`
+	Tenantids              []pgtype.UUID        `json:"tenantids"`
+	Externalids            []pgtype.UUID        `json:"externalids"`
+	Seenats                []pgtype.Timestamptz `json:"seenats"`
+	Keys                   []string             `json:"keys"`
+	Payloads               [][]byte             `json:"payloads"`
+	Additionalmetadatas    [][]byte             `json:"additionalmetadatas"`
+	Scopes                 []*string            `json:"scopes"`
+	TriggeringWebhookNames []*string            `json:"triggeringWebhookName"`
```
abelanger5: Must have missed this in a previous PR review, but we typically use `pgtype.Text` to support nullable instead of a `*string`

api/v1/server/handlers/v1/webhooks/create.go
abelanger5: yeah I'm not sure, this seems reasonable - don't see anything helpful in https://github.com/oapi-codegen/oapi-codegen?tab=readme-ov-file#how-does-oapi-codegen-handle-anyof-allof-and-oneof

api/v1/server/middleware/webhook_rate_limit.go
abelanger5: Is there no way to use the hatchetmiddleware which has a signature of `Middleware(r *middleware.RouteInfo) echo.HandlerFunc`? The `RouteInfo` will give us a more stable operation ID (which is still brittle, but perhaps not as brittle)

api-contracts/openapi/paths/v1/webhooks/webhook.yaml
```diff
@@ -0,0 +1,264 @@
...
+    x-resources: ["tenant", "v1-webhook"]
```
abelanger5, line 91: I think it may actually be better to remove `x-resources` from here and do all of the population at the handler level. That would get around all of the weird `maybeTenant` and casting and make it more straightforward.

api/v1/server/handlers/v1/webhooks/receive.go
```diff
@@ -0,0 +1,417 @@
...
+
+			algorithm := webhook.AuthHmacAlgorithm.V1IncomingWebhookHmacAlgorithm
+			encoding := webhook.AuthHmacEncoding.V1IncomingWebhookHmacEncoding
+
+			expectedSignature, err := computeHMACSignature(webhookPayload, decryptedSigningSecret, algorithm, encoding)
+
+			if err != nil {
+				return false, &ValidationError{
+					Code:      Http500,
+					ErrorText: fmt.Sprintf("failed to compute HMAC signature: %s", err),
```
abelanger5, line 416: Do we want to return unauthorized here instead? Just thinking that we might not want to get swarmed with 500s

abelanger5 (approved):
Added some comments but nothing major, overall looks very good 💯

api-contracts/openapi/paths/v1/webhooks/webhook.yaml
mrkaye97: hah, I'm not sure, maybe? do you know of any?

pkg/repository/v1/sqlcv1/olap-overwrite.sql.go
mrkaye97: ah yeah I can fix this, thanks

api/v1/server/handlers/v1/webhooks/receive.go
mrkaye97: good catch 👍

api/v1/server/middleware/webhook_rate_limit.go
mrkaye97: maybe we can? I'll look into that quickly 🙌

api-contracts/openapi/paths/v1/webhooks/webhook.yaml
abelanger5: can't find any examples online, should be fine

PR #1992, opened 2025-07-14
Fix: Remove internal replay batching for now

mrkaye97:
# Description

Removing batching, running replays serially to relieve db pressure

## Type of change

### DIFF
--- a/internal/services/admin/v1/server.go
+++ b/internal/services/admin/v1/server.go
@@ -264,47 +264,12 @@ func (a *AdminServiceImpl) ReplayTasks(ctx context.Context, req *contracts.Repla
 		taskIdInsertedAtRetryCountToExternalId[record] = task.WorkflowRunExternalID
 	}
 
-	workflowRunIdToTasksToReplay := make(map[pgtype.UUID][]v1.TaskIdInsertedAtRetryCount)
-	for _, item := range tasksToReplay {
-		workflowRunIdToTasksToReplay[item.WorkflowRunId] = append(
-			workflowRunIdToTasksToReplay[item.WorkflowRunId],
-			item.TaskIdInsertedAtRetryCount,
-		)
-	}
-
-	var batches [][]v1.TaskIdInsertedAtRetryCount
-	var currentBatch []v1.TaskIdInsertedAtRetryCount
-	batchSize := 100
-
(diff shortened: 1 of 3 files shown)


internal/services/admin/v1/server.go
```diff
@@ -264,47 +264,12 @@ func (a *AdminServiceImpl) ReplayTasks(ctx context.Context, req *contracts.Repla
...
-	}
-
 	replayedIds := make([]string, 0)
 
-	for _, batch := range batches {
+	for _, taskToReplay := range tasksToReplay {
 		// send the payload to the tasks controller, and send the list of tasks back to the client
 		toReplay := tasktypes.ReplayTasksPayload{
-			Tasks: batch,
+			Tasks: []v1.TaskIdInsertedAtRetryCount{taskToReplay.TaskIdInsertedAtRetryCount},
```
abelanger5: With the way this is written, they will likely still be processed in parallel by the consumer of this message, since we have a QOS of 100 by default on RabbitMQ. 

I think we likely need a bigger change in the method which actually calls `replayTasks` to iterate through the batch of messages one by one.

abelanger5 (changes requested):

pkg/repository/v1/task.go
```diff
@@ -2422,13 +2423,25 @@ func makeEventTypeArr(status sqlcv1.V1TaskEventType, n int) []sqlcv1.V1TaskEvent
...
+}
+
 func (r *TaskRepositoryImpl) ReplayTasks(ctx context.Context, tenantId string, tasks []TaskIdInsertedAtRetryCount) (*ReplayTasksResult, error) {
 	tx, commit, rollback, err := sqlchelpers.PrepareTx(ctx, r.pool, r.l, 30000)
 
 	if err != nil {
 		return nil, err
 	}
 
+	err = r.queries.AdvisoryLock(ctx, tx, hash(tenantId))
```
abelanger5: Have we confirmed that we don't have a collision with how we're computing some other advisory lock (like in the concurrency logic)? Might make sense to prefix `replay_` to the tenant ID here or smth

abelanger5 (changes requested):

pkg/repository/v1/task.go
mrkaye97: good call, will fix that

abelanger5 (approved):

PR #1998, opened 2025-07-15
fix: improve baseline wait latency to rely less on polling

abelanger5:
# Description

Improves latency for returning task results in the v1 dispatcher by getting rid of the serialized iterator. 

## Type of change


mnafees (approved):
LGTM

PR #2000, opened 2025-07-15
Fix: Unified task / workflow run getter

mrkaye97:
# Description

Getting rid of uses of `V1TaskGet` (only used one place, results in meh performance on the FE b/c of duplicate queries + confusing SDK behavior). Added a couple things to the `V1WorkflowRunDetails` to make it have parity with the existing task getter, and then updated to use that everywhere.

## Type of change


api/v1/server/oas/transformers/v1/tasks.go
```diff
@@ -50,33 +50,43 @@ func ToTaskSummary(task *sqlcv1.PopulateTaskRunDataRow) gen.V1TaskSummary {
...
+		parentTaskExternalId = &parsedId
+	}
+
 	return gen.V1TaskSummary{
+		ActionId:           &task.ActionID,
+		AdditionalMetadata: &additionalMetadata,
+		Attempt:            &attempt,
+		DisplayName:        task.DisplayName,
+		Duration:           durationPtr,
+		ErrorMessage:       &task.ErrorMessage.String,
```
mrkaye97, line 67: alphabetized these (and the ones below)

mrkaye97:
Bailing on this, just going to make a fix SDK side for now

PR #2001, opened 2025-07-15
Fix: Replay bugs

mrkaye97:
# Description

Attempting to fix a couple replay bugs:

1. Optimistically showing one toast to say the request was submitted
2. Batching by workflow run id
3. Deduplicating ids before returning to the FE

## Type of change

### DIFF
--- a/internal/services/admin/v1/server.go
+++ b/internal/services/admin/v1/server.go
@@ -254,6 +254,7 @@ func (a *AdminServiceImpl) ReplayTasks(ctx context.Context, req *contracts.Repla
 				RetryCount: task.RetryCount,
 			},
 			WorkflowRunExternalId: task.WorkflowRunID,
+			TaskExternalId:        task.ExternalID,
 		}
 
 		if _, exists := existingReplays[record]; exists {
@@ -326,7 +327,7 @@ func (a *AdminServiceImpl) ReplayTasks(ctx context.Context, req *contracts.Repla
 		}
 
 		for _, task := range batch {
-			replayedIds = append(replayedIds, task.WorkflowRunExternalId.String())
+			replayedIds = append(replayedIds, task.TaskExternalId.String())
 		}
 
 		time.Sleep(200 * time.Millisecond)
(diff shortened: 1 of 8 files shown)


internal/services/admin/v1/server.go
```diff
@@ -252,16 +252,20 @@ func (a *AdminServiceImpl) ReplayTasks(ctx context.Context, req *contracts.Repla
...
-		tasksToReplay = append(tasksToReplay, TaskIdInsertedAtRetryCountWithWorkflowRunId{
-			TaskIdInsertedAtRetryCount: record,
-			WorkflowRunId:              task.WorkflowRunExternalID,
-		})
 
-		taskIdInsertedAtRetryCountToExternalId[record] = task.WorkflowRunExternalID
+		if _, exists := taskIdInsertedAtRetryCountToExternalId[record]; !exists {
+			tasksToReplay = append(tasksToReplay, TaskIdInsertedAtRetryCountWithWorkflowRunId{
+				TaskIdInsertedAtRetryCount: record,
+				WorkflowRunId:              task.WorkflowRunExternalID,
```
abelanger5: We already have the `WorkflowRunExternalID` here, so why do we need to modify `TaskIdInsertedAtRetryCount` and add it there?

mrkaye97: 🤦 good call, thanks

internal/services/admin/v1/server.go
```diff
@@ -240,40 +235,46 @@ func (a *AdminServiceImpl) ReplayTasks(ctx context.Context, req *contracts.Repla
...
-			InsertedAt: task.InsertedAt,
-			RetryCount: task.RetryCount,
+		record := tasktypes.TaskIdInsertedAtRetryCountWithExternalId{
+			TaskIdInsertedAtRetryCount: v1.TaskIdInsertedAtRetryCount{
+				Id:         task.ID,
+				InsertedAt: task.InsertedAt,
+				RetryCount: task.RetryCount,
+			},
+			WorkflowRunExternalId: task.WorkflowRunID,
+			TaskExternalId:        task.ExternalID,
```
mrkaye97, line 257: added this to return to the FE

internal/services/admin/v1/server.go
```diff
@@ -240,40 +235,46 @@ func (a *AdminServiceImpl) ReplayTasks(ctx context.Context, req *contracts.Repla
...
+		}
+
+		if _, exists := existingReplays[record]; exists {
+			continue
 		}
```
mrkaye97, line 262: needed to dedupe the result of `FlattenExternalIds`

abelanger5 (approved):

PR #2010, opened 2025-07-16
Feat: CEL Debug Endpoint

mrkaye97:
# Description

Adds a CEL debug endpoint + corresponding SDK clients for running arbitrary CEL expressions

Successful eval (no error):

```python
from hatchet_sdk import Hatchet

hatchet = Hatchet()

response = hatchet.cel.debug(
    expression="""
        input.foo == 'bar' &&
        additional_metadata.key == 'value' &&
        payload.filter_key == 'filter_value'
    """,
    input={"foo": "bar"},
    additional_metadata={"key": "value"},
    filter_payload={"filter_key": "filter_value"},
)


print(response)
```

```python
result=CELSuccess(status='success', output=True)
```

Erroring eval:

```python
response = hatchet.cel.debug(
    expression="x==2",
    input={"foo": "bar"},
)

print(response)
```

```python
result=CELFailure(status='failure', error="failed to compile expression: ERROR: <input>:1:1: undeclared reference to 'x' (in container '')\n | x==2\n | ^")
```

## Type of change


api/v1/server/handlers/v1/cel/debug.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/labstack/echo/v4"
+)
+
+func (c *V1CELService) V1CelDebug(ctx echo.Context, request gen.V1CelDebugRequestObject) (gen.V1CelDebugResponseObject, error) {
+	additionalMetadata := make(map[string]interface{})
+	if request.Body.AdditionalMetadata != nil {
+		additionalMetadata = *request.Body.AdditionalMetadata
+	}
+
+	filterPayload := make(map[string]interface{})
```
mnafees, line 16: If this remains empty should we simply return a message?

mnafees (approved):

api/v1/server/handlers/v1/cel/debug.go
mrkaye97: it's optional, so it's okay for it to be empty (when you create a filter, you don't need to supply a payload, but you can if you want to) - not sure if that answers the question!

abelanger5 (approved):

PR #2012, opened 2025-07-16
Feat: OLAP Table for CEL Eval Failures

mrkaye97:
# Description

Adding a table in the OLAP db for temporarily storing CEL evaluation failures + wiring up writes. We can also add this to the debug endpoint + webhooks once it merges

Will need to update the webhooks + debug endpoint PRs once this goes in

## Type of change


api/v1/server/handlers/v1/cel/debug.go
```diff
@@ -25,6 +31,19 @@ func (c *V1CELService) V1CelDebug(ctx echo.Context, request gen.V1CelDebugReques
 	),
 	)
 
+	if err != nil {
```
abelanger5: Isn't the point of the debug endpoint to return a readable error to the user here? Why would we ingest this into the table?

abelanger5 (changes requested):

api/v1/server/handlers/v1/cel/debug.go
mrkaye97: 🤦 yeah I guess no need, will remove this one

abelanger5 (approved):

PR #2019, opened 2025-07-17
Fix: Multiple tenants for task & dag status updates

mrkaye97:
# Description

Update statuses for many tenants in one shot instead of one at a time

## Type of change

### DIFF
--- a/internal/services/controllers/v1/olap/process_dag_status_updates.go
+++ b/internal/services/controllers/v1/olap/process_dag_status_updates.go
@@ -2,66 +2,63 @@ package olap
 
 import (
 	"context"
-	"strings"
 
 	msgqueue "github.com/hatchet-dev/hatchet/internal/msgqueue/v1"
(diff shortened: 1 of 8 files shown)


internal/services/controllers/v1/olap/process_dag_status_updates.go
```diff
@@ -68,50 +83,54 @@ func (o *OLAPControllerImpl) updateDAGStatuses(ctx context.Context, tenantId str
 	}
 
 	if o.prometheusMetricsEnabled {
-		workflowNames, err := o.repo.Workflows().ListWorkflowNamesByIds(ctx, tenantId, workflowIds)
-		if err != nil {
-			return false, err
-		}
+		for _, tenantId := range tenantIds {
```
mrkaye97: This isn't great, but I figured it's about equivalent to what we have now processing tenants one at a time anyways

abelanger5 (approved):

PR #2020, opened 2025-07-17
Fix: Improve `UpdateDAGStatuses` and `UpdateTaskStatuses`

mrkaye97:
# Description

Some small optimizations for the `UpdateDAGStatuses` query

1. Using PKs for more joins, orders, etc.
2. Wrote a "setup" query to do partition pruning by first selected the min inserted at to consider, and then using that to prune partitions so we need to scan fewer of them

## Type of change

### DIFF
--- a/pkg/repository/v1/sqlcv1/olap.sql
+++ b/pkg/repository/v1/sqlcv1/olap.sql
@@ -807,6 +807,27 @@ SELECT
 FROM
     updated_tasks t;
 
+
+-- name: FindMinInsertedAtForDAGStatusUpdates :one
+WITH tenants AS (
+    SELECT UNNEST(
+        find_matching_tenants_in_task_status_updates_tmp_partition(
+            @partitionNumber::int,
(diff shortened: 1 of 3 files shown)


pkg/repository/v1/sqlcv1/olap.sql
```diff
@@ -711,8 +731,9 @@ WITH tenants AS (
     JOIN
         updatable_events e ON
```
abelanger5, line 732: Do we want to change this to a `WHERE...IN` as well?

abelanger5 (approved):

pkg/repository/v1/sqlcv1/olap.sql
mrkaye97: I was thinking we couldn't because we use some columns from here?

abelanger5: yep, you're right

PR #2026, opened 2025-07-19
create typed input validator access on base workflow

szvsw:
# Description

Since `WorkflowConfig.input_validator` has type `type[BaseModel]` and is not generic, currently you can't recover the input validator class from a workflow/standalone instance in a way that the type checker correctly infers the type.  This allows you to safely access it.

If there are tests/docs to add, just point me to where they should be added!

## Type of change

## What's Changed

- [x] updates `runnables.workflow.BaseWorkflow`


szvsw:
@mrkaye97 bumped the version and updated changelog, lmk if it looks okay.

mrkaye97 (approved):
Thanks, looks great!

PR #2027, opened 2025-07-21
fix: only reacquire lease if expiring

grutt:
# Description

Fixes Excessive UPDATE operations on the Lease table causing high database load.
Every active worker/queue/concurrency strategy gets sent to AcquireOrExtendLeases() every second, triggering unnecessary lease updates.

The Scale
Frequency: Updates every 1 second
Scope: All active resources (workers + queues + concurrency strategies)
SQL Impact: ON CONFLICT ... DO UPDATE fires for every resource every second
Example: 10 workers + 5 queues = 15 updates/second = 1.3M updates/day

## Type of change

## What's Changed

- [x] only write the lease update if close to expiration


abelanger5:
closing in favor of #2034

PR #2032, opened 2025-07-21
Use `gofumpt` for Go formatting

mnafees:
# Description

We do not have a standard formatting process in place for Go code.

## Type of change

## What's Changed

- [x] Add support for gofumpt via `task pre-commit-run`


mrkaye97 (approved):
lgtm! would also prefer a line break before checking errors though, if we can configure that :P

PR #2035, opened 2025-07-21
[hotfix] Fix duration calculation of DAGs and single tasks

mnafees:
# Description

We seemed to have introduced some regression in PR #2019 due to which the duration calculation was wrong.

## Type of change

### DIFF
--- a/internal/services/controllers/v1/olap/process_dag_status_updates.go
+++ b/internal/services/controllers/v1/olap/process_dag_status_updates.go
@@ -59,9 +59,6 @@ func (o *OLAPControllerImpl) runDAGStatusUpdates(ctx context.Context) func() {
 func (o *OLAPControllerImpl) notifyDAGsUpdated(ctx context.Context, rows []v1.UpdateDAGStatusRow) error {
 	tenantIdToPayloads := make(map[pgtype.UUID][]tasktypes.NotifyFinalizedPayload)
 	tenantIdToWorkflowIds := make(map[string][]pgtype.UUID)
-	dagIds := make([]int64, 0, len(rows))
-	dagInsertedAts := make([]pgtype.Timestamptz, 0, len(rows))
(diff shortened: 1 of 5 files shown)


internal/services/controllers/v1/olap/process_dag_status_updates.go
```diff
@@ -74,38 +71,62 @@ func (o *OLAPControllerImpl) notifyDAGsUpdated(ctx context.Context, rows []v1.Up
...
 					return err
 				}
 
-				for i, row := range rows {
+				dagDurations := make(map[int64]*sqlcv1.GetDagDurationsByDagIdsRow)
+				for i, duration := range dagDurationsArray {
+					if i < len(tenantDagIds) {
+						dagDurations[tenantDagIds[i]] = duration
+					}
+				}
```
mrkaye97: should we return the map out of the OLAP repo method instead? (this is what I implemented in my other PR if you want to copy + paste)

internal/services/controllers/v1/olap/process_dag_status_updates.go
```diff
@@ -74,38 +71,62 @@ func (o *OLAPControllerImpl) notifyDAGsUpdated(ctx context.Context, rows []v1.Up
...
+
+				for _, row := range tenantRows {
 					if row.ReadableStatus == sqlcv1.V1ReadableStatusOlapCOMPLETED || row.ReadableStatus == sqlcv1.V1ReadableStatusOlapFAILED || row.ReadableStatus == sqlcv1.V1ReadableStatusOlapCANCELLED {
 						workflowName := workflowNames[row.WorkflowId]
 						if workflowName == "" {
 							continue
 						}
 
-						dagDuration := dagDurations[i]
+						dagDuration := dagDurations[row.DagId]
```
mrkaye97: I think we need to use the `row.ExternalID` here (also see that other PR)

abelanger5 (approved):

PR #2036, opened 2025-07-21
Fix: filter lookup not retaining scope information

mrkaye97:
# Description

Fixing an issue where multiple messages are read off the queue with events that reference the same workflow, but many different scopes. We'd create a single mapping with a workflow id to a list of filters, but this was incorrect because it'd intermingle the scopes from the different events. Instead, we need to key on both the workflow id _and_ the scope.

## Type of change

### DIFF
--- a/pkg/repository/v1/sqlcv1/filters.sql
+++ b/pkg/repository/v1/sqlcv1/filters.sql
@@ -91,7 +91,7 @@ GROUP BY workflow_id
 
 -- name: ListFiltersForEventTriggers :many
 WITH inputs AS (
-    SELECT
+    SELECT DISTINCT
         UNNEST(sqlc.narg(workflowIds)::UUID[]) AS workflow_id,
         UNNEST(sqlc.narg(scopes)::TEXT[]) AS scope
 )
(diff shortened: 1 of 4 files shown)


pkg/repository/v1/trigger.go
```diff
@@ -241,16 +246,12 @@ func (r *TriggerRepositoryImpl) TriggerFromEvents(ctx context.Context, tenantId
 		return nil, fmt.Errorf("failed to list workflows for events: %w", err)
 	}
 
-	workflowEventKeyToIncomingEventKey := make(map[string]string)
```
mrkaye97, line 244: drive by - removing this, it's not needed

pkg/repository/v1/sqlcv1/filters.sql
```diff
@@ -91,7 +91,7 @@ GROUP BY workflow_id
 
 -- name: ListFiltersForEventTriggers :many
 WITH inputs AS (
-    SELECT
+    SELECT DISTINCT
```
mrkaye97, line 94: deduping here to handle the case where there are duplicate id, scope pairs which would cause the join in the main query to blow up

abelanger5 (approved):

PR #2037, opened 2025-07-22
Add cron related metadata to cron workflows

mnafees:
# Description

We want a way to be able to identify when a workflow was triggered as part of a cron job with additional metadata

## Type of change

## What's Changed

- [x] Additional metadata in cron workflows v1 including cron expression, name, and parent ID

image


internal/services/ticker/cron_v1.go
```diff
@@ -14,15 +15,25 @@ import (
...
+
+	metadata := map[string]any{
+		"hatchet_cron_expression": cron,
+		"hatchet_cron_parent_id":  cronParentId,
+	}
 
-	if additionalMetadata != nil {
-		additionalMetaBytes, err = json.Marshal(additionalMetadata)
+	if cronName != nil {
+		metadata["hatchet_cron_name"] = *cronName
```
mrkaye97: nit: for events, we use a double underscore (e.g. `hatchet__event_id`) - would be good to be consistent with that here!

internal/services/ticker/cron_v1.go
```diff
@@ -14,15 +15,25 @@ import (
...
 func (t *TickerImpl) runCronWorkflowV1(ctx context.Context, tenantId string, workflowVersion *dbsqlc.GetWorkflowVersionForEngineRow, cron, cronParentId string, cronName *string, input []byte, additionalMetadata map[string]interface{}, priority *int32) error {
-	var additionalMetaBytes []byte
-	var err error
+	if additionalMetadata == nil {
+		additionalMetadata = make(map[string]interface{})
+	}
+
+	metadata := map[string]any{
+		"hatchet_cron_expression": cron,
+		"hatchet_cron_parent_id":  cronParentId,
```
mrkaye97: same here

mrkaye97: also I think the `parentId` might not be meaningful to the user - can you try cross referencing this with the `id` that you see on `metadata.id` in the network tab if you load the crons page to see if they're the same? I think this `parentId` might refer to `WorkflowTriggers.id` and not to the cron itself (might need to pass the cron's id through to here)

mrkaye97 (approved):
looks great!

PR #2040, opened 2025-07-23
[Python]: Fixing logging bugs, fixing duped sleep key bug

mrkaye97:
# Description

### Added

- Adds support for filters and formatters in the logger that's passed to the Hatchet client.
- Adds a flag to disable log capture.

### Changed

- Fixes a bug in `aio_sleep_for` and the `SleepCondition` that did not allow duplicate sleeps to be awaited correctly.

## Type of change


mrkaye97:
Note to self: the sleep key might break replays

grutt (approved):

PR #2044, opened 2025-07-23
Allow member roles to be changed by owners and admins

mnafees:
# Description

We should allow member roles to be changed from the dashboard.

## Type of change


grutt (approved):

api/v1/server/handlers/tenants/update_member.go
```diff
@@ -0,0 +1,74 @@
...
+				apierrors.NewAPIErrors("only an owner can change a role to owner"),
+			), nil
+		}
+
+		// Cannot change role of an owner
+		if memberToUpdate.Role == dbsqlc.TenantMemberRoleOWNER {
+			return gen.TenantMemberUpdate400JSONResponse(
+				apierrors.NewAPIErrors("only an owner can change the role of another owner"),
+			), nil
+		}
```
mrkaye97, line 48: what's the difference between these two?

api/v1/server/handlers/tenants/update_member.go
```diff
@@ -0,0 +1,74 @@
...
+			apierrors.NewAPIErrors("you cannot change your own role"),
+		), nil
+	}
+
+	// construct the database query
+	updateOpts := &repository.UpdateTenantMemberOpts{
+		Role: repository.StringPtr(string(request.Body.Role)),
+	}
+
+	// update the member
```
mrkaye97: something of a stylistic nit, but I generally don't like comments like these (and similar above - "construct the database query"). IMO it should be clear from the code what these are doing (and it is)

frontend/app/src/next/hooks/use-members.tsx
```diff
@@ -92,11 +99,39 @@ export function MembersProvider({ children }: { children: React.ReactNode }) {
...
+          title: 'Error updating member role',
+
+          variant: 'destructive',
+          error,
+        });
+        throw error;
+      }
+    },
+    onSuccess: async () => {
+      await membersQuery.refetch();
```
mrkaye97: I know we've done this in a bunch of places so it's fine here too, but we're doing to need to change our use of `refetch` to `queryClient.invalidateQueries` in many places as some point, so it might be preferable to do that here now so it's less work for the future 😅

frontend/app/src/pages/main/tenant-settings/members/components/update-member-form.tsx
```diff
@@ -0,0 +1,124 @@
...
+  DialogHeader,
+  DialogTitle,
+} from '@/components/ui/dialog';
+import { TenantMember, TenantMemberRole } from '@/lib/api';
+
+const schema = z.object({
+  role: z.enum([
+    TenantMemberRole.OWNER,
+    TenantMemberRole.ADMIN,
+    TenantMemberRole.MEMBER,
```
mrkaye97: nit: we might be able to use `nativeEnum` with the generated type here?

mrkaye97: (not 100% sure)

frontend/app/src/pages/main/tenant-settings/members/components/update-member-form.tsx
```diff
@@ -0,0 +1,124 @@
+import { cn } from '@/lib/utils';
```
mrkaye97, line 1: ah - we actually probably don't need to be updating V0 pages, just FYI (sorry, this is confusing and keeps biting me too 😬 )

we're planning to remove these soon + I haven't been making any updates to them since people are migrating off. Not a big deal and fine to leave this, just FYI

mrkaye97 (approved):
nice work 🙌

api/v1/server/handlers/tenants/update_member.go
mnafees: This entire check only checks when the person executing the request is not an OWNER. In the first case, we are simply checking that they are not trying to make someone as OWNER. In the second case, we are checking that they are unable to change an existing OWNER's role to something else.

frontend/app/src/next/components/members/members-table.tsx
```diff
@@ -24,6 +24,13 @@ import { RemoveMemberForm } from '@/next/pages/authenticated/dashboard/settings/
 import { RevokeInviteForm } from '@/next/pages/authenticated/dashboard/settings/team/components/revoke-invite-form';
```
mrkaye97: agh crap, I missed this before, sorry 😬 

this is yet another thing we don't want to update - these `/next` files are for the new FE that we decided to scrap. we should just remove them, but we definitely don't want to make anything currently in production rely on them since they'll be harder to remove then

frontend/app/src/next/hooks/use-members.tsx
```diff
@@ -87,8 +96,35 @@ export function MembersProvider({ children }: { children: React.ReactNode }) {
       }
     },
     onSuccess: async () => {
-      await membersQuery.refetch();
-      await invitesQuery.refetch();
+      await queryClient.invalidateQueries({ queryKey: ['tenant-members:list', tenantId] });
```
mrkaye97: same here re: `next` - I was wondering how this worked and then I noticed the hook it was using

frontend/app/src/next/lib/can/features/members.permissions.ts
```diff
@@ -81,4 +81,52 @@ export const members: PermissionSet = {
         message: 'You do not have permission to invite members',
       };
     },
+  updateRole:
```
mrkaye97: also not sure if we have `can` in the non-`next` part of the project

abelanger5 (approved):

PR #2047, opened 2025-07-23
Feat: Payload Store Repository

mrkaye97:
# Description

Initial setup of an external payload store, which allows the engine to offload payloads to an external store after a provided amount of time. The external store just needs to implement to the interface provided, so it could be e.g. S3, or also something like Dynamo, a separate PG DB, etc.

Only setting up payload writes for the `v1_task` table for now, but soon we'll want to do `v1_task_event`, `v1_dag`, etc.

## Type of change


cmd/hatchet-migrate/migrate/migrations/20250722162812_v1_0_30.sql
```diff
@@ -0,0 +1,24 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TYPE v1_payload_type AS ENUM ('WORKFLOW_INPUT', 'TASK_OUTPUT');
+
+CREATE TABLE v1_payload (
+    tenant_id UUID NOT NULL,
+    task_id BIGINT NOT NULL,
+    task_inserted_at TIMESTAMPTZ NOT NULL,
```
mrkaye97: unsure if this is a good idea or not - I was trying to think of cases where we'd want to tie payloads to non-task things but was struggling to think of any

pkg/repository/v1/repository.go
```diff
@@ -19,47 +19,45 @@ type Repository interface {
...
+		tasks:        newTaskRepository(shared, taskRetentionPeriod, maxInternalRetryCount),
+		scheduler:    newSchedulerRepository(shared),
+		matches:      newMatchRepository(shared),
+		olap:         newOLAPRepository(shared, olapRetentionPeriod, true),
+		logs:         newLogLineRepository(shared),
+		workers:      newWorkerRepository(shared),
+		workflows:    newWorkflowRepository(shared),
+		ticker:       newTickerRepository(shared),
+		filters:      newFilterRepository(shared),
+		payloadStore: shared.payloadStore,
```
mrkaye97, line 72: I was pretty unhappy with this. The reason for it is we need to be able to access the payload store inside of the match repo, the tasks repo, the olap repo, etc., but we also need it on the dispatcher. Would love better ideas here, since I think this also will break the ability to override but I'm not sure about that

pkg/repository/v1/payloadstore.go
```diff
@@ -0,0 +1,144 @@
...
+
+type StorePayloadOpts struct {
+	TaskId         int64
+	TaskInsertedAt pgtype.Timestamptz
+	Type           sqlcv1.V1PayloadType
+	Payload        []byte
+}
+
+func (p *payloadStoreRepositoryImpl) Store(ctx context.Context, tenantId string, payloads []StorePayloadOpts) error {
+	tx, commit, rollback, err := sqlchelpers.PrepareTx(ctx, p.pool, p.l, 5000)
```
mrkaye97: basic idea here is we can change this to have any backend (e.g. S3 with a cache, etc.) so that we can plug whatever we need in. For S3, we might use e.g. `taskId/taskInsertedAt/type.json` as the key or something like that

pkg/repository/v1/payloadstore.go
```diff
@@ -0,0 +1,450 @@
...
+
+	if err != nil {
+		return false, err
+	}
+
+	if err := commit(ctx); err != nil {
+		return false, err
+	}
+
+	return true, nil
```
abelanger5: We're always returning `true` here which seems incorrect: this is signaling the operation to continue looping so this is effectively creating an infinite loop

pkg/repository/v1/payloadstore.go
```diff
@@ -0,0 +1,450 @@
...
+	for opts, payload := range payloads {
+		externalStoreOpts = append(externalStoreOpts, StorePayloadOpts{
+			Id:         opts.Id,
+			InsertedAt: opts.InsertedAt,
+			Type:       opts.Type,
+			Payload:    payload,
+		})
+	}
+
+	err = p.offloadToExternal(ctx, tx, tenantId, externalStoreOpts...)
```
abelanger5: We don't want external writes in the middle of a transaction - this risks the transaction living for a very long time which will cause a number of issues. 

I think a better strategy is to use two transactions with a lease. First transaction acquires the lease, then we flush, second transaction releases the lease and deletes the rows. Depending on the number of external writers we'd like to support we could either use a single row on a separate lease table, or lease each row individually.

pkg/repository/v1/payloadstore.go
```diff
@@ -0,0 +1,450 @@
...
+}
+
+type ExternalStore interface {
+	Store(ctx context.Context, tenantId string, payloads ...StorePayloadOpts) (map[RetrievePayloadOpts]ExternalPayloadLocationKey, error)
+	BulkRetrieve(ctx context.Context, tenantId string, keys ...ExternalPayloadLocationKey) (map[ExternalPayloadLocationKey][]byte, error)
+}
+
+type PayloadStoreOption func(*payloadStoreRepositoryImpl)
+
+func WithExternalStore(store ExternalStore, externalStoreLocationName string, nativeStoreTTL time.Duration) PayloadStoreOption {
```
abelanger5: Instead of `WithExternalStore` here, we could add `OverwriteExternalStore` as a method to the `PayloadStoreRepository`, which allows us to overwrite the external store after we've initialized the repository which is what we want.

abelanger5 (commented):
Added some comments

sql/schema/v1-core.sql
```diff
@@ -1628,3 +1628,38 @@ CREATE TABLE v1_durable_sleep (
...
+    offload_at TIMESTAMPTZ NOT NULL,
+    payload_id BIGINT NOT NULL,
+    payload_inserted_at TIMESTAMPTZ NOT NULL,
+    payload_type v1_payload_type NOT NULL,
+    operation v1_payload_wal_operation NOT NULL,
+
+    offload_process_lease_id UUID,
+    offload_process_lease_expires_at TIMESTAMPTZ,
+
+    PRIMARY KEY (offload_at, payload_id, payload_inserted_at, payload_type, tenant_id),
```
abelanger5: Shouldn't tenant_id be second in the index? Otherwise it won't be used when querying the payload WAL

pkg/repository/v1/sqlcv1/payload-store.sql
```diff
@@ -0,0 +1,138 @@
...
+
+-- name: PollPayloadWALForRecordsToOffload :many
+WITH to_update AS (
+    SELECT *
+    FROM v1_payload_wal
+    WHERE
+        tenant_id = @tenantId::UUID
+        AND offload_at < NOW()
+        AND offload_process_lease_id IS NULL OR offload_process_lease_expires_at < NOW()
+    ORDER BY offload_at, payload_id, payload_inserted_at, payload_type
```
abelanger5: Shouldn't we order by all pkeys? (especially if changing the order)

abelanger5 (changes requested):

internal/queueutils/pool.go
```diff
@@ -51,6 +52,22 @@ func (p *OperationPool) SetTenants(tenants []*dbsqlc.Tenant) {
...
+	}
+
+	p.ops.Range(func(key, value interface{}) bool {
+		if _, ok := partitionMap[key.(string)]; !ok {
+			p.ops.Delete(key)
+		}
+
+		return true
+	})
+}
```
mrkaye97: not sure what the right way to do this is, so added this for now as a hack

internal/services/controllers/v1/task/process_payload_wal.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+func (tc *TasksControllerImpl) processPayloadWAL(ctx context.Context, partitionNumberString string) (bool, error) {
+	parsedPartitionNumber, err := strconv.Atoi(partitionNumberString)
+
+	if err != nil {
+		return false, err
+	}
+
+	return tc.repov1.Payloads().ProcessPayloadWAL(ctx, int32(parsedPartitionNumber))
```
mrkaye97: same here - wasn't sure how to pass the partitions through properly so did this as a hack, can change it to whatever makes more sense

internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -258,14 +288,40 @@ func (d *DispatcherImpl) handleTaskBulkAssignedTask(ctx context.Context, msg *ms
 
 				currInput.Parents = readableIdToData
 
-				task.Input = currInput.Bytes()
+				// Didn't need this before, which I think is an indication of a bug somewhere
+				if currInput.Triggers == nil {
+					currInput.Triggers = make(map[string]map[string]interface{})
+				}
```
mrkaye97: need to circle back to this (mostly a note to self)

pkg/repository/v1/payloadstore.go
```diff
@@ -0,0 +1,482 @@
...
+	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/rs/zerolog"
+)
+
+type StorePayloadOpts struct {
+	Id         int64
+	InsertedAt pgtype.Timestamptz
+	Type       sqlcv1.V1PayloadType
+	Payload    []byte
```
mrkaye97, line 20: I was thinking of making this `PayloadContent` instead of a byte array, but it makes it harder to implement this interface (since most of `PayloadContent` is sort of internal to this file), so I wasn't sure about that. but it's too easy to write a random byte array to the db right now IMO

pkg/repository/v1/payloadstore.go
```diff
@@ -0,0 +1,482 @@
...
+		Types:       payloadTypes,
+		Payloads:    payloadData,
+	})
+
+	if err != nil {
+		return fmt.Errorf("failed to write payloads: %w", err)
+	}
+
+	// only need to write to the WAL if we have an external store configured
+	if p.externalStoreEnabled {
```
mrkaye97, line 127: I'm a bit nervous about this short circuiting - feels like it'll make this harder to maintain in the future, but I also didn't want to have a ton of WAL overhead if it's just a no-op

pkg/repository/v1/payloadstore.go
```diff
@@ -0,0 +1,482 @@
...
+
+	return p.externalStore.Store(ctx, payloads...)
+}
+
+func (p *payloadStoreRepositoryImpl) ProcessPayloadWAL(ctx context.Context, partitionNumber int32) (bool, error) {
+	// no need to process the WAL if external store is not enabled
+	if !p.externalStoreEnabled {
+		return false, nil
+	}
+
```
mrkaye97, line 252: more short circuiting here, same concerns

pkg/repository/v1/payloadstore.go
```diff
@@ -0,0 +1,482 @@
...
+			Location:            PayloadLocationExternal,
+			ExternalLocationKey: &key,
+		}
+
+		marshalledContent, err := payloadContent.Marshal()
+		if err != nil {
+			return false, fmt.Errorf("failed to marshal payload content: %w", err)
+		}
+
+		values = append(values, marshalledContent)
```
mrkaye97: this is a footgun - since we can write any byte array to the db, it's totally valid Go to just pass the payload itself instead of creating a new `PayloadContent` and then marshalling it

cmd/hatchet-migrate/migrate/migrations/20250827162812_v1_0_41.sql
```diff
@@ -0,0 +1,157 @@
...
+    SELECT to_char(targetDate, 'YYYYMMDD') INTO targetDateStr;
+    SELECT to_char(targetDate + INTERVAL '1 day', 'YYYYMMDD') INTO targetDatePlusOneDayStr;
+    SELECT lower(format('%s_%s', targetTableName, targetDateStr)) INTO newTableName;
+    -- exit if the table exists
+    IF EXISTS (SELECT 1 FROM pg_tables WHERE tablename = newTableName) THEN
+        RETURN 0;
+    END IF;
+
+    EXECUTE
+        format('CREATE TABLE %s (LIKE %s INCLUDING INDEXES INCLUDING CONSTRAINTS)', newTableName, targetTableName);
```
mrkaye97: the only change here is adding `INCLUDING CONSTRAINTS`

pkg/repository/v1/sqlcv1/payload-store.sql
```diff
@@ -0,0 +1,153 @@
...
+    inline_content
+)
+
+SELECT
+    i.tenant_id,
+    i.id,
+    i.inserted_at,
+    i.type,
+    i.location,
+    CASE WHEN i.external_location_key = '' OR i.location = 'EXTERNAL' THEN NULL ELSE i.external_location_key END,
```
mrkaye97, line 55: any thoughts on this? I kinda prefer it to needing the overwrite, since the key will never be empty

cmd/hatchet-migrate/migrate/migrations/20250827162812_v1_0_41.sql
```diff
@@ -0,0 +1,157 @@
...
+CREATE TABLE v1_payload_wal (
+    tenant_id UUID NOT NULL,
+    offload_at TIMESTAMPTZ NOT NULL,
+    payload_id BIGINT NOT NULL,
+    payload_inserted_at TIMESTAMPTZ NOT NULL,
+    payload_type v1_payload_type NOT NULL,
+    operation v1_payload_wal_operation NOT NULL,
+
+    offload_process_lease_id UUID,
+    offload_process_lease_expires_at TIMESTAMPTZ,
```
abelanger5: Do we need a lease on each row? would a lease which represents each partition of the WAL not be sufficient?

cmd/hatchet-migrate/migrate/migrations/20250827162812_v1_0_41.sql
```diff
@@ -0,0 +1,157 @@
...
+
+    PRIMARY KEY (offload_at, tenant_id, payload_id, payload_inserted_at, payload_type),
+    CONSTRAINT "v1_payload_wal_payload" FOREIGN KEY (payload_id, payload_inserted_at, payload_type, tenant_id) REFERENCES v1_payload (id, inserted_at, type, tenant_id) ON DELETE CASCADE
+) PARTITION BY HASH (tenant_id);
+
+SELECT create_v1_hash_partitions('v1_payload_wal'::TEXT, 4);
+
+CREATE INDEX idx_payload_wal_lease_expiry ON v1_payload_wal (tenant_id, offload_process_lease_id) WHERE offload_process_lease_id IS NOT NULL;
+CREATE INDEX idx_payload_wal_lease_acquisition ON v1_payload_wal (tenant_id, offload_process_lease_id) WHERE offload_process_lease_id IS NULL;
```
abelanger5: Aren't these complements of each other? why don't we just have one partition (assuming we want to keep the row-level leasing)

internal/queueutils/pool.go
abelanger5: what is the point of this?

abelanger5: isn't this just going to delete all operations which aren't using partitioning? this is usually a shared operation pool for each controller IIRC

internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -226,12 +226,42 @@ func (d *DispatcherImpl) handleTaskBulkAssignedTask(ctx context.Context, msg *ms
 			continue
 		}
 
+		retrivePayloadOpts := make([]v1.RetrievePayloadOpts, len(bulkDatas))
```
abelanger5:
```suggestion
		retrievePayloadOpts := make([]v1.RetrievePayloadOpts, len(bulkDatas))
```

internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -226,12 +226,42 @@ func (d *DispatcherImpl) handleTaskBulkAssignedTask(ctx context.Context, msg *ms
...
+				InsertedAt: task.InsertedAt,
+				Type:       sqlcv1.V1PayloadTypeTASKINPUT,
+				TenantId:   task.TenantID,
+			}]
+
+			if input == nil || !ok {
+				// If the input wasn't found in the payload store,
+				// fall back to the input stored on the task itself.
+				input = task.Input
+			}
```
abelanger5, line 261: doesn't the payload store repository use this fallback?

internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -226,12 +226,42 @@ func (d *DispatcherImpl) handleTaskBulkAssignedTask(ctx context.Context, msg *ms
...
+				InsertedAt: task.InsertedAt,
+				Type:       sqlcv1.V1PayloadTypeTASKINPUT,
+				TenantId:   task.TenantID,
+			}
+		}
+
+		inputs, err := d.repov1.Payloads().BulkRetrieve(ctx, retrivePayloadOpts...)
+
+		if err != nil {
+			d.l.Error().Err(err).Msgf("could not bulk retrieve inputs for %d tasks", len(bulkDatas))
```
abelanger5, line 243: If we can't retrieve inputs, probably need to requeue

cmd/hatchet-migrate/migrate/migrations/20250827162812_v1_0_41.sql
mrkaye97: yeah this (same as the comment below) was a bit of a relic of before partitioning. I don't think we can just lease the whole WAL partition though, because what if there are way more rows in it than we want to process at once? maybe I'm misunderstanding something here, but my thought was we still need a row-level lease assuming we don't just select everything out of the partition every time

internal/queueutils/pool.go
mrkaye97: well so for our other jobs, we have this same thing but with tenants (`SetTenants`), and I wasn't sure how to spawn N jobs, one per partition, where we _aren't_ doing it by tenant, so I added this as a hack. Not sure what the correct way to do this is, though, or if we even need it at all. But yeah basically was just looking for a way to pass the partition id into `runProcessPayloadWAL`

internal/services/dispatcher/dispatcher_v1.go
mrkaye97: thanks

cmd/hatchet-migrate/migrate/migrations/20250827162812_v1_0_41.sql
abelanger5: Correct me if I'm wrong, but we only want one reader/writer of the WAL partition at a time, which can be represented with an advisory lock or a row lock on a different table like `wal_partition_leases`, instead of a row-level lease.

internal/queueutils/pool.go
abelanger5: instead of using the operation pool, perhaps just create an operation per partition?

cmd/hatchet-migrate/migrate/migrations/20250827162812_v1_0_41.sql
mrkaye97: maybe I'm getting confused at the use of "partition" - I'm thinking of a partition as a partition of the WAL in the db, where we could read from each partitioned table separately and not have lock contention if we had N processes processing the WAL concurrently (one per PG partition). It sounds like maybe you're referring to a different partition?

internal/services/dispatcher/dispatcher_v1.go
mrkaye97: not sure I follow the question here - do you mean this is redundant?

cmd/hatchet-migrate/migrate/migrations/20250827162812_v1_0_41.sql
abelanger5: No, I agree with that - we're trying to have only N processes leasing each partition at a time, one per partition. So instead of having a row-level lease, we could have a lease which represents each partition. Like `lease_0` can represent a lease for reading/writing partition 0, `lease_1` can represent reading/writing partition 1, etc. That way we don't need to lease every row individually

internal/services/dispatcher/dispatcher_v1.go
abelanger5: yes, isn't calling `BulkRetrieve` on the payload store safe to do without needing to fall back each time?

abelanger5: since it also checks the `v1_task` table directly?

cmd/hatchet-migrate/migrate/migrations/20250827162812_v1_0_41.sql
mrkaye97: Ahh, okay, I think I see what you mean, I'll make some edits here and report back

mrkaye97: alright pushed a change there - removed the lease related columns and added an advisory lock based on the partition id, which I think is what you're after?

internal/services/dispatcher/dispatcher_v1.go
mrkaye97: I don't think so? this is for handling the case where e.g. a DAG started before the upgrade and then the second task runs afterwards, in which case I think there might be no data in the payloads table yet. let me know if you're thinking of something different here!

internal/msgqueue/v1/postgres/msgqueue.go
```diff
@@ -252,7 +252,7 @@ func (p *PostgresMessageQueue) Subscribe(queue msgqueue.Queue, preAck msgqueue.A
 		return eg.Wait()
 	}
 
-	op := queueutils.NewOperationPool(p.l, 60*time.Second, "postgresmq", queueutils.OpMethod(func(ctx context.Context, id string) (bool, error) {
+	op := queueutils.NewOperationPool(p.l, 60*time.Second, "postgresmq", queueutils.OpMethod[string](func(ctx context.Context, id string) (bool, error) {
```
abelanger5, line 255: I'm surprised the `[string]` is necessary? Should be inferred from the method

abelanger5 (approved):

internal/msgqueue/v1/postgres/msgqueue.go
mrkaye97: I thought that too, not sure if it's a rough edge or something 🤷

PR #2057, opened 2025-07-27
[Python]: Flag to enable / disable gRPC fork support, modify valid task return type

mrkaye97:
# Description

### Added

- Adds a new config option `grpc_enable_fork_support` to allow users to enable or disable gRPC fork support. This is useful for environments where gRPC fork support is not needed or causes issues. Previously was set to `False` by default, which would cause issues with e.g. Gunicorn setups. Can also be set with the `HATCHET_CLIENT_GRPC_ENABLE_FORK_SUPPORT` environment variable.

### Changed

- Changes `ValidTaskReturnType` to allow `Mapping[str, Any]` instead of `dict[str, Any]` to allow for more flexible return types in tasks, including using `TypedDict`.

## Type of change

### DIFF
--- a/sdks/python/hatchet_sdk/connection.py
+++ b/sdks/python/hatchet_sdk/connection.py
@@ -60,7 +60,7 @@ def new_conn(config: ClientConfig, aio: bool) -> grpc.Channel | grpc.aio.Channel
 
     # Set environment variable to disable fork support. Reference: https://github.com/grpc/grpc/issues/28557
     # When steps execute via os.fork, we see `TSI_DATA_CORRUPTED` errors.
-    os.environ["GRPC_ENABLE_FORK_SUPPORT"] = "False"
+    os.environ["GRPC_ENABLE_FORK_SUPPORT"] = config.grpc_enable_fork_support
 
     if config.tls_config.strategy == "none":
         conn = start.insecure_channel(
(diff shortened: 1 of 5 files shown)


sdks/python/hatchet_sdk/connection.py
```diff
@@ -60,7 +60,7 @@ def new_conn(config: ClientConfig, aio: bool) -> grpc.Channel | grpc.aio.Channel
 
     # Set environment variable to disable fork support. Reference: https://github.com/grpc/grpc/issues/28557
     # When steps execute via os.fork, we see `TSI_DATA_CORRUPTED` errors.
-    os.environ["GRPC_ENABLE_FORK_SUPPORT"] = "False"
+    os.environ["GRPC_ENABLE_FORK_SUPPORT"] = config.grpc_enable_fork_support
```
abelanger5: If a user enables fork support, do we want to set `GRPC_POLL_STRATEGY=poll` or `GRPC_POLL_STRATEGY=epoll1`?

From reference thread: https://github.com/grpc/grpc/issues/28557#issuecomment-1022710189

mrkaye97: seems like maybe we should? based on `Fork support is only compatible with the epoll1 and poll polling strategies` 

I can add that

abelanger5 (approved):

PR #2058, opened 2025-07-27
Deploy HyperDX locally via docker-compose and add traces to task controller

mnafees:
# Description

Add local Jaegar as part of docker compose infra. Also add tracing to partition table creation from tasks controller.

## Type of change

### DIFF
--- a/docker-compose.infra.yml
+++ b/docker-compose.infra.yml
@@ -34,14 +34,17 @@ services:
       - prometheus
 
   jaeger:
-    image: jaegertracing/all-in-one:latest
+    image: cr.jaegertracing.io/jaegertracing/jaeger:latest
     container_name: hatchet-jaeger
     ports:
       - "16686:16686"  # Jaeger UI
-      - "4317:4317"  # gRPC
-      - "4318:4318"  # HTTP
+      - "4317:4317"    # OTLP gRPC
+      - "4318:4318"    # OTLP HTTP
     environment:
-      - COLLECTOR_OTLP_ENABLED=true
+      - SPAN_STORAGE_TYPE=badger
+      - BADGER_EPHEMERAL=false
+      - BADGER_DIRECTORY_VALUE=/badger/data
+      - BADGER_DIRECTORY_KEY=/badger/keys
     volumes:
       - hatchet_jaeger_data:/badger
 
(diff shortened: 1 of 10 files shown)


docker-compose.infra.yml
```diff
@@ -33,6 +33,19 @@ services:
...
 
+  jaeger:
+    image: jaegertracing/all-in-one:latest
+    container_name: hatchet-jaeger
+    ports:
+      - "16686:16686"  # Jaeger UI
+      - "4317:4317"  # gRPC
+      - "4318:4318"  # HTTP
+    environment:
+      - COLLECTOR_OTLP_ENABLED=true
```
mrkaye97: we might want to set `badger.span-store-ttl` to some relatively small value initially? 

```
SPAN_STORAGE_TYPE=badger
BADGER_SPAN_STORE_TTL=24h
```

from here: https://www.jaegertracing.io/docs/1.19/deployment/cli/

mrkaye97 (commented):
nice! left one question

internal/services/controllers/v1/task/controller.go
```diff
@@ -252,80 +253,113 @@ func (tc *TasksControllerImpl) Start() (func() error, error) {
 
 	ctx, cancel := context.WithCancel(context.Background())
 
+	spanContext, span := telemetry.NewSpan(ctx, "tasks-controller-start")
```
mrkaye97: maybe we should shoot for having a consistent format? I like this idea: `"OLAPControllerImpl.createTablePartition"`

internal/services/controllers/v1/task/controller.go
```diff
@@ -252,80 +253,113 @@ func (tc *TasksControllerImpl) Start() (func() error, error) {
...
 		),
 	)
 
 	if err != nil {
+		err := fmt.Errorf("could not schedule step run timeout: %w", err)
+
 		cancel()
-		return nil, fmt.Errorf("could not schedule step run timeout: %w", err)
+		span.RecordError(err)
+		span.End()
```
mrkaye97, line 272: I'm a bit surprised we need to call these two lines manually

mnafees: Generally you will do `defer span.End()` right after getting a new span inside a function but in this particular `Start()` function we are doing things little differently and depend on the `cleanup()` function to take care of such things.

As for `span.RecordErr()`, we do need to call it whenever we want to register errors and link it to the current span.

mrkaye97 (approved):

PR #2061, opened 2025-07-29
Update worker active status only when needed

mnafees:
# Description

We seem to be returning the entire `Worker` whenever we try to update worker active status.

## Type of change

## What's Changed

- [x] Do not return anything when updating worker active status
- [x] Prevent redundant updates in the span of 100 ms

### DIFF
--- a/pkg/repository/postgres/dbsqlc/workers.sql
+++ b/pkg/repository/postgres/dbsqlc/workers.sql
@@ -250,7 +250,7 @@ WHERE
   "webhookId" = @webhookId::uuid
 RETURNING *;
 
--- name: UpdateWorkerActiveStatus :one
+-- name: UpdateWorkerActiveStatus :exec
 UPDATE "Worker"
 SET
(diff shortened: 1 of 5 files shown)


pkg/repository/postgres/dbsqlc/workers.sql
```diff
@@ -250,18 +250,18 @@ WHERE
...
 WHERE
     "id" = @id::uuid
     AND (
-        "lastListenerEstablished" IS NULL
-        OR "lastListenerEstablished" <= sqlc.narg('lastListenerEstablished')::timestamp
-        )
-RETURNING *;
+        "isActive" != @isActive::boolean
+        OR "lastListenerEstablished" IS NULL
+        OR "lastListenerEstablished" < sqlc.narg('lastListenerEstablished')::timestamp - INTERVAL '100 milliseconds'
```
mrkaye97, line 263: what's the reason for adding the interval here?

PR #2069, opened 2025-07-31
[Python] Feat: More use of `logger.exception`

mrkaye97:
# Description

Attempting to improve exception context propagation further

## Type of change


jfaust:
I can't comment on the line, because it's out of the diff, but `runner.py:181` needs to change here as well.

mrkaye97:
> I can't comment on the line, because it's out of the diff, but `runner.py:181` needs to change here as well.

lovely, thanks!

PR #2073, opened 2025-07-31
Add GRPC callback interceptor and correlation IDs to respective API and GRPC handlers

mnafees:
# Description

API and GRPC handlers should be able to "emit" a correlation ID which can be used for various purposes down the execution pipeline.

## Type of change


api/v1/server/handlers/v1/workflow-runs/trigger.go
```diff
@@ -122,6 +122,8 @@ func (t *V1WorkflowRunsService) V1WorkflowRunCreate(ctx echo.Context, request ge
 		return nil, err
 	}
 
+	ctx.Set("correlationId", sqlchelpers.UUIDToStr(rawWorkflowRun.WorkflowRun.ExternalID))
```
abelanger5: I'd be strongly in favor of searching for a specific key in the additional metadata as well, and setting that as the correlation ID if it exists. Note that I think correlation ID and resource ID are distinct and we should probably store both

api/v1/server/handlers/v1/workflow-runs/trigger.go
```diff
@@ -122,6 +122,16 @@ func (t *V1WorkflowRunsService) V1WorkflowRunCreate(ctx echo.Context, request ge
 		return nil, err
 	}
 
+	if request.Body.AdditionalMetadata != nil {
+		correlationId, ok := (*request.Body.AdditionalMetadata)["correlationId"].(string)
```
abelanger5: Probably safer to do:

```go
if correlationIdInterface, exists := (*request.Body.AdditionalMetadata)["correlationId"]; exists {
        if correlationId, ok := correlationIdInterface.(string); ok {

        }
    }
```

For all of these - since additional metadata is a `map[string]interface{}` and some old SDKs were technically able to send JSON as the interface value

internal/services/admin/server.go
```diff
@@ -167,6 +175,12 @@ func (a *AdminServiceImpl) bulkTriggerWorkflowV0(ctx context.Context, req *contr
 		if err != nil {
 			return nil, fmt.Errorf("could not queue workflow run: %w", err)
 		}
+
+		var corrId *string
+		if req.Workflows[i].AdditionalMetadata != nil {
+			corrId = extractCorrelationId(*req.Workflows[i].AdditionalMetadata)
+		}
+		grpcmiddleware.TriggerCallback(ctx, workflowRunId, "workflow-run", corrId)
```
abelanger5: Why aren't we setting these in `ctx` in this case?

api/v1/server/handlers/workflows/create_scheduled.go
```diff
@@ -41,6 +41,14 @@ func (t *WorkflowService) ScheduledWorkflowRunCreate(ctx echo.Context, request g
 		), nil
 	}
 
+	correlationId, ok := request.Body.AdditionalMetadata["correlationId"].(string)
```
abelanger5: I'm also worried about panics here, there doesn't seem to be enough guarding?

abelanger5 (changes requested):

api/v1/server/handlers/api-tokens/revoke.go
```diff
@@ -24,5 +25,8 @@ func (a *APITokenService) ApiTokenUpdateRevoke(ctx echo.Context, request gen.Api
 		return nil, err
 	}
 
+	ctx.Set(string(constants.ResourceIdKey), apiToken.ID.String())
+	ctx.Set(string(constants.ResourceTypeKey), string(constants.ResourceTypeApiToken))
```
abelanger5:
```suggestion
	ctx.Set(string(constants.ResourceTypeKey), constants.ResourceTypeApiToken.String())
```

thought that was the point of adding the string method?

api/v1/server/handlers/workflows/create_scheduled.go
```diff
@@ -41,6 +42,17 @@ func (t *WorkflowService) ScheduledWorkflowRunCreate(ctx echo.Context, request g
...
+	correlationIdInterface, ok := (request.Body.AdditionalMetadata)[string(constants.CorrelationIdKey)]
+	if ok {
+		correlationId, ok := correlationIdInterface.(string)
+		if ok {
+			ctx.Set(string(constants.CorrelationIdKey), correlationId)
+		}
+	}
+
+	ctx.Set(string(constants.ResourceIdKey), scheduled.ID.String())
+	ctx.Set(string(constants.ResourceTypeKey), "scheduled-workflow")
```
abelanger5: missed constant?

abelanger5 (changes requested):

abelanger5 (approved):

PR #2077, opened 2025-08-01
Fix: Scheduled runs race w/ idempotency key check

mrkaye97:
# Description

Attempt to fix race condition causing scheduled runs to trigger multiple times by adding an idempotency keys table to act kind of like a lock, preventing multiple of the same run from being triggered

## Type of change


pkg/repository/v1/idempotency.go
```diff
@@ -0,0 +1,73 @@
...
+func (r *idempotencyRepository) EvictExpiredIdempotencyKeys(context context.Context, tenantIds []pgtype.UUID) error {
+	return r.queries.CleanUpExpiredIdempotencyKeys(context, r.pool, tenantIds)
+}
+
+type KeyClaimantPair struct {
+	IdempotencyKey      IdempotencyKey
+	ClaimedByExternalId pgtype.UUID
+}
+
+func claimIdempotencyKeys(context context.Context, queries *sqlcv1.Queries, pool *pgxpool.Pool, tenantId string, claims []KeyClaimantPair) (map[IdempotencyKey]WasAlreadyClaimed, error) {
```
mrkaye97: what's the right way to do this? I want this to be available inside of the other repos but don't want to need to put this repo on the shared repo ideally

pkg/repository/v1/sqlcv1/idempotency-keys.sql
```diff
@@ -0,0 +1,78 @@
...
+            FROM incoming_claims
+        )
+        AND claimed_by_external_id IS NULL
+        AND expires_at > NOW()
+), to_update AS (
+    -- Cases to handle:
+    -- 1. What happens if we get two incoming records for the same key but different claimed_by_external_id? (basically two runs trying to claim the same key)
+    --   1a. What about if we have two different idempotency keys for the same key value with different expirations?
+    --       e.g. key = 'abc', expires_at = 'x' and key = 'abc', expires_at = 'y'?
+    -- ✅ 2. What happens if we get two incoming records for the same key and same claimed_by_external_id?
```
mrkaye97: will remove this / reformat the comment, just added to help me think through it 😅

cmd/hatchet-migrate/migrate/migrations/20250904151226_v1_0_41.sql
```diff
@@ -0,0 +1,23 @@
...
+    expires_at TIMESTAMPTZ NOT NULL,
+    claimed_by_external_id UUID,
+
+    inserted_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    PRIMARY KEY (tenant_id, expires_at, key)
+);
+
+CREATE UNIQUE INDEX v1_idempotency_key_unique_tenant_key ON v1_idempotency_key (tenant_id, key);
```
mrkaye97, line 17: not 100% sure about this - do we want to allow multiple of the same key with different `expires_at`? or make the keys tenant-unique?

sql/schema/v1-core.sql
```diff
@@ -1628,3 +1628,19 @@ CREATE TABLE v1_durable_sleep (
...
+    expires_at TIMESTAMPTZ NOT NULL,
+    claimed_by_external_id UUID,
+
+    inserted_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    PRIMARY KEY (tenant_id, expires_at, key)
+);
+
+CREATE UNIQUE INDEX v1_idempotency_key_expires_at_idx ON v1_idempotency_key (tenant_id, key);
```
abelanger5: The name of this index seems incorrect?

cmd/hatchet-migrate/migrate/migrations/20250904151226_v1_0_41.sql
abelanger5: keys being tenant unique makes sense to me

abelanger5 (approved):

sql/schema/v1-core.sql
mrkaye97: whoops yeah, thanks - will fix this

PR #2078, opened 2025-08-01
Fix: Propagate priority through to DAG subtasks

mrkaye97:
# Description

Propagates priority through to DAG subtasks

## Type of change


cmd/hatchet-migrate/migrate/migrations/20250801183812_v1_0_35.sql
```diff
@@ -0,0 +1,11 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE v1_match
+ADD COLUMN trigger_priority INTEGER;
```
mrkaye97, line 4: we have one table (`v1_task`) where priority is nullable, so I don't think we can make this safely non-null. we could provide a default here but I don't think it'll make a difference since these rows are ephemeral 🤷

abelanger5 (approved):

PR #2086, opened 2025-08-05
feat(ruby): initial commit

grutt:
# Description

Initial minimal ruby sdk

## Type of change

## What's Changed
- [ ] Consistent configuration loading with python. Note: many params are unused for now
- [ ] Generated Rest API
- [ ] Runs Feature Client
- [ ] Events Feature Client
- [ ] Initial e2e tests


sdks/ruby/src/.rubocop.yml
```diff
@@ -0,0 +1,8 @@
+AllCops:
+  TargetRubyVersion: 3.1
+
+Style/StringLiterals:
+  EnforcedStyle: double_quotes
+
+Style/StringLiteralsInInterpolation:
```
mrkaye97, line 7: there are probably some rubocop recommendations for which rules to enable somewhere which might be useful here

sdks/ruby/src/lib/hatchet-sdk.rb
```diff
@@ -0,0 +1,86 @@
...
+    # @option options [Logger] :logger Custom logger instance (default: Logger.new($stdout))
+    # @option options [Integer] :listener_v2_timeout Timeout for listener v2 in milliseconds
+    # @option options [Integer] :grpc_max_recv_message_length Maximum gRPC receive message length (default: 4MB)
+    # @option options [Integer] :grpc_max_send_message_length Maximum gRPC send message length (default: 4MB)
+    # @option options [Hash] :worker_preset_labels Hash of preset labels for workers
+    # @option options [Boolean] :enable_force_kill_sync_threads Enable force killing of sync threads (default: false)
+    # @option options [Boolean] :enable_thread_pool_monitoring Enable thread pool monitoring (default: false)
+    # @option options [Integer] :terminate_worker_after_num_tasks Terminate worker after this many tasks
+    # @option options [Boolean] :disable_log_capture Disable log capture (default: false)
+    # @option options [Boolean] :grpc_enable_fork_support Enable gRPC fork support (default: false)
```
mrkaye97, line 48: some of these might not be applicable to ruby? e.g. `enable_thread_pool_monitoring`, `terminate_worker_after_num_tasks`, `disable_log_capture`, `grpc_enable_fork_support`, `OpenTelemetryConfig`, etc.?

sdks/ruby/src/lib/hatchet-sdk.rb
```diff
@@ -0,0 +1,86 @@
...
+    #     token: "eyJhbGciOiJIUzI1NiJ9...",
+    #     namespace: "my_app",
+    #     worker_preset_labels: { "env" => "production", "version" => "1.0.0" }
+    #   )
+    def initialize(**options)
+      @config = Config.new(**options)
+    end
+
+    def rest_client
+        @rest_client ||= Hatchet::Clients.rest_client(@config)
```
mrkaye97, line 69: not bad!

sdks/ruby/src/lib/hatchet/clients.rb
```diff
@@ -0,0 +1,71 @@
...
+      # @return [Hatchet::Clients::Rest::ApiClient] Configured REST API client
+      # @raise [LoadError] if REST client hasn't been generated yet
+      #
+      # @example Create a REST client
+      #   config = Hatchet::Config.new(token: "your-jwt-token")
+      #   rest_client = Hatchet::Clients.rest_client(config)
+      #   workflows_api = Hatchet::Clients::Rest::WorkflowApi.new(rest_client)
+      def rest_client(config)
+        unless rest_available?
+          raise LoadError, "REST client not available. Run `rake api:generate` to generate it from the OpenAPI spec."
```
mrkaye97, line 32: this seems like it should either never happen or be some sort of fatal error, but I'm not sure this is the right error message for a user-facing thing?

sdks/ruby/src/lib/hatchet/features/runs.rb
```diff
@@ -0,0 +1,423 @@
...
+      # @raise [Hatchet::Error] If the API request fails or returns an error
+      # @example Poll with default settings (1s interval, no timeout)
+      #   result = runs.poll("workflow-run-123")
+      # @example Poll with custom interval and timeout
+      #   result = runs.poll("workflow-run-123", interval: 2.0, timeout: 30.0)
+      # @since 0.0.0
+      def poll(workflow_run_id, interval: 1.0, timeout: nil)
+        start_time = Time.now
+        
+        loop do
```
grutt, line 355: this is blocking.. need to think about this

sdks/ruby/src/lib/hatchet/clients/rest.rb
```diff
@@ -0,0 +1,126 @@
...
+
+require 'fileutils'
+require 'rbconfig'
+
+# Integration file for generated Hatchet REST API client
+# This file loads the generated REST client and makes it available under the Hatchet::Clients::Rest namespace
+
+begin
+  # Set up load paths for the generated client
+  rest_lib_path = File.expand_path("rest/lib", __dir__)
```
mrkaye97, line 12: will these paths also work when the library is packaged up? I'm not too familiar with this, but this seems a bit odd to me to be using file paths in one of the clients

sdks/ruby/src/lib/hatchet/clients/rest.rb
```diff
@@ -0,0 +1,126 @@
...
+  # Create an alias so hatchet-sdk-rest/ paths resolve to the actual location
+  # This is a bit of a hack, but necessary because the generator expects gem-style paths
+  hatchet_sdk_rest_base = File.expand_path("rest/lib/hatchet-sdk-rest", __dir__)
+  $LOAD_PATH.unshift(hatchet_sdk_rest_base) unless $LOAD_PATH.include?(hatchet_sdk_rest_base)
+  
+  # Create a symlink in the load path to make hatchet-sdk-rest/ paths work
+  fake_gem_path = File.expand_path("rest/lib/hatchet-sdk-rest/hatchet-sdk-rest", __dir__)
+  unless File.exist?(fake_gem_path)
+    FileUtils.mkdir_p(File.dirname(fake_gem_path))
+    # On Unix systems, create a symlink; on Windows, copy the files
```
mrkaye97, line 24: we probably don't need windows support?

sdks/ruby/src/lib/hatchet/clients/rest.rb
```diff
@@ -0,0 +1,126 @@
...
+              config.scheme = hatchet_config.server_url.start_with?('https') ? 'https' : 'http'
+            end
+            
+            # Set timeout if available
+            if hatchet_config.listener_v2_timeout
+              config.timeout = hatchet_config.listener_v2_timeout / 1000.0 # Convert ms to seconds
+            end
+            
+            config
+          end
```
mrkaye97: why do we split the host and scheme here?

sdks/ruby/src/lib/hatchet/clients/rest.rb
```diff
@@ -0,0 +1,126 @@
...
+  # This allows the main SDK to load without errors even before generation
+  warn "REST client not fully loaded: #{e.message}" if ENV["DEBUG"]
+  
+  module Hatchet
+    module Clients
+      module Rest
+        # Placeholder classes that will raise helpful errors
+        class ApiClient
+          def initialize(*)
+            raise LoadError, "REST client not generated. Run `rake api:generate` to generate it."
```
mrkaye97, line 101: same here re: maybe not the best user-facing error / how would this even happen?

sdks/ruby/src/lib/hatchet/config.rb
```diff
@@ -0,0 +1,480 @@
...
+    end
+
+    def parse_json_array(value)
+      return nil if value.nil? || value.empty?
+
+      JSON.parse(value)
+    rescue JSON::ParserError
+      nil
+    end
+  end
```
mrkaye97, line 134: probably can remove this for now unless we're adding an instrumentor here?

sdks/ruby/src/lib/hatchet/config.rb
```diff
@@ -0,0 +1,480 @@
...
+  # @example Configuration via environment variables
+  #   ENV["HATCHET_CLIENT_TOKEN"] = "your-jwt-token"
+  #   ENV["HATCHET_CLIENT_HOST_PORT"] = "localhost:7070"
+  #   config = Config.new  # Will load from environment
+  class Config
+    DEFAULT_HOST_PORT = "localhost:7070"
+    DEFAULT_SERVER_URL = "https://app.dev.hatchet-tools.com"
+    DEFAULT_GRPC_MAX_MESSAGE_LENGTH = 4 * 1024 * 1024 # 4MB
+
+    ENV_FILE_NAMES = [".env", ".env.hatchet", ".env.dev", ".env.local"].freeze
```
mrkaye97, line 164: this is for pydantic-settings in Python to read from, although fine by me to have the same here too

sdks/ruby/src/lib/hatchet/config.rb
```diff
@@ -0,0 +1,480 @@
...
+    # @!attribute healthcheck
+    #   @return [HealthcheckConfig] Healthcheck configuration
+    # @!attribute otel
+    #   @return [OpenTelemetryConfig] OpenTelemetry configuration
+    attr_accessor :token, :tenant_id, :host_port, :server_url, :namespace,
+                  :logger, :listener_v2_timeout, :grpc_max_recv_message_length,
+                  :grpc_max_send_message_length, :worker_preset_labels,
+                  :enable_force_kill_sync_threads, :enable_thread_pool_monitoring,
+                  :terminate_worker_after_num_tasks, :disable_log_capture,
+                  :grpc_enable_fork_support, :tls_config, :healthcheck, :otel
```
mrkaye97: same here re: python-specific config. healthcheck is another one

sdks/ruby/src/lib/hatchet/features/events.rb
```diff
@@ -0,0 +1,266 @@
...
+      # @raise [ArgumentError] If required parameters are missing
+      # @raise [Hatchet::Error] If the API request fails or returns an error
+      # @example Creating a simple event
+      #   response = events.create(
+      #     key: "user-login",
+      #     data: { user_id: 123, action: "login" },
+      #     additional_metadata: { ip_address: "192.168.1.1" }
+      #   )
+      # @since 0.0.2
+      def create(key:, data:, additional_metadata: nil, priority: nil, scope: nil, namespace: nil)
```
mrkaye97, line 61: this should be `push` I think to be consistent

sdks/ruby/src/spec/hatchet/features/events_spec.rb
```diff
@@ -0,0 +1,487 @@
...
+    it "returns the replay response" do
+      result = events_client.replay(event_ids: ["event-1"])
+      expect(result).to eq(replay_response)
+    end
+  end
+
+  describe "private methods" do
+    describe "#apply_namespace" do
+      it "applies default namespace from config" do
+        config_with_namespace = Hatchet::Config.new(token: valid_token, tenant_id: "test-tenant", namespace: "test_")
```
grutt: need to remove tenant_id from config options

sdks/ruby/src/lib/hatchet/features/events.rb
```diff
@@ -0,0 +1,266 @@
...
+        event_key = apply_namespace(key, namespace)
+        
+        event_request = HatchetSdkRest::CreateEventRequest.new(
+          key: event_key,
+          data: data,
```
mrkaye97, line 66: nit: I think we call this either `input` or `payload` in other places

sdks/ruby/src/lib/hatchet/features/events.rb
```diff
@@ -0,0 +1,266 @@
...
+      #   )
+      def push(event_key, payload, additional_metadata: nil, namespace: nil, priority: nil)
+        create(
+          key: event_key,
+          data: payload,
+          additional_metadata: additional_metadata,
+          priority: priority,
+          namespace: namespace
+        )
+      end
```
mrkaye97, line 98: wait, why do we have both `create` and `push` 🤔 - maybe we can just delete the `create` method

sdks/ruby/src/lib/hatchet/features/events.rb
```diff
@@ -0,0 +1,266 @@
...
+      def replay(event_ids: nil, keys: nil, since: nil, until_time: nil)
+        replay_request = HatchetSdkRest::ReplayEventRequest.new(
+          event_ids: event_ids,
+          keys: keys,
+          since: since&.utc&.iso8601,
+          until: until_time&.utc&.iso8601
+        )
+        
+        @event_api.event_update_replay(@config.tenant_id, replay_request)
+      end
```
mrkaye97: I think these might not do anything in V1 - would be good to check

sdks/ruby/src/lib/hatchet/features/runs.rb
```diff
@@ -0,0 +1,423 @@
...
+      def initialize(since:, until_time: nil, statuses: nil, workflow_ids: nil, additional_metadata: nil)
+        @since = since
+        @until_time = until_time
+        @statuses = statuses
+        @workflow_ids = workflow_ids
+        @additional_metadata = additional_metadata
+      end
+    end
+
+    # Options for bulk cancel and replay operations
```
mrkaye97, line 21: nit: I'd remove these types of comments

sdks/ruby/src/lib/hatchet/features/runs.rb
```diff
@@ -0,0 +1,423 @@
...
+          external_ids: @ids,
+          filter: v1_task_filter
+        )
+      end
+
+      private
+
+      def maybe_additional_metadata_to_kv(metadata)
+        return nil unless metadata
+        metadata.map { |k, v| { key: k.to_s, value: v.to_s } }
```
mrkaye97, line 63: do we want to force these to strings?

sdks/ruby/src/lib/hatchet/features/runs.rb
```diff
@@ -0,0 +1,423 @@
...
+            statuses: statuses,
+            until: until_time.utc.iso8601,
+            additional_metadata: maybe_additional_metadata_to_kv(additional_metadata),
+            workflow_ids: workflow_ids,
+            worker_id: worker_id,
+            parent_task_external_id: parent_task_external_id,
+            triggering_event_external_id: triggering_event_external_id
+          }
+        )
+      end
```
mrkaye97, line 248: this should do pagination by default - it's copying the Python code, but `list_with_pagination` was added to avoid a breaking change. we can probably remove that method here and just have pagination happen automatically in the `list`

sdks/ruby/src/lib/hatchet/features/runs.rb
```diff
@@ -0,0 +1,423 @@
...
+        ranges
+      end
+
+      # Convert additional metadata hash to key-value array format expected by API
+      #
+      # @param metadata [Hash<String, String>, nil] Metadata hash
+      # @return [Array<Hash>, nil] Array of {key: string, value: string} objects
+      def maybe_additional_metadata_to_kv(metadata)
+        return nil unless metadata
+        metadata.map { |k, v| { key: k.to_s, value: v.to_s } }
```
mrkaye97, line 419: same here re: cast to string

sdks/ruby/src/scripts/generate.sh
```diff
@@ -0,0 +1,167 @@
...
+    
+    echo -e "${GREEN}✅ Ruby REST client generated${NC}"
+}
+
+# Function to apply custom patches
+apply_patches() {
+    echo -e "${BLUE}🔧 Applying custom patches...${NC}"
+    
+    echo -e "${GREEN}✅ Custom patches applied${NC}"
+}
```
mrkaye97: rm this since it's not doing anything?

sdks/ruby/src/scripts/generate.sh
```diff
@@ -0,0 +1,167 @@
...
+    echo -e "${BLUE}🔍 Validating OpenAPI specification...${NC}"
+    
+    if [ ! -f "$OPENAPI_SPEC" ]; then
+        echo -e "${RED}❌ OpenAPI spec not found at: $OPENAPI_SPEC${NC}"
+        exit 1
+    fi
+    
+    openapi-generator-cli validate -i "$OPENAPI_SPEC"
+    echo -e "${GREEN}✅ OpenAPI specification is valid${NC}"
+}
```
mrkaye97: I think we can assume it's valid

mrkaye97 (commented):
left a bunch of comments on misc. things - will go through the tests more later

sdks/ruby/src/lib/hatchet-sdk.rb
grutt: yeah had a note on this in the pr comment. i'll go through and remove some

sdks/ruby/src/lib/hatchet-sdk.rb
grutt: best feature lol

sdks/ruby/src/lib/hatchet/clients.rb
grutt: good flag. i need to sink into this. basically should never hit

sdks/ruby/src/lib/hatchet/clients/rest.rb
grutt: i think this can be removed.. i made some changes to the default loader.

sdks/ruby/src/lib/hatchet/features/runs.rb
grutt: should be string or int?

mrkaye97: can't it be anything? since we just write it as jsonb? maybe I'm wrong about that idk

mrkaye97: it won't work in the frontend if it's not a string though I don't think (unless pg has some magic type matching capabilities in json)

sdks/ruby/src/lib/hatchet/clients/rest.rb
mrkaye97: nice 👍

mrkaye97 (approved):

PR #2089, opened 2025-08-06
[HAT-429] Go SDK v1

mnafees:
# Description

New Go SDK v1 without generics.

## Type of change


sdks/go/client.go
```diff
@@ -0,0 +1,221 @@
...
+	maxRuns := config.slots
+	if config.durableSlots > maxRuns {
+		maxRuns = config.durableSlots
+	}
+
+	workerOpts := []worker.WorkerOpt{
+		worker.WithClient(c.v0Client),
+		worker.WithName(name),
+		worker.WithMaxRuns(maxRuns),
+	}
```
mrkaye97: this is a bit confusing, but in Python / TS we're basically running two workers side by side - one for durable tasks and another for non-durable tasks. I think we need to do the same here. we basically have two entirely separate action registries, etc., and then each one gets assigned the proper number of slots as is passed in

sdks/go/client.go
```diff
@@ -0,0 +1,221 @@
...
+			})
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		// Register on failure function if exists
+		if onFailureFn != nil {
+			// The on-failure action ID is derived from the workflow name
+			onFailureActionID := strings.ToLower(fmt.Sprintf("%s:on-failure", req.Name))
```
mrkaye97: I think we're using an underscore as the separator for this in the other SDKs

sdks/go/client.go
```diff
@@ -0,0 +1,221 @@
...
+
+// RunNoWait executes a workflow with the provided input without waiting for completion.
+// Returns a workflow run reference that can be used to track the run status.
+func (c *Client) RunNoWait(ctx context.Context, workflowName string, input any) (*v0Client.Workflow, error) {
+	return c.v0Client.Admin().RunWorkflow(workflowName, input)
+}
+
+// RunBulk executes multiple workflow instances with different inputs.
+// Returns workflow run IDs that can be used to track the run statuses.
+func (c *Client) RunBulk(ctx context.Context, workflowName string, inputs []any) ([]string, error) {
```
mrkaye97: nit: in Python we're calling this `RunMany`, and also for all of these, the `run` flavors return the result, and the `no_wait` flavors return a ref

sdks/go/client.go
mnafees: TS SDK: https://github.com/hatchet-dev/hatchet/blob/main/sdks/typescript/src/v1/client/worker/worker-internal.ts#L909

Python SDK: https://github.com/hatchet-dev/hatchet/blob/main/sdks/python/hatchet_sdk/runnables/workflow.py#L1019

I guess we should consolidate the naming for sure 😅

sdks/go/client.go
```diff
@@ -0,0 +1,221 @@
...
+)
+
+// Client provides the main interface for interacting with Hatchet.
+type Client struct {
+	v0Client v0Client.Client
+}
+
+// NewClient creates a new Hatchet client.
+// Configuration options can be provided to customize the client behavior.
+func NewClient(opts ...v0Client.ClientOpt) (*Client, error) {
```
chen-openmart, line 23: What is the difference here between `v1.NewHatchetClient(opt)`
```
import (
	v1 "github.com/hatchet-dev/hatchet/pkg/v1"
)

func main() {
    v1hatchet, _ := v1.NewHatchetClient(opt)
}
```

mnafees: @chen-openmart The initialization of the client itself is the same as the older v1 implementation. The main differences are around using reflection instead of Go generics to overcome the limitations around semantics.

sdks/go/client.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// NewStandaloneTask creates a workflow containing a single task.
+// This is a convenience method for simple workflows.
+func (c *Client) NewStandaloneTask(name string, fn any, options ...TaskOption) *Workflow {
+	return NewStandaloneTask(name, fn, c.v0Client, options...)
+}
+
+// Run executes a workflow with the provided input and waits for completion.
+func (c *Client) Run(ctx context.Context, workflowName string, input any) error {
```
abelanger5: Since `Run` is blocking, we should definitely return a result object here

sdks/go/workflow.go
```diff
@@ -0,0 +1,608 @@
...
+// The function parameter must have the signature:
+//
+//	func(ctx Context, input T) (T, error)
+//
+// For durable tasks, use:
+//
+//	func(ctx DurableContext, input T) (T, error)
+//
+// Function signatures are validated at runtime using reflection.
+func (w *Workflow) NewTask(name string, fn any, options ...TaskOption) *Workflow {
```
abelanger5: What is the thinking behind when I should use `NewTask` versus `AddTask`?

sdks/go/workflow.go
```diff
@@ -0,0 +1,608 @@
...
+// WithSkipIf sets a condition that will skip the task if met.
+func WithSkipIf(condition condition.Condition) TaskOption {
+	return func(config *taskConfig) {
+		config.skipIf = condition
+	}
+}
+
+// Task represents a task reference for building DAGs.
+type Task struct {
+	NamedTask create.NamedTask
```
abelanger5: Do we just want to give it a `Name()` method instead? I think we needed this pattern for generics but I don't see why we need it atm

sdks/go/workflow.go
```diff
@@ -0,0 +1,608 @@
...
+		if f64, ok := val.(float64); ok {
+			return reflect.ValueOf(int64(f64)), true
+		}
+	}
+
+	return reflect.Value{}, false
+}
+
+// Workflow represents a workflow definition that can contain multiple tasks.
+type Workflow struct {
```
abelanger5, line 109: Would also be good if `Workflow` had a `Name()` method since namespaces can modify the name

abelanger5 (changes requested):

abelanger5 (approved):

PR #2093, opened 2025-08-06
add wsl note for python

szvsw:
# Description

Currently, workers on Windows raise an attribute error. This can be confusing for someone following the docs, as I do not think it mentions anywhere else that windows workers are not really supported.  A user might not realize that they are doing nothing wrong and have not misconfigured anything, as the error is somewhat cryptic (it's about signaling).

## Type of change

## What's Changed

- [x] Added a recommendation to either use WSL or Docker containers when building with Hatchet on windows.


mrkaye97 (approved):
Thanks!

PR #2097, opened 2025-08-06
Fix: Call `PopulateTaskRunData` sequentially

mrkaye97:
# Description

For some reason the planner (esp. with Timescale) is really bad at figuring out which partitions to scan and how to join results together for this query, but running it for a single task is fast 🤷 

## Type of change

### DIFF
--- a/pkg/repository/v1/olap.go
+++ b/pkg/repository/v1/olap.go
@@ -705,8 +705,10 @@ func (r *OLAPRepositoryImpl) ListTasksByDAGId(ctx context.Context, tenantId stri
 }
 
 func (r *OLAPRepositoryImpl) ListTasksByIdAndInsertedAt(ctx context.Context, tenantId string, taskMetadata []TaskMetadata) ([]*sqlcv1.PopulateTaskRunDataRow, error) {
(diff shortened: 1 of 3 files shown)


pkg/repository/v1/olap.go
```diff
@@ -605,23 +606,30 @@ func (r *OLAPRepositoryImpl) ListTasks(ctx context.Context, tenantId string, opt
 		return nil, 0, err
 	}
 
-	taskIds := make([]int64, 0)
-	taskInsertedAts := make([]pgtype.Timestamptz, 0)
+	tasksWithData := make([]*sqlcv1.PopulateTaskRunDataRow, 0)
+	span.SetAttributes(attribute.KeyValue{
+		Key:   "PopulateTaskRunData.num_tasks_to_populate",
+		Value: attribute.Int64Value(int64(len(rows))),
+	})
```
mrkaye97: any opinions on what naming convention to use for attributes?

mrkaye97:
~this broke some deduping, need to fix that~

abelanger5 (approved):

Issue #2095, opened 2025-08-06
Upgrade prometheus-client dependency to >0.22


srochiramani:
**Describe the issue**
Currently prometheus-client is pinned to 0.21 which is blocking upgrading other libraries to the latest version. In my case, its livekit. Is there a reason the python SDK is pinned to this version? 

**Environment**
- SDK: Python

**Expected behavior**
Pinned dependency has a larger window of versions for prometheus-client

srochiramani:
Looks like there was a PR to make this change -> #2046. Would be great to get this in soon.

srochiramani:
@abelanger5 thoughts on this? Upon trying this in a fork, everything seems okay -- curious if we can bump this up in the next release.

mrkaye97:
hey @srochiramani - thanks! mind re-opening that PR and making the constraint `>=0.21.1`? I think that should work well!

srochiramani:
@mrkaye97 done at #2098  -- the previous PR wasn't mine.

mrkaye97:
closed by #2098 !

PR #2098, opened 2025-08-06
Update prometheus-client version to >=0.21.1

srochiramani:
Update prometheus-client version to >=0.21.1 (instead of being pinned to v0.21.1)

# Description

This PR relaxes the version constraint of prometheus-client from ^0.21.1 to >=0.22. The reason for this change is that the previous constraint was too strict and caused dependency conflicts in Poetry when used alongside other packages that require a newer version. Since prometheus-client>=0.22 is backward compatible, this update is safe and avoids unnecessary version resolution issues.

Fixes # (issue)
#2095

## Type of change
- [X] Chore (changes which are not directly related to any business logic)


mrkaye97:
looks great, thank you! we'll also need to bump to a new patch version of the Hatchet SDK - sorry for forgetting to mention that

srochiramani:
@mrkaye97 thanks -- done.

mrkaye97 (approved):
looks great, thank you for the contribution!

PR #2110, opened 2025-08-11
[HAT-432] Enforce task priorities to be between 1 and 3

mnafees:
# Description

We only support priorities between 1 and 3 (inclusive) and need to enforce it on the API, gRPC layers.

## Type of change


cmd/hatchet-migrate/migrate/migrations/20250811130258_v1_0_36.sql
```diff
@@ -0,0 +1,108 @@
+-- +goose Up
+-- +goose StatementBegin
+
+UPDATE v1_task
+SET priority = 3
+WHERE priority > 3;
```
mrkaye97: I think this is a pretty risky statement to run in a migration and we probably shouldn't do it like this (there must be a safer way, will think about it). two issues here afaict:

1. The priority is not in an index, so we'll need to scan the whole table to find matching rows, and then we'll acquire a lock on all the rows with priority > 3 (which I think should be very small so probably not a big deal). the first part could take a pretty long time
2. Since we're only going to lock the rows that are found by the where clause and iirc we're using snapshot isolation, there's nothing preventing someone from inserting a new row with priority 4 in the meantime which would make adding the check constraint downstream fail, and then the migration would fail. I think to avoid this the best thing to do would be to deploy the change that prevents writes of priority > 4 on the Go side first so we know that none of those are being written, and then enforce the constraint. but I think here we'll do that in the other order

on the other hand though, we also need to move a lot of data around for the lookup tables and such, but the concern here I think is that this table is critical to task processing and the OLAP tables aren't (although the lookup table might be). I'm not 100% sure here, just being cautious 🤷 

just my two cents!

mrkaye97: Actually one other thing - IIRC we use priority 4 internally for replays, and I'm not sure if that'll work here. We'd need to test that out

internal/services/admin/server.go
```diff
@@ -413,6 +413,10 @@ func (a *AdminServiceImpl) ScheduleWorkflow(ctx context.Context, req *contracts.
 		return nil, status.Errorf(codes.InvalidArgument, "Invalid request: %s", err)
 	}
 
+	if req.Priority != nil && (*req.Priority < 1 || *req.Priority > 3) {
+		return nil, status.Errorf(codes.InvalidArgument, "priority must be between 1 and 3, got %d", *req.Priority)
+	}
+
```
mrkaye97, line 419: I know Alexander and I disagree on this 😅, IMO it's harder to maintain validation logic that's duplicated all over multiple layers of the app, but I think he's said in the past he'd prefer to just have it in multiple places for safety and the like. Don't feel super strongly, just a thought since if we wanted to change this rule we'd need to change it in many places in the code and also in the db

cmd/hatchet-migrate/migrate/migrations/20250811130258_v1_0_36.sql
mnafees: @mrkaye97 Thanks for the comments! Point number 2 makes sense to me. I can update the PR to only contain the Go changes and then create a follow up PR with the SQL changes.
As for point number 1 - would it make sense to do a paginated query for `v1_task`?

internal/services/admin/server.go
mnafees: Haha, I often like the Go way of error handling which is this way and to have multiple (similar) checks.

cmd/hatchet-migrate/migrate/migrations/20250811130258_v1_0_36.sql
mrkaye97: I'm not sure 😅 - this is kinda similar in intention to the migration that partitions the lookup table because that also is expensive, needs to take out locks, etc., but I don't think we've really decided on a strategy that we like yet.

I do kinda wonder if we could create an index concurrently on the priority, then use it to update, then drop it 😅 

honestly, it might just be fine, but just trying to be cautious here since this table could easily tens or hundreds of millions of tasks in it

internal/services/admin/server.go
mrkaye97: haha, sounds good to me 😅

mrkaye97 (approved):
🙌

abelanger5 (approved):

PR #2111, opened 2025-08-11
Fix: Deadlocking on DAG concurrency

mrkaye97:
# Description

Attempting to fix a deadlocking issue by simplifying to `FOR UPDATE` locks into a single CTE

Note: Not sure about performance here

## Type of change


pkg/repository/v1/sqlcv1/concurrency-overwrite.sql.go
abelanger5, line 1: Looks like we haven't updated both cancel in progress and cancel newest, just cancel newest? I think this would impact both strategies.

abelanger5 (changes requested):

abelanger5 (approved):

PR #2114, opened 2025-08-11
Refactor: Runs list state management + bug fixes part I

mrkaye97:
# Description

Lots of changes and misc. fixes here:

1. Set up a provider for managing runs state so we don't need to drill props everywhere
2. Use query params with a table key so we can store multiple tables in the URL - each key stores a minified JSON representation of the filter state
3. Fixes flashing when opening the additional meta popover
4. Fixes a bunch of query param setting issues which caused the back button to not work
5. Cleans up the action buttons on the runs table to be more consolidated so they're less likely to overflow on smaller screens
6. Adds back the freeze button to pause run fetching
7. Removes a bunch of options from the child runs view in the side sheet since I don't think it makes much sense to e.g. the trigger button
8. Removes the link to the children, since now that the child view in the side sheet works I don't think we really need it anymore
9. Renames the `**/tasks/**` paths to use `workflows` instead, since the workflow is really still the top level abstraction and calling them tasks was confusing IMO
10. Reworks the bulk cancel / replay modal a bit to be more nicely styled
11. Fixes the custom time filter
12. Makes the workflow and status filters multi-select
13. Fixed some issues with z indices on the side sheet temporarily (will move this over to the full side panel thing soon, hopefully)
14. Fixes the width of the column selector so the additional metadata doesn't overflow
15. Hides the parent external id from the column selector
16. Removes client-side sorting from the events table since it's not useful anyways
17. Changed `/workers/all` -> `/workers` and added a redirect

## Type of change


frontend/app/src/components/v1/molecules/combobox/combobox.tsx
```diff
@@ -164,7 +164,7 @@ export function Combobox({
           )}
         </Button>
       </PopoverTrigger>
-      <PopoverContent className="w-[300px] p-2" align="start">
+      <PopoverContent className="w-[300px] p-2 z-[70]" align="start">
```
mrkaye97, line 173: this was a bug - it was causing the filters to not pop up on top of the side sheet

abelanger5 (approved):

PR #2122, opened 2025-08-12
Chore: Regen Snippets

mrkaye97:
# Description

what it says on the tin

## Type of change


mnafees (approved):
codegen !FTW

PR #2127, opened 2025-08-13
[WIP][HAT-435] Make sure we do not return expired queue items according to tenant data retention period

mnafees:
# Description

Fixes # (issue)

## Type of change

## What's Changed

### DIFF
--- a/pkg/repository/v1/sqlcv1/queue.sql
+++ b/pkg/repository/v1/sqlcv1/queue.sql
@@ -81,6 +81,8 @@ SELECT
     *
 FROM
     v1_queue_item qi
+JOIN
+    "Tenant" t ON qi.tenant_id = t.id
 WHERE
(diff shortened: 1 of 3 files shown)


pkg/repository/v1/sqlcv1/queue.sql
```diff
@@ -90,6 +92,8 @@ WHERE
     )
     -- Added to ensure that the index is used
     AND qi.priority >= 1 AND qi.priority <= 4
+    -- Filter out those that are older than tenant's retention period
+    AND qi.task_inserted_at >= NOW() - t."dataRetentionPeriod"::interval
```
mrkaye97, line 96: any idea if this will change the plan for this query meaningfully? the task inserted at isn't part of the PK (the other columns in this where clause all are) but it's used in a separate index too

I think this is probably fine though given we'll use the PK for everything else in the where clause 🤷

PR #2128, opened 2025-08-13
Allow RabbitMQ to be used with Hatchet Lite

mnafees:
# Description

Right now we only allow the `postgres` message queue kind in Hatchet Lite. This PR keeps the default while allows the user to override the env var to use RabbitMQ.

## Type of change


frontend/docs/pages/self-hosting/hatchet-lite.mdx
```diff
@@ -70,21 +78,65 @@ Then run `docker-compose -f docker-compose.hatchet.yml up` to get the Hatchet Li
...
       - "8888:8888"
       - "7077:7077"
+    depends_on:
+      postgres:
+        condition: service_healthy
     environment:
+      LITE_MSGQUEUE_KIND: rabbitmq
+      LITE_MSGQUEUE_RABBITMQ_URL: amqp://user:password@rabbitmq:5672
+      # Refer to https://docs.hatchet.run/self-hosting/configuration-options
+      # for a list of all supported environment variables
```
mrkaye97: if we have this here, then if people copy-paste this they'll end up using Rabbit by default right? I don't know if we want that

mnafees: This is part of the tab `With Postgres + RabbitMQ` so this is expected.
image

mrkaye97: oooo thanks, missed that 😅

mrkaye97 (approved):

abelanger5 (approved):

abelanger5 (approved):

PR #2131, opened 2025-08-13
Feat: Webhook fixes / improvements

mrkaye97:
# Description

* Adding Slack as an integration + adding (basic) support for a challenge
* Adding support for updating the expression
* Adding headers to the CEL env
* Docs

## Type of change


sql/schema/v1-core.sql
```diff
@@ -522,7 +522,7 @@ CREATE TYPE v1_incoming_webhook_hmac_algorithm AS ENUM ('SHA1', 'SHA256', 'SHA51
 CREATE TYPE v1_incoming_webhook_hmac_encoding AS ENUM ('HEX', 'BASE64', 'BASE64URL');
 
 -- Can add more sources in the future
-CREATE TYPE v1_incoming_webhook_source_name AS ENUM ('GENERIC', 'GITHUB', 'STRIPE');
+CREATE TYPE v1_incoming_webhook_source_name AS ENUM ('GENERIC', 'GITHUB', 'STRIPE', 'SLACK', 'DISCORD');
```
abelanger5: `DISCORD` here seems like an error?

api/v1/server/handlers/v1/webhooks/receive.go
```diff
@@ -239,11 +254,104 @@ func (vr ValidationError) ToResponse() (gen.V1WebhookReceiveResponseObject, erro
...
+			}
+		}
+
+		algorithm := webhook.AuthHmacAlgorithm.V1IncomingWebhookHmacAlgorithm
+		encoding := webhook.AuthHmacEncoding.V1IncomingWebhookHmacEncoding
+		decryptedSigningSecret, err := w.config.Encryption.Decrypt(webhook.AuthHmacWebhookSigningSecret, "v1_webhook_hmac_signing_secret")
+
+		if err != nil {
+			return false, &ValidationError{
+				Code:      Http500,
```
abelanger5, line 320: Should this be a 500 response?

api/v1/server/handlers/v1/webhooks/receive.go
```diff
@@ -239,11 +254,104 @@ func (vr ValidationError) ToResponse() (gen.V1WebhookReceiveResponseObject, erro
...
+			}
+		}
+
+		sigBaseString := fmt.Sprintf("v0:%d:%s", timestamp, webhookPayload)
+
+		hash, err := computeHMACSignature([]byte(sigBaseString), decryptedSigningSecret, algorithm, encoding)
+
+		if err != nil {
+			return false, &ValidationError{
+				Code:      Http500,
```
abelanger5, line 331: Same here?

frontend/docs/pages/home/webhooks.mdx
```diff
@@ -5,7 +5,7 @@ import UniversalTabs from "@/components/UniversalTabs";
 
 # Webhooks
```
grutt, line 6: Let's add a Preview badge and a notification saying 

> This feature is currently in development and might change. Reach out for feedback or if you encounter any problems registering any external webhooks.

abelanger5 (changes requested):

sql/schema/v1-core.sql
mrkaye97: ah yeah thanks, forgot to remove this one when I was unwinding that

api/v1/server/handlers/v1/webhooks/receive.go
mrkaye97: I think so? same for the one below too (failing to compute the signature, although maybe that one is more debatable and could be e.g. 403)

abelanger5: ah gotcha, yeah this one looks like a correct 500, the one below does not (we'd fail if the signatures don't match right? definitely feels like a 403 instead)

mrkaye97: ah yeah sorry, we will fail and return 403 if the signatures don't match, but that error is just checking if we fail to compute the signature, not if it doesn't match the provided one

abelanger5: got it, seems reasonable then

abelanger5 (approved):

PR #2148, opened 2025-08-15
[hotfix] Move Go SDK v1 examples for now

mnafees:


mrkaye97 (approved):
good for now 👍 I think there was a different place inside of pkg where these used to go, but we can move them over there later

PR #2150, opened 2025-08-15
Fix: More examples / snippets fixes + rework

mrkaye97:
# Description

Simplifying the snippet generation a bunch. The basic idea here is:

1. I replaced the whole `snips` directory with a zero-dependency Python script that we can run wherever, which will just generate the snippets and examples and dump them out to the correct places (right now just the `docs` dir, but we can also put them in the app later if we want)
2. Instead of making a whole nested directory of all the snippets, it's now just one TS file
3. Changed the "generate and sync" setup where we'd write to `/snips` and then copy to `/app` and `/docs` to just write to the other places directly
4. The github URLs and such are created generation-side now, so the `Snippet` component + what we need to do on build is much simpler
5. Updated all the refs in the docs to use the new structure
6. Updated the GHA in CI to use a label to tag these PRs so we only ever have one open at a time, and to ping the person who's PR generated the docs to approve + merge

## Type of change


sdks/python/examples/dag/worker.py
```diff
@@ -38,6 +38,7 @@ async def step3(input: EmptyModel, ctx: Context) -> RandomSum:
     return RandomSum(sum=one + two)
 
 
+# 👀 cool thing!
```
mrkaye97: todo: check if this generates properly

abelanger5 (approved):
Nice! One minor thing, I actually liked the code path, I think it frames and contextualizes the component nicely. Is there a reason we can't have that with these changes?

Before:

Image

After: 

Image

mrkaye97:
> Nice! One minor thing, I actually liked the code path, I think it frames and contextualizes the component nicely. Is there a reason we can't have that with these changes?
> 
> Before:
> 
> Image
> After:
> 
> Image

oh, whoops - yeah, just missed that! I'll add it back, it must be an optional param or something

PR #2154, opened 2025-08-17
Go SDK `StartBlocking` takes a context, other fixes

mnafees:


sdks/go/client.go
```diff
@@ -93,12 +88,20 @@ func (c *Client) NewWorker(name string, options ...WorkerOption) (*Worker, error
...
 	for _, workflow := range regularWorkflows {
-		req, _, _, onFailureFn := workflow.Dump()
+		req, regularActions, _, onFailureFn := workflow.Dump()
 		err := nonDurableWorker.RegisterWorkflowV1(req)
 		if err != nil {
 			return nil, err
 		}
 
+		// Register all regular task actions
+		for _, namedFn := range regularActions {
```
abelanger5, line 105: Note that this still isn't exactly correct, since we're only looking at `regularWorkflows`, but a workflow can have both durable and non-durable tasks. So we're failing to register any non-durable tasks which are on a "durable" workflow. We shouldn't think of workflows as durable or non-durable, we should think of tasks as durable and non-durable.

abelanger5 (changes requested):

abelanger5 (approved):

PR #2155, opened 2025-08-17
fix: move rate limited queue items off the main queue

abelanger5:
# Description

Currently, items in a queue which are being rate-limited maintain their position in the queue. This means that in cases with heavy use of dynamic rate limits, the head of the queue gets blocked (up to the `SERVER_SINGLE_QUEUE_LIMIT`, which I believe is 100 by default). This results in many scheduling timeouts, even for items which are not necessarily being rate limited. While this behavior was intended with the v1 refactor, it's undocumented and unexpected thus resembles a bug. 

Some other drive-by fixes:
- Renames the confusingly named `package v2` in the v1 scheduler
- Adds some more traces/spans/attributes to a few components which can be helpful for tracing

## Type of change


cmd/hatchet-migrate/migrate/migrations/20250817183956_v1_0_37.sql
```diff
@@ -0,0 +1,45 @@
+-- +goose Up
+-- +goose StatementBegin
+-- v1_rate_limited_qis represents a queue item that has been rate limited and removed from the v1_queue_item table.
+CREATE TABLE v1_rate_limited_qis (
```
mrkaye97: nit: can we name this `v1_rate_limited_queue_items` instead for clarity?

cmd/hatchet-migrate/migrate/migrations/20250817183956_v1_0_37.sql
```diff
@@ -0,0 +1,45 @@
...
+    autovacuum_vacuum_threshold='25',
+    autovacuum_analyze_threshold='25',
+    autovacuum_vacuum_cost_delay='10',
+    autovacuum_vacuum_cost_limit='1000'
+);
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+DROP TABLE IF EXISTS v1_rate_limited_qis;
```
mrkaye97: nit: don't think we'll need `IF EXISTS` here

sql/schema/v1-core.sql
```diff
@@ -362,6 +362,7 @@ CREATE TABLE v1_task_expression_eval (
 );
 
 -- CreateTable
+-- NOTE: changes to v1_queue_item should be reflected in v1_rate_limited_qis
```
mrkaye97: 🙌

pkg/repository/v1/scheduler_queue.go
```diff
@@ -212,6 +215,29 @@ func (d *queueRepository) MarkQueueItemsProcessed(ctx context.Context, r *Assign
 		})
 	}
 
+	// remove rate limited queue items from the queue and place them in the v1_rate_limited_qis table
+	// we only do this if the requeue_after time is at least 2 seconds in the future, to avoid thrashing
+	qisToMoveToRateLimited := make([]int64, 0, len(r.RateLimited))
+	qisToMoveToRateLimitedRQAfter := make([]pgtype.Timestamptz, 0, len(r.RateLimited))
+
+	for _, row := range r.RateLimited {
+		if row.NextRefillAt != nil && row.NextRefillAt.UTC().After(time.Now().UTC().Add(2*time.Second)) {
```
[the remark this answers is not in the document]
mrkaye97: I sorta agree with this 🤷

mrkaye97 (approved):
looks great, nice idea!

PR #2159, opened 2025-08-18
Fix: Matches deadlock

mrkaye97:
# Description

Fixing a matches deadlock caused by `SaveSatisfiedMatchConditions` and `GetSatisfiedMatchConditions` locking `v1_match` and `v1_match_condition` in different orders

## Type of change


pkg/repository/v1/sqlcv1/matches.sql
```diff
@@ -97,21 +104,12 @@ WITH input AS (
         (v1_match_condition.v1_match_id, v1_match_condition.id) = (c.v1_match_id, c.id)
     RETURNING
         v1_match_condition.v1_match_id, v1_match_condition.id
-), distinct_match_ids AS (
-    SELECT
-        DISTINCT v1_match_id
```
abelanger5, line 102: I imagine this DISTINCT is important, since the tuples are (match_id, condition_id, data) we can definitely select multiple matches

abelanger5 (changes requested):

pkg/repository/v1/sqlcv1/matches.sql
mrkaye97: do you think making `locked_matches` use `SELECT DISTINCT` would work?

abelanger5: yes, or doing that out of the return statement should also work

mrkaye97: gotcha, can do! I was thinking it wouldn't really matter since we only use the resulting match ids to pass into an `=ANY(@matchIds)` in `SaveSatisfiedMatchConditions`, so I think the logic will be the same with or without the distinct and this would be more performant on this query (although potentially worse on the next query if there are dupes in the array)

not sure if I'm misunderstanding something here though, just what I thought from reading through initially

abelanger5: yeah looks like it might not matter downstream, should be good either way

abelanger5 (dismissed):

mrkaye97:
@abelanger5 closing this since I think it's superseded?

PR #2161, opened 2025-08-19
Fix: Check if workflow run is nil before cast

mrkaye97:
# Description

Fixes a panic

## Type of change


mrkaye97:
closing b/c I don't think this helps much

PR #2163, opened 2025-08-19
Feat: Run `ANALYZE` on a few tables once a day

mrkaye97:
# Description

Starting with these, figure we can add whatever other tables we want easily. `ANALYZE` should improve our query performance on the partitioned tables significantly

The reason for this is that, as it turns out, autovacuum does not affect the partitioned table itself, only the partitions. This was resulting in the planner making bad choices for how to plan + execute queries (especially ones involving `UNNEST` with many parameters where we'd need to scan many partitions), ending up significantly mis-estimating how many rows different CTEs would return and similar, causing very poor performance. The idea here is that running an analyze would improve these estimations, leading to much improved query performance

## Type of change

### DIFF
--- a/internal/services/controllers/v1/olap/analyze.go
+++ b/internal/services/controllers/v1/olap/analyze.go
@@ -1,0 +1,28 @@
+package olap
+
+import (
+	"context"
+)
+
+func (oc *OLAPControllerImpl) runAnalyze(ctx context.Context) func() {
+	return func() {
+		oc.l.Debug().Msgf("analyze: running analyze on partitioned tables")
+
+		tenant, err := oc.p.GetInternalTenantForController(ctx)
+
+		if err != nil {
+			oc.l.Error().Err(err).Msg("could not get internal tenant")
+			return
+		}
+
+		if tenant == nil {
+			return
+		}
+
+		err = oc.repo.OLAP().AnalyzeOLAPTables(ctx)
+
+		if err != nil {
+			oc.l.Error().Err(err).Msg("could not analyze OLAP tables")
+		}
+	}
+}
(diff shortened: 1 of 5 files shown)


pkg/repository/v1/sqlcv1/olap.sql
```diff
@@ -16,6 +16,15 @@ SELECT
...
 
+-- name: AnalyzeV1RunsOLAP :exec
+ANALYZE v1_runs_olap;
+
+-- name: AnalyzeV1TasksOLAP :exec
+ANALYZE v1_tasks_olap;
+
+-- name: AnalyzeV1DAGsOLAP :exec
+ANALYZE v1_dags_olap;
+
```
mrkaye97, line 27: I was thinking we'd want to run these separately to limit how much time each takes, but can combine them too

internal/services/controllers/v1/olap/analyze.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"context"
+)
+
+func (oc *OLAPControllerImpl) runAnalyze(ctx context.Context) func() {
+	return func() {
+		oc.l.Debug().Msgf("analyze: running analyze on partitioned tables")
+
+		tenant, err := oc.p.GetInternalTenantForController(ctx)
```
abelanger5: Note that we are moving away from using the internal tenant logic and instead favoring advisory locks on the database level, because the internal tenant risks duplicate operations running when the internal tenant moves between controllers.

abelanger5 (changes requested):

internal/services/controllers/v1/olap/analyze.go
mrkaye97: gotcha, thanks! you're thinking the strategy here should be to acquire an advisory lock, run these, and then release it?

abelanger5: yes. safest if we can do this in the same TX as the analyze, relatively safe if the advisory lock has a statement timeout if we need to do this in multiple txs

abelanger5 (approved):
💯

PR #2169, opened 2025-08-20
Use port 7070 for internal GRPC broadcast address

andersthuesen:
# Description

Use port 7070 for internal GRPC broadcast address

## Type of change


abelanger5 (approved):
Thanks @andersthuesen! This looks correct.

PR #2171, opened 2025-08-20
feat: analytics events

grutt:
# Description

Adds additional analytics events for tracking user and tenant activation

## Type of change

### DIFF
--- a/pkg/analytics/analytics.go
+++ b/pkg/analytics/analytics.go
@@ -1,17 +1,17 @@
 package analytics
 
 type Analytics interface {
-	// Enqueue queues an analytics event for processing
-	// @param event The name of the event to track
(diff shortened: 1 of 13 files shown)


pkg/analytics/analytics.go
```diff
@@ -1,12 +1,23 @@
 package analytics
 
 type Analytics interface {
-	Enqueue(event string, userId string, tenantId *string, data map[string]interface{})
+	// Enqueue queues an analytics event for processing
+	// @param event The name of the event to track
```
mnafees: These seem like JavaDoc comments

abelanger5 (approved):
Agreed with Nafees that the javadoc-style comments aren't particularly idiomatic and seem unnecessary

PR #2172, opened 2025-08-20
Misc. Python Fixes

mrkaye97:
# Description

### Added

- Adds back an optional `cel-python` dependency for v0 compatibility, allowing users to use the v0 client with the v0-compatible features in the SDK.
- Adds `dependencies` to the `mock_run` methods on the `Standalone`.
- Removes `aiostream` dependency that was unused.
- Removes `aiohttp-retry` dependency that was unused.

## Type of change

### DIFF
--- a/sdks/python/pyproject.toml
+++ b/sdks/python/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "hatchet-sdk"
-version = "1.17.1"
+version = "1.17.2"
 description = ""
 authors = ["Alexander Belanger <alexander@hatchet.run>"]
 readme = "README.md"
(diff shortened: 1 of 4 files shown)


sdks/python/pyproject.toml
```diff
@@ -21,18 +21,21 @@ protobuf = "^5.29.5"
...
 opentelemetry-sdk = { version = "^1.28.0", optional = true }
 opentelemetry-instrumentation = { version = ">=0.49b0", optional = true }
 opentelemetry-distro = { version = ">=0.49b0", optional = true }
 opentelemetry-exporter-otlp = { version = "^1.28.0", optional = true }
 opentelemetry-exporter-otlp-proto-http = { version = "^1.28.0", optional = true }
-prometheus-client = ">=0.21.1"
-pydantic-settings = "^2.7.1"
+
+## CEL Extra (for v0 compatibility)
+cel-python = { version = "^0.4.0", optional = true }
```
mrkaye97, line 38: turned out that removing this was a breaking change for v0 (no linters there), but I don't want to re-add it b/c it causes dependency resolution issues, so just adding it as an extra for now and we can remove this in a few weeks when we nuke the v0 code

abelanger5 (approved):

PR #2180, opened 2025-08-21
[hotfix] Disallow `nil` input in Go SDK

mnafees:
# Description

Disallow `nil` input for task.

## Type of change

### DIFF
--- a/sdks/go/workflow.go
+++ b/sdks/go/workflow.go
@@ -329,6 +329,14 @@ func (t *Task) GetName() string {
 //
 // Function signatures are validated at runtime using reflection.
 func (w *Workflow) NewTask(name string, fn any, options ...TaskOption) *Task {
+	if name == "" {
+		panic("task name cannot be empty")
+	}
+
+	if fn == nil {
+		panic("task '" + name + "' has a nil input function")
+	}
+
 	config := &taskConfig{}
 
 	for _, opt := range options {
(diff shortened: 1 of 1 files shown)


sdks/go/workflow.go
```diff
@@ -339,10 +347,10 @@ func (w *Workflow) NewTask(name string, fn any, options ...TaskOption) *Task {
 	fnType := fnValue.Type()
 
 	if fnType.Kind() != reflect.Func {
-		panic("task function must be a function")
+		panic("task input must be a function")
 	}
 	if fnType.NumIn() != 2 {
-		panic("task function must have exactly 2 parameters: (ctx Context, input T)")
+		panic("task input must have exactly 2 parameters: (ctx hatchet.Context, input T)")
```
mrkaye97: not sure I follow the reason for this change - shouldn't this be the function?

sdks/go/workflow.go
```diff
@@ -356,10 +364,10 @@ func (w *Workflow) NewTask(name string, fn any, options ...TaskOption) *Task {
 	fnType := fnValue.Type()
 
 	if fnType.Kind() != reflect.Func {
-		panic("task function must be a function")
+		panic("task input must be a function")
```
mrkaye97: same here I think 😅 - sorry! agree the naming is a bit confusing though. I'd usually say "`fn` must be a function" if we really need the `any`, but is it really not possible to not use `any` here? I'm a bit surprised by that

mrkaye97: e.g. can we do:

```
func (w *Workflow) NewTask(name string, fn func(any, hatchet.Context) map[string]any, options ...TaskOption) *Task {
```

at least?

mrkaye97 (approved):

PR #2181, opened 2025-08-21
Add tenant prometheus metric to add reason for task scheduling time out

mnafees:
# Description

A new tenant metric called `hatchet_tenant_task_scheduling_timed_out` has been added which will contain the workflow and task names with the reason for why a task had a scheduling time out 
- `no workers available`
- `rate limited`

## Type of change


pkg/scheduling/v1/queuer.go
```diff
@@ -54,24 +68,43 @@ func newQueuer(conf *sharedConfig, tenantId pgtype.UUID, queueName string, s *Sc
 		defaultLimit = conf.singleQueueLimit
 	}
 
+	cacheSize := 1000
+	if conf.singleQueueLimit > 1000 {
+		cacheSize = conf.singleQueueLimit
+	}
```
abelanger5: The `singleQueueLimit` has nothing to do with the actual number of queues, it's the amount of rows we read out of each queue, so it doesn't make sense to have these values be correlated

pkg/scheduling/v1/queuer.go
```diff
@@ -410,6 +497,7 @@ func (q *Queuer) ack(r *assignResults) {
 	for _, rateLimitedItem := range r.rateLimited {
```
abelanger5, line 500: Since we are now making these statements dependent on ordering (i.e. we must process `r.rateLimited` last for this to work), can we add a warning here to indicate that the ordering shouldn't be changed?

abelanger5 (changes requested):

abelanger5 (changes requested):
@mnafees there have been some changes to the `queuer.go` files since this PR. In particular adding the rate limits which are requeued in a separate table. Could you merge with main and take a close look at the changes to ensure correctness?

mnafees:
Closing this due to staleness. Might revisit later.

PR #2183, opened 2025-08-21
[FE]: Docs-in-app, Part I

mrkaye97:
# Description

Bringing back docs-in-app + the in-app side panel. Basically opens the side panel in the places where we'd open the sheet before. The side panel is resizeable (and the size you drag to will save), and it can display (right now) event details, task run details, and docs. 

Once this goes in, we'll need to go through and add docs links to all of the empty states everywhere + anywhere else it's appropriate, and I'll also probably try to replace some more modals with side panels.

## Type of change

#### Open

Screenshot 2025-08-28 at 3 16 46 PM

#### Dragged wider, metrics + actions scrolling

Screenshot 2025-08-28 at 3 17 02 PM

#### Empty State, Docs Button

Screenshot 2025-08-28 at 3 17 26 PM

#### Docs view

Screenshot 2025-08-28 at 3 19 44 PM

## Events (will do more other places later)

Screenshot 2025-08-28 at 3 23 49 PM

TODO:

- [x] Fix overflow states (scrolling?) on toolbar filters, table widths, side panel header, etc.
- [x] Look for other modals to replace with side panel

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,6 +13,12 @@ jobs:
     steps:
       - name: Clone repository
         uses: actions/checkout@v5
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
+      - name: Generate docs snippets and examples
+        working-directory: frontend/snippets
+        run: python3 generate.py
       - name: Build frontend
         run: docker build -f ./build/package/frontend.dockerfile .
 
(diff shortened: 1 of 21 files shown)


.github/workflows/build.yml
```diff
@@ -85,6 +91,12 @@ jobs:
     steps:
       - name: Clone repository
         uses: actions/checkout@v5
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
+      - name: Generate docs snippets and examples
+        working-directory: frontend/snippets
+        run: python3 generate.py
```
mrkaye97, line 99: needing to copy this is a bit unfortunate but not the end of the world imo

abelanger5 (approved):

PR #2185, opened 2025-08-22
Add `Into()` to extract task result and `StandaloneTask`s

mnafees:
# Description

Introduces `WorkflowResult`, `TaskResult` and `TaskResult.Into()`

## Type of change


abelanger5:
See #2198 for another example of a type conversion bug, not sure if it's entirely related or not - but I'd really advocate that we just use `encoding/json` for all of this.

abelanger5 (approved):

PR #2186, opened 2025-08-22
feat:  improved onboarding part 1

grutt:
# Description

Adds attribution, onboarding questions, and tenant environments.

## Type of change


api-contracts/openapi/components/schemas/tenant.yaml
```diff
@@ -67,6 +70,13 @@ TenantAlertingSettings:
     - maxAlertingFrequency
   type: object
 
+TenantEnvironment:
```
mrkaye97, line 74: nit: I've been liking adding comments here + in the corresponding SQL file saying something like `IMPORTANT: Keep these values in sync with {{ the other place }}`

cmd/hatchet-migrate/migrate/migrations/20250821131856_v1_0_39.sql
```diff
@@ -0,0 +1,17 @@
+-- +goose Up
+-- +goose StatementBegin
+-- enum for environment
+CREATE TYPE "TenantEnvironment" AS ENUM ('local', 'development', 'production');
+
+ALTER TABLE "Tenant"
+    ADD COLUMN "onboardingData" JSONB,
```
mrkaye97, line 7: nit: not sure if we have a preference, but I've been liking making these JSON columns non-null with a default of `{}`, but not sure if that's easier to reason about than `null` here 🤷

frontend/app/src/pages/onboarding/create-tenant/components/hear-about-us-form.tsx
```diff
@@ -0,0 +1,132 @@
...
+    { value: 'linkedin', label: 'LinkedIn', icon: FaLinkedin },
+    { value: 'twitter', label: 'Twitter', icon: FaTwitter },
+    { value: 'blog', label: 'Blog/Article', icon: FileText },
+    { value: 'colleague', label: 'Word of Mouth', icon: Users },
+    { value: 'github', label: 'GitHub', icon: Github },
+    { value: 'conference', label: 'Event', icon: Calendar },
+    { value: 'other', label: 'Other', icon: HelpCircle },
+  ];
+
+  // Convert value to array if it's a string for backward compatibility
```
mrkaye97: what's the backwards compat here? not sure I'm understanding this, since it's net-new

frontend/app/src/pages/onboarding/create-tenant/components/step-progress.tsx
```diff
@@ -0,0 +1,44 @@
+import { Badge } from '@/components/ui/badge';
+
+export function StepProgress({
+  steps,
+  currentStep,
+  onStepClick,
+}: {
+  steps: any[];
```
mrkaye97: is it possible to not use `any` here? or if we only use it to index over, maybe we can pass `numSteps` as a `number` instead?

frontend/app/src/pages/onboarding/create-tenant/components/tenant-create-form.tsx
```diff
@@ -1,132 +1,239 @@
...
+    switch (environment) {
+      case 'local':
+        return '-local';
+      case 'development':
+        return '-dev';
+      case 'production':
+        return '-prod';
+      default:
+        return '-dev'; // Default to dev if no environment selected
+    }
```
mrkaye97: is the environment here really one of those enum values? or can it actually be anything? if it's an enum, I think we should use the enum type here + have one of those `exhaustivenessCheck` things I've been putting other places

api-contracts/openapi/components/schemas/tenant.yaml
```diff
@@ -67,6 +70,13 @@ TenantAlertingSettings:
...
+  enum:
+    - "local"
+    - "development"
+    - "production"
```
mrkaye97, line 78: do we want `staging` or `qa` or anything else? I think one possible footgun here is people will want to call their environments different things

frontend/app/src/pages/onboarding/create-tenant/components/tenant-create-form.tsx
```diff
@@ -1,132 +1,239 @@
...
+      label: 'Local Dev',
+      icon: Monitor,
+      description: 'Testing and development on your local machine',
+    },
+    {
+      value: 'development',
+      label: 'Development',
+      icon: Settings,
+      description: 'Shared development environment or staging',
+    },
```
mrkaye97, line 151: this is a bit confusing to me - I'd think of `local` and `dev` as being the same, and `staging` as separate (likely also depends on the team)

frontend/app/src/pages/onboarding/create-tenant/components/what-building-form.tsx
```diff
@@ -0,0 +1,139 @@
...
+      value: 'internal-automations',
+      label: 'Internal Automations',
+      icon: Workflow,
+    },
+    { value: 'webhooks', label: 'Webhooks', icon: Webhook },
+    { value: 'other', label: 'Other', icon: HelpCircle },
+  ];
+
+  // Convert value to array if it's a string for backward compatibility
+  const selectedValues = Array.isArray(value) ? value : value ? [value] : [];
```
mrkaye97, line 46: same qq here re: backwards compat

frontend/app/src/pages/onboarding/create-tenant/components/what-building-form.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+  // Check if "other" is selected and extract the custom value
+  const otherSelection = selectedValues.find((v) => v.startsWith('other'));
+  const isOtherSelected = !!otherSelection;
+  const otherValue = otherSelection
+    ? otherSelection.replace('other: ', '')
+    : '';
```
mrkaye97: this all looks pretty similar to the other component - can we factor it out?

frontend/app/src/pages/onboarding/create-tenant/index.tsx
```diff
@@ -1,17 +1,52 @@
...
+import { HearAboutUsForm } from './components/hear-about-us-form';
+import { WhatBuildingForm } from './components/what-building-form';
+import { StepProgress } from './components/step-progress';
+import { OnboardingStepConfig, OnboardingFormData } from './types';
 
 export default function CreateTenant() {
+  const [searchParams] = useSearchParams();
+  const navigate = useNavigate();
+
+  // Get step from URL parameter, default to 0
```
mrkaye97: nit: don't think we need this comment

frontend/app/src/pages/onboarding/create-tenant/index.tsx
```diff
@@ -20,6 +55,20 @@ export default function CreateTenant() {
...
+        (window as any).posthog.capture('onboarding_completed', {
+          hear_about_us: Array.isArray(formData.hearAboutUs)
+            ? formData.hearAboutUs.join(', ')
+            : formData.hearAboutUs,
+          what_building: Array.isArray(formData.whatBuilding)
+            ? formData.whatBuilding.join(', ')
+            : formData.whatBuilding,
+          tenant_id: tenant.data.metadata.id,
+        });
+      }
```
mrkaye97: I've had a lot of success in the past making a hook called `useAnalytics` or something that just has a `trackEvent` function it exports, which takes e.g. the event name + payload - figure that might be helpful here so we can worry about the `window` stuff in one place and not need to duplicate it all over

mrkaye97 (commented):
left some comments! looks good broadly, nice work on this 🙌

api-contracts/openapi/components/schemas/tenant.yaml
grutt: good call

cmd/hatchet-migrate/migrate/migrations/20250821131856_v1_0_39.sql
grutt: i think null is fine in this case

frontend/app/src/pages/onboarding/create-tenant/components/step-progress.tsx
grutt: made a small change to be a bit more clear, but i actually think that this might be the only acceptable use of any since we literally expect steps to be _any_ array.

frontend/app/src/pages/onboarding/create-tenant/components/tenant-create-form.tsx
grutt: fair.. the intent is marking local tenants per dev vs shared dev... any ideas on how to make this more clear?

mrkaye97: not really 😅, I sorta think this will devolve into people wanting to use their own names, but I'd think that allowing for local / dev / staging / qa / prod might cover 90% of cases 🤷

WDYT?

frontend/app/src/pages/onboarding/create-tenant/components/what-building-form.tsx
grutt: good call

frontend/app/src/pages/onboarding/create-tenant/index.tsx
grutt: :P i think i had this for next, added something simple here

mrkaye97 (approved):
🥳

PR #2190, opened 2025-08-22
fix: match and cancel newest/in progress deadlocks

abelanger5:
# Description

Builds off the deadlock fixes in #2159 but with a different strategy for resolving `RunParentCancelNewest/RunParentCancelInProgress` deadlocking with `CreateTasks` on the `v1_workflow_concurrency_slot` table. 

The crux of the issue is that `CreateTasks` is called for each distinct step ID separately, because the `pgx` driver panics when multi-dimensional arrays are called with a different dimension between entries (still not sure why). 

Although we are correctly ordering `v1_workflow_concurrency_slot` in the trigger that gets called after `CreateTasks`, because we are effectively ordering by the `step_id` _first_, we are placing locks in different orders between `RunParentCancelNewest/RunParentCancelInProgress` and the transaction which is inserting tasks. 

Why are we locking `v1_workflow_concurrency_slot` after `CreateTasks`? When a task is created in a non-`QUEUED` state (for example, created in a cancelled state), we might need to release the workflow-level concurrency slot. 

This PR moves the logic for cleaning up the `v1_workflow_concurrency_slot` to a query which runs after all of the `CreateTask` queries have executed. 

It also explicitly locks all `v1_workflow_concurrency_slot` rows which are being processed by the concurrency strategy. We have effectively been doing this, but the way it is implemented is another source of potential deadlocking, because we were locking these in two batches: the slots which were going to be cancelled, and the ones which were going to be run. This leads to inconsistent ordering from the `CreateTasks` locking order which could also cause a deadlock. Instead we lock them all upfront. 

## Type of change

### DIFF
--- a/cmd/hatchet-migrate/migrate/migrations/20250822193023_v1_0_39.sql
+++ b/cmd/hatchet-migrate/migrate/migrations/20250822193023_v1_0_39.sql
@@ -1,0 +1,304 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION v1_task_insert_function()
+RETURNS TRIGGER AS $$
+DECLARE
+    rec RECORD;
+BEGIN
+    WITH new_slot_rows AS (
+        SELECT
+            id,
+            inserted_at,
+            retry_count,
+            tenant_id,
+            priority,
+            concurrency_parent_strategy_ids[1] AS parent_strategy_id,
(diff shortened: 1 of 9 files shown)


mrkaye97 (approved):
LGTM, tested locally w/ cancel newest + cancel in progress, both task and workflow level just to see

PR #2195, opened 2025-08-25
Feat: Worker slot Prom metrics

mrkaye97:
# Description

Adds worker slot data (used, total) to prom metrics

## Type of change

To do:

- [x] Test it
- [ ] Add docs

### DIFF
--- a/pkg/integrations/metrics/prometheus/tenant.go
+++ b/pkg/integrations/metrics/prometheus/tenant.go
@@ -23,6 +23,7 @@ const (
 	TenantCancelledTasksTotal          TenantHatchetMetric = "hatchet_tenant_cancelled_tasks"
 	TenantReassignedTasksTotal         TenantHatchetMetric = "hatchet_tenant_reassigned_tasks"
 	TenantUsedWorkerSlotsTotal         TenantHatchetMetric = "hatchet_tenant_used_worker_slots"
+	TenantAvailableWorkerSlotsTotal    TenantHatchetMetric = "hatchet_tenant_available_worker_slots"
 	TenantWorkerSlotsTotal             TenantHatchetMetric = "hatchet_tenant_worker_slots"
 )
 
@@ -141,6 +142,14 @@ var (
 		prometheus.GaugeOpts{
 			Name: string(TenantUsedWorkerSlotsTotal),
 			Help: "Number of used slots per worker pool",
+		},
+		[]string{"worker"},
+	)
+
+	TenantAvailableWorkerSlots = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: string(TenantAvailableWorkerSlotsTotal),
+			Help: "Number of available slots per worker pool",
 		},
 		[]string{"worker"},
 	)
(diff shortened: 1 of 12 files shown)


pkg/integrations/metrics/prometheus/tenant.go
```diff
@@ -126,4 +129,28 @@ var (
 		Name: string(TenantReassignedTasksTotal),
 		Help: "The total number of tasks that were reassigned to a worker",
 	}, []string{"tenant_id"})
+
+	TenantWorkerSlots = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: string(TenantWorkerSlotsTotal),
+			Help: "Total number of slots per worker pool",
+		},
+		[]string{"worker_id", "worker_name"},
```
mnafees: We should add the `tenant_id` label here which is used by the prom metrics endpoint to filter over.

pkg/integrations/metrics/prometheus/tenant.go
```diff
@@ -126,4 +129,28 @@ var (
...
+	)
+
+	TenantUsedWorkerSlots = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: string(TenantUsedWorkerSlotsTotal),
+			Help: "Number of used slots per worker pool",
+		},
+		[]string{"worker_id", "worker_name"},
```
mnafees: Same label here as well

pkg/integrations/metrics/prometheus/tenant.go
```diff
@@ -126,4 +129,28 @@ var (
...
+	)
+
+	TenantAvailableWorkerSlots = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: string(TenantAvailableWorkerSlotsTotal),
+			Help: "Number of available slots per worker pool",
+		},
+		[]string{"worker_id", "worker_name"},
```
mnafees: And here

pkg/scheduling/v1/scheduler.go
```diff
@@ -874,10 +876,12 @@ func (s *Scheduler) getSnapshotInput(mustSnapshot bool) (*SnapshotInput, bool) {
 				workerSlotUtilization[workerId] = &SlotUtilization{
 					UtilizedSlots:    0,
 					NonUtilizedSlots: 0,
+					TotalSlots:       0,
 				}
 			}
 
 			uniqueSlots[slot] = true
+			workerSlotUtilization[workerId].TotalSlots++
```
mnafees: Do we need to store total slots at all? Can we simply just pass the total of utilized + non-utilized to the prom metric?

mnafees (changes requested):

pkg/integrations/metrics/prometheus/tenant.go
mrkaye97: nice, thanks! will fix that

pkg/scheduling/v1/scheduler.go
mrkaye97: happy to remove this if that's good for you!

mnafees (approved):

PR #2204, opened 2025-08-26
feat : soft delete tenant from UI

grutt:
# Description

Tenant owners can delete tenants

## Type of change

## What's Changed

- [x] Adds UI affordance to delete tenant from the overview screen
- [x] API endpoint to delete the tenant
- [x] SQL filters for soft-deleted tenants


mrkaye97 (commented):
I'm a little nervous about this because of how many places we e.g. join to the `Tenant` table, etc. - is there a reason for preferring soft deletes over hard deletes here?

grutt:
1. i do want to have some "undo" period baked in for the worst case
2. we've had issues in the past cleaning up orphaned relations so we've gone the soft delete route

I don't think there's significant risk with filter, but I can go through to ensure i didn't miss anything

abelanger5:
Closing since we have the archiving feature, we'll need to rethink this with organizations - @mnafees can you take a look at some point?

PR #2207, opened 2025-08-26
fix: remove rate limited items from in memory buffer

abelanger5:
# Description

Removes queue items which have been rate limited from the in-memory buffer in `queuer.go`. 

## Type of change

### DIFF
--- a/pkg/scheduling/v1/queuer.go
+++ b/pkg/scheduling/v1/queuer.go
@@ -225,7 +225,7 @@ func (q *Queuer) loopQueue(ctx context.Context) {
 
 				countMu.Lock()
 				count += numFlushed
-				processedQiLength += len(ar.assigned) + len(ar.unassigned) + len(ar.schedulingTimedOut) + len(ar.rateLimited)
+				processedQiLength += len(ar.assigned) + len(ar.unassigned) + len(ar.schedulingTimedOut) + len(ar.rateLimited) + len(ar.rateLimitedToMove)
 				countMu.Unlock()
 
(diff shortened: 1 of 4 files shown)


pkg/scheduling/v1/queuer.go
```diff
@@ -407,6 +407,11 @@ func (q *Queuer) ack(r *assignResults) {
 		delete(q.unassigned, schedulingTimedOutItem.ID)
 	}
 
+	for _, rateLimitedItemToMove := range r.rateLimitedToMove {
+		delete(q.unacked, rateLimitedItemToMove.qi.ID)
+		delete(q.unassigned, rateLimitedItemToMove.qi.ID)
+	}
```
abelanger5, line 413: This is the crucial fix - we needed delete the items from `q.unassigned` so they aren't just read in again on requeue

pkg/scheduling/v1/rate_limit.go
```diff
@@ -11,7 +11,7 @@ import (
 	v1 "github.com/hatchet-dev/hatchet/pkg/repository/v1"
 )
 
-const MAX_RATE_LIMIT_UPDATE_FREQUENCY = 990 * time.Millisecond // avoid boundary conditions on 1 second polls
+const MAX_RATE_LIMIT_UPDATE_FREQUENCY = 500 * time.Millisecond // avoid boundary conditions on 1 second polls
```
abelanger5, line 14: Drive by: noticed issues in the 1-second rate limits and lowering this should fix it.

mrkaye97 (approved):

PR #2222, opened 2025-08-28
Feat: Add bulk log ingest gRPC endpoint

mrkaye97:
# Description

Adding a bulk log ingest endpoint so we can buffer + flush server side instead of making a zillion requests. Also removed all my old debug logs. Will need to make a corresponding change in Cloud for Loki

## Type of change


mrkaye97:
closing, can do this later

PR #2227, opened 2025-08-29
Fix: Partition Core Tables

mrkaye97:
# Description

Adding three migrations for partitioning existing core tables (`v1_dag_data`, `v1_dag_to_task`, and `v1_lookup_table`).

For the lookup table, the steps are:

1. Create the new partitioned table and its partitions + a trigger to insert into it
2. Backfill the old data
3. Cut over

It's a bit different than with Timescale since we need to create the partitions manually. I was thinking it'd make sense to just have one big partition for all the old data, which will all get dropped in one shot after a week or so, but we could also loop through and create many. Definitely makes me nervous to need the three steps in this one since it's really not easily rerunnable / reversible and it's in the OSS

For the other two, we just attach the existing data as a partition to a new table, which should be pretty fast

## Type of change


cmd/hatchet-migrate/migrate/migrations/20250829173445_v1_0_41.sql
```diff
@@ -0,0 +1,526 @@
...
+            autovacuum_vacuum_scale_factor = ''0.1'',
+            autovacuum_analyze_scale_factor=''0.05'',
+            autovacuum_vacuum_threshold=''25'',
+            autovacuum_analyze_threshold=''25'',
+            autovacuum_vacuum_cost_delay=''10'',
+            autovacuum_vacuum_cost_limit=''1000''
+        )', newTableName);
+    EXECUTE
+        format('ALTER TABLE %s ATTACH PARTITION %s FOR VALUES FROM (''%s'') TO (''%s'')', targetTableName, newTableName, startDateStr, endDateStr);
+END$$;
```
mrkaye97: Important: I don't know of a way to make this all `IF NOT EXISTS`-able, so we won't be able to rerun this safely AFAICT

cmd/hatchet-migrate/migrate/migrations/20250829173445_v1_0_41.sql
```diff
@@ -0,0 +1,526 @@
...
+DO $$
+DECLARE
+    startDateStr varchar;
+    endDateStr varchar;
+    targetTableName CONSTANT varchar := 'v1_lookup_table_partitioned';
+    newTableName varchar;
+BEGIN
+    SELECT '19700101' INTO startDateStr;
+    SELECT TO_CHAR(date_trunc('week', (NOW() - INTERVAL '8 days')::DATE), 'YYYYMMDD') INTO endDateStr;
+    SELECT LOWER(FORMAT('%s_%s', targetTableName, startDateStr)) INTO newTableName;
```
abelanger5: Just to confirm, anything older than 8 days goes into the same partition?

cmd/hatchet-migrate/migrate/migrations/20250829173445_v1_0_41.sql
abelanger5: I think it's safe to check for whether we have the table already before `CREATE TABLE IF NOT EXISTS`, and if we do just return early? since these statements are all committed in the same tx

mrkaye97: yeah, that would work! how about the partitioning though? I guess we could query the info schema?

cmd/hatchet-migrate/migrate/migrations/20250829173445_v1_0_41.sql
mrkaye97: yeah this was my version of "from the beginning of time until last week" just to make sure we catch everything

mrkaye97:
@abelanger5 I need to do another self review here but I think these should be good to take a look through at this point at least

mrkaye97:
closing this since it's way stale

PR #2237, opened 2025-09-02
`TaskOutput` takes in a string arg

mnafees:
# Description

Go SDK `TaskOutput()` takes a string for ease of use when tasks are defined in other places.

## Type of change


mrkaye97 (approved):
I think this is fine but also a bit of a bummer - it's really nice in Python / TS to get type safety by passing the `Task` itself, but that also relies on generics and such so maybe it's a no-op here anyways 🤷

PR #2241, opened 2025-09-03
Fix: Partition core `v1_dag_to_task`

mrkaye97:
# Description

Partitioning `v1_dag_to_task` - this one was much simpler, since we can just create a partition using all of the existing data and attach it to the new table

## Type of change


cmd/hatchet-migrate/migrate/migrations/20250903152758_v1_0_41.sql
```diff
@@ -0,0 +1,56 @@
...
+    task_inserted_at TIMESTAMPTZ NOT NULL,
+    PRIMARY KEY (dag_id, dag_inserted_at, task_id, task_inserted_at)
+);
+
+INSERT INTO v1_dag_to_task_original
+SELECT * FROM v1_dag_to_task;
+
+DROP TABLE v1_dag_to_task;
+ALTER TABLE v1_dag_to_task_original RENAME TO v1_dag_to_task;
+ALTER INDEX v1_dag_to_task_original_pkey RENAME TO v1_dag_to_task_pkey;
```
mrkaye97, line 72: this is mostly just for me testing on local

mrkaye97:
@abelanger5 any idea how long `ATTACH PARTITION` might take? I didn't find much online about it - I think it'll need to revalidate any constraints, so I'm not 100% sure this is safe to do

cmd/hatchet-migrate/migrate/migrations/20250903152758_v1_0_41.sql
```diff
@@ -0,0 +1,57 @@
...
+CREATE TABLE v1_dag_to_task_partitioned (
+    dag_id BIGINT NOT NULL,
+    dag_inserted_at TIMESTAMPTZ NOT NULL,
+    task_id BIGINT NOT NULL,
+    task_inserted_at TIMESTAMPTZ NOT NULL,
+    PRIMARY KEY (dag_id, dag_inserted_at, task_id, task_inserted_at)
+) PARTITION BY RANGE(dag_inserted_at);
+
+SELECT create_v1_range_partition('v1_dag_to_task_partitioned', NOW()::DATE);
+SELECT create_v1_range_partition('v1_dag_to_task_partitioned', (NOW() + INTERVAL '1 day')::DATE);
```
abelanger5: Shouldn't there be corresponding code changes to maintain this partition?

Also: would weekly partitions make sense here, or daily?

mrkaye97: yup, thanks 🤦 keep forgetting this when going from cloud to the oss

re: weekly vs daily - I guess weekly is fine? it'd be easier on the planner

cmd/hatchet-migrate/migrate/migrations/20250903152758_v1_0_41.sql
```diff
@@ -0,0 +1,66 @@
...
+            new_table_name,
+            TO_CHAR((NOW() + INTERVAL '1 day')::DATE, 'YYYYMMDD')
+        );
+END $$;
+
+ALTER TABLE v1_dag_to_task_partitioned RENAME TO v1_dag_to_task;
+ALTER INDEX v1_dag_to_task_partitioned_pkey RENAME TO v1_dag_to_task_pkey;
+
+SELECT create_v1_range_partition('v1_dag_to_task', (NOW() + INTERVAL '1 day')::DATE);
+ANALYZE v1_dag_to_task;
```
abelanger5, line 53: Let's move this analyze to above the `ALTER TABLE` (and do it on the `v1_dag_to_task_partitioned` table) and split into multiple statements?

abelanger5 (changes requested):

cmd/hatchet-migrate/migrate/migrations/20250903152758_v1_0_41.sql
mrkaye97: yep, will do!

mrkaye97: argh - so the idea here was to run this one in one tx, since we just need to attach the old data as a partition. maybe we can run the analyze as a one-off before? we could also do the four separate statements thing, just much more complicated

cmd/hatchet-migrate/migrate/migrations/20250903152758_v1_0_41.sql
```diff
@@ -0,0 +1,66 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE v1_dag_to_task_partitioned (
```
abelanger5, line 6: Are we forgetting to run the ANALYZE cron on this parent partition table?

Also, mentioned this elsewhere but let's make that ANALYZE cron run more frequently, like once per hour.

cmd/hatchet-migrate/migrate/migrations/20250903152758_v1_0_41.sql
abelanger5: yep running it as a one-off before makes sense to me

abelanger5 (changes requested):

cmd/hatchet-migrate/migrate/migrations/20250903152758_v1_0_41.sql
mrkaye97: yep, alright will add that and will look for other partitioned tables to analyze too

mrkaye97: alright so I added an analyze here and also for a handful of the other tables (set it up to run once every three hours for now)

also removed the analyze from the migration so the tx takes less time

abelanger5: oh sorry, I do think we still need analyze, just not in the main tx. perhaps the easiest path is just to run it in a separate tx after the migration?

mrkaye97: I was thinking we'd just run it manually, but if you want, we can split this up into multiple statements and manually begin / commit each - WDYT?

abelanger5: yeah let's do the latter

mrkaye97:
merging these together

PR #2258, opened 2025-09-07
feat: optimistic scheduling

abelanger5:
# Description

Features:
- Made some improvements to listening for a task completed event for a single-task workflow by hooking into an existing tenant message `task-completed`. We can similarly add `task-failed` and `task-cancelled` in the future. 

Drawbacks:
- Increases the complexity of scheduling as the paths for optimistic scheduling are quite different from the regular path, since we protect everything with a single transaction

Limitations:
- This won't be turned on in HA mode and as `n` engines are horizontally scales the chances of optimistic scheduling reduce by `1/n` - we only use local schedulers when they have a lease on a tenant. We will need to build out a sticky load balancing strategy to take advantage of optimistic scheduling in HA setups. 

## Type of change


cmd/hatchet-engine/engine/run.go
```diff
@@ -878,6 +888,75 @@ func runV1Config(ctx context.Context, sc *server.ServerConfig) ([]Teardown, erro
...
+			return nil, fmt.Errorf("could not start scheduler (v1): %w", err)
+		}
+
+		teardown = append(teardown, Teardown{
+			Name: "schedulerv1",
+			Fn:   cleanup,
+		})
+
+		localScheduler = sv1
+	}
```
abelanger5, line 958: Just moved this up and added the `localScheduler` assignment, but otherwise no code changes.

cmd/hatchet-migrate/migrate/migrations/20250907161856_v1_0_41.sql
abelanger5, line 1: This will need to be updated before final merge, as we'll probably get this deployed after payloads and partitions.

cmd/hatchet-migrate/migrate/migrations/20250907161856_v1_0_41.sql
```diff
@@ -0,0 +1,299 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION v1_task_insert_function()
+RETURNS TRIGGER AS $$
+DECLARE
+    rec RECORD;
+BEGIN
+    -- Only insert if there's a single task with initial_state = 'QUEUED' and concurrency_strategy_ids is not null
+    IF (SELECT COUNT(*) FROM new_table WHERE initial_state = 'QUEUED' AND concurrency_strategy_ids[1] IS NOT NULL) > 0 THEN
```
abelanger5, line 9: Added this IF statement to avoid cascading triggers on concurrency tables

cmd/hatchet-migrate/migrate/migrations/20250907161856_v1_0_41.sql
```diff
@@ -0,0 +1,299 @@
...
+        step_timeout,
+        COALESCE(priority, 1),
+        sticky,
+        desired_worker_id,
+        retry_count
+    FROM new_table
+    WHERE initial_state = 'QUEUED' AND concurrency_strategy_ids[1] IS NULL;
+
+    -- Only insert into v1_dag and v1_dag_to_task if dag_id and dag_inserted_at are not null
+    IF (SELECT COUNT(*) FROM new_table WHERE dag_id IS NOT NULL AND dag_inserted_at IS NOT NULL) = 0 THEN
```
abelanger5: Similarly, added this IF statement to avoid extra statements in this trigger for single-task workflows

internal/services/controllers/v1/olap/signal/signal.go
abelanger5, line 1: I had to consolidate signaling logic between the local scheduler (which creates tasks) and the tasks controller, and this was the cleanest way to do it. I also figured it'd give us an easier way to hook into two-phase commits to keep the tasks controller and OLAP controller in sync.

internal/services/dispatcher/server.go
```diff
@@ -209,11 +209,11 @@ func (worker *subscribedWorker) CancelStepRun(
 	})
 }
 
-func (s *DispatcherImpl) Register(ctx context.Context, request *contracts.WorkerRegisterRequest) (*contracts.WorkerRegisterResponse, error) {
+func (d *DispatcherImpl) Register(ctx context.Context, request *contracts.WorkerRegisterRequest) (*contracts.WorkerRegisterResponse, error) {
```
abelanger5, line 212: Drive-by, but the golint rules were complaining about using `s *DispatcherImpl` in some places and `d *DispatcherImpl` in other places, and I agree it makes the code harder to read, so changed all locations to use `d *DispatcherImpl`.

internal/services/dispatcher/server_v1.go
```diff
@@ -1318,11 +1330,37 @@ func (s *DispatcherImpl) isMatchingWorkflowRunV1(msg *msgqueue.Message, acks *wo
...
+					EventTimestamp: timestamppb.New(time.Now()),
+					Results: []*contracts.StepRunResult{
+						{
+							StepRunId:      payload.ExternalId,
+							StepReadableId: payload.StepReadableId,
+							JobRunId:       payload.ExternalId,
+							Output:         &output,
+						},
+					},
+				})
```
abelanger5, line 1359: This is the core change in this file, if we have a single-task workflow and we get a task-completed event, we send it immediately to the worker. 

I think there's a chance of race conditions here, because we don't want to send a message to a worker before its been fully written/acknowledged by the database, which we risk here. 

So if we were to fail to write the `task-completed` event to the database, we won't successfully release the task, and it may have a different status after we hit a timeout/reassignment than the parent workflow would see. 

Yet another case where 2PC is necessary - or perhaps just moving the processing of the task-completed event to the gRPC layer like we used to have. I need to think a bit more about it.

pkg/repository/v1/shared.go
```diff
@@ -25,17 +27,19 @@ type taskExternalIdTenantIdTuple struct {
...
-	m                         *metered.Metered
+	pool                        *pgxpool.Pool
+	v                           validator.Validator
+	l                           *zerolog.Logger
+	queries                     *sqlcv1.Queries
+	queueCache                  *cache.Cache
+	stepExpressionCache         *cache.Cache
+	tenantIdWorkflowNameCache   *expirable.LRU[string, *sqlcv1.ListWorkflowsByNamesRow]
+	stepsInWorkflowVersionCache *expirable.LRU[string, []*sqlcv1.ListStepsByWorkflowVersionIdsRow]
+	stepIdLabelsCache           *expirable.LRU[string, []*sqlcv1.GetDesiredLabelsRow]
```
abelanger5, line 38: I would like to start moving towards usage of the `expirable` package instead of our home-built cache, because of nicer typing/LRU support - I think we can fully make this transition after we deprecate v0.

internal/services/admin/v1/server.go
```diff
@@ -435,6 +438,53 @@ func (i *AdminServiceImpl) generateExternalIds(ctx context.Context, tenantId str
 }
 
 func (i *AdminServiceImpl) ingest(ctx context.Context, tenantId string, opts ...*v1.WorkflowNameTriggerOpts) error {
+	if i.localScheduler != nil {
+		localWorkerIds := map[string]struct{}{}
+
+		if i.localDispatcher != nil {
```
mrkaye97, line 444: is this an exact copy of the implementation in `internal/services/admin/server_v1.go`?

internal/services/controllers/v1/olap/signal/signal.go
```diff
@@ -0,0 +1,724 @@
...
+		pubBuffer: pubBuffer,
+	}
+}
+
+func (s *OLAPSignaler) SignalDAGsCreated(ctx context.Context, tenantId string, dags []*v1.DAGWithData) error {
+	// notify that tasks have been created
+	// TODO: make this transactionally safe?
+	for _, dag := range dags {
+		dagCp := dag
+		msg, err := tasktypes.CreatedDAGMessage(tenantId, dagCp)
```
mrkaye97, line 46: are we copying this because it gets mutated? if so, I'd much prefer to not mutate it 😅

internal/services/controllers/v1/olap/signal/signal.go
```diff
@@ -0,0 +1,724 @@
...
+	if err != nil {
+		return err
+	}
+
+	// notify that tasks have been cancelled
+	// TODO: make this transactionally safe?
+	for _, task := range tasks {
+		msg, err := tasktypes.MonitoringEventMessageFromInternal(tenantId, tasktypes.CreateMonitoringEventPayload{
+			TaskId:         task.ID,
+			RetryCount:     task.RetryCount,
```
mrkaye97, line 275: think it's worth consolidating these loops for performance?

internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -386,6 +387,187 @@ func (d *DispatcherImpl) handleTaskBulkAssignedTask(ctx context.Context, msg *ms
...
+		return true
+	})
+
+	return workerIds
+}
+
+// Note: this is very similar to handleTaskBulkAssignedTask, with some differences in what's sync vs run in a goroutine
+// In this method, we wait until all tasks have been sent to the worker before returning
+func (d *DispatcherImpl) HandleLocalAssignments(ctx context.Context, tenantId, workerId string, tasks []*schedulingv1.AssignedItemWithTask) error {
+	// we set a timeout of 25 seconds because we don't want to hold the semaphore for longer than the visibility timeout (30 seconds)
```
mrkaye97, line 406: is this 30 second visibility timeout hard-coded somewhere? it seems risky to hard-code this 25 sec if that's modifiable

internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -386,6 +387,187 @@ func (d *DispatcherImpl) handleTaskBulkAssignedTask(ctx context.Context, msg *ms
...
+			currInput := &v1.V1StepRunData{}
+
+			if task.Input != nil {
+				err := json.Unmarshal(task.Input, currInput)
+
+				if err != nil {
+					d.l.Warn().Err(err).Msg("failed to unmarshal input")
+					continue
+				}
+			}
```
mrkaye97, line 448: just FYI, a bunch of this will conflict with the payloads changes

internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -386,6 +387,187 @@ func (d *DispatcherImpl) handleTaskBulkAssignedTask(ctx context.Context, msg *ms
...
+				err := w.StartTaskFromBulk(ctx, tenantId, task)
+
+				if err != nil {
+					multiErr = multierror.Append(
+						multiErr,
+						fmt.Errorf("could not send action for task %s to worker %s (%d / %d): %w", sqlchelpers.UUIDToStr(task.ExternalID), workerId, i+1, len(workers), err),
+					)
+				} else {
+					success = true
+					break
```
mrkaye97, line 509: why do we `break` here on success?

internal/services/ingestor/ingestor_v1.go
```diff
@@ -138,27 +225,33 @@ func (i *IngestorImpl) ingestReplayedEventV1(ctx context.Context, tenant *dbsqlc
...
+		return nil, fmt.Errorf("expected 1 event to be ingested, got %d", len(events))
+	}
+
+	return events[0], nil
 }
 
-func eventToTaskV1(tenantId, eventExternalId, key string, data, additionalMeta []byte, priority *int32, scope *string, triggeringWebhookName *string) (*msgqueue.Message, error) {
-	payloadTyped := tasktypes.UserEventTaskPayload{
-		EventExternalId:         eventExternalId,
+func eventToPayload(tenantId, key string, data, additionalMeta []byte, priority *int32, scope *string, triggeringWebhookName *string) tasktypes.UserEventTaskPayload {
```
mrkaye97, line 243: nit: I think this naming is a little confusing even if technically correct given our abuse of the term "payload"

pkg/repository/v1/scheduler_optimistic.go
```diff
@@ -0,0 +1,359 @@
...
+	return &optimisticSchedulingRepositoryImpl{
+		sharedRepository: shared,
+	}
+}
+
+func (r *optimisticSchedulingRepositoryImpl) StartTx(ctx context.Context) (*OptimisticTx, error) {
+	return r.PrepareOptimisticTx(ctx)
+}
+
+func (r *optimisticSchedulingRepositoryImpl) TriggerFromEvents(ctx context.Context, tx *OptimisticTx, tenantId string, opts []EventTriggerOpts) ([]*sqlcv1.V1QueueItem, *TriggerFromEventsResult, error) {
```
mrkaye97, line 31: this and `TriggerFromNames` are all copy and paste right? are there any changes here?

pkg/repository/v1/trigger.go
```diff
@@ -1761,3 +1755,84 @@ func (r *sharedRepository) processWorkflowExpression(ctx context.Context, expres
...
+			continue
+		}
+
+		workflowNamesToLookup = append(workflowNamesToLookup, name)
+	}
+
+	// look up the workflow versions for the workflow names
+	workflowVersions, err := r.queries.ListWorkflowsByNames(ctx, tx, sqlcv1.ListWorkflowsByNamesParams{
+		Tenantid:      sqlchelpers.UUIDFromStr(tenantId),
+		Workflownames: workflowNamesToLookup,
```
mrkaye97, line 1777: should we check if `workflowNamesToLookup` is empty before running this query?

pkg/repository/v1/trigger.go
```diff
@@ -1761,3 +1755,84 @@ func (r *sharedRepository) processWorkflowExpression(ctx context.Context, expres
...
+			res[id] = steps
+			continue
+		}
+
+		workflowVersionsToLookup = append(workflowVersionsToLookup, sqlchelpers.UUIDFromStr(id))
+	}
+
+	steps, err := r.queries.ListStepsByWorkflowVersionIds(ctx, tx, sqlcv1.ListStepsByWorkflowVersionIdsParams{
+		Tenantid: sqlchelpers.UUIDFromStr(tenantId),
+		Ids:      workflowVersionsToLookup,
```
mrkaye97, line 1816: same here maybe?

mrkaye97 (commented):
left some questions! I'm a little nervous about the amount of complexity this adds to some already-complex parts of the codebase, but if it's worth the performance gains then 🤷

internal/services/dispatcher/dispatcher_v1.go
abelanger5: we're technically looping over individual worker sessions, and only one of those will be valid, the others will fail when trying to send to the worker (or they should). this is an edge case when a worker reconnects and we haven't determined that the old connection has been interrupted yet.

pkg/repository/v1/scheduler_optimistic.go
abelanger5: yeah there are some small changes in how we're handling pre/post commits and treating the transactions

pkg/repository/v1/scheduler_optimistic.go
```diff
@@ -0,0 +1,359 @@
...
+	// get the queue items for the tasks that were created
+	taskIds := make([]int64, 0, len(tasks))
+	taskInsertedAts := make([]pgtype.Timestamptz, 0, len(tasks))
+	retryCounts := make([]int32, 0, len(tasks))
+
+	for _, task := range tasks {
+		taskIds = append(taskIds, task.ID)
+		taskInsertedAts = append(taskInsertedAts, task.InsertedAt)
+		retryCounts = append(retryCounts, task.RetryCount)
+	}
```
abelanger5, line 244: double check that tasks array includes the tasks created as part of each DAG

abelanger5:
closing in favor of #2867

PR #2274, opened 2025-09-09
[hotfix] Do not panic when webhook worker listener returns error

mnafees:
# Description

Instead of calling `panic()`, we should log the error and exit from the goroutine when the webhook worker listener returns an error.

## Type of change


mrkaye97 (approved):
thanks!

PR #2275, opened 2025-09-09
Org UI feedback improvements

mnafees:
# Description

This PR fixes a couple of things
- Alignment of tenant indicator in the org selector
- Org name change
- More uniformity across all tables in the org view
- Timestamps properly user-readable and formatted

## Type of change

### DIFF
--- a/frontend/app/src/pages/onboarding/create-tenant/index.tsx
+++ b/frontend/app/src/pages/onboarding/create-tenant/index.tsx
@@ -135,15 +135,15 @@ export default function CreateTenant() {
       setTimeout(() => {
         if (result.type === 'cloud') {
           const tenant = result.data as OrganizationTenant;
-          navigate(`/tenants/${tenant.id}/onboarding/get-started`);
+          window.location.href = `/tenants/${tenant.id}/onboarding/get-started`;
           return;
         }
 
         const tenant = result.data as Tenant;
         if (tenant.version === TenantVersion.V1) {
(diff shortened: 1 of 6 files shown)


frontend/app/src/pages/onboarding/create-tenant/index.tsx
```diff
@@ -132,16 +134,16 @@ export default function CreateTenant() {
...
+          navigate(`/tenants/${tenant.metadata.id}/onboarding/get-started`);
         } else {
-          // For regular tenants, use the existing onboarding flow
-          const tenant = result.data as any;
-          if (tenant.version === TenantVersion.V1) {
-            window.location.href = `/tenants/${tenant.metadata.id}/onboarding/get-started`;
-          } else {
-            window.location.href = `/onboarding/get-started?tenant=${tenant.metadata.id}`;
-          }
+          navigate(`/onboarding/get-started?tenant=${tenant.metadata.id}`);
```
mrkaye97: there was a reason for the hard-redirects here, but I'm struggling to remember what it was. it might be good to use `window.location.href = ` for this one just to be safe. we can change it when we remove the old v0 code

mrkaye97 (commented):
left one comment + a couple small UI things:

1. when you click the X to close the orgs page, it looks like we replace the browser history, so going back doesn't go back to the orgs page again
2. the pencil icon flashes when you either confirm or cancel editing the name of an org, probably missing a loading state there or something

mrkaye97 (approved):

PR #2276, opened 2025-09-09
Feat: Filters UI, Events page refactor, Misc. other fixes

mrkaye97:
# Description

* Adding a UI for filters to see what's there
* Updated the filters endpoint to handle pagination properly
* Refactored the events page a bit to have a `useEvents` hook and make better use of query param state + the new `usePagination` hook
* Also removed the run id (always empty) from the scheduled runs page

## Type of change

#### Empty State

Screenshot 2025-09-09 at 2 10 35 PM

#### Many filters
Screenshot 2025-09-10 at 2 54 08 PM

#### Create form

Screenshot 2025-09-10 at 2 58 17 PM

#### Details panel

Screenshot 2025-09-10 at 2 58 26 PM

#### Edit mode

Screenshot 2025-09-10 at 2 58 39 PM


api/v1/server/handlers/v1/filters/list.go
```diff
@@ -30,22 +30,33 @@ func (t *V1FiltersService) V1FilterList(ctx echo.Context, request gen.V1FilterLi
 		scopeParams = append(scopeParams, *scopes...)
 	}
 
-	filters, err := t.config.V1.Filters().ListFilters(
+	filterLimit := int64(20000)
```
abelanger5, line 33: This seems like a pretty big default limit? can't imagine sending an API consumer 20k rows :)

abelanger5 (approved):

api/v1/server/handlers/v1/filters/list.go
mrkaye97: yeah, I just left this to be the same as what it was in the SQL query before this change 🤷 the FE + SDKs will always send a value

PR #2280, opened 2025-09-10
Do not show archived tenants anywhere

mnafees:
# Description

The server will return 403s for all archived tenants so no need to show them in the UI at all.

## Type of change

### DIFF
--- a/frontend/app/src/hooks/use-organizations.ts
+++ b/frontend/app/src/hooks/use-organizations.ts
@@ -8,6 +8,8 @@ import {
   ManagementTokenDuration,
   OrganizationMember,
   OrganizationForUser,
+  Organization,
+  TenantStatusType,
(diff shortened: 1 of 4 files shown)


frontend/app/src/hooks/use-organizations.ts
```diff
@@ -28,6 +30,36 @@ export function useOrganizations() {
...
+        } catch (error) {
+          console.warn(
+            `Failed to fetch detailed org data for ${org.metadata.id}:`,
+            error,
+          );
+          return null;
+        }
+      });
+      const results = await Promise.all(orgPromises);
+      return results.filter((org): org is Organization => org !== null);
```
mrkaye97: do you think we should have a bulk endpoint for this instead so we don't need to make N API calls here?

abelanger5 (approved):

mrkaye97 (approved):

PR #2297, opened 2025-09-13
Fix `GetDetails` in `Runs` feature client of Go SDK v1

mnafees:
# Description

`GetDetails` in Go SDK v1 always returned a 404 since it called an older endpoint that no longer works for v1. For the older v1 SDK, change the signature and deprecate it. For the newer one, just get rid of the method.

## Type of change

### DIFF
--- a/pkg/v1/features/runs.go
+++ b/pkg/v1/features/runs.go
@@ -22,6 +22,7 @@ type RunsClient interface {
 	GetStatus(ctx context.Context, runId string) (*rest.V1WorkflowRunGetStatusResponse, error)
 
 	// GetDetails retrieves detailed information about a workflow run by its ID.
+	// Deprecated: Use Get instead.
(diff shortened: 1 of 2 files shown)


pkg/v1/features/runs.go
```diff
@@ -22,7 +22,8 @@ type RunsClient interface {
 	GetStatus(ctx context.Context, runId string) (*rest.V1WorkflowRunGetStatusResponse, error)
 
 	// GetDetails retrieves detailed information about a workflow run by its ID.
-	GetDetails(ctx context.Context, runId string) (*rest.WorkflowRunGetShapeResponse, error)
+	// Deprecated: Use Get instead.
+	GetDetails(ctx context.Context, runId string) (*rest.V1WorkflowRunGetResponse, error)
```
mrkaye97, line 26: I assume this is fine since it doesn't work and not many people are using the V1 SDK yet, but this is a breaking change right?

abelanger5 (approved):

PR #2309, opened 2025-09-17
Fix: DAG details rendering in side panel, backwards compatible event list API

mrkaye97:
# Description

Couple of small fixes:

1. Making the v0 event list API compatible with the v1 tables by checking the tenant version (basically just copied the v1 implementation for the part, figure we'll delete soon)
2. Fixing an issue with the side panel not rendering properly because of some kind of react state race or rerender race

[Easier to review events changes without whitespace](https://github.com/hatchet-dev/hatchet/pull/2309/files?diff=unified&w=1)

## Type of change

### DIFF
--- a/api/v1/server/handlers/events/list.go
+++ b/api/v1/server/handlers/events/list.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/jackc/pgx/v5/pgtype"
 	"github.com/labstack/echo/v4"
 
 	"github.com/hatchet-dev/hatchet/api/v1/server/oas/apierrors"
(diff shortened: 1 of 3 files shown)


api/v1/server/handlers/events/list.go
```diff
@@ -16,115 +17,229 @@ import (
...
+			if err != nil {
+				return nil, err
+			}
 
-	if request.Params.EventIds != nil {
-		eventIds := make([]string, len(*request.Params.EventIds))
+			rows = append(rows, *eventData)
+		}
+	case dbsqlc.TenantMajorEngineVersionV1:
+		since := time.Now().Add(-time.Hour * 24)
```
abelanger5, line 135: Would it make sense to set this to the retention window or is that risky?

abelanger5 (approved):

api/v1/server/handlers/events/list.go
mrkaye97: 🤷 I figure it doesn't really matter since it'll just get removed soon, I can do that if you think it's worth it

PR #2316, opened 2025-09-18
Fix: Payloads OLAP backwards compat

mrkaye97:
# Description

Fixing a backwards compatibility issue by falling back to the `input` on the OLAP side if the `payload` is null

## Type of change

### DIFF
--- a/pkg/repository/v1/olap.go
+++ b/pkg/repository/v1/olap.go
@@ -1302,7 +1302,7 @@ func (r *OLAPRepositoryImpl) writeTaskBatch(ctx context.Context, tenantId string
 
 		// fall back to input if payload is empty
 		// for backwards compatibility
-		if payload == nil || len(payload) == 0 {
+		if len(payload) == 0 {
 			payload = task.Input
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/repository/v1/olap.go
```diff
@@ -1298,6 +1298,14 @@ func (r *OLAPRepositoryImpl) writeTaskBatch(ctx context.Context, tenantId string
 	params := make([]sqlcv1.CreateTasksOLAPParams, 0)
 
 	for _, task := range tasks {
+		payload := task.Payload
+
+		// fall back to input if payload is empty
+		// for backwards compatibility
+		if payload == nil || len(payload) == 0 {
```
abelanger5: Should we add a check if `len(task.Input) == 0` as well, and set it to some placeholder, or would that never happen?

abelanger5 (changes requested):

pkg/repository/v1/olap.go
mrkaye97: We could set it to `'{}'` as a placeholder if you think it's worth it, but I can't imagine that's possible? This was the original version (using the `task.Input` directly): https://github.com/hatchet-dev/hatchet/pull/2047/files#diff-f03a2522a98ff2b4aaa7198abc2aecc84ef2af42a2e5212a5a321d49c8fac681L1317

abelanger5 (approved):

PR #2321, opened 2025-09-21
fix: `skip locked` on queue updates

abelanger5:
# Description

When we upsert queues across many transactions, it becomes a single point of contention due to multiple `INSERT...ON CONFLICT` statements. 

This PR calls `SKIP LOCKED` on any existing queues which have a lock (we only update queues to set `last_active`, so we don't really risk anything by skipping the lock in this case), and only calls `INSERT` on queues that don't exist. 

From what I can tell, it's not possible to do this without 3 CTEs + 1 statement, which is unfortunate. 

## Type of change

### DIFF
--- a/pkg/repository/v1/sqlcv1/queue.sql
+++ b/pkg/repository/v1/sqlcv1/queue.sql
@@ -1,22 +1,35 @@
 -- name: UpsertQueues :exec
 WITH ordered_names AS (
     SELECT unnest(@names::text[]) AS name
-    ORDER BY name
+    ORDER BY name ASC
+), existing_queues AS (
+    SELECT tenant_id, name, last_active
+    FROM v1_queue
+    WHERE tenant_id = $1
+      AND name = ANY(@names::text[])
+), locked_existing_queues AS (
+    SELECT *
+    FROM v1_queue
+    WHERE
+        tenant_id = $1
+        AND name IN (SELECT name FROM existing_queues)
+    ORDER BY name ASC
+    FOR UPDATE SKIP LOCKED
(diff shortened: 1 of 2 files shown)


pkg/repository/v1/sqlcv1/queue.sql
```diff
@@ -1,22 +1,35 @@
 -- name: UpsertQueues :exec
 WITH ordered_names AS (
     SELECT unnest(@names::text[]) AS name
-    ORDER BY name
+    ORDER BY name ASC
+), existing_queues AS (
+    SELECT tenant_id, name, last_active
+    FROM v1_queue
+    WHERE tenant_id = $1
```
mrkaye97, line 8: intentional to be putting the `$1` placeholder here instead of `@tenantId::UUID`?

pkg/repository/v1/sqlcv1/queue.sql
```diff
@@ -1,22 +1,35 @@
...
+    WHERE
+        tenant_id = $1
+        AND name IN (SELECT name FROM existing_queues)
+    ORDER BY name ASC
+    FOR UPDATE SKIP LOCKED
+), names_to_insert AS (
+    SELECT on1.name
+    FROM ordered_names on1
+    LEFT JOIN existing_queues eq ON eq.name = on1.name
+    WHERE eq.name IS NULL
```
mrkaye97, line 22: nit: I usually prefer the `NOT EXISTS` or `NOT IN` way of doing this but I don't think it's a huge deal

mrkaye97 (approved):

PR #2354, opened 2025-09-28
feat: add pagination support for V1LogLineList

### LINKED ISSUE #2312: Python: hatchet.logs.list() returns an object with a pagination member, but no way to use it
`V1LogLineList` has
```
pagination: Optional[PaginationResponse] = None
```

But `hatchet.logs.list()` only takes a task ID:
```
def list(self, task_run_id: str) -> V1LogLineList:
```

jishnundth:
# Description

add pagination support for v1 LogLineList, for the api server, python and ts sdks. 
refactor UI to support pagination on logs

Fixes #2312 

## Type of change

## What's Changed

- [x] refactor openapi schema for v1 listLogs
- [x] refactor ListLogLines repository logic to return count of loglines
- [x] add v1 logLine SQL query to fetch the count of loglines based on query params
- [x] auto generated changes for python and typescript sdks 
- [x] documentation change is required here: https://docs.hatchet.run/sdks/python/feature-clients/logs#list 
- [x] hatchet UI changes to support pagination on logs


api-contracts/openapi/paths/v1/tasks/tasks.yaml
```diff
@@ -456,6 +456,20 @@ listLogs:
...
+        schema:
+          type: integer
+          format: int64
+      - description: The number to limit by
+        in: query
+        name: limit
+        required: false
+        schema:
+          type: integer
+          format: int64
```
mrkaye97, line 465: we'll need to add a `since` parameter here too - we use Loki for our managed cloud logs backend, and Loki doesn't allow limit + offset pagination (will need to be limit + since instead)

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/step-run-logs.tsx
```diff
@@ -1,10 +1,142 @@
...
+  isLoading,
+  onFirstPage,
+  onPrevPage,
+  onNextPage,
+  onLastPage,
+  onPageSizeChange,
+}: LogsPaginationProps) {
+  return (
+    <div className="flex items-center justify-between px-2 mt-4 border-t pt-4">
+      <div className="flex-1 text-sm text-gray-600 dark:text-gray-400"></div>
```
mrkaye97: I think we should use `useInfiniteQuery` for this and paginate automatically on scroll. WDYT? Then we can use the existing component, at least probably

pkg/repository/v1/sqlcv1/log_line.sql
```diff
@@ -31,3 +31,12 @@ ORDER BY
...
 OFFSET COALESCE(sqlc.narg('offset'), 0);
+
+
+-- name: CountLogLines :one
+SELECT COUNT(*) AS total
+FROM v1_log_line l
+WHERE
+    l.tenant_id = @tenantId::uuid
+    AND l.task_id = @taskId::bigint
+    AND l.task_inserted_at = @taskInsertedAt::timestamptz;
```
mrkaye97: we have a pattern other places where we put this into a CTE and add a `LIMIT 20000`, and then do the count after, for performance. E.g.:

```sql
WITH filtered_logs AS (
    SELECT *
    FROM v1_log_line
    WHERE
        tenant_id = @tenantId::UUID
        AND task_id = @taskId::BIGINT
        AND task_inserted_at = @taskInsertedAt::TIMESTAMPTZ
    LIMIT 20000
)

SELECT COUNT(*)
FROM filtered_logs
;
```

mrkaye97 (commented):
hey @jishnundth - thanks for putting in this change! looks really good. I left a few small comments. also, would you mind posting some screenshots of the frontend?

api-contracts/openapi/paths/v1/tasks/tasks.yaml
jishnundth: Got it. Does the offset need to be removed then? (Offset + limit) does seem to work in self-hosted environments.

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/step-run-logs.tsx
jishnundth: Sure, that's better. I also had concerns regarding the UX when using pagination with buttons, as the user might need to keep clicking 'Next' to see the live logs of a running task.

This is how it looks at the moment. 

image

api-contracts/openapi/paths/v1/tasks/tasks.yaml
mrkaye97: It does not! The idea will be that we'll do what you've done here in the OSS (self-hosted), but we'll also need to pass the oldest previous timestamp through from the frontend (in addition to the offset) for the cloud version. So basically we'll just need to compute that on the fly on the FE. Does that make sense? Happy to try to clarify more!

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/step-run-logs.tsx
mrkaye97: Gotcha! Yeah, so I think that with `useInfiniteQuery`, we could just flatten the results of the query and then have infinite scroll instead of needing pagination here

pkg/repository/v1/sqlcv1/log_line.sql
mrkaye97: Sorry, edited this a bit! Forgot the `LIMIT` 😅

api-contracts/openapi/paths/v1/tasks/tasks.yaml
jishnundth: Please clarify if I'm wrong:

1. In addition to _offset_, add _since_ on V1LogLines Query Params on all rest-clients
2. Then on the FE, we will only need (limit + since) for pagination to work, right? Why do we need the offset on the FE?

The pagination flow would be:
Page 1: { limit: 100, since: undefined } → get latest logs
Page 2: { limit: 100, since: <timestamp_of_last_log> } → get older logs than the oldest log in page 1

Got little confused since you mentioned this:  _from the frontend (in addition to the offset) for the cloud version_

mrkaye97: Pretty much!

For 2), the reason is that I suspect that using `since` for pagination could be quite inefficient over a large number of logs, since there's no index on the `createdAt` column on the logs table. With that said, it's probably fine to just do a sequential scan there anyways, since a single task likely won't have tens of thousands of logs. So with that in mind, yeah, I'd say removing the offset and doing everything with the `since` sounds good to me!

Sorry for the back and forth on this, lots to consider 😅 Let me know if you have more questions!

jishnundth: No worries at all, appreciate the thorough explanation!

jishnundth:
@mrkaye97 PTAL, I've added the mentioned changes.

I also made one additional change:

refetchInterval with useInfiniteQuery was fetching all the existing logs again after each refresh interval. This meant it was reloading all the log pages one by one, which seemed inefficient. To fix this, I implemented a custom polling logic that only fetches the latest logs based on the last timestamp of the already fetched logs, and merging the new ones with old ones. Logs are chronological, refreshing old logs are unnecessary. 

Also, on the FE we’re using (since + limit), so only nextPage (1 or nil) is passed in the pagination segment. This is used by useInfiniteQuery to validate the existence of the next page. The rest of the pagination(currpage & totalpages) parameters seemed meaningless when querying logs only with since.

api-contracts/openapi/paths/v1/tasks/tasks.yaml
```diff
@@ -456,6 +456,27 @@ listLogs:
           format: uuid
           minLength: 36
           maxLength: 36
+      - description: The number to skip
+        in: query
+        name: offset
+        required: false
+        schema:
+          type: integer
+          format: int64
```
mrkaye97: maybe we can remove the offset at this point?

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/step-run-logs.tsx
```diff
@@ -1,33 +1,155 @@
...
+  const getLogsQuery = useInfiniteQuery<
+    V1LogLineList,
+    Error,
+    InfiniteData<V1LogLineList>,
+    string[],
+    string | undefined
+  >({
+    queryKey: ['v1Tasks', 'getLogs', taskRun?.metadata.id],
+    queryFn: async ({ pageParam }) => {
+      const params: any = {
```
mrkaye97: we shouldn't use `any` here if it can be avoided

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/step-run-logs.tsx
```diff
@@ -1,33 +1,155 @@
...
+        taskRun?.metadata.id || '',
+        params,
+      );
+      return { data: response.data, lastTimestamp };
+    },
+    enabled: !!taskRun,
+    refetchInterval: taskRun?.status === V1TaskStatus.RUNNING ? 1000 : 5000,
+    refetchOnWindowFocus: false,
+  });
+
```
mrkaye97: maybe I'm misunderstanding something, but I'm not quite following the need for this - since the log lines from the API are ordered by created at ascending, I'd think that we can do something like what's suggested [in the docs here](https://tanstack.com/query/v4/docs/framework/react/guides/infinite-queries#refetchpage)

mrkaye97:
> @mrkaye97 PTAL, I've added the mentioned changes.
> 
> I also made one additional change:
> 
> refetchInterval with useInfiniteQuery was fetching all the existing logs again after each refresh interval. This meant it was reloading all the log pages one by one, which seemed inefficient. To fix this, I implemented a custom polling logic that only fetches the latest logs based on the last timestamp of the already fetched logs, and merging the new ones with old ones. Logs are chronological, refreshing old logs are unnecessary.
> 
> Also, on the FE we’re using (since + limit), so only nextPage (1 or nil) is passed in the pagination segment. This is used by useInfiniteQuery to validate the existence of the next page. The rest of the pagination(currpage & totalpages) parameters seemed meaningless when querying logs only with since.

Thanks for the changes! Added a couple more comments

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/step-run-logs.tsx
jishnundth: https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-v5#removed-refetchpage-in-favor-of-maxpages

Deprecated in v4, we're using v5. We cannot use this max_pages because, we're always querying forward. where the database query is like since >created timestamp. Can I use an _until_ queryparam as well, to make this work with maxPages.

So that the pagination can be triggered on both ways. Using max pages with current condition, user won't be able to scroll up, and will be stuck with last x pages. WDYT ?

jishnundth:
@mrkaye97 PTAL, the mentioned changes are added. 

[https://github.com/hatchet-dev/hatchet/pull/2354#discussion_r2395445105](url)

I had to add until query param, to support bidirectional pagination on infinitequery, and also removed pagination from v1loglinelist, as it doesn't make sense anymore. 

Let me know if any additional change is required.

mrkaye97:
Great work on this! Looks like a really nice implementation :) 

One thing is that we'll need to leave the pagination info, but you can make it empty (use a zero value or some such), because removing it is a breaking change for the SDKs, even if nobody is using it since it wasn't doing anything before.

I'll pull these changes to my local tomorrow and toy around a bit, and then leave another review!

mrkaye97:
Alright so I pulled the changes and toyed around a bit! Looks great - the API seems to be working really well. What I did was just create a task that logs ~1000 times, and looked for:

1. How do the logs look in the UI as they're still coming in / how does scrolling work as the logs are still coming in
2. How does scrolling work once the logs have _stopped_ coming in

A couple things here:

1. Scrolling is jumping around a lot. I think what should happen is that we fetch the first page, and then as you scroll, we fetch more pages incrementally. But it seems like what's happening is we're fetching multiple pages initially for some reason (I see multiple requests in pretty quick succession in the network tab), and then we're auto-scrolling to the bottom of the fetched logs. 
2. As logs are coming in still, it seems like the scroll is jumping around almost randomly. I'm not sure what's up with this, I think it might have to do with the component refetching, etc.
3. When I scroll to the bottom of the logs, the scroll jumps back up to the previous chunk.

Another thing I'm noticing is once the task is in a terminal state (e.g. `COMPLETED`), we don't need to keep refetching automatically, since by that point, we know that no more logs can come in, so once we've scrolled to the end, that's all we'll get. 

WDYT?

jishnundth:
Hey @mrkaye97, the inconsistent scroll behavior has been fixed, it seems to be pretty smooth now. Also, the pagination has been added back. Please test once more, and let me know.

mrkaye97:
Nice! This looks a lot better - I'm still noticing a couple of things. One is that if I log e.g. 1000 times and scroll to the bottom of the logs (where we're out of pages), it somehow seems to jump back to the `waiting for logs` loading state. I'm not sure how that happens really 🤷 

Another thing is that it appears to be fetching every time I get to the bottom of a "page", but IMO that defeats the purpose of infinite scroll. I think the goal should be that we pre-fetch the first few hundred, and then when you get to line e.g. 200 we fetch 400-500, then at 300 we fetch 500-600, and so on, so that as you scroll it looks like everything's already been fetched and there's no lag.

Sorry for being such a stickler here - it's a complicated UI! Thanks for the work and the patience 😄 Let me know what you think!

jishnundth:
@mrkaye97 Apologies for the delay. Please take a look at the new changes. 
The change is added as you mentioned, 
```
       scrollheight > 50%  and scrolldirection is down -> fetch next page. 
       scrollheight < 40%  and scrolldirection is up -> fetch previous page 
```

We're storing 2 pages on the client, to make the scroll smooth, the infinitescroll is very bumpy if only one page is stored at a time. (using maxPages of InfinteQuery)

The UI behavior is very smooth now, similar to dumping the entire logs and scrolling it. Please test and let me know

mrkaye97:
Hey @jishnundth - great! Thanks :)

I’m on vacation for the next week or so, I’ll take a look when I get back!

mrkaye97:
hey @jishnundth - thanks for the changes! couple things:

1. ~I'm still seeing the lag when we go to fetch the next set of logs. When I looked at the network tab, it's still fetching the next batch on-demand instead of pre-fetching it and then just displaying the next set of rows as we scroll so it goes more smoothly~
2. ~When I log a non-multiple of 100 number of lines, the last set in the batch seems to not show up (e.g. if I log 120, I only see the first 100 logs and not the last 20)~
3. ~When I get to the last page, the last request returns nothing and then I again see the `Waiting for logs...` view~
4. Looks like we have some merge conflicts here - would be great to resolve those!

~Let me know if you want some help here btw, I'm happy to get this over the line 😄~

Sorry about that! I realized I hadn't pulled your latest changes 🤦 I thought I did.

Anyways, great work - the functionality looks really nice now. I'll take a closer look through the code now!

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/step-run-logs.tsx
```diff
@@ -1,32 +1,204 @@
...
 
 export function StepRunLogs({ taskRun }: { taskRun: V1TaskSummary }) {
-  const getLogsQuery = useQuery({
-    ...queries.v1Tasks.getLogs(taskRun?.metadata.id || ''),
+  interface PageBoundary {
+    rowFirstTimestamp: string;
+    rowLastTimestamp: string;
+    fetchedNext?: boolean;
+    fetchedPrevious?: boolean;
+  }
```
mrkaye97: can we move this interface out of the component?

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/step-run-logs.tsx
```diff
@@ -1,32 +1,204 @@
...
+  const pageBoundariesRef = useRef<Record<number, PageBoundary>>({});
+  const currentPageNumberRef = useRef<number>(0);
+  const isRefetchingRef = useRef<boolean>(false);
+  const lastScrollTopRef = useRef<number>(0);
+  const lastScrollPercentageRef = useRef<number>(0);
+  const lastPageNumberRef = useRef<number>(0);
```
mrkaye97: nit: we don't need the type hints here - the type checker can infer them from the defaults

mrkaye97 (commented):
left a couple more minor comments, but this looks great! One more small UI thing (not a big deal, but would be nice to fix if it's easy) is that I wasn't able to scroll back _up_ from e.g. the 1000th log while a task was still running, but once it completed I could scroll up again. I think this might have to do something with the auto-refetch but I'm not sure

Great work!

Edit: actually, I'm not sure how I got scrolling up to work 😅 it seems like it only works some of the time

jishnundth:
@mrkaye97 As you mentioned, scrolling up is broken when a task is running. As a solution for this, I've implemented the following.
1. When a task is running, all the page logs will be stored on FE. By adjusting max_pages to unlimited. Due, to this, bidirectional scroll is possible when task is running as well. 
2. This created another issue, default refetch by useinfinitequery was refetching the entire set of pages, which is unnecessary in our case, as older logs doesn't change. So, created a custom refetch logic that only refetch the latest page. 
3. Also, added a reset logic to remove log entries from cache for a completed task. The issue was when user scrolled maybe upto 5th page, and closed the task panel, and reopened it again, it was again showing the 5th page. And scrolling was not working. Since the pagination tracking variables are reset. 

Tracking pages and page numbers on FE was not working for a running task, that's why I've finalized the above solution. The above solution has minimal database overhead, and the UI also feels pretty smooth now. 

I've tested it countless times now:
For a completed task: https://jam.dev/c/d4878cd6-a2d0-47e5-b4bd-9a636ed27d94 (max_pages: 2)
For a running task: https://jam.dev/c/5a981311-e8ee-413e-8abf-8b9fcc147fbd (max_pages: unlimited)

Please test once, from your end, and let me know whether the results are as expected. Apologies for dragging this for too long.

mrkaye97 (approved):
Looks great, nice work on this! I'll work on getting it shipped to our cloud + push a new tag soon - I need to make a corresponding change for our cloud deployments

thanks for the great contribution!

PR #2368, opened 2025-09-30
feat: gzip compression for large payloads, persistent OLAP writes

mrkaye97:
# Description

- [X] Enables `gzip` compression for payloads over 5kb - pulled this into a constant but I don't think it's worth exposing as a config var yet
- [X] Allows the configuration of `persistent` RabbitMQ messages coming out of the publishing buffers - this was previously causing transient messages to be written to a durable queue which is incorrect and could lead to data loss on RabbitMQ

### DIFF
--- a/internal/msgqueue/v1/rabbitmq/gzip.go
+++ b/internal/msgqueue/v1/rabbitmq/gzip.go
@@ -7,11 +7,6 @@ import (
 	"io"
 )
 
-const (
-	// MinCompressionSize is the minimum payload size (in bytes) that will be compressed
-	MinCompressionSize = 5 * 1024 // 5KB
-)
-
 type CompressionResult struct {
 	Payloads       [][]byte
 	WasCompressed  bool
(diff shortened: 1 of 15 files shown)


internal/msgqueue/v1/rabbitmq/gzip.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"bytes"
+	"compress/gzip"
+	"fmt"
+	"io"
+)
+
+const (
+	// MinCompressionSize is the minimum payload size (in bytes) that will be compressed
+	MinCompressionSize = 5 * 1024 // 5KB
```
mrkaye97: think it's worth an env var for this?

internal/services/dispatcher/server_v1.go
```diff
@@ -614,26 +614,6 @@ func (s *DispatcherImpl) handleTaskCompleted(inputCtx context.Context, task *sql
...
-		}
-
-		err = s.pubBuffer.Pub(inputCtx, msgqueue.OLAP_QUEUE, olapMsg, false)
-
-		if err != nil {
-			s.l.Error().Err(err).Msg("could not publish to OLAP queue")
-		}
-	}()
-
 	msg, err := tasktypes.CompletedTaskMessage(
```
mrkaye97, line 617: intentional to leave this change in?

mnafees (approved):
LGTM 👍🏽

internal/services/dispatcher/server_v1.go
abelanger5: yep

mrkaye97 (commented):
🚢

PR #2370, opened 2025-09-30
Feat: Task Event and DAG Payloads

mrkaye97:
# Description

Using the payload store for task event and DAG data. A couple decisions here:

1. The existing `created_at` in the task events table is a `TIMESTAMP(3)`, so I just added a new `inserted_at` - it's a bummer, but the id + inserted at will uniquely identify the event and we don't need to worry about more timestamp conversions this way.
2. Needed to add an `external_id` on the task event so that we can retrieve the data after we write the payload. This is non-breaking, it's nullable for now, and is only needed going forward so not a big deal from my PoV.
3. Changed the defaults for the three dual writes env vars to `true` from `false`, which seemed safer to me for people who are self-hosting

Everything else was pretty straightforward since there's only one place writing into this table and a handful of reads.

## Type of change


cmd/hatchet-migrate/migrate/migrations/20251007191452_v1_0_47.sql
```diff
@@ -0,0 +1,9 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE v1_task_event ADD COLUMN external_id UUID;
```
mrkaye97: need a unique identifier on write so that we can access the autogenerated `id` somehow

cmd/hatchet-migrate/migrate/migrations/20251007191452_v1_0_47.sql
```diff
@@ -0,0 +1,15 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TYPE v1_payload_type ADD VALUE IF NOT EXISTS 'TASK_EVENT_DATA';
+ALTER TABLE v1_task_event
+    ADD COLUMN external_id UUID,
+    ADD COLUMN inserted_at TIMESTAMPTZ DEFAULT NOW();
```
mrkaye97: need these for storing payloads (the `inserted_at`, since `created_at` is a `TIMESTAMP(3)` and I didn't want to mess with changing the type, and `external_id` so we have a unique ID which we can use to look up the data by)

pkg/config/server/server.go
```diff
@@ -598,10 +598,11 @@ type ServerConfig struct {
 }
 
 type PayloadStoreConfig struct {
-	EnablePayloadDualWrites        bool          `mapstructure:"enablePayloadDualWrites" json:"enablePayloadDualWrites,omitempty" default:"false"`
-	WALPollLimit                   int           `mapstructure:"walPollLimit" json:"walPollLimit,omitempty" default:"1000"`
-	WALProcessInterval             time.Duration `mapstructure:"walProcessInterval" json:"walProcessInterval,omitempty" default:"15s"`
-	ExternalCutoverProcessInterval time.Duration `mapstructure:"externalCutoverProcessInterval" json:"externalCutoverProcessInterval,omitempty" default:"15s"`
+	EnablePayloadDualWrites          bool          `mapstructure:"enablePayloadDualWrites" json:"enablePayloadDualWrites,omitempty" default:"true"`
+	EnableTaskEventPayloadDualWrites bool          `mapstructure:"enableTaskEventPayloadDualWrites" json:"enableTaskEventPayloadDualWrites,omitempty" default:"true"`
```
mrkaye97, line 602: @abelanger5 wdyt of defaulting these to true? I think if they're false, people who are self-hosting will have failures if they deploy + try to replay, etc.

pkg/repository/v1/task.go
```diff
@@ -2578,16 +2609,27 @@ func (r *sharedRepository) createTaskEvents(
...
 		retryCounts[i] = task.RetryCount
 		eventTypesStrs[i] = string(eventTypes[i])
 
-		if len(eventDatas[i]) == 0 {
+		externalId := sqlchelpers.UUIDFromStr(uuid.NewString())
+		externalIds[i] = externalId
+
+		// important: if we don't set this to `eventDatas[i]` and instead allow it to be nil optionally
+		// we'll get errors downstream when we try to read the payload back and parse it in `registerChildWorkflows`
+		// because it'll try to unmarshal the `nil` value.
```
mrkaye97, line 2629: this is risky, but I'm not sure if we want to "fix" it downstream by swallowing that error or not 🤷

cmd/hatchet-migrate/migrate/migrations/20251007231452_v1_0_48.sql
```diff
@@ -0,0 +1,15 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TYPE v1_payload_type ADD VALUE IF NOT EXISTS 'TASK_EVENT_DATA';
+ALTER TABLE v1_task_event
+    ADD COLUMN external_id UUID,
+    ADD COLUMN inserted_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP;
```
abelanger5: We've double checked that this doesn't try to update all existing rows to `CURRENT_TIMESTAMP`? Seeing some contradictory information online: https://stackoverflow.com/questions/38916063/add-timestamp-column-with-default-now-for-new-rows-only

internal/services/dispatcher/v1/server.go
```diff
@@ -168,12 +168,24 @@ func (d *DispatcherServiceImpl) ListenForDurableEvent(server contracts.V1Dispatc
 			return fmt.Errorf("could not find external id for task %d, signal key %s", e.TaskID, e.EventKey.String)
 		}
 
+		payload, err := d.repo.Payloads().Retrieve(ctx, v1.RetrievePayloadOpts{
```
abelanger5: Looks like an `n+1` - ideally we could do this out of `d.repo.Tasks().ListSignalCompletedEvents` with a bulk read.

pkg/config/server/server.go
abelanger5: yep, sounds good

pkg/repository/v1/task.go
```diff
@@ -3355,21 +3418,52 @@ func (r *TaskRepositoryImpl) ListTaskParentOutputs(ctx context.Context, tenantId
...
+	}
+
+	workflowRunIdsToOutputs := make(map[string][]*TaskOutputEvent)
+
+	for retrieveOpts, workflowRunId := range retrieveOptsToWorkflowRunId {
+		wrId := sqlchelpers.UUIDToStr(workflowRunId)
+		payload, ok := payloads[retrieveOpts]
+
+		if !ok {
+			r.l.Error().Msgf("ListenForDurableEvent: task %s with ID %d and inserted_at %s has empty payload, falling back to input", wrId, retrieveOpts.Id, retrieveOpts.InsertedAt.Time)
```
abelanger5:
```suggestion
			r.l.Error().Msgf("ListTaskParentOutputs: task %s with ID %d and inserted_at %s has empty payload, falling back to input", wrId, retrieveOpts.Id, retrieveOpts.InsertedAt.Time)
```

sql/schema/v1-core.sql
```diff
@@ -327,6 +327,7 @@ CREATE TYPE v1_task_event_type AS ENUM (
 -- CreateTable
 CREATE TABLE v1_task_event (
     id bigint GENERATED ALWAYS AS IDENTITY,
+    inserted_at timestamptz NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
abelanger5: Seeing `NOT NULL` here which doesn't seem to match the migration?

abelanger5 (changes requested):
Nice! Few small things but looks like we're getting there 💯

sql/schema/v1-core.sql
mrkaye97: ah whoops, thanks, I keep doing this 🤦

internal/services/dispatcher/v1/server.go
mrkaye97: good call - let me change this V1TaskEvent struct to be V1TaskEventWithPayload like we have for the task and refactor a bit

cmd/hatchet-migrate/migrate/migrations/20251007231452_v1_0_48.sql
mrkaye97: nice catch, it does indeed - let me try to hack around this

pkg/repository/v1/task.go
```diff
@@ -2843,21 +2923,19 @@ func (r *TaskRepositoryImpl) ReplayTasks(ctx context.Context, tenantId string, t
 			}
 		}
 
-		input, err := r.payloadStore.Retrieve(ctx, RetrievePayloadOpts{
+		retrieveOpt := RetrievePayloadOpts{
```
mrkaye97, line 2926: @abelanger5 FYI, this was N+1 also - I just removed the singleton `Retrieve` method too

abelanger5 (approved):
Looks good to me!

PR #2375, opened 2025-10-01
Feat: Immediate Payload Offloads

mrkaye97:
# Description

Adds a second WAL-type table `v1_payload_cutover_queue_item` (since this is basically a queue of things to cut over to the external store), which we write into when we're writing the external keys to the payloads table and then we use to update the payloads table to use the external store and then delete from.

The idea here is that now the individual WAL records will be very short-lived (15 sec, maybe less), and we'll offload the payloads quickly to S3 so that we can access them basically immediately from the OLAP repo. Then, we can cut over the payloads table to use the offloaded payloads whenever we want to.

## Type of change


pkg/repository/v1/sqlcv1/payload-store.sql
```diff
@@ -104,16 +104,13 @@ WITH tenants AS (
...
-WHERE
-    offload_at < NOW()
-    AND tenant_id = ANY(SELECT tenant_id FROM tenants)
-ORDER BY offload_at, payload_id, payload_inserted_at, payload_type, tenant_id
-FOR UPDATE
+WHERE tenant_id = ANY(SELECT tenant_id FROM tenants)
+ORDER BY offload_at, tenant_id, payload_id, payload_inserted_at, payload_type
 LIMIT @pollLimit::INT
-
+FOR UPDATE SKIP LOCKED
```
mrkaye97, line 110: I think `SKIP LOCKED` makes sense here - no reason to be waiting for locks AFAICT

sql/schema/v1-core.sql
```diff
@@ -1644,21 +1644,18 @@ CREATE TABLE v1_payload (
 
     PRIMARY KEY (tenant_id, inserted_at, id, type),
     CHECK (
-        (location = 'INLINE' AND external_location_key IS NULL)
+        location = 'INLINE'
```
mrkaye97, line 1649: relaxing this check constraint to allow external keys before we've cut over to using the external store, so the S3 write can happen separately from the cutover

sql/schema/v1-core.sql
```diff
@@ -1684,14 +1693,38 @@ BEGIN
         'SELECT ARRAY(
             SELECT DISTINCT e.tenant_id
             FROM %I e
-            WHERE e.offload_at < NOW()
```
mrkaye97, line 1687: now we want to poll all payloads in the WAL - we'll be able to remove the `offload_at` later

pkg/repository/v1/sqlcv1/payload-store.sql
```diff
@@ -104,16 +104,13 @@ WITH tenants AS (
 
 SELECT *
 FROM v1_payload_wal
-WHERE
-    offload_at < NOW()
-    AND tenant_id = ANY(SELECT tenant_id FROM tenants)
-ORDER BY offload_at, payload_id, payload_inserted_at, payload_type, tenant_id
-FOR UPDATE
+WHERE tenant_id = ANY(SELECT tenant_id FROM tenants)
+ORDER BY offload_at, tenant_id, payload_id, payload_inserted_at, payload_type
```
mrkaye97: I think that the mis-ordering of this `ORDER BY` was causing really poor performance on this query

sql/schema/v1-core.sql
```diff
@@ -1660,14 +1660,28 @@ CREATE TABLE v1_payload_wal (
...
 
-    PRIMARY KEY (offload_at, tenant_id, payload_id, payload_inserted_at, payload_type),
-    CONSTRAINT "v1_payload_wal_payload" FOREIGN KEY (payload_id, payload_inserted_at, payload_type, tenant_id) REFERENCES v1_payload (id, inserted_at, type, tenant_id) ON DELETE CASCADE
+    -- todo: we probably should shuffle this index around - it'd make more sense now for the order to be
+    -- (tenant_id, offload_at, payload_id, payload_inserted_at, payload_type)
+    -- so we can filter by the tenant id first, then order by offload_at
+    PRIMARY KEY (offload_at, tenant_id, payload_id, payload_inserted_at, payload_type)
 ) PARTITION BY HASH (tenant_id);
 
 CREATE INDEX v1_payload_wal_payload_lookup_idx ON v1_payload_wal (payload_id, payload_inserted_at, payload_type, tenant_id);
```
mrkaye97, line 1671: bummer here - I'd like to drop this index, but I don't think we can drop an index on a partitioned table concurrently, so I think we'll need to wait for the WAL to be much smaller and then run that drop (and also drop the index below here, and ideally also shuffle the PK around) later

cmd/hatchet-migrate/migrate/migrations/20251007220652_v1_0_47.sql
```diff
@@ -0,0 +1,11 @@
...
+-- +goose Up
+-- https://stackoverflow.com/a/70958260
+CREATE INDEX CONCURRENTLY IF NOT EXISTS v1_payload_wal_0_poll_idx ON v1_payload_wal_0 (tenant_id, offload_at);
+CREATE INDEX CONCURRENTLY IF NOT EXISTS v1_payload_wal_1_poll_idx ON v1_payload_wal_1 (tenant_id, offload_at);
+CREATE INDEX CONCURRENTLY IF NOT EXISTS v1_payload_wal_2_poll_idx ON v1_payload_wal_2 (tenant_id, offload_at);
+CREATE INDEX CONCURRENTLY IF NOT EXISTS v1_payload_wal_3_poll_idx ON v1_payload_wal_3 (tenant_id, offload_at);
+CREATE INDEX IF NOT EXISTS v1_payload_wal_poll_idx ON v1_payload_wal (tenant_id, offload_at);
+
+-- +goose Down
+DROP INDEX IF EXISTS v1_payload_wal_poll_idx;
```
mrkaye97, line 11: per laurenz, this is a good way to go about this 🤷 

I also changed the polling queries to just order by offload at / cut over at, since with `SKIP LOCKED` I don't think we need to be worried about deadlocks (pretty sure?) and the offload time is the only thing that really matters for ordering. So because of that, was thinking that an index with just tenant + offload at is fine (and the cut over table will use the PK for that)

abelanger5 (approved):

internal/services/controllers/v1/task/controller.go
```diff
@@ -364,7 +365,25 @@ func (tc *TasksControllerImpl) Start() (func() error, error) {
...
+		span.SetStatus(codes.Error, "could not run process payload WAL")
+		span.End()
+
+		return nil, wrappedErr
+	}
+
+	_, err = tc.s.NewJob(
+		gocron.DurationJob(tc.repov1.Payloads().ExternalCutoverProcessInterval()),
+		gocron.NewTask(
+			tc.runProcessPayloadExternalCutovers(ctx),
```
mnafees:
```suggestion
			tc.runProcessPayloadExternalCutovers(spanContext),
```

mnafees (changes requested):

internal/services/controllers/v1/task/controller.go
mrkaye97: good catch, thanks!

mnafees (approved):

PR #2378, opened 2025-10-02
Cleanup job for old and invalid entries

mnafees:
# Description

This PR introduces a cron job that would run every day to clean up old and invalid entries from core tables.

## Type of change


pkg/repository/v1/sqlcv1/tasks.sql
```diff
@@ -914,3 +914,53 @@ ANALYZE v1_task;
...
 -- name: AnalyzeV1TaskEvent :exec
 ANALYZE v1_task_event;
+
+-- name: CleanupV1TaskRuntime :one
+WITH trs as (
+    SELECT vtr.task_id, vtr.task_inserted_at, vtr.retry_count
+    FROM v1_task_runtime vtr
+    LEFT JOIN v1_task vt ON vtr.task_id = vt.id
+        AND vtr.task_inserted_at = vt.inserted_at
+    WHERE vt.id IS NULL
```
mrkaye97: nit: I usally prefer the `WHERE NOT EXISTS (...)` way of doing this - I think it's supposed to be a little better for performance but I haven't tested

pkg/repository/v1/sqlcv1/tasks.sql
```diff
@@ -914,3 +914,53 @@ ANALYZE v1_task;
...
+    LIMIT sqlc.arg(batch_size)::int
```
mrkaye97: nit for consistency with other places:
```suggestion
    LIMIT @batchSize::int
```

pkg/repository/v1/sqlcv1/tasks.sql
```diff
@@ -914,3 +914,53 @@ ANALYZE v1_task;
...
+    )
+    order by task_id ASC
+    FOR UPDATE SKIP LOCKED
+)
+DELETE FROM v1_task_runtime
+WHERE (task_id, task_inserted_at, retry_count) IN (
+    SELECT task_id, task_inserted_at, retry_count
+    FROM locked_trs
+)
+RETURNING (SELECT COUNT(*) FROM locked_trs) as deleted_count;
```
mrkaye97: do we need to first lock these and then delete them? or can we just delete them?

mrkaye97: I think the `DELETE` should return the number of deleted rows automatically, so we probably don't explicitly need to have this `RETURNING`

pkg/repository/v1/task.go
```diff
@@ -3444,3 +3447,92 @@ func (r *TaskRepositoryImpl) AnalyzeTaskTables(ctx context.Context) error {
...
+			break
+		}
+
+		r.l.Debug().Int64("deleted", deleted).Int64("total", totalDeleted).Msg("cleaned up v1_task_runtime batch")
+	}
+
+	if totalDeleted > 0 {
+		r.l.Info().Int64("total_deleted", totalDeleted).Msg("cleaned up v1_task_runtime")
+	}
+
```
mrkaye97: I think maybe we should do the same thing we're doing in some other places where we return a flag for `shouldContinue` instead of having the loop inside of this. e.g. the `ProcessPayloadWAL` thing does this, might be worth taking a look there!

pkg/repository/v1/sqlcv1/tasks.sql
mnafees: cc'ing @abelanger5 for an explanation on this

mnafees: @mrkaye97 So we are doing this to make sure they are locked in a stable order before deletion to avoid deadlocks.

mrkaye97: sounds good!

pkg/repository/v1/sqlcv1/queue.sql
```diff
@@ -451,3 +451,30 @@ SELECT
...
+), locked_qis AS (
+    SELECT task_id, task_inserted_at, retry_count
+    FROM v1_queue_item
+    WHERE (task_id, task_inserted_at, retry_count) IN (
+        SELECT task_id, task_inserted_at, retry_count
+        FROM qis
+    )
+    order by id ASC
+    FOR UPDATE SKIP LOCKED
+)
```
mrkaye97: Sorry for being a PITA 😅 Can we simplify these two CTEs into a single one, or is there a reason for two? E.g. (untested):

```suggestion
WITH locked_qis as (
    SELECT qi.task_id, qi.task_inserted_at, qi.retry_count
    FROM v1_queue_item qi
    WHERE NOT EXISTS (
        SELECT 1
        FROM v1_task vt
        WHERE qi.task_id = vt.id
            AND qi.task_inserted_at = vt.inserted_at
    )
    ORDER BY id ASC
    LIMIT @batchSize::int
    FOR UPDATE SKIP LOCKED
)
```

pkg/repository/v1/sqlcv1/tasks.sql
```diff
@@ -918,3 +918,57 @@ ANALYZE v1_task;
...
+), locked_trs AS (
+    SELECT task_id, task_inserted_at, retry_count
+    FROM v1_task_runtime
+    WHERE (task_id, task_inserted_at, retry_count) IN (
+        SELECT task_id, task_inserted_at, retry_count
+        FROM trs
+    )
+    order by task_id ASC
+    FOR UPDATE SKIP LOCKED
+)
```
mrkaye97: same qq here I think

pkg/repository/v1/sqlcv1/tasks.sql
```diff
@@ -918,3 +918,57 @@ ANALYZE v1_task;
...
+), locked_cs AS (
+    SELECT task_id, task_inserted_at, task_retry_count
+    FROM v1_concurrency_slot
+    WHERE (task_id, task_inserted_at, task_retry_count) IN (
+        SELECT task_id, task_inserted_at, task_retry_count
+        FROM cs
+    )
+    order by task_id, task_inserted_at, task_retry_count, strategy_id
+    FOR UPDATE SKIP LOCKED
+)
```
mrkaye97: and here - let me know if I'm misunderstanding and there's a reason for separate CTEs!

pkg/repository/v1/task.go
```diff
@@ -3626,3 +3630,66 @@ func (r *TaskRepositoryImpl) AnalyzeTaskTables(ctx context.Context) error {
...
+	// Process one batch of v1_queue_item
+	result, err := r.queries.CleanupV1QueueItem(ctx, tx, batchSize)
+	if err != nil {
+		return false, fmt.Errorf("error cleaning up v1_queue_item: %v", err)
+	}
+
+	if result.RowsAffected() > 0 {
+		r.l.Debug().Int64("deleted", result.RowsAffected()).Msg("cleaned up v1_queue_item batch")
+		shouldContinue = true
```
mrkaye97, line 3663: I think this logic might be wrong but I'm not sure - we want `shouldContinue` to be `true` when there are _more_ rows to process, so if `RowsAffected() == batchSize` (at least I'm pretty sure). I don't think this is too problematic since if there's nothing to process it'll stop, but I think it might be a subtle bug

Let me know if I'm misunderstanding!

mrkaye97 (commented):
should we also add one for `v1_workflow_concurrency_slot`?

mrkaye97 (approved):

PR #2383, opened 2025-10-04
fix: Fix checking value in Golang ValidateJSONB

icbd:
# Description

https://github.com/hatchet-dev/hatchet/blob/2a08cbf77b98da6c21c1417ec25dfc42f075208f/sdks/python/hatchet_sdk/worker/runner/runner.py#L501-L506

This comment has explained the significance of this method, but the implementation of this method has problems.
This PR fixes the value of this check.

Fixes # (issue)

When the object returned normally contains "\u0000", the task will be reset to failure.

Source Code: 
https://github.com/hatchet-dev/hatchet/blob/5bf9f97720aba5a8095b69fc69e510ce424db4f1/internal/services/dispatcher/server_v1.go#L557-L562

When the object returned error contains "\u0000", the error content is displayed normally (No check for ValidateJSONB). 

## Type of change

## What's Changed

### DIFF
--- a/internal/services/dispatcher/server_v1.go
+++ b/internal/services/dispatcher/server_v1.go
@@ -554,7 +554,8 @@ func (s *DispatcherImpl) sendStepActionEventV1(ctx context.Context, request *con
 		s.l.Warn().Msg("retry count is nil, using task's current retry count")
 	}
 
-	if request.EventType == contracts.StepActionEventType_STEP_EVENT_TYPE_COMPLETED {
+	if request.EventType == contracts.StepActionEventType_STEP_EVENT_TYPE_COMPLETED ||
+		request.EventType == contracts.StepActionEventType_STEP_EVENT_TYPE_FAILED {
 		if err := repository.ValidateJSONB([]byte(request.EventPayload), "taskOutput"); err != nil {
 			request.EventPayload = err.Error()
 			request.EventType = contracts.StepActionEventType_STEP_EVENT_TYPE_FAILED
--- a/pkg/repository/jsonb.go
+++ b/pkg/repository/jsonb.go
@@ -1,8 +1,8 @@
 package repository
 
 import (
-	"bytes"
 	"fmt"
+	"strings"
 )
 
 func ValidateJSONB(jsonb []byte, fieldName string) error {
@@ -10,7 +10,7 @@ func ValidateJSONB(jsonb []byte, fieldName string) error {
 		return nil
 	}
 
-	if bytes.Contains(jsonb, []byte("\u0000")) {
+	if strings.Contains(string(jsonb), "\\u0000") {
 		return fmt.Errorf("encoded jsonb contains invalid null character \\u0000 in field `%s`", fieldName)
 	}
 
--- a/pkg/repository/jsonb.go
+++ b/pkg/repository/jsonb.go
@@ -1,8 +1,8 @@
 package repository
 
 import (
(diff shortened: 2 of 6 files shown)


internal/services/dispatcher/server_v1.go
```diff
@@ -554,11 +554,9 @@ func (s *DispatcherImpl) sendStepActionEventV1(ctx context.Context, request *con
 		s.l.Warn().Msg("retry count is nil, using task's current retry count")
 	}
 
-	if request.EventType == contracts.StepActionEventType_STEP_EVENT_TYPE_COMPLETED {
```
icbd, line 557: Need to double check here

mrkaye97: I think we probably only want this for completed / failed?

baodong-openmart: Updated

baodong-openmart: I don't have a full development environment, so I've only run unit tests on this change. 

Could you help run an E2E test to confirm the change? Thanks

mrkaye97: yeah, will do tomorrow AM - CI should be a pretty good indicator though 👍

pkg/repository/jsonb.go
```diff
@@ -1,16 +1,16 @@
...
-	"strings"
 )
 
 func ValidateJSONB(jsonb []byte, fieldName string) error {
 	if len(jsonb) == 0 {
 		return nil
 	}
 
-	if strings.Contains(string(jsonb), "\u0000") {
+	if bytes.Contains(jsonb, []byte("\u0000")) {
```
mrkaye97: fyi @icbd this didn't work so I reverted it - not completely sure why though tbh, it seemed fine at first glance

mrkaye97: I think this broke a unit test btw, but I validated the behavior with an E2E test. Let me know if you disagree!

internal/services/dispatcher/server_v1.go
icbd: i think the core of this PR is changing  `\\u0000` to \u0000

Using bytes just improves performance.

I'm not sure if a couple new commits are directly related to this PR.
If it's not relevant, we can submit it first and rebase it.

WDYT？

mrkaye97: For me, the issue here isn't that the `Validate` method isn't working (it is) but we just weren't calling it every place we needed to. Is that not your understanding? It's easy to verify - if you change the `string` thing to `bytes`, those python tests will fail and you'll see errors on the engine. I'm not sure why it doesn't work, but it doesn't 🤷

Also, I'm not too worried about the bytes -> string conversion for performance here, although it's a fair point

icbd: https://github.com/hatchet-dev/hatchet/pull/2383/commits/b0bee2b9bf68547c7fc02d1b9ce8d80d26af5d74

This commit is what I want, everything else is optional.

JSON encoder will filter and resolve NUL（JSON encoder transfers \u0000 to `\\u0000`

\u0000 is one character，means NUL，pg will also reject it.

`\\u0000` are six characters，are normal string.

so，you may check the invalid character NUL，instead of a normal str

mrkaye97:
closing as stale, feel free to reopen + fix the unit test there if you like!

PR #2391, opened 2025-10-07
fix: optimize concurrency slot trigger method

abelanger5:
# Description

Optimizes the concurrency slot trigger to avoid slow deletions. The PG query planner doesn't seem to like the `JOIN` conditions on `v1_dag_to_task`, `v1_concurrency_slot`, and `v1_task` simultaneously. It first selects the `v1_concurrency_slot` rows and then calls `Merge Append` on `v1_task` rows in an inefficient manner. 

## Type of change


mrkaye97 (approved):
nice, looks good - sorta bizarre that the planner couldn't figure this out but 🤷

PR #2395, opened 2025-10-07
Feat: Pausable Crons

mrkaye97:
# Description

Adding pause support for crons. Needs a bit of cleanup still, but basically is a button in the crons table which propagates down to the `WorkflowTriggerCronRef`

## Type of change


frontend/app/src/pages/main/v1/recurring/index.tsx
```diff
@@ -1,5 +1,158 @@
...
+import { useCrons } from './hooks/use-crons';
+import {
+  CronColumn,
+  workflowKey,
+  metadataKey,
+} from './components/recurring-columns';
+import { TriggerWorkflowForm } from '../workflows/$workflow/components/trigger-workflow-form';
 
-export default function Crons() {
-  return <CronsTable />;
```
mrkaye97, line 4: no significant changes beyond adding the enable button / column here, just deleted the other file since it wasn't doing anything

pkg/repository/v1/sqlcv1/workflows.sql
```diff
@@ -208,6 +208,17 @@ INSERT INTO "WorkflowTriggerEventRef" (
...
+WITH previous_trigger AS (
+    SELECT "enabled"
+    FROM "WorkflowTriggerCronRef" c
+    JOIN "WorkflowTriggers" t ON t."id" = c."parentId"
+    JOIN "WorkflowVersion" wv ON wv."id" = t."workflowVersionId"
+    WHERE
+        wv."id" = sqlc.narg('oldWorkflowVersionId')::uuid
+        AND "cron" = @cronTrigger::text
+        AND "name" = sqlc.narg('name')::text
+)
```
mrkaye97, line 220: @abelanger5 verifying this is okay to do - the goal here is to move the `enabled` flag onto the new cron ref from the old one. main risk seems to be the possibility of this returning multiple rows, which I don't think should happen but I'm not sure

pkg/repository/v1/sqlcv1/workflows.sql
```diff
@@ -391,14 +405,21 @@ WHERE
 
 -- name: MoveCronTriggerToNewWorkflowTriggers :exec
 WITH triggersToUpdate AS (
-    SELECT cronTrigger."id" FROM "WorkflowTriggerCronRef" cronTrigger
+    SELECT
+        cronTrigger."id",
+        cronTrigger."enabled"
+    FROM "WorkflowTriggerCronRef" cronTrigger
```
mrkaye97, line 411: same here - checking this is fine! I think it should be

pkg/repository/v1/sqlcv1/workflows.sql
```diff
@@ -216,17 +227,20 @@ INSERT INTO "WorkflowTriggerCronRef" (
...
+    "enabled"
+)
+SELECT
+    @workflowTriggersId::uuid AS "parentId",
+    @cronTrigger::text AS "cron",
+    sqlc.narg('name')::text AS "name",
+    sqlc.narg('input')::jsonb AS "input",
+    sqlc.narg('additionalMetadata')::jsonb AS "additionalMetadata",
+    gen_random_uuid() AS "id",
+    COALESCE(sqlc.narg('method')::"WorkflowTriggerCronRefMethods", 'DEFAULT') AS "method",
```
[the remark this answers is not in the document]
mrkaye97: 🤦

pkg/repository/v1/sqlcv1/workflows.sql
abelanger5: Yeah shouldn't be possible given the unique constraint we have on the cron - but perhaps you can return `LIMIT 1` to be safe?

abelanger5 (approved):

pkg/repository/v1/sqlcv1/workflows.sql
mrkaye97: I was nervous that `LIMIT 1` would mask an error - I'd sorta prefer it failed loudly 😅

PR #2396, opened 2025-10-07
Fix: Change noisy error log to debug

mrkaye97:
# Description

What it says on the tin - don't think this log is very helpful since it usually is just indicating that the task exceeded its retention window and was dropped

## Type of change


abelanger5 (commented):
This is an indication that data is in an inconsistent state on the scheduler, and is going to keep sending bad payloads over the MQ - I think it's quite useful actually, we just need to fix the underlying cause

mnafees:
https://github.com/hatchet-dev/hatchet/pull/2378 should be fixing the underlying issue I believe?

mrkaye97:
ah yeah true, alright will just close this - opened in frustration because I was trying to parse some logs yesterday and was flooded with these again 😬

PR #2398, opened 2025-10-08
Use `UTC` for all pgx connections and check for database TZ

### LINKED ISSUE #2355: [BUG] hatchet-lite does not work with external postgres
Following the instructions, it seems like I can't queue any jobs. Followed instructions exactly as the docs. The dashboard works, I created an API key, connected a worker but the job doesn't run. I used the simple hello world example.

I see this in the docker logs:
```
2025-08-22T08:18:04.244Z WRN API host=localhost:8080 latency=11.705958 method=GET remote_ip=::1 service=server status=404 uri=/api/v1/stable/workflow-runs/1cbe498f-be27-4a47-a8f6-494c1d96113d user_agent=OpenAPI-Generator/1.0.0/python

2025-08-22T08:18:05.59Z WRN API host=localhost:8080 latency=3.439333 method=GET remote_ip=::1 service=server status=404 uri=/api/v1/stable/workflow-runs/1cbe498f-be27-4a47-a8f6-494c1d96113d user_agent=OpenAPI-Generator/1.0.0/python
```

On top of this the latest python sdk (1.17.2) doesn't work. It doesn't install:
```
    raise InvalidVersion(f"Invalid version: {version!r}")
pip._vendor.packaging.version.InvalidVersion: Invalid version: 'otel'
```

So i'm using 1.17.1.

worker.py
```
hatchet = Hatchet(debug=True, config=ClientConfig(token=HATCHET_CLIENT_TOKEN, tls_config=ClientTLSConfig(strategy='none')))

class SimpleInput(BaseModel):
    message: str
 
@hatchet.task(name="SimpleTask", input_validator=SimpleInput)
def simple(input: SimpleInput, ctx: Context) -> dict[str, str]:
    return {
      "transformed_message": input.message.lower(),
    }


def main():
    worker = hatchet.worker('test-worker', workflows=[simple], slots=1)
    worker.start()
 
if __name__ == "__main__":
    main()
```

runner.py
```
simple.run(SimpleInput(message="HeLlO WoRlD"))
```

This should be easy to replicate. Specifically it seems to impact connecting Hatchet Lite to existing Postgres instances for some reason.

Posted this on [discord](https://discord.com/channels/1088927970518909068/1408364901008937010) a month ago, was successfully reproduced and then everyone involved stopped responding...

mnafees:
# Description

By default we now default to the UTC timezone for all pgx connections. This PR also adds a new flag `DATABASE_ENFORCE_UTC_TIMEZONE` with default `true` which will error out with a helpful message for the user.

Fixes https://github.com/hatchet-dev/hatchet/issues/2355

## Type of change

### DIFF
--- a/pkg/config/loader/loader.go
+++ b/pkg/config/loader/loader.go
@@ -141,12 +141,13 @@ func (c *ConfigLoader) InitDataLayer() (res *database.Layer, err error) {
 		return nil, err
 	}
 
-	// ref: https://github.com/jackc/pgx/issues/1549
 	config.AfterConnect = func(ctx context.Context, conn *pgx.Conn) error {
 		// Set timezone to UTC for all connections
 		if _, err := conn.Exec(ctx, "SET TIME ZONE 'UTC'"); err != nil {
 			return err
 		}
+
+		// ref: https://github.com/jackc/pgx/issues/1549
 
 		t, err := conn.LoadType(ctx, "v1_readable_status_olap")
 		if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/config/loader/loader.go
```diff
@@ -819,3 +845,37 @@ func loadInternalClient(l *zerolog.Logger, conf *server.InternalClientTLSConfigF
...
+	}
+	defer tempConn.Close(context.Background())
+
+	var dbTimezone string
+	if err := tempConn.QueryRow(context.Background(), "SHOW timezone").Scan(&dbTimezone); err != nil {
+		return fmt.Errorf("could not query %s timezone: %w", dbLabel, err)
+	}
+
+	// Accept both "UTC" and "Etc/UTC" as valid UTC timezones
+	if dbTimezone != "UTC" && dbTimezone != "Etc/UTC" {
```
abelanger5, line 864: Is this case sensitive/do we have a link to these options in the PG docs or source code?

mnafees: Could find https://doxygen.postgresql.org/findtimezone_8c.html#a724e4f891f171442adc75deed1923b02 which is the same as what the IANA recommends so I think we should be good?

abelanger5 (approved):

PR #2409, opened 2025-10-10
Fix OTel span attribute naming convention

mnafees:
# Description

Fixes span naming according to convention. Also relocates `/internal/telemetry` to `/pkg/telemetry` for broader usage.

## Type of change


mrkaye97 (approved):
nice! is the plan to change the names of the spans themselves too?

mnafees:
@mrkaye97 yes sir, as part of a following PR!

PR #2410, opened 2025-10-10
Feat: OLAP Payloads

mrkaye97:
# Description

Creates a `v1_payloads_olap` table and replicates data into it at the same time as we're writing other OLAP tables (tasks, dags, task events, events), and then also pubs a message with the external key of the payload to use when the core db has the WAL record for that payload processed. In practice, this means that in an ideal state, payloads will only live in the OLAP db for a very short amount of time before being offloaded to and read from S3.

A couple of important things here:

1. First of all, I think we made a bit of a mistake using id + inserted at + type as the key in the core payloads table, so I went for using an `external_id` here in hopes that maybe we'll eventually translate the core table to that too. I think it keeps things much simpler, but the inconsistency between here and the core db is unfortunate
2. Cleaned up some method names and signatures a bit, and removed a little bit of unused code
3. Rewired the API to pull payloads from the payloads table / S3 instead of directly from the other tables (task, dag, etc.)
4. Refactored the OLAP repo a bit to pass `FoobarWithPayload` around instead of `sqlcv1.Foobar` to make it a bit easier to reason about - the hope here is that in the near future we can stop reading the inline payloads in all these tables and then simplify that pretty significantly

## Type of change


internal/services/controllers/v1/olap/controller.go
```diff
@@ -364,11 +354,36 @@ func (tc *OLAPControllerImpl) handleBufferedMsgs(tenantId, msgId string, payload
...
+
+	msgs := msgqueue.JSONConvert[v1.OLAPPayloadsToOffload](payloads)
+
+	for _, msg := range msgs {
+		for _, payload := range msg.Payloads {
+			// todo: should we sample here?
+			// if !tc.sample(payload.ExternalLocationKey) {
+			// 	tc.l.Debug().Msgf("skipping payload offload external id %s", payload.ExternalId)
+			// 	continue
+			// }
```
abelanger5: Yeah assuming that the external id here is the workflow run external id, this should work I think

internal/services/controllers/v1/olap/controller.go
```diff
@@ -364,11 +354,36 @@ func (tc *OLAPControllerImpl) handleBufferedMsgs(tenantId, msgId string, payload
 		return tc.handleFailedWebhookValidation(context.Background(), tenantId, payloads)
 	case "cel-evaluation-failure":
 		return tc.handleCelEvaluationFailure(context.Background(), tenantId, payloads)
+	case "offload-payload":
```
abelanger5, line 369: What happens if we miss this offload-payload event?

internal/services/dispatcher/server_v1.go
```diff
@@ -588,6 +589,7 @@ func (s *DispatcherImpl) handleTaskStarted(inputCtx context.Context, task *sqlcv
 		task.ID,
 		retryCount,
 		request,
+		pgtype.UUID{},
```
abelanger5: Why is this sometimes `pgtype.UUID{}` and other times a randomly generated UUID?

internal/services/controllers/v1/olap/controller.go
mrkaye97: nothing really - if we miss the event, we just use the inline payload and never update the row, so the main risk is we miss _every_ event and disk usage goes up, but if it's a handful there should be no big effect on anything

internal/services/dispatcher/server_v1.go
mrkaye97: ah yeah so I'm using this as a nil value since I wasn't sure if using a pointer to a uuid was a code smell, although I guess I could also use a string.

re: random uuid - the reason is that for completed / failed events, we need to generate a uuid on the dispatcher so that we can write that same UUID for the task event in both `v1_task_event` and `v1_task_events_olap`, so this is what we do in `handleTaskCompleted` and `handleTaskFailed`. everywhere else, the UUID can be generated farther downstream for the core table, since we only need to tie completed / failed events to the OLAP db and nothing else (don't need to offload payloads for any of the other types of events)

not sure if that's clear - lmk if you want more info here. what do you think for passing a `nil`? make it a `*string` instead of `pgtype.UUID`?

internal/services/controllers/v1/olap/controller.go
mrkaye97: should work meaning good to uncomment this? if yes, sounds good 👍

abelanger5: yep should be good to uncomment

api/v1/server/handlers/v1/workflow-runs/get.go
```diff
@@ -88,7 +93,13 @@ func (t *V1WorkflowRunsService) getWorkflowRunDetails(
 		return nil, err
 	}
 
-	result, err := transformers.ToWorkflowRunDetails(taskRunEvents, workflowRun, shape, tasks, stepIdToTaskExternalId, workflowVersion)
+	externalIdToPayload, err := t.config.V1.OLAP().ReadPayloads(ctx, tenantId, externalIds...)
```
mrkaye97: I don't love drilling this down into the transformer but I didn't want to modify a bunch of the structs we pass around to include payloads here since it'd be a much bigger change. I think we'll need to refactor this stuff anyways, so I figured the tech debt here was fine

internal/services/controllers/v1/olap/controller.go
```diff
@@ -646,7 +672,76 @@ func (tc *OLAPControllerImpl) handleCreateMonitoringEvent(ctx context.Context, t
...
+		}] = opt.ExternalID
+
+		offloadToExternalOpts = append(offloadToExternalOpts, v1.OffloadToExternalStoreOpts{
+			StorePayloadOpts: &v1.StorePayloadOpts{
+				Id:         dummyId,
+				InsertedAt: sqlchelpers.TimestamptzFromTime(dummyInsertedAt),
+				ExternalId: opt.ExternalID,
+				Type:       sqlcv1.V1PayloadTypeTASKEVENTDATA,
+				Payload:    opt.Output,
+				TenantId:   tenantId,
```
mrkaye97, line 708: this logic is really janky - we have to do the same thing for the events too. the basic issue is they will never be updated from the core db, so we need to do the offloading / cutovers on the OLAP side. the issue is that if we want to respect the same TTL as in the core repo, we'd need another set of WAL tables here which IMO is overkill.

@abelanger5 WDYT of this? I'm not sure if there's another alternative other than either something like this or having WAL tables for these in the OLAP code

abelanger5: Yeah this seems alright to me

abelanger5 (approved):

PR #2413, opened 2025-10-13
Propagate `RunId` for all `Run()` variants

mnafees:
# Description

Right now the `RunId` is only available via `WorkflowRef`. This PR propagates the run ID for use even from the `Run()` variants as well as `TaskResult`.

## Type of change


mrkaye97 (approved):
nice!

PR #2417, opened 2025-10-14
feat: stateful polling intervals

abelanger5:
# Description

Right now, we have a number of operations which poll the database for every tenant in the system, every second. On large Hatchet instances with many tenants and a significant number of inactive tenants, this drains a lot of database connections and resources. 

This PR adds stateful intervals which are tracked in the database in a new `v1_operation_interval_settings` table. The idea is that for each operation, we increase the polling interval if there's no activity on the tenant. 

I've started with just a handful of task controller operations which aren't super critical but can drain resources pretty quickly. The idea is that we can reuse this `Interval` struct in a number of places in the future (like the concurrency/scheduling queues). 

## Type of change


internal/operation/interval.go
```diff
@@ -0,0 +1,179 @@
...
+	l               *zerolog.Logger
+	repo            v1.IntervalSettingsRepository
+	gauge           IntervalGauge
+	operationId     string
+	resourceId      string // tenant ID, queue name, etc.
+	maxJitter       time.Duration
+	startInterval   time.Duration
+	currInterval    time.Duration
+	maxInterval     time.Duration
+	noRowsCount     int
```
abelanger5: Should probably rename this `noActivityCount`

internal/operation/interval.go
```diff
@@ -0,0 +1,179 @@
...
+	maxJitter, startInterval, maxInterval time.Duration,
+	incBackoffCount int,
+	gauge IntervalGauge,
+) *Interval {
+	if maxInterval < 0 {
+		maxInterval = time.Minute
+	}
+
+	// read the current interval from the database
+	// TODO: don't use context.Background()
```
abelanger5: Address TODO

internal/services/controllers/v1/task/controller.go
```diff
@@ -55,10 +56,10 @@ type TasksControllerImpl struct {
...
 	opsPoolJitter                            time.Duration
-	timeoutTaskOperations                    *queueutils.OperationPool[string]
-	reassignTaskOperations                   *queueutils.OperationPool[string]
-	retryTaskOperations                      *queueutils.OperationPool[string]
-	emitSleepOperations                      *queueutils.OperationPool[string]
+	timeoutTaskOperations                    *operation.OperationPool
+	reassignTaskOperations                   *operation.OperationPool
+	retryTaskOperations                      *operation.OperationPool
+	emitSleepOperations                      *operation.OperationPool
 	processPayloadWALOperations              *queueutils.OperationPool[int64]
```
abelanger5, line 64: Note: I'm skipping the partition ones for now, because they don't benefit from this, but I'll move everything over to the `operation` package in a future PR.

internal/services/controllers/v1/task/controller.go
```diff
@@ -272,78 +305,6 @@ func (tc *TasksControllerImpl) Start() (func() error, error) {
 
 	spanContext, span := telemetry.NewSpan(ctx, "TasksControllerImpl.Start")
 
-	_, err = tc.s.NewJob(
```
abelanger5, line 275: One nice thing about this new package is we track crons internal to the operation, so we don't need to use `gocron` here

internal/services/controllers/v1/task/process_reassignments.go
```diff
@@ -9,33 +9,10 @@ import (
 	tasktypes "github.com/hatchet-dev/hatchet/internal/services/shared/tasktypes/v1"
 	"github.com/hatchet-dev/hatchet/internal/telemetry"
 	"github.com/hatchet-dev/hatchet/pkg/integrations/metrics/prometheus"
-	"github.com/hatchet-dev/hatchet/pkg/repository/postgres/dbsqlc"
 	"github.com/hatchet-dev/hatchet/pkg/repository/postgres/sqlchelpers"
 	"github.com/hatchet-dev/hatchet/pkg/repository/v1/sqlcv1"
 )
 
-func (tc *TasksControllerImpl) runTenantReassignTasks(ctx context.Context) func() {
```
abelanger5, line 17: We've also moved this logic into the operation pool

internal/operation/pool.go
```diff
@@ -0,0 +1,145 @@
...
+	hasInterval             bool
+	intervalRepo            v1.IntervalSettingsRepository
+	intervalMaxJitter       time.Duration
+	intervalStartInterval   time.Duration
+	intervalMaxInterval     time.Duration
+	intervalIncBackoffCount int
+	intervalGauge           IntervalGauge
+}
+
+func NewOperationPool(p *partition.Partition, ql *zerolog.Logger, operationId string, timeout time.Duration, description string, method OpMethod, fs ...func(*OperationPool)) *OperationPool {
```
abelanger5, line 51: Need to return cleanup method / clean up when shutting down

internal/operation/interval.go
```diff
@@ -0,0 +1,179 @@
...
+	currInterval, err := repo.ReadInterval(context.Background(), operationId, resourceId)
+
+	if err != nil {
+		currInterval = 0
+	}
```
mrkaye97, line 60: do we want to log this error just in case?

mrkaye97 (approved):

mnafees (approved):

PR #2420, opened 2025-10-15
[hotfix] Avoid throwing error logs from ratelimit MW for invalid API routes

mnafees:
# Description

Avoid polluting logs from invalid API route invocations. This PR gets rid of error logs such as 
```
{"time":"2025-10-15T15:33:26.428891+02:00","level":"ERROR","prefix":"echo","file":"ratelimit.go","line":"37","message":"no matching operation was found"}
```

## Type of change


mrkaye97 (commented):
thanks for the fix! I was thinking part of the problem here too is that it's using the logger from the echo context and not the same logging setup that we have everywhere else. WDYT?

mnafees:
@mrkaye97 The syntax of the log line itself changes due to using Echo's logger but it will still be caught in our OTel setup since we filter for the substring `ERR`.

mrkaye97:
> @mrkaye97 The syntax of the log line itself changes due to using Echo's logger but it will still be caught in our OTel setup since we filter for the substring `ERR`.

sounds good! is it not possible to pass e.g. the zerolog logger we use everywhere else so we can have consistent formatting?

mrkaye97 (approved):

PR #2426, opened 2025-10-16
fix: small scheduler optimizations

abelanger5:
# Description

Two small optimizations on the scheduler:
- [X] Increments the snapshot counter properly so that we only snapshot every 20 iterations (currently we snapshot every time)
- [X] Adds a wrapping context to the `replenish` method so we get errors if database reads are too slow

## Type of change


pkg/scheduling/v1/scheduler.go
```diff
@@ -410,6 +412,8 @@ func (s *Scheduler) loopSnapshot(ctx context.Context) {
 			}
 
 			s.exts.ReportSnapshot(sqlchelpers.UUIDToStr(s.tenantId), in)
+
+			count++
```
mrkaye97, line 416: 😅

mrkaye97 (approved):

PR #2427, opened 2025-10-16
Fix nil error in `handleTaskBulkAssignedTask`

mnafees:
# Description

Fixes an issue when bulk retrieve for payload inputs returns an error by making sure to continue to the next iteration like the other operations in the same function do.

## Type of change


internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -240,10 +240,12 @@ func (d *DispatcherImpl) handleTaskBulkAssignedTask(ctx context.Context, msg *ms
 		inputs, err := d.repov1.Payloads().BulkRetrieve(ctx, retrievePayloadOpts...)
 
 		if err != nil {
-			d.l.Error().Err(err).Msgf("could not bulk retrieve inputs for %d tasks", len(bulkDatas))
 			for _, task := range bulkDatas {
 				requeue(task)
 			}
+
+			d.l.Error().Err(err).Msgf("could not bulk retrieve inputs for %d tasks", len(bulkDatas))
+			continue
```
mrkaye97: I might need to look at the git history for this... do we want to `continue` here?

mnafees: It is leading to a nil map error in the following lines. We can make sure that `BulkRetrieve` always returns a non-nil map if the logic requires this to not continue the parent iteration.

mrkaye97 (approved):

PR #2433, opened 2025-10-20
New tenant task stats endpoint

mnafees:
# Description

This PR introduces a new endpoint `/api/v1/tenants/{tenant}/task-stats` which returns the current stats for queued and running tasks in the form
```json
{
  "task1": {
    "queued": {
      "total": 23,
      "concurrency": [
        {
          "expression": "CEL EXPR",
          "type": "GROUP_ROUND_ROBIN",
          "keys": {
            "key1": 1,
            "key2": 3
          }
        }
      ],
      "queues": {
        "task1:task1": 1,
        "task1:task11": 2
      }
    },
    "running": {
      "total": 23,
      "concurrency": [
        {
          "expression": "CEL EXPR",
          "type": "GROUP_ROUND_ROBIN",
          "keys": {
            "key1": 1,
            "key2": 3
          }
        }
      ]
    }
  },
  "task2": {}
}
```

## Type of change


pkg/repository/v1/sqlcv1/tasks.sql
```diff
@@ -976,3 +976,101 @@ SELECT
...
+        v1_task t ON qi.task_id = t.id AND qi.task_inserted_at = t.inserted_at
+    JOIN
+        "Workflow" w ON t.workflow_id = w.id
+    WHERE
+        qi.tenant_id = @tenantId::uuid
+        AND t.tenant_id = @tenantId::uuid
+        AND w."deletedAt" IS NULL
+        AND (t.concurrency_keys IS NULL OR cardinality(t.concurrency_keys) = 0)
+    GROUP BY
+        w.name
```
abelanger5: What is the purpose of this? If the point is to query for items in the concurrency queues, this doesn't accomplish that. We need to query `v1_concurrency_slot` directly. 

Also, let's not forget about `v1_rate_limit_queue_item` and `v1_retry_queue_item` as potential sources of queued tasks.

pkg/repository/v1/sqlcv1/tasks.sql
```diff
@@ -976,3 +976,101 @@ SELECT
...
+        COUNT(*) as count
+    FROM
+        v1_queue_item qi
+    JOIN
+        v1_task t ON qi.task_id = t.id AND qi.task_inserted_at = t.inserted_at
+    JOIN
+        "Workflow" w ON t.workflow_id = w.id
+    WHERE
+        qi.tenant_id = @tenantId::uuid
+        AND t.tenant_id = @tenantId::uuid
```
abelanger5: Need to be a bit careful with these if they're not in the pkey - have we checked how this performs on a very large database?

pkg/repository/v1/sqlcv1/tasks.sql
abelanger5: On second glance it looks like the purpose is to join the items with concurrency keys with items that don't have a concurrency key. This is not the correct way to list items that are in a queued state in the concurrency queues. The correct way is to find `v1_concurrency_slots` where `is_filled=false`, which implies that they are queued in the concurrency queues. If they are in a filled slot, they are no longer queued in that concurrency queue.

abelanger5 (changes requested):

pkg/repository/v1/task.go
```diff
@@ -3705,3 +3707,114 @@ func (r *TaskRepositoryImpl) Cleanup(ctx context.Context) (bool, error) {
...
+// TaskStat represents the statistics for a single task step
+type TaskStat struct {
+	Queued  *TaskStatusStat `json:"queued,omitempty"`
+	Running *TaskStatusStat `json:"running,omitempty"`
+}
+
+// TaskStatusStat represents statistics for a specific task status (queued or running)
+type TaskStatusStat struct {
+	Total       int64             `json:"total"`
+	Queues      []string          `json:"queues,omitempty"`
```
abelanger5: I was thinking queues would be a map of queue names to counts, no?

api-contracts/openapi/components/schemas/tenant.yaml
```diff
@@ -403,3 +403,44 @@ TenantStepRunQueueMetrics:
...
+
+TaskStatusStat:
+  properties:
+    total:
+      type: integer
+      format: int64
+    queues:
+      type: array
+      items:
+        type: string
```
abelanger5: See comment in the repository about this - I think this should be map of queue names to counts

abelanger5 (changes requested):

abelanger5 (approved):

PR #2435, opened 2025-10-20
Do not lowercase workflow names in the TypeScript SDK

mnafees:
# Description

The TS SDK forcefully lowercases the workflow names while registering it on the worker. This breaks other operations if the workflow name contains uppercase letters.

## Type of change


abelanger5 (commented):
Is this a breaking change? What happens if someone is running an API for triggering workflows on an older version version of the SDK vs their workers? What happens to queued runs if the SDK version is updated on a new set of workers? I'd like to make sure we've thought through all edge cases here

mrkaye97 (commented):
I think this is a breaking change, unfortunately - if someone has WorkflowA queued and then updates their SDK and reboots the worker, that run will stay queued until it times out (at least I'm pretty sure - I think the engine is case-sensitive).

I'm not really sure what the best way around this is, but at the very least it shouldn't be a patch version

mnafees:
Closing this for now and should revisit later with better testing and a proper rollout plan.

PR #2448, opened 2025-10-27
Retry RMQ messages indefinitely with aggressive logging after 5 retries

mnafees:
# Description

Modified RabbitMQ message queue retry behavior to continue retrying failed messages indefinitely instead of dropping them after a configured retry limit. Messages that fail more than 5 times now trigger aggressive error logging with full context.

## Type of change

### DIFF
--- a/internal/msgqueue/rabbitmq/rabbitmq.go
+++ b/internal/msgqueue/rabbitmq/rabbitmq.go
@@ -643,12 +643,13 @@ func (t *MessageQueueImpl) subscribe(
 							// message was rejected before
 							deathCount := xDeath[0].(amqp.Table)["count"].(int64)
 
-							t.l.Error().
-								Int64("death_count", deathCount).
-								Str("message_id", msg.ID).
-								Str("tenant_id", msg.TenantID()).
-								Int("max_death_count", t.maxDeathCount).
-								Msg("message has been rejected")
+							if deathCount > 5 {
+								t.l.Error().
+									Int64("death_count", deathCount).
+									Str("message_id", msg.ID).
+									Str("tenant_id", msg.TenantID()).
+									Msgf("message has been retried for %d times", deathCount)
+							}
 
 							if t.enableMessageRejection && deathCount > int64(t.maxDeathCount) {
 								t.l.Error().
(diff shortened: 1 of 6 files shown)


internal/msgqueue/rabbitmq/rabbitmq.go
```diff
@@ -626,15 +626,12 @@ func (t *MessageQueueImpl) subscribe(
...
-							if deathCount > int64(msg.Retries) {
-								t.l.Debug().Msgf("message %s has been rejected %d times, not requeuing", msg.ID, deathCount)
-
-								// acknowledge so it's removed from the queue
-								if err := rabbitMsg.Ack(false); err != nil {
-									t.l.Error().Msgf("error acknowledging message: %v", err)
-								}
-
-								return
+							if deathCount > 5 {
```
abelanger5, line 646: Let's add a config var to:
1. Enable rejecting messages after a certain number of deaths
2. Set the max death count

We are going to want this if we need to purge messages in an emergency, for example.

abelanger5 (changes requested):

abelanger5:
Also, we are now failing the integration test:

```
        	Error:      	message exceeded maximum retry count
        	Test:       	TestDeadLetteringExceedRetriesFailure
```

Let's run the integration test with the config var I was referencing above?

internal/msgqueue/v1/rabbitmq/rabbitmq.go
```diff
@@ -757,16 +773,25 @@ func (t *MessageQueueImpl) subscribe(
...
 					deathCount := xDeath[0].(amqp.Table)["count"].(int64)
 
-					t.l.Debug().Msgf("message has been rejected %d times", deathCount)
+					t.l.Error().
+						Int64("death_count", deathCount).
+						Str("message_id", msg.ID).
+						Str("tenant_id", msg.TenantID).
+						Int("num_payloads", len(msg.Payloads)).
+						Int("max_death_count", t.maxDeathCount).
+						Msg("message has been rejected")
```
abelanger5: Nit: this is a confusing message. The error can instead be `message has been retried %d times` and perhaps we should only log this after a significant number of retries, for example 3 or 5 retries?

abelanger5 (changes requested):

abelanger5 (approved):

PR #2457, opened 2025-10-29
feat: add health and metrics api on typescript sdk worker

### LINKED ISSUE #2390: [FEAT] Better observibility for Typescript workers with Healthchecks to deploy inside a docker container
I am using the self hosted version of hatchet. I have an express webapp that is being deployed to a managed container service. I have done a similar setup for the hatchet workers and deploying it inside of a container. It works however as far as I can see, healthchecks is only supported on the python version of the SDK. This is a huge issue as I want to have alerting setup if the container goes down. 

Are there any plans to bring this to the typescript sdk? I am **_naively_** imagining this shouldn't be too much work as it is just a healthcheck endpoint :). If not, would there be a way to define this ourselves?

On the same "feature parity" and observability topic, ideally we would like to have the opentelemtry instrumentation for the typescript SDK as well. Are there any plans for this?

jishnundth:
## Description
Added health and metrics api on typescript SDK worker. The behavior is consistent with the current health and metrics API in other SDKs.

Partly Fixes #2390 : OpenTelemetry changes for TS SDK is not added. 

## Type of change

## What's Changed

- [x] feat: health and metrics API on typescript sdk worker
- [x] add: prom-client to fetch metrics data
- [x] add: track health status of worker across different states
- [x] tests: config tests for new health_enabled and health_port env variables

### DIFF
--- a/sdks/typescript/package.json
+++ b/sdks/typescript/package.json
@@ -85,6 +85,7 @@
     "long": "^5.3.1",
     "nice-grpc": "^2.1.12",
     "nice-grpc-common": "^2.0.2",
+    "prom-client": "^15.1.3",
     "protobufjs": "^7.4.0",
     "qs": "^6.14.0",
     "semver": "^7.7.1",
--- a/sdks/typescript/package.json
+++ b/sdks/typescript/package.json
@@ -85,12 +85,14 @@
     "long": "^5.3.1",
     "nice-grpc": "^2.1.12",
     "nice-grpc-common": "^2.0.2",
-    "prom-client": "^15.1.3",
     "protobufjs": "^7.4.0",
     "qs": "^6.14.0",
     "semver": "^7.7.1",
     "yaml": "^2.7.1",
     "zod": "^3.24.2"
+  },
+  "optionalDependencies": {
+    "prom-client": "^15.1.3"
   },
   "packageManager": "pnpm@10.16.1"
 }
--- a/sdks/typescript/package.json
+++ b/sdks/typescript/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hatchet-dev/typescript-sdk",
-  "version": "1.9.8",
+  "version": "1.10.0",
   "description": "Background task orchestration & visibility for developers",
   "types": "dist/index.d.ts",
   "files": [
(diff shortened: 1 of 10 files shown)


sdks/typescript/package.json
```diff
@@ -85,6 +85,7 @@
     "long": "^5.3.1",
     "nice-grpc": "^2.1.12",
     "nice-grpc-common": "^2.0.2",
+    "prom-client": "^15.1.3",
```
mrkaye97: would it be possible to make this dependency optional? it was a mistake to not do that in the Python SDK

sdks/typescript/src/v1/client/worker/health-server.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  private async handleRequest(req: IncomingMessage, res: ServerResponse): Promise<void> {
+    const url = req.url || '';
+
+    if (url === '/health' && req.method === 'GET') {
+      await this.handleHealth(res);
+    }
+    else if (url === '/metrics' && req.method === 'GET') {
+      this.handleMetrics(res);
```
mrkaye97: any reason to not `await` this one?

sdks/typescript/src/v1/client/worker/worker-internal.ts
```diff
@@ -784,17 +845,38 @@ export class V1Worker {
...
+        this.setStatus(WorkerStatus.UNHEALTHY);
+        return;
+      }
+    }
+
     // ensure all workflows are registered
     await Promise.all(this.registeredWorkflowPromises);
 
     if (Object.keys(this.action_registry).length === 0) {
+      this.setStatus(WorkerStatus.HEALTHY);
```
mrkaye97: not 100% sure if we need this, since if there are no actions then IIRC the worker will just not start up?

mrkaye97 (commented):
left a few small comments - thanks for putting in this change!

sdks/typescript/src/v1/client/worker/worker-internal.ts
jishnundth: Aah, you’re right. I’ll remove it.

jishnundth:
@mrkaye97 PTAL at the mentioned changes

mrkaye97 (commented):
looks great, thanks for the work here! you'll need to also bump the minor version in the `package.json`, and then I'll approve 🚀

jishnundth:
> looks great, thanks for the work here! you'll need to also bump the minor version in the `package.json`, and then I'll approve 🚀

Done✅

mrkaye97 (approved):

PR #2460, opened 2025-10-30
Logs for liveness and readiness endpoints + PG conn stats

mnafees:
# Description

Adds logging to the liveness and readiness endpoints. Also add PG connection stats.

## Type of change


api/v1/server/handlers/metadata/health.go
```diff
@@ -1,41 +1,84 @@
...
+func (u *MetadataService) logHealthErrors(errs []error) {
+	for _, err := range errs {
+		u.config.Logger.Err(err).Msg("health check failed")
 	}
+}
 
-	if !u.config.EngineRepository.Health().IsHealthy() {
-		return nil, fmt.Errorf("engine repository is not healthy")
+func (u *MetadataService) LivenessGet(ctx echo.Context, request gen.LivenessGetRequestObject) (gen.LivenessGetResponseObject, error) {
+	allErrors := u.collectHealthErrors()
```
abelanger5: Let's also set a context timeout which is shorter than our default readiness/liveness probe timeouts, so we can guarantee that we log something

abelanger5 (approved):

PR #2474, opened 2025-10-31
feat: add grpc otel spans, better tx debugging

abelanger5:
# Description

Adds improved debugging for transactions that error out when we time out acquiring a connection from the pool, and adds opentelemetry for gRPC. 

## Type of change

### DIFF
--- a/internal/services/dispatcher/server_v1.go
+++ b/internal/services/dispatcher/server_v1.go
@@ -412,7 +412,7 @@ func (s *DispatcherImpl) subscribeToWorkflowRunsV1(server contracts.Dispatcher_S
 		}
 
 		for _, event := range events {
-			err := sendEvent(ctx, event)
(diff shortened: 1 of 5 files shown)


internal/services/dispatcher/server_v1.go
```diff
@@ -405,7 +412,7 @@ func (s *DispatcherImpl) subscribeToWorkflowRunsV1(server contracts.Dispatcher_S
 		}
 
 		for _, event := range events {
-			err := sendEvent(event)
+			err := sendEvent(ctx, event)
```
mnafees: Should this be using the ctx from sendEventSpan?

abelanger5: Nice catch!

mnafees (approved):

PR #2479, opened 2025-11-02
generate_external_ids_for_child_workflows

grutt:
# Description

Adds otel spans for debugging

## Type of change


abelanger5:
These should be enabled with the recent gRPC otel changes

PR #2492, opened 2025-11-05
Immediate Payload Offloads OLAP Wiring

mrkaye97:
# Description

Adding support for immediate payload offloads on the OLAP side. Basically boils down to making `PutPayloads` write the payloads to S3 immediately if the external store is enabled and the WAL is disabled, and then use those keys for persisting payloads to the db so we never actually write the contents.

## Type of change

### DIFF
--- a/pkg/repository/v1/sqlcv1/olap.sql
+++ b/pkg/repository/v1/sqlcv1/olap.sql
@@ -1626,7 +1626,8 @@ WITH inputs AS (
         UNNEST(@insertedAts::TIMESTAMPTZ[]) AS inserted_at,
         UNNEST(@payloads::JSONB[]) AS payload,
         UNNEST(@tenantIds::UUID[]) AS tenant_id,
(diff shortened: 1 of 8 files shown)


mnafees (approved):

pkg/repository/v1/sqlcv1/olap.sql
```diff
@@ -1644,7 +1645,7 @@ SELECT
     i.inserted_at,
     i.location,
     CASE
-        WHEN i.location = 'EXTERNAL' THEN i.payload
```
abelanger5, line 1647: I'm confused by what this used to be - we were writing the payload as the key?

abelanger5 (approved):

pkg/repository/v1/sqlcv1/olap.sql
mrkaye97: copy paste bug with no effect since it'd never be written as `EXTERNAL` 🤦

PR #2496, opened 2025-11-06
Fix: Propagate parent id through to `V1TaskSummary` properly

mrkaye97:
# Description

Fixing a small API QoL bug

## Type of change

### DIFF
--- a/api/v1/server/oas/transformers/v1/tasks.go
+++ b/api/v1/server/oas/transformers/v1/tasks.go
@@ -45,10 +45,11 @@ func ToTaskSummary(task *v1.TaskWithPayloads) gen.V1TaskSummary {
 		durationPtr = &duration
 	}
 
-	var parentTaskExternalId uuid.UUID
+	var parentTaskExternalId *uuid.UUID
 
 	if task.ParentTaskExternalID.Valid {
-		parentTaskExternalId = uuid.MustParse(sqlchelpers.UUIDToStr(task.ParentTaskExternalID))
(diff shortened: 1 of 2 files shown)


mnafees (approved):

api/v1/server/oas/transformers/v1/tasks.go
```diff
@@ -45,6 +45,12 @@ func ToTaskSummary(task *v1.TaskWithPayloads) gen.V1TaskSummary {
 		durationPtr = &duration
 	}
 
+	var parentTaskExternalId uuid.UUID
```
mnafees: Let us make this a pointer by default so that we do not pass all zero UUIDs from the API in case of an invalid UUID.

api/v1/server/oas/transformers/v1/workflow_runs.go
```diff
@@ -69,6 +69,11 @@ func WorkflowRunDataToV1TaskSummary(task *v1.WorkflowRunData, workflowIdsToNames
 	}
 	attempt := retryCount + 1
 
+	var parentTaskExternalId uuid.UUID
```
mnafees: Same as above.

mnafees (changes requested):

api/v1/server/oas/transformers/v1/tasks.go
mrkaye97: good call 👍

mnafees (approved):

PR #2515, opened 2025-11-12
Batch poll for user cron workflow runs

mnafees:
# Description

Batch process user crons instead of all at once.

## Type of change

### DIFF
--- a/pkg/repository/postgres/dbsqlc/tickers.sql
+++ b/pkg/repository/postgres/dbsqlc/tickers.sql
@@ -110,6 +110,7 @@ WITH latest_workflow_versions AS (
 ),
 active_cron_schedules AS (
     SELECT
+        cronSchedule."id",
         cronSchedule."parentId",
         versions."id" AS "workflowVersionId",
         triggers."tenantId" AS "tenantId"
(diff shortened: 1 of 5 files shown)


pkg/repository/postgres/dbsqlc/tickers.sql
```diff
@@ -131,7 +132,10 @@ active_cron_schedules AS (
             )
             OR "tickerId" = @tickerId::uuid
         )
+    ORDER BY cronSchedule."id" ASC
```
mrkaye97, line 135: any idea how the query plan looks for this? I think the only index on this table is on (parentId, cron, key), although based on the where clauses I figure we're doing a sequential scan here anyways

pkg/repository/postgres/dbsqlc/tickers.sql
```diff
@@ -131,7 +132,10 @@ active_cron_schedules AS (
...
     FOR UPDATE SKIP LOCKED
+    LIMIT @batchSize::integer
+    OFFSET COALESCE(sqlc.narg('offset')::integer, 0)
```
mrkaye97, line 138: just flagging that I'm a little nervous about doing limit-offset pagination since it can be pretty inefficient, but it's probably fine here if the number of batches is relatively small

mnafees:
Superseded by https://github.com/hatchet-dev/hatchet/pull/2524

PR #2517, opened 2025-11-12
Fix: Event lookup table PK

mrkaye97:
# Description

Removes the tenant id from the PK of the event lookup table so that queries stop timing out. Basically the same as our usual process of creating a new table with a trigger, then backfilling, then cutting over.

## Type of change


cmd/hatchet-migrate/migrate/migrations/20251112202112_v1_0_54.sql
```diff
@@ -0,0 +1,203 @@
...
+        )::timestamp AS range_start
+    FROM pg_inherits
+    JOIN pg_class parent ON pg_inherits.inhparent = parent.oid
+    JOIN pg_class child ON pg_inherits.inhrelid = child.oid
+    WHERE parent.relname = 'v1_event_lookup_table_olap'
+    ORDER BY range_start
+)
+
+SELECT create_v1_weekly_range_partition('v1_event_lookup_table_olap_new'::text, range_start::DATE)
+FROM existing_partitions;
```
mrkaye97, line 31: thought this was a good trick - using the existing partitions to create the new ones, so we never miss any

PR #2523, opened 2025-11-14
Attempt to fix pgx multi dimensional slice reflection error #1

mnafees:
# Description

Make sure the concurrency key results are set to `FAILED` in case of failures so that the cardinality is maintained which leads to errors from pgx's multi dimensional slice reflection issues.

## Type of change


abelanger5 (approved):
Looks good!

PR #2533, opened 2025-11-17
Feat: Initial cross-domain identify setup

mrkaye97:
# Description

Setting up cross-domain identify in PostHog following [this guide](https://posthog.com/tutorials/cross-domain-tracking). Should just need to pass the session id and distinct id as hash params in the URL on redirect now.

One thing I wasn't clear on is if it's possible to get either one or the other, but not both, of the params, but I think it's not based on:

> You need to pass users' distinct_id and session_id between PostHog initializations

## Type of change

### DIFF
--- a/frontend/app/src/components/molecules/analytics-provider.tsx
+++ b/frontend/app/src/components/molecules/analytics-provider.tsx
@@ -13,19 +13,19 @@ const AnalyticsProvider: React.FC<
 > = ({ user, children }) => {
   const meta = useApiMeta();
 
-  // allows for cross-domain tracking with PostHog
-  // see: https://posthog.com/tutorials/cross-domain-tracking
-  // for setup instructions and more details
-  const hashParams = new URLSearchParams(window.location.hash.substring(1));
-  const distinctId = hashParams.get('distinct_id');
-  const sessionId = hashParams.get('session_id');
-
   const [loaded, setLoaded] = React.useState(false);
 
   const { tenant } = useTenant();
   const { identify } = useAnalytics();
 
   const config = useMemo(() => {
+    if (import.meta.env.DEV) {
+      return {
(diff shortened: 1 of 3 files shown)


frontend/app/src/components/molecules/analytics-provider.tsx
```diff
@@ -19,6 +23,13 @@ const AnalyticsProvider: React.FC<
   const { identify } = useAnalytics();
 
   const config = useMemo(() => {
+    if (import.meta.env.DEV) {
+      return {
+        apiKey: import.meta.env.VITE_PUBLIC_POSTHOG_KEY,
+        apiHost: import.meta.env.VITE_PUBLIC_POSTHOG_HOST,
+      };
+    }
```
mrkaye97, line 31: did this for local dev 🤷 figure it's fine to leave in but can also remove it

frontend/app/src/hooks/use-analytics.tsx
```diff
@@ -61,6 +64,11 @@ export function useAnalytics(): UseAnalyticsReturn {
       } catch (error) {
         console.warn('Analytics identify failed:', error);
       }
+
+      // important: clear out the distinct_id and session_id from local storage
+      // after identifying the user so we don't re-bootstrap over and over
+      localStorage.removeItem(POSTHOG_DISTINCT_ID_LOCAL_STORAGE_KEY);
+      localStorage.removeItem(POSTHOG_SESSION_ID_LOCAL_STORAGE_KEY);
```
mrkaye97, line 71: I think this is the best place to do this, but happy to hear other suggestions here :) I think the main risk is that if we don't clear these values, we keep bootstrapping the same session every time, but we can't remove them immediately after we initialize posthog because we're more likely to lose the data if there's a page crash, someone navigates away, etc. This felt like a good middle ground

abelanger5 (approved):

PR #2544, opened 2025-11-19
[Go SDK] Resubscribe and get a new listener stream when gRPC connections fail

mnafees:
# Description

Using the cached listener stream when the gRPC server has been restarted or issues a cancelled error leads to unexpectedly losing out workflow run events. This PR fixes the issue at hand.

## Type of change


pkg/client/listener.go
```diff
@@ -196,26 +196,33 @@ func (l *WorkflowRunsListener) RemoveWorkflowRun(
...
+
+		if retryErr := l.retrySubscribe(context.Background()); retryErr != nil {
+			l.l.Error().Err(retryErr).Msg("failed to resubscribe after send failure")
+		}
+
 		time.Sleep(DefaultActionListenerRetryInterval)
 	}
 
-	return fmt.Errorf("could not send to the worker after %d retries", DefaultActionListenerRetryCount)
+	return fmt.Errorf("could not send to the worker after %d retries, please restart the worker and try again", DefaultActionListenerRetryCount)
```
grutt: this is confusing since it might not be a worker on the subscription end. should we be casing on this on the client directly?

grutt (approved):

abelanger5 (approved):

PR #2548, opened 2025-11-20
Fix invalid memory dereference in `insertTasks` when no step config exists for a task

mnafees:
# Description

Do not panic due to invalid memory dereference in `insertTasks()` when step config does not exist for a task.

## Type of change


mrkaye97 (approved):
so this looks fine to me, but how does this happen in practice?

mnafees:
@mrkaye97 `ListStepsByIds` cases on if the workflow is not deleted and if it is then the respective step config will not be returned.

mnafees:
Superseded by https://github.com/hatchet-dev/hatchet/pull/2976

PR #2549, opened 2025-11-20
Initialize concurrency keys slice for replayed tasks

mnafees:
# Description

We initialize concurrency keys in a certain fashion in `insertTasks` but fail to do so in `replayTasks` which can lead to the multi-dimensional slice being nil or with a different cardinality as expected.

## Type of change

### DIFF
--- a/pkg/repository/v1/task.go
+++ b/pkg/repository/v1/task.go
@@ -1826,13 +1826,13 @@ func (r *sharedRepository) insertTasks(
 
 		concurrencyKeys[i] = make([]string, 0)
 
-		if strats, ok := concurrencyStrats[task.StepId]; ok {
-			// we write any parent strategy ids to the task regardless of initial state, as we need to know
-			// when to release the parent concurrency slot
-			taskParentStrategyIds := make([]pgtype.Int8, 0)
-			taskStrategyIds := make([]int64, 0)
-			emptyConcurrencyKeys := make([]string, 0)
+		// we write any parent strategy ids to the task regardless of initial state, as we need to know
+		// when to release the parent concurrency slot
+		taskParentStrategyIds := make([]pgtype.Int8, 0)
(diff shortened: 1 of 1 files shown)


pkg/repository/v1/task.go
```diff
@@ -2260,6 +2260,18 @@ func (r *sharedRepository) replayTasks(
...
+
+			for range strats {
+				emptyConcurrencyKeys = append(emptyConcurrencyKeys, "")
+			}
+
+			concurrencyKeys[i] = emptyConcurrencyKeys
+		} else {
+			concurrencyKeys[i] = make([]string, 0)
+		}
+
```
abelanger5: this looks like a bug to me. no matter what, on replay we're setting the concurrency keys to an empty string? this means that those tasks won't have the correct concurrency keys assigned, right?

pkg/repository/v1/task.go
```diff
@@ -1826,13 +1826,13 @@ func (r *sharedRepository) insertTasks(
...
-		taskStrategyIds := make([]int64, 0)
-		emptyConcurrencyKeys := make([]string, 0)
-
 		if strats, ok := concurrencyStrats[task.StepId]; ok {
+			// we write any parent strategy ids to the task regardless of initial state, as we need to know
+			// when to release the parent concurrency slot
+			taskParentStrategyIds := make([]pgtype.Int8, 0)
+			taskStrategyIds := make([]int64, 0)
+			emptyConcurrencyKeys := make([]string, 0)
+
```
abelanger5: This seems like it might introduce a bug? If we don't have a step in the `concurrencyStrats` map, then we never set `parentStrategyIds[i]` to an empty slice, so it's nil instead? I'm not sure if that would lead to an issue or not.

abelanger5 (changes requested):

abelanger5 (approved):

PR #2575, opened 2025-11-26
CI: Only run full Python matrix on changes to Python SDK

mrkaye97:
# Description

Small CI tweak to only run the full Python matrix on changes to the SDK itself, and otherwise just test against 3.13 if the SDK hasn't changed

## Type of change


mrkaye97:
closing for now since the path filtering logic is too brittle

PR #2586, opened 2025-12-01
Feat: Job for payload cutovers to external

mrkaye97:
# Description

Adds a job that processes old payload partitions, offloads the data to an external store, and then finally swaps in the new partition for the existing one so that we no longer store the payloads inline. The basic idea is that we:

1. Try to acquire an advisory lock, and if we do, then create a new table, set up some triggers on the source partition to replicate any changes over
2. Backfill the existing data in batches (write the batches to external, store the key, write the key to the new partition)
3. Swap in the new partition for the existing one. Importantly, we use a hack here where we create the partition's check constraint for the bounds manually so that we don't need to validate it at attach time, to limit how long we have to hold the exclusive lock on the parent table for.

## Type of change


internal/services/controllers/v1/task/controller.go
```diff
@@ -333,25 +331,10 @@ func (tc *TasksControllerImpl) Start() (func() error, error) {
...
-
-		return nil, wrappedErr
-	}
-
-	_, err = tc.s.NewJob(
-		gocron.DurationJob(tc.repov1.Payloads().ExternalCutoverProcessInterval()),
+		// for local dev, set a short process interval and switch
+		// the two lines below so that the duration job is used
+		// gocron.DurationJob(tc.repov1.Payloads().ExternalCutoverProcessInterval()),
+		gocron.DailyJob(1, gocron.NewAtTimes(gocron.NewAtTime(3, 0, 0))),
```
abelanger5: Don't we want to be running this continuously (like once per minute) so that if the controller we're running on crashes, we pick it up quickly?

internal/services/controllers/v1/task/process_payload_wal.go
```diff
@@ -4,38 +4,14 @@ import (
...
 		tc.l.Debug().Msgf("processing payload external cutovers")
 
-		partitions := []int64{0, 1, 2, 3}
-
-		tc.processPayloadExternalCutoversOperations.SetPartitions(partitions)
-
-		for _, partitionId := range partitions {
-			tc.processPayloadExternalCutoversOperations.RunOrContinue(partitionId)
-		}
+		tc.processPayloadExternalCutoversOperations.RunOrContinue(0)
```
abelanger5: Don't think we need to use the `operations` here - we should just be able to call the function directly and use the gocron option `gocron.WithSingletonMode(gocron.LimitModeReschedule)`

pkg/repository/v1/payloadstore.go
```diff
@@ -160,117 +141,30 @@ func (p *payloadStoreRepositoryImpl) Store(ctx context.Context, tx sqlcv1.DBTX,
...
-			externalLocationKeys = append(externalLocationKeys, "")
-
-			if p.externalStoreEnabled {
-				offloadAts = append(offloadAts, pgtype.Timestamptz{Time: payload.InsertedAt.Time.Add(*p.inlineStoreTTL), Valid: true})
-			} else {
-				offloadAts = append(offloadAts, pgtype.Timestamptz{Time: time.Now(), Valid: true})
-			}
-		}
-	} else {
-		// WAL_ENABLED = false: Skip inline, go directly to external store, we will:
```
abelanger5, line 197: Hmm, I think we might want to keep the immediate offload logic (as much as it pains me with the additional complexity) - it's a very useful "break-glass" type feature in case we're filling up an NVMe disk quickly, since on busy databases it can take many hours to get a replica ready.

pkg/repository/v1/payloadstore.go
```diff
@@ -393,315 +271,300 @@ func (p *payloadStoreRepositoryImpl) retrieve(ctx context.Context, tx sqlcv1.DBT
...
+				tableName,
+			),
+			tenantIds,
+			ids,
+			insertedAts,
+			externalIds,
+			types,
+			locations,
+			externalLocationKeys,
+		)
```
abelanger5: nit: move this into a separate method? also the `fmt.Sprintf` is setting off alarm bells here, it looks fine though.

pkg/repository/v1/payloadstore.go
```diff
@@ -393,315 +271,300 @@ func (p *payloadStoreRepositoryImpl) retrieve(ctx context.Context, tx sqlcv1.DBT
...
+		if err != nil {
+			return fmt.Errorf("failed to copy offloaded payloads into temp table: %w", err)
 		}
 
-		tenantIdToPayloads[updatedPayload.TenantID.String()] = append(tenantIdToPayloads[updatedPayload.TenantID.String()], OLAPPayloadToOffload{
-			ExternalId:          updatedPayload.ExternalID,
-			ExternalLocationKey: updatedPayload.ExternalLocationKey.String,
-		})
-	}
+		offset += int64(len(payloads))
```
abelanger5: Just making extra sure: there's no correctness bug when using the offset and limit here, since we're iterating over unique rows?

abelanger5: Also: what happens if old data gets inserted? We might end up re-iterating over the same rows multiple times? Might be worth double-checking that it's safe from a:
1. External store perspective (if we try to process a row that's already externally stored, we don't just write an empty payload accidentally)
2. From a payload write perspective

pkg/repository/v1/payloadstore.go
```diff
@@ -393,315 +271,300 @@ func (p *payloadStoreRepositoryImpl) retrieve(ctx context.Context, tx sqlcv1.DBT
...
-	externalLocationKeys := make([]string, 0, len(retrieveOptsToStoredKey))
+	if err := commit(ctx); err != nil {
+		rollback()
+		return fmt.Errorf("failed to commit copy offloaded payloads transaction: %w", err)
+	}
 
-	for _, opt := range retrieveOpts {
-		offloadAt, exists := retrieveOptsToOffloadAt[opt]
+	for true {
+		tx, commit, rollback, err = sqlchelpers.PrepareTx(ctx, p.pool, p.l, 10000)
```
abelanger5: would be a little careful of overwriting tx and rollback here - would probably be safer to refactor this into a new method with those values scoped to the function.

pkg/repository/v1/payloadstore.go
```diff
@@ -393,315 +271,300 @@ func (p *payloadStoreRepositoryImpl) retrieve(ctx context.Context, tx sqlcv1.DBT
...
+		tenantIds := make([]pgtype.UUID, 0, len(payloads))
+		ids := make([]int64, 0, len(payloads))
+		insertedAts := make([]pgtype.Timestamptz, 0, len(payloads))
+		externalIds := make([]pgtype.UUID, 0, len(payloads))
+		types := make([]string, 0, len(payloads))
+		locations := make([]string, 0, len(payloads))
+		externalLocationKeys := make([]string, 0, len(payloads))
+
+		for r, k := range retrieveOptsToKey {
+			// qq: do we need conflict resolution here? I think the `COPY` is probably fine
```
abelanger5: remove comment?

pkg/repository/v1/payloadstore.go
```diff
@@ -393,315 +271,300 @@ func (p *payloadStoreRepositoryImpl) retrieve(ctx context.Context, tx sqlcv1.DBT
...
-}
+		if err := commit(ctx); err != nil {
+			return fmt.Errorf("failed to commit copy offloaded payloads transaction: %w", err)
+		}
 
-func (p *payloadStoreRepositoryImpl) ProcessPayloadExternalCutovers(ctx context.Context, partitionNumber int64) (bool, error) {
-	// no need to cut over if external store is not enabled
-	if !p.externalStoreEnabled {
-		return false, nil
+		if len(payloads) < int(p.externalCutoverBatchSize) {
```
abelanger5: This comparison also makes me a little nervous, not sure about a better way to do it though. 

I think it'll be fine, but is there an easy way we can roughly estimate data integrity between the two tables before we do the cutover? Not sure if counting the rows would be too slow, perhaps we could sample a set of rows across the tables.

sql/schema/v1-core.sql
```diff
@@ -1756,6 +1756,272 @@ BEGIN
...
+        source_partition_name,
+        temp_table_name || '_iat_chk_bounds'
+    );
+
+    RAISE NOTICE 'Successfully swapped partition %', source_partition_name;
+    RETURN source_partition_name;
+END;
+$$;
+
+CREATE TABLE v1_payload_cutover_job_offset (
```
abelanger5: Threw me off a bit that this is different from the order of tables in the migration file, let's just keep them the same?

abelanger5 (changes requested):

internal/services/controllers/v1/task/controller.go
mrkaye97: good call, should be fine to do this since we have this `is_completed` flag too 👍

internal/services/controllers/v1/task/process_payload_wal.go
mrkaye97: ah nice thanks! didn't know about that

pkg/repository/v1/payloadstore.go
mrkaye97: so I think this is inevitable, unfortunately. afaict, there's no way to avoid the case where a new payload is inserted in a chunk that we've already scanned while we're processing other rows. this will mean that we re-process some rows, which is why we need some conflict resolution (I just have `ON CONFLICT DO NOTHING`). we could probably avoid this with some logic to diff out already-processed records or something but I think that query would be too expensive and not worth it, and this will work the majority of the time. WDYT?

for the external store, I'll check to make sure there's no risk to re-processing there 👍

pkg/repository/v1/payloadstore.go
mrkaye97: I think maybe a count query is fine? I'll try adding that and see

pkg/repository/v1/payloadstore.go
mrkaye97: re: sprintf - I know this is a sql injection risk, but I'm not sure there's another way to dynamically determine which table to query 🤷

pkg/repository/v1/payloadstore.go
```diff
@@ -141,30 +142,109 @@ func (p *payloadStoreRepositoryImpl) Store(ctx context.Context, tx sqlcv1.DBTX,
...
 
-	for _, payload := range payloads {
-		tenantId := sqlchelpers.UUIDFromStr(payload.TenantId)
-		uniqueKey := PayloadUniqueKey{
-			ID:         payload.Id,
-			InsertedAt: payload.InsertedAt,
-			TenantId:   tenantId,
-			Type:       payload.Type,
+	if p.enableImmediateOffloads && p.externalStoreEnabled {
+		// WAL_ENABLED = false: Skip inline, go directly to external store, we will:
```
abelanger5: remove comment?

abelanger5 (approved):

pkg/repository/v1/payloadstore.go
mrkaye97: whoops yeah, copy paste 😬

PR #2587, opened 2025-12-01
Cross-Domain Tracking and Analytics Refactoring

undrash:
# Description

Implemented cross-domain tracking and refactored the analytics implementation in the app.

## Type of change

## What's Changed

- Cross-domain tracking: Users clicking `cloud.onhatchet.run` links in docs are now tracked as the same session in the cloud app
- Refactored PostHog provider in app to use posthog-js/react instead of raw script injection
- Global link handler in docs intercepts all links (Markdown and HTML) to append tracking params

> NOTE: Initially I implemented a markdown parser that decorates the markdown link with the tracking IDs, but then I realized that in some markdown files there are HTML snippets used with HTML links (<a> tags) - this led me to implement the global link handler so the different types of links don't fall through the cracks.


frontend/docs/mdx-components.tsx
```diff
@@ -0,0 +1,14 @@
+// MDX components for Nextra 4
+import { Callout, Cards, Steps, Tabs, FileTree } from "nextra/components";
```
mrkaye97, line 2: thanks for moving this 👍

frontend/app/src/components/molecules/analytics-provider.tsx
```diff
@@ -1,110 +0,0 @@
...
-posthog.init('${config.apiKey}',{
-  api_host:'${config.apiHost}',
-  ${bootstrapConfig}
-  session_recording: {
-      maskAllInputs: true,
-      maskTextSelector: "*"
-  }
-})
```
mrkaye97, line 82: 🥳

frontend/docs/components/CrossDomainLinkHandler.tsx
```diff
@@ -0,0 +1,81 @@
...
+      event.preventDefault();
+
+      const newHref = appendTrackingParams(href, sessionId, distinctId);
+
+      // Respect the original link's target attribute
+      const linkTarget = anchor.getAttribute("target");
+
+      if (linkTarget === "_blank") {
+        // Preserve rel attribute if present, otherwise use safe defaults
+        const rel = anchor.getAttribute("rel") || "noopener";
```
mrkaye97, line 86: do you think we need to check if the domain we're headed to is one of ours? not sure if as-is this will send posthog info in the hash if we e.g. link to Github

undrash: There is a `shouldHandleLink` function call a little bit above on line `52` that checks whether it is a target domain we're interested in tracking.

It does a pretty simple check right now, but it should work 😛 
```
const CROSS_DOMAIN_TARGETS = ["cloud.onhatchet.run"];

function shouldHandleLink(href: string): boolean {
  return CROSS_DOMAIN_TARGETS.some((target) => href.includes(target));
}
```

mrkaye97: ah nice alright, missed that - maybe we should include `hatchet.run` and all of its subdomains? also maybe the same with `onhatchet.run` too (since right now we have a lot of foobar.onhatchet.run

undrash: That makes sense, I'll do that!

mrkaye97: awesome, thanks 👍

undrash: The `hatchet.run` will be working out of the box since it's the same domain - with the current config all of the subdomains are tracked. Will do the same for onhatchet also, so we target all it's domains

mrkaye97 (approved):

PR #2591, opened 2025-12-02
fix: ensure that slow worker doesn't interrupt dispatcher, guard large RabbitMQ pubs

abelanger5:
# Description

1. If a slow worker or client is connected and receives a cancellation via `handleTaskCancelled`, there's a small risk that sending the cancellation signal risks a RabbitMQ channel consumer timeout, which results in the dispatcher resetting its connection and potentially dropping messages. 

    This makes the stream sends at least partially context-aware (though since the stream's context is managed elsewhere, so we don't have a great way to prevent the message from sending). 

2. Fixes large payload pubs (larger than 16 MB on RabbitMQ `4.x`). Instead of seeing an async channel exception after a large pub, this returns an error immediately instead of sending to Rabbit. 

## Type of change

### DIFF
--- a/internal/msgqueue/v1/rabbitmq/rabbitmq.go
+++ b/internal/msgqueue/v1/rabbitmq/rabbitmq.go
@@ -259,31 +259,11 @@ func (t *MessageQueueImpl) SendMessage(ctx context.Context, q msgqueue.Queue, ms
 	ctx, span := telemetry.NewSpan(ctx, "MessageQueueImpl.SendMessage")
 	defer span.End()
 
-	totalSize := getPayloadSize(msg.Payloads)
-
-	if totalSize > t.maxPayloadSize {
-		err := fmt.Errorf("message size %d bytes exceeds maximum allowed size of %d bytes", totalSize, t.maxPayloadSize)
-		span.RecordError(err)
-		span.SetStatus(codes.Error, "message size exceeds maximum allowed size")
-		return err
-	}
-
-	if totalSize > maxSizeErrorLogThreshold {
-		t.l.Error().
-			Int("message_size_bytes", totalSize).
-			Int("num_messages", len(msg.Payloads)).
-			Str("tenant_id", msg.TenantID).
-			Str("queue_name", q.Name()).
-			Str("message_id", msg.ID).
-			Msg("sending a very large message, this may impact performance")
-	}
-
 	span.SetAttributes(
 		attribute.String("MessageQueueImpl.SendMessage.queue_name", q.Name()),
 		attribute.String("MessageQueueImpl.SendMessage.tenant_id", msg.TenantID),
 		attribute.String("MessageQueueImpl.SendMessage.message_id", msg.ID),
 		attribute.Int("MessageQueueImpl.SendMessage.num_payloads", len(msg.Payloads)),
-		attribute.Int("MessageQueueImpl.SendMessage.total_size_bytes", totalSize),
 	)
 
 	err := t.pubMessage(ctx, q, msg)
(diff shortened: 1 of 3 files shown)


internal/msgqueue/v1/rabbitmq/rabbitmq.go
```diff
@@ -320,28 +328,40 @@ func (t *MessageQueueImpl) pubMessage(ctx context.Context, q msgqueue.Queue, msg
...
+
+		poolCh, err := t.pubChannels.Acquire(acquireCtx)
+
+		if err != nil {
+			acquireSpan.RecordError(err)
+			acquireSpan.SetStatus(codes.Error, "error acquiring publish channel")
+			acquireSpan.End()
+			t.l.Error().Msgf("[pubMessage] cannot acquire channel: %v", err)
+			return err
+		}
```
abelanger5, line 346: This error is always going to be a timeout when acquiring the channel, which means that the channel is full and retrying is unlikely to help, and will probably make the problem worse. So intentional that we're returning an error. Adding a comment.

internal/msgqueue/v1/rabbitmq/rabbitmq.go
```diff
@@ -320,28 +328,40 @@ func (t *MessageQueueImpl) pubMessage(ctx context.Context, q msgqueue.Queue, msg
...
-	acquireSpan.End()
+		pub = poolCh.Value()
 
-	pub := poolCh.Value()
+		if pub.IsClosed() {
+			poolCh.Destroy()
+			acquireErr = fmt.Errorf("channel is closed")
+			pub = nil
+			continue
+		}
```
abelanger5, line 359: This is the error we'd actually like to retry on - if there's a channel exception (which can be async, after the pub gets sent), we might be acquiring a closed channel.

internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -42,7 +42,18 @@ func (worker *subscribedWorker) StartTaskFromBulk(
 	action.ActionType = contracts.ActionType_START_STEP_RUN
 	action.ActionPayload = string(inputBytes)
 
-	return worker.sendToWorker(ctx, action)
+	err := worker.sendToWorker(ctx, action)
+
+	if err != nil {
+		// if the context is done, we return nil, because the worker took too long to receive the message
+		if errors.Is(err, context.DeadlineExceeded) {
+			return nil
```
abelanger5, line 52: The point of this is to not requeue the message if we hit this failure, because we don't know whether the worker received it or not. one solve is worker-side acks with a bidi stream, but it's a large refactor.

internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -42,7 +42,18 @@ func (worker *subscribedWorker) StartTaskFromBulk(
 	action.ActionType = contracts.ActionType_START_STEP_RUN
 	action.ActionPayload = string(inputBytes)
 
-	return worker.sendToWorker(ctx, action)
+	err := worker.sendToWorker(ctx, action)
+
+	if err != nil {
+		// if the context is done, we return nil, because the worker took too long to receive the message
+		if errors.Is(err, context.DeadlineExceeded) {
```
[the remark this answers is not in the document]
abelanger5: this is intentional, context.Canceled means the stream gets cancelled while sending the message, which should requeue

internal/msgqueue/v1/rabbitmq/rabbitmq.go
```diff
@@ -318,30 +306,65 @@ func (t *MessageQueueImpl) pubMessage(ctx context.Context, q msgqueue.Queue, msg
...
-		return err
-	}
 
-	acquireSpan.End()
+		pub = poolCh.Value()
 
-	pub := poolCh.Value()
+		// we need to case on the channel being closed here, because the channel exception may be async (after a previous pub has been
+		// sent), so we might have acquired a closed channel from the pool
+		if pub.IsClosed() {
```
mrkaye97, line 354: do we also want to check if the channel is clos_ing_, not just closed, here?

internal/msgqueue/v1/rabbitmq/rabbitmq.go
```diff
@@ -318,30 +306,65 @@ func (t *MessageQueueImpl) pubMessage(ctx context.Context, q msgqueue.Queue, msg
...
 				msg.ID, compressionResult.OriginalSize, compressionResult.CompressedSize, compressionResult.CompressionRatio*100)
 		}
+
+		totalSize := getMessageSize(msg)
+
+		if totalSize > t.maxPayloadSize {
+			err := fmt.Errorf("message size %d bytes exceeds maximum allowed size of %d bytes", totalSize, t.maxPayloadSize)
+			span.RecordError(err)
+			span.SetStatus(codes.Error, "message size exceeds maximum allowed size")
+			return err
```
mrkaye97, line 316: is this _is_ being caused by large payloads, is there a risk here that we just retry the same send over and over?

PR #2600, opened 2025-12-04
feat: dlq for dispatcher queues

abelanger5:
# Description

Adds a dead-lettering mechanism to dispatcher queues. Previously dispatcher queues were exclusive and auto-deleted, so in the case of an abrupt shutdown, assignment messages from the scheduler were lost. This modifies dispatcher queues to be exclusive and expirable, meaning they'll shut down after some period of inactivity / no consumers (currently set to 10 minutes). 

This allows us to attach a DLQ to these queues using the message expiry. Messages will expire after 20 seconds in the dispatcher queues and get sent to the dispatcher's DLQ (this is a little confusing because each DLQ also has another DLQ for retries). 20 seconds seems like a reasonable number because we cannot guarantee that a worker is active after 30 seconds, so a backlog of greater than 30 seconds on the dispatcher is problematic anyway.  

## Type of change

### DIFF
--- a/internal/msgqueue/v1/postgres/msgqueue.go
+++ b/internal/msgqueue/v1/postgres/msgqueue.go
@@ -3,6 +3,7 @@ package postgres
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"sync"
 	"time"
 
@@ -61,7 +62,7 @@ func WithQos(qos int) MessageQueueImplOpt {
 	}
 }
 
-func NewPostgresMQ(repo repository.MessageQueueRepository, fs ...MessageQueueImplOpt) (func() error, *PostgresMessageQueue) {
+func NewPostgresMQ(repo repository.MessageQueueRepository, fs ...MessageQueueImplOpt) (func() error, *PostgresMessageQueue, error) {
 	opts := defaultMessageQueueImplOpts()
 
 	for _, f := range fs {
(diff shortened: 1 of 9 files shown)


internal/queueutils/backoff.go
```diff
@@ -13,9 +14,17 @@ func SleepWithExponentialBackoff(base, max time.Duration, retryCount int) { // n
...
+		pow = 1 << retryCount
+	}
+
 	// Calculate exponential backoff
-	backoff := base * (1 << retryCount)
-	if backoff > max {
+	backoff := base * pow
+
+	// if backoff / pow does not recover base, we've overflowed
+	if backoff > max || backoff/pow != base {
```
abelanger5, line 27: drive-by - fixes an overflow causing sleep durations to go to 0

internal/msgqueue/v1/rabbitmq/rabbitmq.go
```diff
@@ -152,7 +152,7 @@ func WithMessageRejection(enabled bool, maxDeathCount int) MessageQueueImplOpt {
 }
 
 // New creates a new MessageQueueImpl.
-func New(fs ...MessageQueueImplOpt) (func() error, *MessageQueueImpl) {
+func New(fs ...MessageQueueImplOpt) (func() error, *MessageQueueImpl, error) {
```
abelanger5, line 155: Added an error to the return values because we weren't casing on if `initQueue` failed in this constructor which gets us into an undefined state (leading to a panic)

internal/msgqueue/v1/postgres/msgqueue.go
```diff
@@ -361,8 +368,17 @@ func (p *PostgresMessageQueue) upsertQueue(ctx context.Context, queue msgqueue.Q
 		consumer = &str
 	}
 
+	autoDeleted := queue.AutoDeleted()
+
+	// FIXME: note that this differs from the RabbitMQ implementation, since we auto-delete Postgres MQs after
```
abelanger5, line 373: Not a priority to fix or align timeouts here

internal/msgqueue/v1/rabbitmq/rabbitmq.go
```diff
@@ -609,6 +613,16 @@ func (t *MessageQueueImpl) initQueue(ch *amqp.Channel, q msgqueue.Queue) (string
...
 	}
 
+	if !q.IsDLQ() && q.DLQ() != nil && !q.DLQ().IsAutoDLQ() {
+		args["x-dead-letter-exchange"] = ""
+		args["x-dead-letter-routing-key"] = q.DLQ().Name()
+	}
+
+	if q.IsExpirable() {
+		args["x-message-ttl"] = int32(20000) // 20 seconds
+		args["x-expires"] = int32(600000)    // 10 minutes
```
abelanger5, line 623: this is the queue-level expiry after the last consumer disconnects

internal/services/scheduler/v1/scheduler.go
```diff
@@ -700,3 +710,145 @@ func taskBulkAssignedTask(tenantId string, workerIdsToTaskIds map[string][]int64
...
+
+	taskIds := make([]int64, 0)
+
+	for _, innerMsg := range msgs {
+		for _, tasks := range innerMsg.WorkerIdToTaskIds {
+			taskIds = append(taskIds, tasks...)
+		}
+	}
+
+	toRequeue, err := s.repov1.Tasks().ListTasks(ctx, msg.TenantID, taskIds)
```
mrkaye97: naming nit: the idea here is we send `could not send task to worker` and fail the task, and then either the user retries or it gets picked up by an automated retry right? if yes, I think this naming `toRequeue` is a little confusing

mrkaye97 (approved):
left one small comment but otherwise looks good! I think it's possible that Python failure is flaky, but there might be something wrong with the retry count incrementing logic that's real though, so I'd just make sure

PR #2609, opened 2025-12-05
Fix: Leasing for payload job

mrkaye97:
# Description

Making the payload job acquire a lease before continuing as opposed to an advisory lock so that only one controller can run the job at a time

## Type of change

### DIFF
--- a/pkg/repository/v1/payloadstore.go
+++ b/pkg/repository/v1/payloadstore.go
@@ -2,6 +2,7 @@ package v1
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"sort"
 	"time"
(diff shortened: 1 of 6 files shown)


pkg/repository/v1/payloadstore.go
```diff
@@ -513,23 +518,62 @@ func (p *payloadStoreRepositoryImpl) ProcessPayloadCutoverBatch(ctx context.Cont
...
-	PartitionDateStr string
+	ShouldRun      bool
+	LastOffset     int64
+	PartitionDate  PartitionDate
+	LeaseProcessId pgtype.UUID
 }
 
-func (p *payloadStoreRepositoryImpl) prepareCutoverTableJob(ctx context.Context) (*CutoverJobRunMetadata, error) {
+func (p *payloadStoreRepositoryImpl) acquireOrExtendJobLease(ctx context.Context, tx pgx.Tx, processId pgtype.UUID, partitionDate PartitionDate, offset int64) (*CutoverJobRunMetadata, error) {
+	leaseInterval := 2 * time.Minute
```
mrkaye97, line 528: important: does this seem reasonable? we could also make it an env var - 2 mins seemed like a looong time to need for processing one iteration, so it seemed okay to me

abelanger5: I think it just needs to be longer than the max statement timeout in the process right (looks like 10 seconds)? we can perhaps pull these into the top of the file so it's easier to compare times.

abelanger5 (approved):

pkg/repository/v1/payloadstore.go
mrkaye97: yeah good call, will make that a multiple of the timeout

mrkaye97: oh wait so it's more involved than that b/c the process will create a new tx on every batch, so we'd get N statements (one per batch) with 10s timeouts each. I'm going to leave this as-is for now, I think basically it needs to be longer than the longest reasonable time to complete the job end to end if only one iteration were needed

PR #2613, opened 2025-12-05
Feat: Process all old partitions in a loop

mrkaye97:
# Description

Loops through old partitions and process each one sequentially

## Type of change

### DIFF
--- a/pkg/repository/v1/payloadstore.go
+++ b/pkg/repository/v1/payloadstore.go
@@ -10,6 +10,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/hatchet-dev/hatchet/pkg/repository/postgres/sqlchelpers"
 	"github.com/hatchet-dev/hatchet/pkg/repository/v1/sqlcv1"
+	"github.com/hatchet-dev/hatchet/pkg/telemetry"
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgtype"
 	"github.com/jackc/pgx/v5/pgxpool"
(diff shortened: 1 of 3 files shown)


pkg/repository/v1/sqlcv1/payload-store-overwrite.sql.go
```diff
@@ -113,3 +113,51 @@ func ComparePartitionRowCounts(ctx context.Context, tx DBTX, tempPartitionName,
...
+        child.relname::text AS partition_name,
+        SUBSTRING(pg_get_expr(child.relpartbound, child.oid) FROM 'FROM \(''([^'']+)')::DATE AS lower_bound,
+        SUBSTRING(pg_get_expr(child.relpartbound, child.oid) FROM 'TO \(''([^'']+)')::DATE AS upper_bound
+    FROM pg_inherits
+    JOIN pg_class parent ON pg_inherits.inhparent = parent.oid
+    JOIN pg_class child ON pg_inherits.inhrelid = child.oid
+    WHERE parent.relname = 'v1_payload'
+    ORDER BY child.relname
+)
+
```
mrkaye97, line 129: had to use an override here b/c sqlc doesn't like compiling queries using `pg_inherits` and similar

pkg/repository/v1/payloadstore.go
```diff
@@ -685,6 +677,43 @@ func (p *payloadStoreRepositoryImpl) CopyOffloadedPayloadsIntoTempTable(ctx cont
...
 
+	return nil
+}
+
+func (p *payloadStoreRepositoryImpl) CopyOffloadedPayloadsIntoTempTable(ctx context.Context) error {
+	if !p.externalStoreEnabled {
+		return nil
+	}
+
+	ctx, span := telemetry.NewSpan(ctx, "payload_store_repository_impl.CopyOffloadedPayloadsIntoTempTable")
```
mrkaye97: added some spans too

abelanger5 (approved):

PR #2617, opened 2025-12-08
Feat: batch flush

grutt:
# Description

Introduces configurable batch assignment/flush support across scheduler, repository, API contracts, and SDKs. Adds batch metadata columns, persistence, and runtime tracking along with the orchestration logic to evaluate batch keys, buffer assignments, and flush batches via new scheduler workflows. Includes regenerated contracts/clients and new TypeScript examples with end-to-end coverage.

Fixes #N/A

## Type of change

## What's Changed

- Add migration wiring batch columns on `Step`, queue/task tables, runtime metadata, and new `v1_task_batch_run` tracking table.
- Extend repository APIs to compute batch keys, manage batch runs, update runtimes, and surface runtime data through `ListTasks`.
- Implement scheduler batch buffering/flush machinery with timers, dispatcher/worker coordination, and completion handling.
- Regenerate gRPC/OpenAPI contracts and update Go + TypeScript clients to expose batch configuration and runtime fields.
- Provide examples and tests for batch assignment, including Go scheduler/unit coverage and TypeScript e2e exercises.


internal/services/scheduler/v1/scheduler.go
```diff
@@ -737,15 +1146,355 @@ func (s *Scheduler) handleDeadLetteredMessages(msg *msgqueue.Message) (err error
...
+	bytes, err := json.Marshal(data)
+	if err != nil {
+		return ""
+	}
+
+	return string(bytes)
+}
+
+func (s *Scheduler) getBatchConfig(ctx context.Context, tenantId string, stepId string) (*batchConfig, error) {
+	if cached, ok := s.batchConfigs.Load(stepId); ok {
```
grutt: ttl cache

pkg/auth/token/token_test.go
```diff
@@ -92,9 +92,11 @@ func TestRevokeTenantToken(t *testing.T) {
 			t.Fatal(err.Error())
 		}
 
-		// validate the token
+		// validate the token twice to ensure cache hit does not error
 		_, _, err = jwtManager.ValidateTenantToken(context.Background(), token.Token)
+		assert.NoError(t, err)
```
grutt, line 97: flyby changes here.. im not clear on why these tests stated failing

pkg/repository/cache/cache.go
```diff
@@ -13,6 +13,9 @@ type Cacheable interface {
 	// Get gets a value from the cache with the given key
 	Get(key string) (interface{}, bool)
 
+	// Delete removes a value from the cache with the given key
+	Delete(key string)
+
```
grutt, line 18: flyby related to the token_test changes

pkg/repository/postgres/dbsqlc/workers.sql
```diff
@@ -4,11 +4,13 @@ SELECT
     ww."url" AS "webhookUrl",
     ww."id" AS "webhookId",
     workers."maxRuns" - (
-        SELECT COUNT(*)
-        FROM "SemaphoreQueueItem" sqi
+        SELECT
```
grutt: revert, this is v0

pkg/repository/postgres/dbsqlc/workers.sql
```diff
@@ -46,11 +48,13 @@ SELECT
     sqlc.embed(w),
     ww."url" AS "webhookUrl",
     w."maxRuns" - (
-        SELECT COUNT(*)
-        FROM "SemaphoreQueueItem" sqi
+        SELECT
```
grutt: revert, this is v0

pkg/repository/v1/sqlcv1/batch.go
```diff
@@ -26,6 +26,7 @@ type RegisterBatchBatchResults struct {
 	closed bool
 }
 
+// TODO why?
```
grutt: why is this registered this way?

pkg/repository/v1/sqlcv1/queue.sql
```diff
@@ -238,10 +243,16 @@ WITH input AS (
         t.retry_count,
         t.worker_id,
         @tenantId::uuid,
+        t.batch_key,
         t.timeout_at
     FROM
         updated_tasks t
-    ON CONFLICT (task_id, task_inserted_at, retry_count) DO NOTHING
+    ON CONFLICT (task_id, task_inserted_at, retry_count) DO UPDATE
```
grutt, line 259: contemplate this

pkg/repository/v1/sqlcv1/queue.sql
```diff
@@ -251,6 +262,41 @@ SELECT
...
+    input.task_retry_count,
+    NULL,
+    @tenantId::uuid,
+    CURRENT_TIMESTAMP + convert_duration_to_interval(t.step_timeout),
+    t.batch_key
+FROM
+    input
+JOIN
+    v1_task t ON t.id = input.task_id AND t.inserted_at = input.task_inserted_at
+ON CONFLICT (task_id, task_inserted_at, retry_count) DO UPDATE
```
grutt, line 302: contemplate this

pkg/repository/v1/task.go
```diff
@@ -1576,6 +1668,28 @@ func (r *sharedRepository) releaseTasks(ctx context.Context, tx sqlcv1.DBTX, ten
...
+		if task.BatchID.Valid {
+			id := sqlchelpers.UUIDToStr(task.BatchID)
+			if id != "" {
+				uniqueBatchIDs[id] = struct{}{}
+			}
+		}
+	}
+
+	for batchID := range uniqueBatchIDs {
+		err := r.queries.CompleteTaskBatchRun(ctx, tx, sqlcv1.CompleteTaskBatchRunParams{
```
grutt: is this correct?

pkg/repository/v1/task.go
```diff
@@ -1786,6 +1901,60 @@ func (r *sharedRepository) insertTasks(
...
+			}
+
+			res, err := r.celParser.ParseAndEvalStepRun(stepConfig.BatchKeyExpression.String, cel.NewInput(
+				cel.WithInput(task.Input.Input),
+				cel.WithAdditionalMetadata(additionalMeta),
+				cel.WithWorkflowRunID(task.ExternalId),
+				cel.WithParents(task.Input.TriggerData),
+			))
+
+			// TODO write an event for failed parse
```
grutt, line 1966: todo

pkg/repository/v1/task.go
```diff
@@ -2260,6 +2433,47 @@ func (r *sharedRepository) replayTasks(
...
+				return nil, fmt.Errorf("failed to parse batch key expression (%s): input is nil", stepConfig.BatchKeyExpression.String)
+			}
+
+			res, err := r.celParser.ParseAndEvalStepRun(stepConfig.BatchKeyExpression.String, cel.NewInput(
+				cel.WithInput(task.Input.Input),
+				cel.WithAdditionalMetadata(additionalMeta),
+				cel.WithWorkflowRunID(task.ExternalId),
+				cel.WithParents(task.Input.TriggerData),
+			))
+
```
grutt, line 2497: todo write event to olap

internal/services/scheduler/v1/scheduler.go
```diff
@@ -737,15 +1146,355 @@ func (s *Scheduler) handleDeadLetteredMessages(msg *msgqueue.Message) (err error
...
+func (s *Scheduler) batchFlush(ctx context.Context, req *batchFlushRequest) (err error) {
+	if req == nil || len(req.Items) == 0 {
+		return nil
+	}
+
+	reserved := req.BatchKey != "" && req.MaxRuns > 0
+
+	defer func() {
+		if err != nil && reserved {
+			if completeErr := s.repov1.Tasks().CompleteTaskBatchRun(ctx, req.TenantID, req.BatchID); completeErr != nil {
```
grutt: check for correctness here

grutt:
image

api-contracts/dispatcher/dispatcher.proto
```diff
@@ -174,6 +175,30 @@ message AssignedAction {
...
+
+    // (optional) payload for starting a batch execution
+    optional BatchStartPayload batch_start = 24;
+
+    // (optional) the partition key used for batching
+    optional string batchKey = 25;
+}
+
+message BatchStartPayload {
+    string batchId = 1;
```
abelanger5: Lots of these fields seem duplicated with the `optional` fields above in `AssignedAction`, and we're mixing casing

pkg/auth/token/token_test.go
abelanger5: this has been flaky for a while, haven't had the chance to look into it

pkg/repository/v1/sqlcv1/tasks.sql
```diff
@@ -160,12 +160,81 @@ WHERE
...
+    tr.worker_id AS runtime_worker_id,
+    tr.timeout_at AS runtime_timeout_at
 FROM
-    v1_task
+    v1_task t
+LEFT JOIN
+    v1_task_runtime tr ON tr.task_id = t.id
+    AND tr.task_inserted_at = t.inserted_at
+    AND tr.retry_count = t.retry_count
+    AND tr.tenant_id = t.tenant_id
```
abelanger5, line 212: Slightly worried about this left join, we've seen these query plans go wrong before - I wonder if we just want to do this join application-side when we determine whether the task has a batch_id or not

grutt: 🤔 would adding a `AND t.batch_key IS NOT NULL` here satisfy your concern for the base case. The query planner shouldn't join if its not a batch run

api-contracts/dispatcher/dispatcher.proto
```diff
@@ -174,6 +175,27 @@ message AssignedAction {
...
+    optional string batchId = 21;
+
+    // (optional) the total number of tasks in the batch
+    optional int32 batchSize = 22;
+
+    // (optional) the index of the task within the batch
+    optional int32 batchIndex = 23;
+
+    // (optional) payload for starting a batch execution
+    optional BatchStartPayload batch_start = 24;
```
abelanger5: nit: inconsistent casing

api-contracts/openapi/components/schemas/v1/task.yaml
```diff
@@ -236,6 +236,9 @@ V1TaskEvent:
       $ref: "#/V1TaskEventType"
     message:
       type: string
+    eventPayload:
```
abelanger5, line 239: Perhaps a silly question, but how do we not already have this? We're rendering the worker id somehow right?

internal/services/admin/v1/server.go
```diff
@@ -874,6 +874,33 @@ func getCreateTaskOpts(tasks []*contracts.CreateTaskOpts, kind string) ([]v1.Cre
...
+
+			if stepCp.Batch.BatchKey != nil && *stepCp.Batch.BatchKey != "" {
+				steps[j].BatchConfig.BatchKeyExpression = stepCp.Batch.BatchKey
+			}
+
+			if stepCp.Batch.MaxRuns != nil {
+				steps[j].BatchConfig.MaxRuns = stepCp.Batch.MaxRuns
+			}
+
+		}
```
abelanger5, line 903: just confirming we've tested the validation logic here? I saw the validation on the repository side, but wouldn't hurt to double check

cmd/hatchet-migrate/migrate/migrations/20251219181212_batching_consolidated.sql
```diff
@@ -0,0 +1,1419 @@
...
+    tenant_id UUID NOT NULL,
+    queue TEXT NOT NULL,
+    task_id BIGINT NOT NULL,
+    task_inserted_at TIMESTAMPTZ NOT NULL,
+    external_id UUID NOT NULL,
+    action_id TEXT NOT NULL,
+    step_id UUID NOT NULL,
+    workflow_id UUID NOT NULL,
+    workflow_run_id UUID NOT NULL,
+    schedule_timeout_at TIMESTAMP(3),
```
abelanger5, line 45: well, this is frustrating - it does look like all of our `schedule_timeout_at` fields are this data type, no-op here but we should fix this at some point

cmd/hatchet-migrate/migrate/migrations/20251215195000_v1_batching_consolidated.sql
```diff
@@ -0,0 +1,1419 @@
...
+
+CREATE INDEX IF NOT EXISTS v1_batched_queue_item_lookup_idx
+    ON v1_batched_queue_item (tenant_id ASC, step_id ASC, batch_key ASC, inserted_at ASC);
+
+CREATE INDEX IF NOT EXISTS v1_batched_queue_item_queue_idx
+    ON v1_batched_queue_item (tenant_id ASC, queue ASC, priority DESC, inserted_at ASC);
+
+-- Ensure any historical queue->batch redirect trigger is removed (final schema has no redirect trigger)
+DROP TRIGGER IF EXISTS v1_queue_item_redirect_to_batch ON v1_queue_item;
+DROP FUNCTION IF EXISTS v1_queue_item_redirect_to_batch_fn();
```
abelanger5: this seems specific to development? would be in favor of removing

cmd/hatchet-migrate/migrate/migrations/20251219181212_batching_consolidated.sql
```diff
@@ -0,0 +1,1419 @@
...
+DROP FUNCTION IF EXISTS v1_queue_item_redirect_to_batch_fn();
+
+-- v1_batch_runtime table (renamed from v1_task_batch_run in earlier iterations)
+DO $$
+BEGIN
+    IF to_regclass('v1_task_batch_run') IS NOT NULL AND to_regclass('v1_batch_runtime') IS NULL THEN
+        EXECUTE 'ALTER TABLE v1_task_batch_run RENAME TO v1_batch_runtime';
+    END IF;
+END
+$$;
```
abelanger5: also seems specific to development

cmd/hatchet-migrate/migrate/migrations/20251219181212_batching_consolidated.sql
```diff
@@ -0,0 +1,1419 @@
...
+DO $$
+BEGIN
+    IF EXISTS (
+        SELECT 1
+        FROM pg_constraint c
+        JOIN pg_class t ON t.oid = c.conrelid
+        WHERE t.relname = 'v1_batch_runtime'
+          AND c.conname = 'v1_task_batch_run_pkey'
+    ) THEN
+        EXECUTE 'ALTER TABLE v1_batch_runtime RENAME CONSTRAINT v1_task_batch_run_pkey TO v1_batch_runtime_pkey';
```
abelanger5: why are we doing this? was this also just for development?

cmd/hatchet-migrate/migrate/migrations/20251219181212_batching_consolidated.sql
```diff
@@ -0,0 +1,1419 @@
...
+    END IF;
+END
+$$;
+
+ALTER TABLE IF EXISTS v1_batch_runtime DROP COLUMN IF EXISTS completed_at;
+
+DROP INDEX IF EXISTS v1_task_batch_run_active_key_idx;
```
abelanger5: same deal here, just flagging in case missed

cmd/hatchet-migrate/migrate/migrations/20251219181212_batching_consolidated.sql
```diff
@@ -0,0 +1,1419 @@
...
+CREATE INDEX IF NOT EXISTS v1_batch_runtime_key_idx
+    ON v1_batch_runtime (tenant_id, step_id, batch_key);
+
+-- OLAP enum additions for batching lifecycle events
+ALTER TYPE v1_event_type_olap ADD VALUE IF NOT EXISTS 'BATCH_BUFFERED';
+ALTER TYPE v1_event_type_olap ADD VALUE IF NOT EXISTS 'WAITING_FOR_BATCH';
+ALTER TYPE v1_event_type_olap ADD VALUE IF NOT EXISTS 'BATCH_FLUSHED';
+-- +goose StatementEnd
+
+-- +goose StatementBegin
```
abelanger5, line 118: not seeing why this needs to be a separate tx? functions should be transactionally safe

sql/schema/v1-core.sql
```diff
@@ -423,6 +429,70 @@ CREATE INDEX v1_task_runtime_tenantId_workerId_idx ON v1_task_runtime (tenant_id
...
+        v1_batch_runtime br
+    WHERE
+        (br.tenant_id, br.batch_id) IN (SELECT tenant_id, batch_id FROM deletable);
+
+    RETURN NULL;
+END;
+$$
+LANGUAGE plpgsql;
+
+DROP TRIGGER IF EXISTS after_v1_task_runtime_delete_cleanup_batch_runtime ON v1_task_runtime;
```
abelanger5: similar to migration comments, let's not leak dev logic into the schema

sql/schema/v1-core.sql
```diff
@@ -1048,26 +1166,31 @@ BEGIN
...
+        nt.step_timeout,
+        COALESCE(nt.priority, 1),
+        nt.sticky,
+        nt.desired_worker_id,
+        nt.retry_count,
+        COALESCE(nt.batch_key, t.batch_key)
+    FROM new_table nt
+    LEFT JOIN v1_task t
+        ON t.id = nt.id
+        AND t.inserted_at = nt.inserted_at
```
abelanger5, line 1192: does this left join have a purpose? seems potentially harmful

cmd/hatchet-migrate/migrate/migrations/20251219181212_batching_consolidated.sql
```diff
@@ -0,0 +1,1419 @@
...
+-- Update trigger functions to match current canonical definitions in sql/schema/v1-core.sql (batch_key propagation)
+CREATE OR REPLACE FUNCTION v1_task_insert_function()
```
abelanger5, line 120: Just a note, I reviewed some more schema-level things in the v1-core.sql table, because it was a lot easier to review changes there.

internal/services/scheduler/v1/scheduler.go
```diff
@@ -699,18 +833,428 @@ func (s *Scheduler) notifyAfterConcurrency(ctx context.Context, tenantId string,
...
+
+	bytes, err := json.Marshal(data)
+	if err != nil {
+		return ""
+	}
+
+	return string(bytes)
+}
+
 func (s *Scheduler) handleDeadLetteredMessages(msg *msgqueue.Message) (err error) {
```
abelanger5, line 1259: It looks like we are not handling a dead-lettered `batch-start` message

internal/services/scheduler/v1/scheduler.go
```diff
@@ -699,18 +833,428 @@ func (s *Scheduler) notifyAfterConcurrency(ctx context.Context, tenantId string,
...
+			TriggerReason: flushReason,
+			TriggerTime:   triggeredAt,
+		}
+
+		if meta.MaxRuns > 0 {
+			maxRuns := int(meta.MaxRuns)
+			startPayload.MaxRuns = &maxRuns
+		}
+
+		startMsg, err := tasktypes.StartBatchMessage(tenantId, startPayload)
```
abelanger5, line 1035: What happens if this message is received by the dispatcher after it gets one of the individual tasks? Ordering is not guaranteed on the dispatcher side because of a QOS which is greater than 1, so if the worker expects this message to be received strictly before all of the batch's task assignments, this might not work.

abelanger5: a potential fix could be to include this message as part of the assignment message...

pkg/repository/v1/sqlcv1/tasks.sql
```diff
@@ -910,15 +1002,81 @@ FROM
     input rec;
 
 -- name: RegisterBatch :batchexec
+-- TODO why?
```
abelanger5: address todo?

pkg/repository/v1/sqlcv1/tasks.sql
```diff
@@ -910,15 +1002,81 @@ FROM
...
+    AND batch_id = @batchId::uuid;
+
 -- name: AnalyzeV1Task :exec
 ANALYZE v1_task;
 
 -- name: AnalyzeV1TaskEvent :exec
 ANALYZE v1_task_event;
 
+-- name: AnalyzeV1BatchRuntime :exec
+ANALYZE v1_batch_runtime;
```
abelanger5: don't think this table is partitioned?

pkg/repository/v1/sqlcv1/queue.sql
```diff
@@ -511,3 +583,238 @@ WHERE (task_id, task_inserted_at) IN (
...
+    tenant_id = @tenantId::uuid
+    AND id = ANY(@ids::bigint[]);
+
+-- name: MoveBatchedQueueItems :many
+WITH input AS (
+    SELECT
+        UNNEST(@ids::bigint[]) AS id
+), moved_items AS (
+    DELETE FROM v1_batched_queue_item
+    WHERE id = ANY(SELECT id FROM input)
```
abelanger5: nit: don't need the unnest pattern, can just do `any(@ids::bigint[])`

cmd/hatchet-migrate/migrate/migrations/20251215195000_v1_batching_consolidated.sql
```diff
@@ -0,0 +1,1419 @@
...
+    autovacuum_analyze_threshold = '25',
+    autovacuum_vacuum_cost_delay = '10',
+    autovacuum_vacuum_cost_limit = '1000'
+);
+
+CREATE INDEX IF NOT EXISTS v1_batched_queue_item_lookup_idx
+    ON v1_batched_queue_item (tenant_id ASC, step_id ASC, batch_key ASC, inserted_at ASC);
+
+CREATE INDEX IF NOT EXISTS v1_batched_queue_item_queue_idx
+    ON v1_batched_queue_item (tenant_id ASC, queue ASC, priority DESC, inserted_at ASC);
```
abelanger5: I question both of these indexes. So I see three query patterns on v1_batched_queue_item:
1. (tenant_id, step_id, batch_key) ordered by priority DESC and id ASC - there likely isn't an index that would be used here because of the ORDER BY. I believe you'd want the order by to be last in the index (and perhaps drop the id asc but not sure)
2. (tenant_id, schedule_timeout_at) ordered by id ASC 
3. Querying directly by IDs

Which queries are these indexes supposed to help with?

sdks/python/examples/batch_task/worker.py
```diff
@@ -0,0 +1,42 @@
...
+
+@workflow.batch_task(
+    name="uppercase",
+    batch_size=2,
+    flush_interval_ms=500,
+    batch_key="input.group",
+    max_runs=1,
+)
+async def uppercase(
+    inputs: list[BatchInput], ctxs: list[Context]
```
mrkaye97: important: do you think these should both be lists, where they share an index? or do you think we should zip the lists together into pairs of `(input, ctx)`? IMO, the second is probably better so that you can loop over each input + ctx as a single pair, but could be convinced otherwise. I usually am just nervous about relying on things to be in the same order 😅 

similarly, I wonder if we should have a single `BatchContext` instead of having many contexts, but I'm not quite sure how that'd work

sdks/python/hatchet_sdk/clients/dispatcher/action_listener.py
```diff
@@ -227,6 +232,22 @@ async def _generator(self) -> AsyncGenerator[Action | None, None]:
 
                         action_payload = ActionPayload()
 
+                    batch_start_payload: BatchStartPayload | None = None
+                    if assigned_action.batch_start is not None:
+                        trigger_time = None
+                        if hasattr(
+                            assigned_action.batch_start.triggerTime, "ToDatetime"
+                        ):
```
mrkaye97: we shouldn't do this, because this sort of thing (`hasattr` and `getattr`) can bypass the type checker in unexpected ways. for this, we should check if `batch_start` is not `None`, and then access the `triggerTime` based on that. it's a bit clunky, but it's more type safe

sdks/python/hatchet_sdk/hatchet.py
```diff
@@ -491,6 +491,160 @@ def inner(
...
+        *,
+        name: str | None = None,
+        description: str | None = None,
+        input_validator: None = None,
+        on_events: list[str] | None = None,
+        on_crons: list[str] | None = None,
+        version: str | None = None,
+        sticky: StickyStrategy | None = None,
+        default_priority: int = 1,
+        concurrency: ConcurrencyExpression | list[ConcurrencyExpression] | None = None,
```
mrkaye97, line 506: not sure if we discussed this, but how does concurrency work in this world again? like if the `max_runs` is set to a smaller number than the `batch_size` here, what happens?

sdks/python/hatchet_sdk/hatchet.py
```diff
@@ -491,6 +491,160 @@ def inner(
...
+        schedule_timeout: Duration = timedelta(minutes=5),
+        execution_timeout: Duration = timedelta(seconds=60),
+        retries: int = 0,
+        rate_limits: list[RateLimit] | None = None,
+        desired_worker_labels: dict[str, DesiredWorkerLabel] | None = None,
+        backoff_factor: float | None = None,
+        backoff_max_seconds: int | None = None,
+        default_filters: list[DefaultFilter] | None = None,
+        batch_size: int = ...,
```
mrkaye97: nit: should we call this `max_batch_size` or some such?

sdks/python/hatchet_sdk/hatchet.py
```diff
@@ -491,6 +491,160 @@ def inner(
...
+        flush_interval_ms: int | None = None,
```
mrkaye97: nit: I wonder if making this a `timedelta` and the converting to ms internally is a better strategy here. also, do we have a maximum flush interval?

sdks/python/hatchet_sdk/hatchet.py
```diff
@@ -491,6 +491,160 @@ def inner(
...
+        batch_key: str | None = None,
+        max_runs: int | None = None,
```
mrkaye97: how does this `max_runs` differ from the concurrency config ^^^?

sdks/python/hatchet_sdk/hatchet.py
```diff
@@ -491,6 +491,160 @@ def inner(
...
+        execution_timeout: Duration = timedelta(seconds=60),
+        retries: int = 0,
+        rate_limits: list[RateLimit] | None = None,
+        desired_worker_labels: dict[str, DesiredWorkerLabel] | None = None,
+        backoff_factor: float | None = None,
+        backoff_max_seconds: int | None = None,
+        default_filters: list[DefaultFilter] | None = None,
+        batch_size: int = ...,
+        flush_interval_ms: int | None = None,
+        batch_key: str | None = None,
```
mrkaye97: is this for managing concurrency?

sdks/python/hatchet_sdk/hatchet.py
```diff
@@ -491,6 +491,160 @@ def inner(
...
+                backoff_factor=backoff_factor,
+                backoff_max_seconds=backoff_max_seconds,
+                concurrency=_concurrency,
+                batch_size=batch_size,
+                flush_interval_ms=flush_interval_ms,
+                batch_key=batch_key,
+                max_runs=max_runs,
+            )
+
+            created_task = task_wrapper(func)  # type: ignore[arg-type]
```
mrkaye97: I think we should not need this `type: ignore` here - what error is this suppressing?

sdks/python/hatchet_sdk/runnables/workflow.py
```diff
@@ -1230,6 +1333,12 @@ def __init__(
 
         return_type = get_type_hints(self._task.fn).get("return")
 
+        if getattr(self._task, "is_batch", False):
```
mrkaye97: same here re: type checker - shouldn't need to use `getattr`

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -67,6 +68,114 @@ class WorkerStatus(Enum):
...
+    ) -> None:
+        state = self.batches.get(batch_id)
+
+        if state is None:
+            state = _BatchState(
+                expected_size=expected_size,
+                items={},
+                started=False,
+                batch_key=batch_key,
+            )
```
mrkaye97, line 176: nit: I think we can use a `defaultdict` for this instead of a plain dict so that we don't need to check if `state` is `None`

mrkaye97: although this might not work well with a dynamic `batch_key` idk

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -67,6 +68,114 @@ class WorkerStatus(Enum):
...
+            self.batches[batch_id] = state
+        elif state.expected_size != expected_size and expected_size > 0:
+            raise ValueError(
+                f"Batch metadata mismatch for '{self.action_id}' batch {batch_id}: expected size {state.expected_size}, got {expected_size}"
+            )
```
mrkaye97, line 181: what happens if this raises? do the tasks ever start? or does the engine find out somehow?

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -121,6 +230,11 @@ def __init__(
         self.lifespan_context = lifespan_context
         self.log_sender = log_sender
 
+        self._batch_controllers: dict[str, _BatchController] = {}
+        for action_id, task in self.action_registry.items():
+            if getattr(task, "is_batch", False):
```
mrkaye97: same here re: getattr

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -121,6 +230,11 @@ def __init__(
...
+                self._batch_controllers[action_id] = _BatchController(action_id, task)
```
mrkaye97, line 307: am I understanding this wrong? this loops through the action registry and if something is _not_ a batch task, creates a batch controller for it?

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -238,6 +356,9 @@ async def async_wrapped_action_func(
         ctx_action_key.set(action.key)
         ctx_additional_metadata.set(action.additional_metadata)
 
+        if getattr(task, "is_batch", False):
```
mrkaye97: same here

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -292,6 +413,164 @@ async def async_wrapped_action_func(
...
+        trigger_time = None
+        batch_key = action.batch_key
+
+        if action.batch_start is not None:
+            expected_size = action.batch_start.expected_size
+            trigger_reason = action.batch_start.trigger_reason
+            trigger_time = action.batch_start.trigger_time
+
+        default_batch_size = (
+            controller.task.batch.batch_size  # type: ignore[union-attr]
```
mrkaye97: probably shouldn't need this ignore either

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -67,6 +68,114 @@ class WorkerStatus(Enum):
...
+    expected_size: int
+    items: dict[int, _BatchQueueItem]
+    started: bool
+    trigger_reason: str | None = None
+    trigger_time: datetime | None = None
+    batch_key: str | None = None
+
+
+class _BatchController:
+    def __init__(self, action_id: str, task: Task[Any, Any]) -> None:
```
mrkaye97: is it possible to make this generic so it's not `[Any, Any]`?

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -292,6 +413,164 @@ async def async_wrapped_action_func(
...
+            or action.batch_size is None
+            or action.batch_index is None
+        ):
+            raise RuntimeError(
+                f"Batch metadata missing for task '{action.action_id}', cannot process batch"
+            )
+
+        batch_id = action.batch_id
+        expected_size = action.batch_size
+        index = int(action.batch_index)
```
mrkaye97: nit: this is already an `int` so no need to type cast here

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -292,6 +413,164 @@ async def async_wrapped_action_func(
...
+            on_ready=lambda batch_id: asyncio.create_task(
+                self._maybe_flush_batch(controller, batch_id)
+            ),
+        )
+
+        try:
+            result = await future
+        except asyncio.CancelledError:
+            controller.remove_item(batch_id=batch_id, index=index)
+            raise
```
mrkaye97, line 596: this seems like the meat of it - how does this work? we create a bunch of futures, add callbacks for each item in the batch, and then try to await them resolving?

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -292,6 +413,164 @@ async def async_wrapped_action_func(
...
+
+            if not isinstance(outputs, list):
+                raise ValueError(
+                    f"Batch task '{controller.action_id}' must return a list of outputs"
+                )
+
+            if len(outputs) != len(items):
+                raise ValueError(
+                    f"Batch task '{controller.action_id}' returned {len(outputs)} outputs for {len(items)} inputs"
+                )
```
mrkaye97, line 664: what happens if we raise here? does it bubble up to the engine?

mrkaye97 (commented):
left a bunch of SDK comments - I'm a bit nervous about some of this, mostly just because there are so many code paths that we have tested with the regular `task` type, but we're missing coverage for basically everything when using `batch` afaict, both on the SDK side and on the engine 😬 

another question: do we want to allow for batch tasks to be durable?

api-contracts/openapi/components/schemas/v1/task.yaml
grutt: we have a workerId prop on this object and message is a human readable string... this object is already fairly inconsistent for different event types we probably should have registered some kind of poly type here for different events.

not sure what the best short term fix is

pkg/repository/v1/sqlcv1/tasks.sql
grutt: the todo was to ask you why we register this in this way 😅... 

-- DO NOT USE: dummy query to satisfy sqlc and register Batch calls on DBTX

pkg/repository/v1/sqlcv1/tasks.sql
grutt: yep good catch

internal/services/scheduler/v1/scheduler.go
grutt: good catch, including the message

pkg/repository/v1/sqlcv1/queue.sql
grutt: gotcha, i guess we need this here because of multiple inputs?

```
-- name: MoveRateLimitedQueueItems :many
WITH input AS (
    SELECT
        UNNEST(@ids::bigint[]) AS id,
        UNNEST(@requeueAfter::timestamptz[]) AS requeue_after
), moved_items AS (
```

cmd/hatchet-migrate/migrate/migrations/20251215195000_v1_batching_consolidated.sql
grutt: good catch, these were for old queries that changed over time

internal/services/admin/v1/server.go
grutt: confirmed

sdks/python/hatchet_sdk/hatchet.py
grutt: good question, the batch max_runs is evaluated _after_ task concurrency, so likely this would force it to hit the flush case

sdks/python/hatchet_sdk/hatchet.py
grutt: leaving this open... i need to think on if this is more clear

sdks/python/hatchet_sdk/hatchet.py
grutt: max number of concurrent batches with the same key

sdks/python/hatchet_sdk/hatchet.py
grutt: yes at the batch level... we can talk through this maybe since its clearly unclear :)

sdks/python/hatchet_sdk/worker/runner/runner.py
grutt: yep... each batch item call creates a per-item future, stores it in the controller under the batch/id+index, and then awaits that future. 

when the batch is marked started and the controller has collected expected_size items, _maybe_flush_batch runs the batch function once and resolves (or rejects) every stored future with its corresponding output (or exception).


this might be a problem/leak if we never flush a batch but thats probably a problem anyway....

sdks/python/hatchet_sdk/worker/runner/runner.py
grutt: each task will fail with this error

sdks/python/hatchet_sdk/worker/runner/runner.py
grutt: good catch

sdks/python/hatchet_sdk/worker/runner/runner.py
grutt: the False here is the default, refactoring to not use getattr which is also more readable

sdks/python/hatchet_sdk/worker/runner/runner.py
mrkaye97: yeah, I was going to bring this up - I think we need to be able to evict batch items after a certain amount of time or something. or similarly, like what happens if one item in a batch starts hanging, or someone blocks the event loop? it seems like a big risk for a memory leak

sdks/python/hatchet_sdk/worker/runner/runner.py
grutt: this should fail all tasks in the batch which will then reassign

sdks/python/pyproject.toml
```diff
@@ -125,6 +125,14 @@ enable_error_code = [
...
 ]
 
+[[tool.mypy.overrides]]
+module = ["opentelemetry.*", "wrapt"]
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = ["opentelemetry.instrumentation.instrumentor"]
+disable_error_code = ["attr-defined"]
+
```
mrkaye97: we shouldn't need this I don't think? I saw the error in CI, but I'm unclear on why there are any lockfile changes at all in this PR (and also these changes). I think you should be able to reset these two files and do `poetry install --all-extras` on local and it should work fine. I'm also concerned about the `type: ignore` on the instrumentor, since I think that'll probably break when someone tries to use it

grutt: let me try..

sdks/python/hatchet_sdk/runnables/task.py
```diff
@@ -266,7 +270,12 @@ def to_proto(self, service_name: str) -> CreateTaskOpts:
             batch_proto = TaskBatchConfigProto(batch_max_size=self.batch.batch_max_size)
 
             if self.batch.batch_max_interval is not None:
-                interval_ms = int(self.batch.batch_max_interval.total_seconds() * 1000)
+                batch_max_interval = (
+                    str_to_timedelta(self.batch.batch_max_interval)
```
mrkaye97: not sure I'm following this 😅 why is `self.batch.batch_max_interval` ever a string? shouldn't it be a timedelta?

grutt: because its using our Duration type for consistency...

mrkaye97: oooh, okay, I see 🤔 

the reason for `Duration` is to allow either a Go duration string (`1s`, `1m`, etc.) _or_ a timedelta, but this was mostly for backwards compat with v0. I think we should only allow `timedelta` here

sdks/python/hatchet_sdk/features/scheduled.py
```diff
@@ -183,7 +182,6 @@ def bulk_delete(
         workflow_id: str | None = None,
         parent_workflow_run_id: str | None = None,
         parent_step_run_id: str | None = None,
-        statuses: list[ScheduledRunStatus] | None = None,
```
mrkaye97: I think this is a breaking change? even if it doesn't do anything

sdks/python/hatchet_sdk/runnables/task.py
grutt: done :)

abelanger5 (changes requested):
Still not seeing that the comments on the migration/schema files have been resolved btw

PR #2636, opened 2025-12-10
Fix: python proto generate script path

jishnundth:
# Description
fix: delayed worker example
add: test for delayed worker python example
fix: python sdk proto generator path

## Type of change


mrkaye97:
hey @jishnundth, thanks for the contribution! - not sure if this is complete yet, but as it is, I'm not sure how much value this test adds. we have lots of tests for this code path already (running a workflow + asserting on results). let me know if I'm missing something here though! 

In general, we use these examples to test features on the engine, as opposed to writing tests to make sure the examples themselves work, if that makes any sense

jishnundth:
@mrkaye97 Thank you for your comment. I think there’s a misunderstanding here. As per the design pattern used in the codebase, almost all Python examples have corresponding tests to validate the example logic with the engine, and these tests are also automatically run through the `sdk-python` workflow. Please let me know if I'm wrong here. 

I can see that `schedule()` is tested in multiple places in the Python SDK, but this specific scenario—where a workflow schedules another workflow—doesn’t seem to be covered elsewhere. I’m open to removing this test or closing the PR if needed.

mrkaye97:
Ah yep, that's true! But in general the idea is that many of tests were written as tests of the engine, but this was just an example we wrote for the docs initially, which is why it's "untested". There's nothing mechanically different between scheduling from an e.g. API vs. scheduling from another task, which is why we don't need to test this code path from my PoV

jishnundth:
@mrkaye97 I agree, I've removed the test

mrkaye97:
going to close this since I think we should probably just remove this task - I always use `./generate.sh` from inside of `/sdks/python` to run the generation code 🤷

PR #2637, opened 2025-12-10
Feat: Parallelize replication from PG -> External

mrkaye97:
# Description

Parallelizing replication from PG to the external store by reading chunks of the PK and then reading the data out for each chunk and doing the writes concurrently

## Type of change

### DIFF
--- a/pkg/repository/v1/olap.go
+++ b/pkg/repository/v1/olap.go
@@ -6,12 +6,14 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"maps"
 	"math/rand"
 	"sort"
 	"sync"
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/hashicorp/go-multierror"
 	lru "github.com/hashicorp/golang-lru/v2"
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgtype"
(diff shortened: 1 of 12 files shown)


pkg/repository/v1/payloadstore.go
```diff
@@ -427,65 +429,112 @@ func (p *payloadStoreRepositoryImpl) ProcessPayloadCutoverBatch(ctx context.Cont
...
 	externalIdToPayload := make(map[PayloadExternalId]sqlcv1.ListPaginatedPayloadsForOffloadRow)
-	offloadOpts := make([]OffloadToExternalStoreOpts, 0, len(payloads))
+	numPayloads := 0
+
+	for _, payloadRange := range payloadRanges {
+		wg.Add(1)
+		go func(pr sqlcv1.CreatePayloadRangeChunksRow) {
+			defer wg.Done()
+
+			payloads, err := p.queries.ListPaginatedPayloadsForOffload(ctx, p.pool, sqlcv1.ListPaginatedPayloadsForOffloadParams{
```
mrkaye97, line 475: is there any way around using `p.pool` here? doesn't seem like we'd be able to share a tx across goroutines, not sure if there's some prior art for this sort of thing

pkg/repository/v1/sqlcv1/olap.sql
```diff
@@ -1867,6 +1867,33 @@ SELECT
...
+        @lastTenantId::UUID,
+        @lastExternalId::UUID,
+        @lastInsertedAt::TIMESTAMPTZ
+    ) p
+), with_rows AS (
+    SELECT
+        tenant_id::UUID,
+        external_id::UUID,
+        inserted_at::TIMESTAMPTZ,
+        ROW_NUMBER() OVER (ORDER BY tenant_id, external_id, inserted_at) AS rn
```
mrkaye97, line 1886: any ideas on a more efficient way to do this?

pkg/repository/v1/payloadstore.go
```diff
@@ -427,65 +438,105 @@ func (p *payloadStoreRepositoryImpl) ProcessPayloadCutoverBatch(ctx context.Cont
...
+				Lastid:         payloadRange.ID,
+				Lasttype:       payloadRange.Type,
+			})
 
-	for _, payload := range payloads {
-		externalId := PayloadExternalId(payload.ExternalID.String())
+			if err != nil {
+				returnErr = multierror.Append(err, fmt.Errorf("failed to list paginated payloads for offload"))
+				return
+			}
```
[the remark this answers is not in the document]
mrkaye97: @abelanger5 how big of a concern is this in practice?

pkg/repository/v1/olap.go
```diff
@@ -2708,69 +2710,110 @@ type OLAPCutoverBatchOutcome struct {
...
+	var returnErr error
 
-	tenantIdToOffloadOpts := make(map[TenantID][]StoreOLAPPayloadOpts)
 	externalIdToKey := make(map[PayloadExternalId]ExternalPayloadLocationKey)
 	externalIdToPayload := make(map[PayloadExternalId]sqlcv1.ListPaginatedOLAPPayloadsForOffloadRow)
+	numPayloads := 0
+
+	for _, payloadRange := range payloadRanges {
+		wg.Add(1)
+		go func(pr sqlcv1.CreateOLAPPayloadRangeChunksRow) {
```
mrkaye97: do I need something like `eg.Go` here? wasn't too sure where that was needed vs. not

abelanger5: I think it's slightly easier to manage instead of appending to a multierror, but it doesn't really matter as long as multierror is concurrency safe, which I think it is

pkg/repository/v1/payloadstore.go
abelanger5: there's no way around it unfortunately, queries in a tx need to be strictly serial so having a tx be concurrency safe doesn't make too much sense

pkg/repository/v1/payloadstore.go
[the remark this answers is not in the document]
abelanger5: probably a bit of a concern: https://github.com/hashicorp/go-multierror/issues/93

and a good reason to use eg.Go (based on the discussion above)

abelanger5 (approved):

PR #2638, opened 2025-12-10
Feat: Add oldest queued + running jobs to task stats

mrkaye97:
# Description

Adding an `Oldest` field to the `TaskStatusStat` which reports the oldest queued and running job, for use with e.g. autoscaling

## Type of change


pkg/repository/v1/task.go
```diff
@@ -3812,6 +3814,11 @@ func (r *TaskRepositoryImpl) GetTaskStats(ctx context.Context, tenantId string)
 				result[stepReadableId] = taskStat
 			}
 			statusStat = result[stepReadableId].Running
+
+			if oldest.Valid && (statusStat.Oldest == nil || oldest.Time.Before(*statusStat.Oldest)) {
+				statusStat.Oldest = &oldest.Time
+			}
+
```
mrkaye97: @mnafees do you think these are okay places to put this? I wasn't 100% sure - tested locally and it seemed to work but the flow control here is a bit involved :P

mnafees: What is the intent here?

mrkaye97: adding the timestamps of the oldest queued and running items to the map we return

mnafees: Ah yes, this makes sense. I think we should add a doc about this very important endpoint and how to read it. Multiple people are using it for auto scaling it seems.

mnafees (approved):

mnafees (approved):

PR #2643, opened 2025-12-11
Billing changes

mnafees:
# Description

Make UI and backend changes for new billing on cloud. Also, gets rid of the legacy `WORKFLOW_RUN` usage meter and type.

## Type of change


cmd/hatchet-migrate/migrate/migrations/20251218123817_v1_0_62.sql
mrkaye97, line 1: just fyi, will need to update the migration version here b/c of how many migrations I've been merging :P

frontend/app/src/components/v1/cloud/billing/subscription.tsx
```diff
@@ -68,11 +68,22 @@ export const Subscription: React.FC<SubscriptionProps> = ({
   const subscriptionMutation = useMutation({
     mutationKey: ['user:update:logout'],
     mutationFn: async ({ plan_code }: { plan_code: string }) => {
```
mrkaye97, line 69: is it possible to make these plan codes an enum so we can pattern match? not sure what they look like

frontend/app/src/pages/main/tenant-settings/resource-limits/index.tsx
```diff
@@ -19,15 +19,21 @@ export default function ResourceLimits() {
 
   const billingState = useQuery({
     ...queries.cloud.billing(tenant.metadata.id),
-    enabled: !!cloudMeta?.data.canBill,
+    enabled: !!tenant?.metadata?.id && !!cloudMeta?.data.canBill,
+    retry: false,
```
mrkaye97: we sure we don't want retries here?

frontend/app/src/pages/main/v1/managed-workers/index.tsx
```diff
@@ -35,47 +31,22 @@ export default function ManagedWorkers() {
...
+  //     }
+  //     setPortalLoading(true);
+  //     const link = await cloudApi.billingPortalLinkGet(tenant!.metadata.id);
+  //     window.open(link.data.url, '_blank');
+  //   } catch (e) {
+  //     handleApiError(e as any);
+  //   } finally {
+  //     setPortalLoading(false);
+  //   }
+  // };
```
mrkaye97: nit: should we remove this?

frontend/app/src/pages/main/tenant-settings/resource-limits/index.tsx
```diff
@@ -56,7 +57,7 @@ export default function ResourceLimits() {
           <Separator className="my-4" />
           <Subscription
             hasPaymentMethods={hasPaymentMethods}
-            active={billingState.data?.subscription}
+            active={billingState.data?.currentSubscription}
             plans={billingState.data?.plans}
             coupons={billingState.data?.coupons}
           />
```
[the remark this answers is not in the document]
mnafees: v0 is no longer used and will be removed soon so we can ignore this

frontend/app/src/lib/can/features/managed-compute.ts
```diff
@@ -25,17 +25,7 @@ const replicaLimits = {
 };
 
 export const managedCompute: PermissionSet = {
-  create: () => (context) => {
```
grutt, line 28: why this change?

frontend/app/src/pages/main/v1/managed-workers/index.tsx
```diff
@@ -35,48 +30,6 @@ export default function ManagedWorkers() {
     managedCompute.canCreateWorkerPool(workerPoolCount),
   );
 
-  // stop polling billing if there are payment methods
```
grutt, line 39: same here, we should still require a cc on file for managed

pkg/config/server/server.go
```diff
@@ -313,10 +318,6 @@ type SecurityCheckConfigFile struct {
 type LimitConfigFile struct {
 	DefaultTenantRetentionPeriod string `mapstructure:"defaultTenantRetentionPeriod" json:"defaultTenantRetentionPeriod,omitempty" default:"720h"`
 
-	DefaultWorkflowRunLimit      int           `mapstructure:"defaultWorkflowRunLimit" json:"defaultWorkflowRunLimit,omitempty" default:"1000"`
```
grutt: i think we have this in our docs for OSS so this is technically a breaking change if someone has a workflow run limit limit set but not a task run limit set. 

perhaps we should also set TR limit to WRF env var in this case?

frontend/app/src/components/v1/cloud/billing/subscription.tsx
mnafees: I am waiting on doing this since I am torn between making Autumn the source of truth and finalizing those plans. But going to do so for sure.

frontend/app/src/components/v1/cloud/billing/subscription.tsx
```diff
@@ -21,22 +20,22 @@ import {
 } from '@/lib/api/generated/cloud/data-contracts';
 import { useApiError } from '@/lib/hooks';
 import queryClient from '@/query-client';
-import { ExclamationTriangleIcon } from '@radix-ui/react-icons';
+// import { ExclamationTriangleIcon } from '@radix-ui/react-icons';
```
mrkaye97: rm commented code?

frontend/app/src/components/v1/cloud/billing/subscription.tsx
```diff
@@ -166,98 +213,181 @@ export const Subscription: React.FC<SubscriptionProps> = ({
...
+                    <CardTitle className="text-2xl mb-1">
+                      {currentPlanDetails.name}
+                    </CardTitle>
+                    <div className="text-3xl font-bold mb-2">
+                      $
+                      {(
+                        currentPlanDetails.amountCents /
+                        100 /
+                        (currentPlanDetails.period === 'yearly' ? 12 : 1)
+                      ).toLocaleString()}{' '}
```
mrkaye97: nit: I think there's probably some library we should be using for formatting currencies

mrkaye97 (approved):
left a couple small comments but otherwise lgtm - I think we need to address Gabe's question re: breaking change though

PR #2647, opened 2025-12-11
chore: migrate to tanstack router

grutt:
# Description

Changes frontend app routing to tanstack for typesafetey 

## Type of change


frontend/app/src/lib/breadcrumbs.ts
```diff
@@ -1,4 +1,37 @@
-import { TenantedPath } from '@/router';
+export type TenantedPath =
```
grutt, line 1: we should probably consider doing something else with this component

frontend/app/src/pages/auth/login/index.tsx
```diff
@@ -76,26 +77,26 @@ export default function Login() {
             </div>
             <p className="text-left text-sm text-gray-700 dark:text-gray-300 w-full">
               By clicking continue, you agree to our{' '}
-              <Link
-                to="https://www.iubenda.com/terms-and-conditions/76608149"
+              <a
+                href="https://www.iubenda.com/terms-and-conditions/76608149"
```
grutt: external urls should now be a tags

frontend/app/src/lib/breadcrumbs.ts
mrkaye97: I think we might not even need this anymore - it was mostly to keep track of things for the V0/1 split, but I wonder if there's a simpler way here

mrkaye97:
this is amazing 🙌 thanks for doing it

grutt:
> this is amazing 🙌 thanks for doing it

needed something mindless to do between calls... im sorry, i know how excited you were to do this 😅

frontend/app/src/hooks/use-tenant.tsx
```diff
@@ -53,7 +56,7 @@ export function useTenantDetails() {
 
       setLastTenant(tenant);
       queryClient.clear();
-      navigate(newPath);
+      navigate({ to: newPath });
```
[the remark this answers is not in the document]
mrkaye97: agree with copilot here :P I think there must be a way to do this in a type-safe way if we can narrow the possibilities for `newPath` to a subset of allowed routes

frontend/app/src/main.tsx
```diff
@@ -38,5 +39,6 @@ ReactDOM.createRoot(document.getElementById('root')!).render(
   <QueryClientProvider client={queryClient}>
     <ReactQueryDevtools initialIsOpen={false} />
     <Router />
+    <TanStackRouterDevtools position="bottom-right" router={router} />
```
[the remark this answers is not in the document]
mrkaye97: ➕ here - I think this will show up in production if we don't do this (which is confusing, because it doesn't happen for the react query devtools above here)

frontend/app/src/pages/main/v1/index.tsx
```diff
@@ -161,46 +169,52 @@ function Sidebar({ className, memberships }: SidebarProps) {
...
               <SidebarButtonPrimary
                 key="workers"
                 onNavLinkClick={onNavLinkClick}
-                to={`/tenants/${tenantId}/workers/all`}
+                to={appRoutes.tenantWorkersRoute.to}
+                params={{ tenant: tenantId }}
                 name="Workers"
                 icon={<ServerStackIcon className="mr-2 h-4 w-4" />}
-                prefix={`/tenants/${tenantId}/workers`}
+                prefix="/tenants/$tenant/workers"
```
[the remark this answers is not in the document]
mrkaye97: ➕

frontend/app/src/pages/main/v1/index.tsx
```diff
@@ -214,39 +228,45 @@ function Sidebar({ className, memberships }: SidebarProps) {
               <SidebarButtonPrimary
                 key="tenant-settings"
                 onNavLinkClick={onNavLinkClick}
-                to={`/tenants/${tenantId}/tenant-settings/overview`}
-                prefix={`/tenants/${tenantId}/tenant-settings`}
+                to={appRoutes.tenantSettingsOverviewRoute.to}
+                params={{ tenant: tenantId }}
+                prefix="/tenants/$tenant/tenant-settings"
```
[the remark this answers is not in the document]
mrkaye97: ➕

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/index.tsx
```diff
@@ -102,7 +103,7 @@ async function fetchDAGRun(id: string) {
 }
 
 export default function Run() {
-  const { run } = useParams();
+  const { run } = useParams({ from: appRoutes.tenantTaskRunsRoute.to });
```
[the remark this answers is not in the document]
mrkaye97: I think this might be correct or is at least worth checking, and the naming is confusing :P

mrkaye97 (commented):
left a couple small comments and think there are some small type safety things, but in general looks amazing, very psyched for this 🙌

frontend/app/src/lib/breadcrumbs.ts
grutt: lets figure this out in a different PR

mrkaye97: sounds good

frontend/app/src/lib/breadcrumbs.ts
```diff
@@ -102,7 +102,7 @@ const createRouteLabel = (path: TenantedPath): string => {
       return '';
     default:
       // eslint-disable-next-line no-case-declarations
-      const exhaustiveCheck: never = path;
+      const exhaustiveCheck: never = path as never;
```
mrkaye97: hehe, I think this change is actually flagging a real error - what's the TS error if we remove this type assertion?

frontend/app/src/pages/main/v1/managed-workers/$managed-worker/index.tsx
```diff
@@ -35,10 +35,10 @@ export default function ExpandedWorkflow() {
   const { refetchInterval } = useRefetchInterval();
 
   const params = useParams({ from: appRoutes.tenantManagedWorkerRoute.to });
-  invariant(params['managed-worker']);
+  invariant(params.managedWorker);
```
mrkaye97: totally forgot to mention it, but I think we don't need these `invariant`s everywhere now! this should just type check to `string`, and we can remove these everywhere

frontend/app/src/lib/breadcrumbs.ts
grutt: ha good catch, sneaky webhook path on bad merge

frontend/app/src/lib/breadcrumbs.ts
```diff
@@ -69,7 +101,7 @@ const createRouteLabel = (path: TenantedPath): string => {
       return '';
     default:
       // eslint-disable-next-line no-case-declarations
-      const exhaustiveCheck: never = path;
+      const exhaustiveCheck = path;
```
mrkaye97: this still needs the `: never` hint I think - it should be `const exhaustiveCheck: never = path;` so that if the path is _not_ `never`, it errors

frontend/app/src/pages/main/v1/managed-workers/$managed-worker/index.tsx
```diff
@@ -47,7 +47,7 @@ export default function ExpandedWorkflow() {
   });
 
   const updateManagedWorkerMutation = useMutation({
-    mutationKey: ['managed-worker:update', params['managed-worker']],
+    mutationKey: ['managed-worker:update', params.managedWorker],
     mutationFn: async (data: UpdateManagedWorkerRequest) => {
       invariant(managedWorker);
```
mrkaye97: I think there are a couple more of these - I just grepped for `invariant(`

grutt: yeah some of these are not path related... but probably we can still remove them

mrkaye97: yeah I think we should still be able to? there are only one or two we really need I think

frontend/app/src/lib/breadcrumbs.ts
grutt: yeah i think you're right.. we need a never on at least one side of the assign here

frontend/app/src/pages/main/v1/managed-workers/$managed-worker/index.tsx
grutt: removed all except the ones that break the build lol

mrkaye97 (approved):

PR #2656, opened 2025-12-12
Chore: Add the tailwind plugin for prettier

mrkaye97:
# Description

Adding the [tailwind plugin](https://tailwindcss.com/blog/automatic-class-sorting-with-prettier#how-classes-are-sorted) for prettier to auto-sort classes to do a better job of keeping consistent across the app

## Type of change


frontend/app/.prettierrc
```diff
@@ -1,4 +1,5 @@
 {
   "singleQuote": true,
-  "trailingComma": "all"
+  "trailingComma": "all",
+  "plugins": ["prettier-plugin-tailwindcss"]
```
mrkaye97, line 4: this is the real change

grutt (approved):

PR #2665, opened 2025-12-15
fix: prevent static rate limits from resetting to 0 

### LINKED ISSUE #1834: [BUG] Static rate limits reset their own limits to 0
**Describe the issue**
I'm unable to get static rate limits to work at all. They seem to reset their limit (not their value) to 0 as soon as they're used.

**Environment**
- SDK: Python 1.10.2
- Engine: Cloud
- 
**Expected behavior**
I don't expect the limit to change without a call to `hatchet.rate_limits.put()`

```python
from pydantic import BaseModel

hatchet = Hatchet(debug=True)

class Input(BaseModel):
    foo: int = 100

workflow = hatchet.workflow(name="test", input_validator=Input)

@workflow.task(rate_limits=[
    RateLimit(
        static_key="rate_limit",
        units="input.foo"
    )
])
async def task(input: Input, ctx: Context):
    print(input)
    return {}

if __name__ == "__main__":
    hatchet.rate_limits.put("rate_limit", 1000, RateLimitDuration.MINUTE)
    worker = hatchet.worker(name="test", workflows=[workflow])
    workflow.run_no_wait(Input())
    worker.start()
```

When I run this, the task gets queued:
Image

And if I look at the Rate Limits section, I see:
Image

If I cancel the task and just run:
```
hatchet.rate_limits.put("rate_limit", 1000, RateLimitDuration.MINUTE)
```

The rate limit goes back up:
Image

But as soon as I run the worker again, it drop back to 0.

What am I doing wrong here?

Manancode:
Fixes [1834](https://github.com/hatchet-dev/hatchet/issues/1834#issue-3125874777)

This PR fixes a bug where static rate limits were being reset to 0 when tasks using them were queued

Verified fix locally using the exact reproduction code from issue
Confirmed static rate limits maintain their limitValue after tasks are queued

### DIFF
--- a/pkg/repository/v1/scheduler_queue.go
+++ b/pkg/repository/v1/scheduler_queue.go
@@ -511,9 +511,11 @@ func (d *queueRepository) GetTaskRateLimits(ctx context.Context, queueItems []*s
 			continue
 		}
 
-		upsertRateLimitBulkParams.Keys = append(upsertRateLimitBulkParams.Keys, key)
-		upsertRateLimitBulkParams.Windows = append(upsertRateLimitBulkParams.Windows, getWindowParamFromDurString(duration))
-		upsertRateLimitBulkParams.Limitvalues = append(upsertRateLimitBulkParams.Limitvalues, int32(limitValue)) // nolint: gosec
+		if limitValue > 0 {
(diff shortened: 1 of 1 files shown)


mrkaye97:
closing this as we merged in a different fix in #2714 !

PR #2667, opened 2025-12-15
Feat: New event getter + janky v0 fix

mrkaye97:
# Description

Couple things here:

1. Adds a `v1-event:get` method on the REST API which uses the PK of the OLAP events table properly to try to limit timeouts.
2. Added a really janky `event-with-tenant` populator to try to improve the V0 backwards compat with only minimal breaking SDK changes needed - migration path here would be to change `EventDataGet` -> `EventDataGetWithTenant`, and add the tenant id, and then nothing else would need to change. just a janky hack fix that we can hopefully remove soon 
3. Refactored a bit

## Type of change

### DIFF
--- a/api-contracts/openapi/paths/event/event.yaml
+++ b/api-contracts/openapi/paths/event/event.yaml
@@ -290,13 +290,13 @@ eventData:
 
 eventDataWithTenant:
   get:
-    x-resources: ["tenant", "event"]
+    x-resources: ["tenant", "event-with-tenant"]
     description: Get the data for an event.
     operationId: event-data:get-with-tenant
     parameters:
       - description: The event id
         in: path
-        name: event
+        name: event-with-tenant
         required: true
         schema:
           type: string
--- a/api-contracts/openapi/paths/event/event.yaml
+++ b/api-contracts/openapi/paths/event/event.yaml
@@ -288,6 +288,53 @@ eventData:
     tags:
       - Event
 
+eventDataWithTenant:
+  get:
+    x-resources: ["tenant", "event"]
+    description: Get the data for an event.
+    operationId: event-data:get-with-tenant
+    parameters:
+      - description: The event id
+        in: path
+        name: event
+        required: true
+        schema:
+          type: string
+          format: uuid
+          minLength: 36
+          maxLength: 36
+      - description: The tenant id
+        in: path
+        name: tenant
+        required: true
+        schema:
+          type: string
+          format: uuid
+          minLength: 36
+          maxLength: 36
+    responses:
+      "200":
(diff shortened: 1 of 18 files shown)


api-contracts/openapi/paths/event/event.yaml
```diff
@@ -288,6 +288,53 @@ eventData:
     tags:
       - Event
 
+eventDataWithTenant:
```
mrkaye97, line 291: open to naming ideas here - I didn't want to call it `eventDataV1` or something for obvious reasons

sdks/python/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to Hatchet's Python SDK will be documented in this changelog
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.21.7] - 2025-12-15
```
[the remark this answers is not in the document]
mrkaye97: 😂 UTC

api/v1/server/handlers/events/get_data.go
```diff
@@ -8,7 +8,15 @@ import (
...
-	event := ctx.Get("event").(*dbsqlc.Event)
+	eventInterface := ctx.Get("event")
+	if eventInterface == nil {
+		return nil, echo.NewHTTPError(404, "event not found")
+	}
+
+	event, ok := eventInterface.(*dbsqlc.Event)
+	if !ok {
+		return nil, echo.NewHTTPError(500, "invalid event type in context")
+	}
```
abelanger5, line 19: nit, but these types of errors should never really happen and in this case a panic is fine (you'll see it if you trivially test locally, and a panic in production has the same effect of throwing a 500 with a stack trace). So can condense this to `ctx.Get("event").(*dbsqlc.Event)`

api/v1/server/handlers/events/get_data.go
```diff
@@ -22,3 +30,29 @@ func (t *EventService) EventDataGet(ctx echo.Context, request gen.EventDataGetRe
...
+	eventInterface := ctx.Get("event-with-tenant")
+
+	if eventInterface == nil {
+		return nil, echo.NewHTTPError(404, "event not found")
+	}
+
+	event, ok := eventInterface.(*dbsqlc.Event)
+	if !ok {
+		return nil, echo.NewHTTPError(500, "invalid event type in context")
+	}
```
abelanger5, line 45: same spiel here

abelanger5 (approved):

PR #2669, opened 2025-12-16
Fix: Payload List Index Performance

mrkaye97:
# Description

Fixing a performance issue where the index scan would take ages with a CTE hack to first get a set of candidates, then scan them

## Type of change

### DIFF
--- a/sql/schema/v1-core.sql
+++ b/sql/schema/v1-core.sql
@@ -1917,7 +1917,7 @@ BEGIN
     END IF;
 
     query := format('
-        WITH candidates AS (
+        WITH candidates AS MATERIALIZED (
             SELECT tenant_id, id, inserted_at, external_id, type, location,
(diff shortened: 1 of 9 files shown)


sql/schema/v1-core.sql
```diff
@@ -1916,16 +1917,26 @@ BEGIN
...
 
     query := format('
+        WITH candidates AS MATERIALIZED (
+            SELECT tenant_id, id, inserted_at, external_id, type, location,
+                external_location_key, inline_content, updated_at
+            FROM %I
+            WHERE
+                (tenant_id, inserted_at, id, type) >= ($1, $2, $3, $4)
+            ORDER BY tenant_id, inserted_at, id, type
+            LIMIT $9 * 2
```
abelanger5, line 1933: Add a comment to clarify?

abelanger5 (approved):

PR #2671, opened 2025-12-16
Refactor: Remove inline button styles where possible

mrkaye97:
# Description

Removing a bunch of inline styles from uses of `Button` throughout the app to make styling more consistent

## Type of change


grutt (approved):
Nice! Code looks good and clicked through locally 🚀

PR #2674, opened 2025-12-17
chore: remove webhook workers from dashboard

grutt:
# Description

Webhook workers were deprecated in V1, this removes it from the dashboard. 

## Type of change


mrkaye97 (approved):
😍

PR #2676, opened 2025-12-17
chore:  add cypres e2e testing to dashboard

grutt:
# Description

I've broken tenant switching one too many times, adds cypress e2e tests along with some common failure modes.

## Type of change


.gitguardian.yaml
grutt, line 1: This seems to not be doing anything...

frontend/app/cypress/e2e/auth/register.cy.ts
```diff
@@ -0,0 +1,27 @@
...
+    cy.get('input#password').clear().type(password);
+    cy.contains('button', 'Create Account').click();
+
+    cy.wait('@register').its('response.statusCode').should('eq', 200);
+
+    cy.location('pathname', { timeout: 5000 }).should(
+      'include',
+      '/onboarding/create-tenant',
+    );
+    cy.contains('h1', 'What are you building?', { timeout: 30000 }).should(
```
mrkaye97, line 23: just flagging that this will inevitably break as soon as we change onboarding, but that's okay for now IMO

frontend/app/vite.config.ts
```diff
@@ -20,6 +20,24 @@ export default defineConfig({
...
+      '127.0.0.1',
+    ],
+    proxy: {
+      // The frontend uses relative `/api/v1/...` paths, so proxy `/api` to the API server.
+      '/api': {
+        target: 'http://127.0.0.1:8080',
+        changeOrigin: true,
+      },
+    },
   },
```
mrkaye97, line 42: I'm a bit nervous about this breaking in production - we need to test on staging before deploying :P

mrkaye97 (approved):
couple small comments, but looks great! want to try re-breaking the tenant switcher to see if the tests fail?

frontend/app/cypress/e2e/auth/register.cy.ts
grutt: yep that was intentional :) i want to make sure we build out a new flow when we make these changes. very important flow

frontend/app/vite.config.ts
grutt: this should be dev only afik, we're serving static assets

grutt:
> couple small comments, but looks great! want to try re-breaking the tenant switcher to see if the tests fail?

yep i did, and it failed

PR #2683, opened 2025-12-18
 Send `create:user` Event from OAuth Flow

undrash:
# Description

I noticed that the last `user:create` event that was recorded in PostHog was 4 month ago - while the last `tenant:create` was much more recent. So I started digging around this problem and it looks like you forgot to send the event from the OAuth flow.

## Type of change
- [x] New feature (breaking change -> please update the cloud repo)

## What's Changed

- Send `user:create` from OAuth GitHub
- Send `user:create` from OAuth Google
- Send `provider` metadata prop at basic signup

---

After a discussion with @mnafees in [this Slack thread](https://literally-talk.slack.com/archives/C09CVP99E1Z/p1766082564834589) we concluded that the optimal approach is to send the events from callbacks


pkg/repository/postgres/user.go
```diff
@@ -24,9 +24,9 @@ func NewUserRepository(shared *sharedRepository) repository.UserRepository {
 	}
 }
 
-func (w *userRepository) RegisterCreateCallback(callback repository.UnscopedCallback[*dbsqlc.User]) {
+func (w *userRepository) RegisterCreateCallback(callback repository.UserCreateCallback) {
 	if w.createCallbacks == nil {
-		w.createCallbacks = make([]repository.UnscopedCallback[*dbsqlc.User], 0)
+		w.createCallbacks = make([]repository.UserCreateCallback, 0)
```
mnafees: This is a breaking change. I know we need the OAuth information but just want to note it here. @abelanger5 I think it should be fine but we would need to update the cloud repo as well to adjust this change.

mnafees (approved):

internal/msgqueue/postgres/msgqueue.go
abelanger5, line 1: drive by: linting errors

internal/msgqueue/rabbitmq/rabbitmq.go
abelanger5, line 1: drive by: linting error

internal/msgqueue/rabbitmq/rabbitmq_test.go
abelanger5, line 1: drive by: linting error

internal/services/controllers/metrics/collector.go
abelanger5, line 1: also linting errors

pkg/repository/user.go
```diff
@@ -41,8 +41,14 @@ type UpdateUserOpts struct {
...
 
+type UserCreateCallbackOpts struct {
+	*sqlcv1.User
+
+	CreateOpts *CreateUserOpts
+}
+
 type UserRepository interface {
-	RegisterCreateCallback(callback UnscopedCallback[*sqlcv1.User])
+	RegisterCreateCallback(callback UnscopedCallback[*UserCreateCallbackOpts])
```
abelanger5, line 51: I ended up moving this back to `UnscopedCallback` because I want to use `cb.Do`, which has panic-wrapping, just in case a callback decides to panic somewhere.

abelanger5 (approved):

PR #2684, opened 2025-12-19
fix: typescript tasks canceled status

### LINKED ISSUE #2655: [BUG] Typescript tasks do not retain their canceled status after they exit
When cancelling a typescript task and throwing in the task in response to the cancellation, the task is marked as failed and retried
=> expected the task to be marked as cancelled and not retried

When cancelling a typescript task and simply exiting the task gracefully, the task is marked a succeeded
=> expected the task to be marked as cancelled and not marked as succeeded

This behavior appears to be different to the python sdk which suppresses status update if cancelled.

Latest typescript sdk and hatchet-lite

aashmanVerma:
Add cancellation checks in success/failure callbacks to prevent sending COMPLETED/FAILED events when a task is cancelled. This ensures cancelled tasks are properly marked as CANCELLED and not retried.

# Description

When a TypeScript task is cancelled, the worker was incorrectly sending status updates (COMPLETED or FAILED) to the Hatchet engine. This caused cancelled tasks to be marked as either succeeded or failed (with retries), instead of being properly marked as CANCELLED.

The fix adds cancellation checks in both the `success()` and `failure()` callbacks before sending status events. When a task is cancelled, the worker now suppresses status updates, allowing the engine to properly handle the CANCELLED state. This matches the Python SDK behavior.

Fixes #2655

## Type of change

## What's Changed
- Added cancellation check in `success()` callback to suppress COMPLETED events when task is cancelled
- Added cancellation check in `failure()` callback to suppress FAILED events when task is cancelled
- Aligns TypeScript SDK behavior with Python SDK

### DIFF
--- a/sdks/typescript/src/v1/client/worker/worker-internal.ts
+++ b/sdks/typescript/src/v1/client/worker/worker-internal.ts
@@ -550,10 +550,13 @@ export class V1Worker {
       };
 
       const success = async (result: any) => {
+        if (context.cancelled) {
+          return;
+        }
+
         this.logger.info(`Step run ${action.stepRunId} succeeded`);
 
         try {
-          // Send the action event to the dispatcher
           const event = this.getStepActionEvent(
             action,
             StepActionEventType.STEP_EVENT_TYPE_COMPLETED,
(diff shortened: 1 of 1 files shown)


mrkaye97:
closing in favor of #2865 - thanks for the fix!

PR #2687, opened 2025-12-19
Feat:  bulk management schedules

grutt:
# Description

Adds affordances for rescheduling and deleting scheduled runs in bulk via API and 

## Type of change


mrkaye97:
I think we need a rebase here?

api-contracts/openapi/components/schemas/workflow_run.yaml
```diff
@@ -240,6 +240,136 @@ ScheduledWorkflowsList:
     pagination:
       $ref: "./metadata.yaml#/PaginationResponse"
 
+UpdateScheduledWorkflowRunRequest:
+  type: object
+  properties:
+    triggerAt:
+      type: string
+      format: date-time
```
mrkaye97, line 248: do we want to allow for updating the payload and metadata too?

api-contracts/openapi/components/schemas/workflow_run.yaml
```diff
@@ -240,6 +240,136 @@ ScheduledWorkflowsList:
...
+    parentWorkflowRunId:
+      type: string
+      format: uuid
+      minLength: 36
+      maxLength: 36
+    parentStepRunId:
+      type: string
+      format: uuid
+      minLength: 36
+      maxLength: 36
```
mrkaye97, line 297: I think this is not amazing naming - I know this is supposed to correspond to the `parentId` in the trigger table, but I think it'd be better to call these something else if we can (e.g. just workflow id might be good here). also, I don't think the parent step run makes sense here unless I'm missing something?

mrkaye97: ah I see we have workflow id above - I feel like maybe that's good enough?

api-contracts/openapi/components/schemas/workflow_run.yaml
```diff
@@ -240,6 +240,136 @@ ScheduledWorkflowsList:
...
+      description: A list of metadata key value pairs to filter by
+      type: array
+      items:
+        type: string
+      example: ["key1:value1", "key2:value2"]
+    statuses:
+      description: A list of scheduled run statuses to filter by
+      type: array
+      items:
+        $ref: "#/ScheduledRunStatus"
```
mrkaye97: is there a case where we'd want to delete anything that wasn't in the `SCHEDULED` status? because once it's past that, it's a regular run, right?

api-contracts/openapi/components/schemas/workflow_run.yaml
```diff
@@ -240,6 +240,136 @@ ScheduledWorkflowsList:
...
+  type: object
+  properties:
+    id:
+      type: string
+      format: uuid
+      minLength: 36
+      maxLength: 36
+    triggerAt:
+      type: string
+      format: date-time
```
mrkaye97, line 333: same here re: updating the payload and metadata potentially

api/v1/server/handlers/workflows/bulk_delete_scheduled.go
```diff
@@ -0,0 +1,205 @@
...
+			if len(all) >= int(count) || len(rows) == 0 {
+				break
+			}
+
+			offset += limit
+			opts.Offset = &offset
+		}
+
+		// Convert list results into ids + pre-fill errors for non-API items.
+		ids = make([]openapi_types.UUID, 0, len(all))
```
mrkaye97: we shouldn't be using these openapi types explicitly I don't think - we've been preferring `uuid.UUID` or `pgtype.UUID` everywhere, we should do that here too

api/v1/server/handlers/workflows/bulk_delete_scheduled.go
```diff
@@ -0,0 +1,205 @@
...
+				idCp := id
+				errors = append(errors, gen.ScheduledWorkflowsBulkError{Id: &idCp, Error: "Scheduled workflow not found."})
+				continue
+			}
+			if meta.Method != dbsqlc.WorkflowTriggerScheduledRefMethodsAPI {
+				idCp := id
+				errors = append(errors, gen.ScheduledWorkflowsBulkError{Id: &idCp, Error: "Cannot delete scheduled run created via code definition."})
+				continue
+			}
+			toDelete = append(toDelete, idStr)
```
mrkaye97: I don't think we need this - do we ever actually use the method field? right now I'm pretty sure every scheduled run is created over the API one way or another

mrkaye97 (commented):
left some comments on the API - mind posting a couple of screenshots of the FE?

api-contracts/openapi/components/schemas/workflow_run.yaml
grutt: yeah... this is just to keep it consistent with previous bad decisions. this is if the schedule is out of a parent task

api-contracts/openapi/components/schemas/workflow_run.yaml
grutt: yes, good catch

api/v1/server/handlers/workflows/bulk_delete_scheduled.go
grutt: nice, yeah we removed this in v1

api-contracts/openapi/components/schemas/workflow_run.yaml
grutt: i think no for now... but i can be convinced otherwise

mrkaye97 (approved):

PR #2689, opened 2025-12-19
feat: improved error boundaries

grutt:
## Description

There are inconsistent and unclear error states for simple error boundaries. This PR standardizes the error UI and makes 403/404 states clearer (especially for tenant-scoped navigation).

## Type of change

## What's Changed

- [x] Added a shared error-page layout (`ErrorPageLayout`) and split error UI into dedicated components:
  - [x] `NotFound` (404)
  - [x] `TenantForbidden` (tenant-scoped 403)
  - [x] `GenericError` (fallback)
  - [x] `NewVersionAvailable` (dynamic import failure → prompt reload)
- [x] Standardized **404** handling via TanStack Router `notFoundComponent`:
  - [x] Root (unknown “unnested” routes like `/x`)
  - [x] Authenticated route subtree
  - [x] Tenant route subtree (unknown “nested” routes like `/tenants/<t>/...`)
- [x] Improved **403** for tenant-scoped requests:
  - [x] `tenantRoute` loader checks tenant membership and throws `Response(403)` → `TenantForbidden`
- [x] Added Cypress coverage + support glue:
  - [x] `error-pages.cy.ts` coverage for tenant forbidden + nested/unnested 404
  - [x] Cypress `uncaught:exception` ignore for TanStack Router `Response` throws
 
- [x] 404s for “resource detail” routes (invalid IDs/slugs)
  - [x] **Organization**: `/organizations/$organization`
  - [x] **Workflow details**: `/tenants/$tenant/workflows/$workflow`
  - [x] **Run details**: `/tenants/$tenant/runs/$run`
  - [x] **Worker details**: `/tenants/$tenant/workers/$worker`
  - [x] **Managed worker details**: `/tenants/$tenant/managed-workers/$managedWorker`


grutt:
image

image

frontend/app/src/pages/error/components/new-version-available.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const navigate = useNavigate();
+  const location = useLocation();
+
+  const queryParams = new URLSearchParams(location.search);
+
+  if (!queryParams.has('updated')) {
+    queryParams.set('updated', 'true');
+    const updatedUrl = `${location.pathname}?${queryParams.toString()}`;
+    window.location.href = updatedUrl;
+  }
```
mrkaye97, line 18: where do we actually use this? it seems a little silly to me - if the idea is just that if we detect this parameter, we tell the user to refresh, why don't we refresh for them and not show this state instead?

frontend/app/src/pages/main/v1/managed-workers/$managed-worker/index.tsx
```diff
@@ -38,6 +40,13 @@ export default function ExpandedWorkflow() {
   const managedWorkerQuery = useQuery({
     ...queries.cloud.getManagedWorker(params.managedWorker),
     refetchInterval,
+    retry: (_failureCount, error) => {
+      if (isAxiosError(error) && error.response?.status === 404) {
+        return false;
+      }
```
mrkaye97: we shouldn't retry on other 4xx either right?

frontend/app/src/router.tsx
```diff
@@ -130,7 +139,34 @@ const v1RedirectRoute = createRoute({
 const tenantRoute = createRoute({
   getParentRoute: () => authenticatedRoute,
   path: 'tenants/$tenant',
+  loader: async ({ params }) => {
+    // Ensure the tenant in the URL is one the user actually has access to.
+    // If not, throw a 403 so the global error boundary can show a friendly message.
+    const memberships = await queryClient.fetchQuery({
+      ...queries.user.listTenantMemberships,
+      retry: false,
+    });
```
mrkaye97, line 153: are we going to do this on every page load? just checking b/c I'm not sure

frontend/app/src/router.tsx
```diff
@@ -130,7 +139,34 @@ const v1RedirectRoute = createRoute({
...
+      throw new Response('Forbidden', { status: 403, statusText: 'Forbidden' });
+    }
+
+    // Optionally warm the tenant details cache, since most tenant pages expect it.
+    // If this fails for any reason, let the error boundary handle it.
+    await queryClient.fetchQuery({
+      queryKey: ['tenant:get', params.tenant],
+      queryFn: async () => (await api.tenantGet(params.tenant)).data,
+      retry: false,
+    });
```
mrkaye97, line 169: same here - maybe this should happen in a provider instead?

frontend/app/src/pages/error/components/new-version-available.tsx
grutt: this is veryyyy weird one but basically it catches invalid cloudlflare cache state some of the time

frontend/app/src/router.tsx
grutt: only on initial load afict, should be safe since its using the query client which will cache the request

frontend/app/src/router.tsx
grutt: i think this is ok for now too

frontend/app/src/pages/main/v1/managed-workers/$managed-worker/index.tsx
grutt: yep

mrkaye97 (approved):

PR #2690, opened 2025-12-19
Fix: Last bits of payload job cleanup

mrkaye97:
# Description

Couple more bits of cleanup for the payload job:

1. Removing unneeded rows from the offsets table once the partition has been dropped (no reason to keep them around at that point)
2. Adding a reconciliation step, where at the end we look for any payloads in the source partition that don't exist in the target, and we just try to select and insert them. If there are too many (hard-coded as 5000 for now), we fail. The idea here is to try to limit how often we need to do manual reconciliation. One note on this front is that the reconciliation query can take a while, so I extended the job lease timeout to 10 mins to give some headroom there and to limit the likelihood of these queries snowballing and blowing up CPU.
3. Reduced the memory footprint of the job by ~50% by not duplicating the payload contents in the stuff that we keep in memory from the goroutines that read the payloads, which should help a bunch

One other note on this front was I noticed for some reason that doing the SELECT + INSERT in the same query would hang indefinitely when separating them out didn't. Not sure what to make of that, but I figured I'd just go with that was working here

## Type of change

### DIFF
--- a/pkg/repository/v1/sqlcv1/olap-overwrite.sql.go
+++ b/pkg/repository/v1/sqlcv1/olap-overwrite.sql.go
@@ -452,6 +452,7 @@ type CutoverOLAPPayloadToInsert struct {
 	InsertedAt          pgtype.Timestamptz
 	ExternalID          pgtype.UUID
 	ExternalLocationKey string
+	InlineContent       []byte
 }
 
 type InsertCutOverOLAPPayloadsIntoTempTableRow struct {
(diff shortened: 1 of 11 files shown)


pkg/repository/v1/sqlcv1/olap-overwrite.sql.go
```diff
@@ -523,32 +528,6 @@ func InsertCutOverOLAPPayloadsIntoTempTable(ctx context.Context, tx DBTX, tableN
 	return &insertRow, err
 }
 
-func CompareOLAPPartitionRowCounts(ctx context.Context, tx DBTX, tempPartitionName, sourcePartitionName string) (bool, error) {
```
mrkaye97, line 526: unused (duplicate)

mnafees (approved):
LGTM 🚀

PR #2692, opened 2025-12-21
Use constants for `msgqueue` tenant message IDs

mnafees:
# Description

Use constants for `msgqueue` tenant message IDs for readability and ease of code navigation.

## Type of change


mrkaye97 (approved):
thanks!

PR #2701, opened 2025-12-22
feat: hatchet cli

abelanger5:
# Description

Creates a new CLI for Hatchet with the primary goal of making quickstart easier for new users. 

Currently implements the following:

- `hatchet profile [add|remove|list|show|update]` - allows managing multiple profiles, like `local` and `cloud`
- `hatchet server [start|stop]` - manages a Hatchet instance locally via docker
- `hatchet quickstart` - generates quickstart directory, similar to `create-react-app`  
- `hatchet worker dev` - creates a filewatcher for a Hatchet instance and automatically injects profile token from a selected profile. Uses a `hatchet.yaml` file in the current directory 
- `hatchet worker run` - runs a script for triggering a workflow using `hatchet.yaml`
- `hatchet tui` - creates the TUI interface 

Screenshots:

(Note these screenshots should `hatchet-cli` but this will be installed as `hatchet` in prod)

image

image

image

## Type of change


cmd/hatchet-cli/cli/internal/config/cli/init.go
```diff
@@ -0,0 +1,127 @@
+package cli
+
+import (
+	"os"
+	"path/filepath"
+	"sync"
+
+	"github.com/charmbracelet/log"
+	"github.com/mitchellh/go-homedir"
```
mnafees: I think we should go with https://pkg.go.dev/os#UserHomeDir instead

.github/workflows/cli-release.yaml
```diff
@@ -0,0 +1,73 @@
+name: Release Hatchet CLI
+
+on:
+  push:
+    tags:
+      - "v*" # Trigger on version tags like v0.73.10
```
mrkaye97, line 6: will this also run on alphas?

mrkaye97 (approved):

.github/workflows/cli-release.yaml
abelanger5: yeah, but the goreleaser flag `prerelease: auto` should mark this as a prerelease if it's an alpha tag

cmd/hatchet-cli/cli/internal/config/cli/init.go
abelanger5: nice catch, just updated

PR #2703, opened 2025-12-23
Fix  layout bugs

grutt:
# Description

Fixes scroll and layout issues, adds cypress tests to prevent regressions

## Type of change


grutt:
image
image
image
image

frontend/app/src/pages/auth/hooks/use-cloud.ts
```diff
@@ -0,0 +1,74 @@
+import { cloudApi } from '@/lib/api/api';
+import { APICloudMetadata } from '@/lib/api/generated/cloud/data-contracts';
+import { useApiError } from '@/lib/hooks';
+import { useQuery } from '@tanstack/react-query';
+import { AxiosError } from 'axios';
+import { useCallback, useMemo } from 'react';
+
+export default function useCloud(tenantId?: string) {
```
mrkaye97, line 8: IMO this should be a provider, so we can fetch this once and just have a boolean flag everywhere that we can check. I think it'd be amazing to consolidate the tenant, organization, and this into a single provider if we can, but that's probably work for later

mrkaye97 (approved):

PR #2704, opened 2025-12-23
feat:  improved navigation

grutt:
# Description

Improves the navigation ux

## Type of change

## What's new:
- [x] collapsible, draggable side nav bar 
- [x] lots of refactoring and cleanup
- [x] user and tenant at top right

image
image
image


mrkaye97 (commented):
going through the code now, but I think maybe having the side bar be draggable doesn't make sense? there's not much value in that afaict, since it just collapses once it hits some breakpoint anyways but doesn't really scroll or anything, and nobody would want it to be wider. I think keeping the existing two options (large or small) with no drag makes the most sense

btw, looks really nice 🥳

grutt:
i have an idea for a feature where resizing will come in handy

frontend/app/src/hooks/use-tenant.tsx
```diff
@@ -42,11 +42,39 @@ export function useTenantDetails() {
...
+            key0 === 'user:get:current' ||
+            key0 === 'user:list-tenant-invites' ||
+            key0 === 'user:list:tenant-invites' ||
+            key0 === 'metadata' ||
+            key0 === 'organization:list'
+          ) {
+            return false;
+          }
+
+          return true;
```
mrkaye97: IMO this isn't worth it - it's easier to just clear everything and refetch, it doesn't take very long and keeps the code simpler

frontend/app/src/pages/main/v1/side-nav-items.tsx
```diff
@@ -0,0 +1,214 @@
...
+      to: appRoutes.tenantSettingsIngestorsRoute.to,
+    },
+    {
+      key: 'quickstart',
+      name: 'Quickstart',
+      to: appRoutes.tenantOnboardingGetStartedRoute.to,
+    },
+  ];
+
+  // TODO-DESIGN update icons
```
mrkaye97, line 66: planning to address this later with design?

mrkaye97 (commented):
not sure where in the code these are, but a couple things:

1. looks like we're refetching tenant memberships pretty often, which is causing the spinner in the tenant picker to show every few seconds, which is causing layout shift and is a bit annoying. I think we can probably not show a spinner there at all, those calls are fast anyways
2. if I click the hatchet logo to collapse / expand the sidebar, there's some jumping / lag, which based on the tenant memberships refetching at the same time I think might be getting caused by refetching queries + rerendering when that happens, but I'm not totally sure

frontend/app/src/pages/main/v1/side-nav-items.tsx
grutt: yeah the idea is we can tag in @sebastiangraz on these specific components as needed, but we can ship structural improvements on the fly

mrkaye97 (approved):
lgtm - did you test out how the tenant switcher will look with organizations? I don't know how to do that yet 😅

grutt:
yep i ran against cloud locally

PR #2710, opened 2025-12-24
Publish `COULD_NOT_SEND_TO_WORKER` OLAP event due to worker backlog

mnafees:
# Description

Instead of erroring out when worker backlog size has been attained, we now send a brand new OLAP event called `COULD_NOT_SEND_TO_WORKER`.

## Type of change


cmd/hatchet-migrate/migrate/migrations/20251224145920_v1_0_65.sql
```diff
@@ -0,0 +1,78 @@
...
+);
+
+ALTER TABLE v1_task_events_olap_tmp
+    ALTER COLUMN event_type TYPE v1_event_type_olap_new
+    USING event_type::text::v1_event_type_olap_new;
+
+ALTER TABLE v1_task_events_olap
+    ALTER COLUMN event_type TYPE v1_event_type_olap_new
+    USING event_type::text::v1_event_type_olap_new;
+
```
mrkaye97: any idea if these will lock these tables and perform a sequential scan to change the types? I didn't see anything on that in the docs on a cursory glance, but if it does that might be a non-starter

mrkaye97: oh wait, hold on 😅 are we only adding a value here? if yes, we can 

```
ALTER TYPE v1_event_type_olap ADD VALUE 'COULD_NOT_SEND_TO_WORKER';
```

cmd/hatchet-migrate/migrate/migrations/20251224145920_v1_0_65.sql
```diff
@@ -0,0 +1,13 @@
...
+-- +goose StatementBegin
+
+ALTER TYPE v1_event_type_olap ADD VALUE 'COULD_NOT_SEND_TO_WORKER';
+
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+
+ALTER TYPE v1_event_type_olap DROP VALUE 'COULD_NOT_SEND_TO_WORKER';
```
mrkaye97: hehe, this one won't work - there's no `DROP VALUE`, so to have a down migration here we'll need to create a new enum, cast the type of the column, drop, and rename like you had before. The hope is we never need to actually run the down migration...

mrkaye97 (approved):

PR #2711, opened 2025-12-24
fix: minor fe bugs and nits

grutt:
# Description

Fixes primary scroll area, redundant breadcrumbs, default pylon button location

## Type of change

image
image
image
image
image

simple menu when not enabled
image

### DIFF
--- a/frontend/app/src/components/v1/nav/help-dropdown.tsx
+++ b/frontend/app/src/components/v1/nav/help-dropdown.tsx
@@ -138,9 +138,7 @@ export function HelpDropdown({
 
               setOpen(true);
             }}
-            aria-label={
-              pylon.enabled ? 'Open Support Chat' : 'Open Support Menu'
-            }
+            aria-label="Open Support Chat"
             selected={open}
           />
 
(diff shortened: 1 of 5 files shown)


frontend/app/src/components/v1/nav/help-dropdown.tsx
```diff
@@ -9,20 +10,86 @@ import {
...
+      </DropdownMenuItem>
+      <DropdownMenuItem
+        variant="interactive"
+        onClick={() =>
+          window.open('https://hatchet.run/office-hours', '_blank')
+        }
+      >
+        <BiCalendar className="mr-2" />
+        Schedule Office Hours
+      </DropdownMenuItem>
```
mrkaye97: maybe I'm confused, but why not make these links instead of using an `onClick` handler?

mrkaye97: ah I see these were just moved from below, but I think the question still stands 😅

grutt: haha idk, but good call

mrkaye97 (approved):

PR #2713, opened 2025-12-24
fix: fresh readme logooooo 

grutt:
# Description

Out with the old, in with the new

## Type of change


mrkaye97 (approved):
New logo just dropped 🔥

abelanger5 (approved):
I reversed these, I think it's correct now?

Before:

Image

After: 

Image

PR #2714, opened 2025-12-25
Fix: Static rate limits resetting to zero

### LINKED ISSUE #1834: [BUG] Static rate limits reset their own limits to 0

mrkaye97:
# Description

Fixes an issue where static rate limits reset themselves to zero because we would pass `null` for the `limit` value. Instead, passing a sentinel `-1`, and then checking if the limit value is `>= 0` before upserting

Fixes #1834 

## Type of change

### DIFF
--- a/pkg/v1/task/task.go
+++ b/pkg/v1/task/task.go
@@ -130,6 +130,11 @@ func makeContractTaskOpts(t *TaskShared, taskDefaults *create.TaskDefaults) *con
 			LimitValuesExpr: rateLimit.LimitValueExpr,
 		}
 
+		if rateLimit.LimitValueExpr == nil {
+			negOne := "-1"
+			rateLimit.LimitValueExpr = &negOne
+		}
+
 		if rateLimit.Units != nil {
 			units32 := int32(*rateLimit.Units) // nolint: gosec
 			rlContract.Units = &units32
--- a/pkg/v1/task/task.go
+++ b/pkg/v1/task/task.go
@@ -132,7 +132,7 @@ func makeContractTaskOpts(t *TaskShared, taskDefaults *create.TaskDefaults) *con
 
 		if rateLimit.LimitValueExpr == nil {
 			negOne := "-1"
-			rateLimit.LimitValueExpr = &negOne
+			rlContract.LimitValuesExpr = &negOne
 		}
 
 		if rateLimit.Units != nil {
--- a/pkg/repository/v1/scheduler_queue.go
+++ b/pkg/repository/v1/scheduler_queue.go
@@ -511,9 +511,12 @@ func (d *queueRepository) GetTaskRateLimits(ctx context.Context, queueItems []*s
 			continue
 		}
 
-		upsertRateLimitBulkParams.Keys = append(upsertRateLimitBulkParams.Keys, key)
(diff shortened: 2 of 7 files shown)


pkg/v1/task/task.go
```diff
@@ -130,6 +130,11 @@ func makeContractTaskOpts(t *TaskShared, taskDefaults *create.TaskDefaults) *con
 			LimitValuesExpr: rateLimit.LimitValueExpr,
 		}
 
+		if rateLimit.LimitValueExpr == nil {
+			negOne := "-1"
+			rateLimit.LimitValueExpr = &negOne
+		}
```
mrkaye97, line 136: @mnafees is this the right place to do this? 😅

abelanger5 (approved):

ameya-taliz:
@mrkaye97 - this fixed the issue with static rate limits in the cloud version, but in self-hosted, now static rate limits get set to be -1 whenever a rate limit is consumed, causing all tasks to be queued.

mrkaye97:
hi @ameya-taliz - which version of the engine are you running?

ameya-taliz:
Pulled latest image today, which looks like version 2.35.1:
```
hatchet-engine:
    image: ghcr.io/hatchet-dev/hatchet/hatchet-engine:latest
    command: /hatchet/hatchet-engine --config /hatchet/config
    restart: on-failure
    depends_on:
      setup-config:
        condition: service_completed_successfully
      migration:
        condition: service_completed_successfully
    ports:
      - "7077:7070"
    environment:
      DATABASE_URL: "postgres://hatchet:hatchet@postgres:5432/hatchet"
      SERVER_GRPC_BIND_ADDRESS: "0.0.0.0"
      SERVER_GRPC_INSECURE: "t"
    volumes:
      - hatchet_certs:/hatchet/certs
      - hatchet_config:/hatchet/config
```

mrkaye97:
ah gotcha - the `latest` tag is very old (months). the most recent tags we've pushed is `v0.74.2`, so pulling that should work!

ameya-taliz:
confirmed that worked - thank you for the quick response! Happy New Year!

mrkaye97:
great! happy new year to you too :)

PR #2722, opened 2025-12-26
Fix: Goroutine to periodically extend lease during reconciliation

mrkaye97:
# Description

Adds a goroutine that runs a ticker that periodically extends the lease while reconciliation is running so that we don't have zombie jobs running expensive queries and snowballing

## Type of change


pkg/repository/v1/olap.go
```diff
@@ -3250,7 +3294,7 @@ func (p *OLAPRepositoryImpl) ProcessOLAPPayloadCutovers(ctx context.Context, ext
 	}
 
 	mostRecentPartitionToOffload := pgtype.Date{
-		Time:  time.Now().Add(-1 * *inlineStoreTTL),
+		Time:  time.Now().Add(-1 * (*inlineStoreTTL + 12*time.Hour)),
```
mrkaye97, line 3297: drive by - this will make our offsets 36h instead of 24h, so the job will start running for e.g. 12/24/2025 at 12:00:00UTC on 12/26/2025, instead of at midnight

abelanger5 (approved):

PR #2725, opened 2025-12-27
Feat: Run detail getter on the engine

mrkaye97:
# Description

Adds a run detail getter to the engine which returns an object in this shape:

```
enum RunStatus {
    QUEUED = 0;
    RUNNING = 1;
    COMPLETED = 2;
    FAILED = 3;
    CANCELLED = 4;
}

message TaskRunDetail {
    string external_id = 1; // the external id (uuid) of the task run
    RunStatus status = 2; // the status of the task run
    optional string error = 3; // (optional) error message from the task run, if any
    optional bytes output = 4; // (optional) the output payload for the task run
    string readable_id = 5; // the readable id of the task
}

message GetRunDetailsResponse {
    bytes input = 1; // the input payload for the workflow run
    RunStatus status = 2; // the status of the workflow run
    map<string, TaskRunDetail> task_runs = 3; // map of task run external ids to their details
}
```

The idea here is you could poll the top level `status` until it's either completed, failed, or cancelled, and then traverse the individual tasks in the DAG to get their outputs, statuses, etc.

Also wrote some Python tests to make sure that the statuses propagated through correctly and such

## Type of change


api-contracts/v1/workflows.proto
```diff
@@ -67,6 +68,15 @@ enum RateLimitDuration {
     YEAR = 6;
 }
 
+enum RunStatus {
+    QUEUED = 0;
+    RUNNING = 1;
+    COMPLETED = 2;
+    FAILED = 3;
+    CANCELLED = 4;
+}
```
mrkaye97, line 77: this is a bit icky since we already have this type on the OpenAPI side, but I wasn't sure if there's a better way

internal/services/admin/v1/server.go
```diff
@@ -407,6 +407,128 @@ func (a *AdminServiceImpl) TriggerWorkflowRun(ctx context.Context, req *contract
...
+}
+
+func (a *AdminServiceImpl) all(statuses []string, target string) bool {
+	for _, status := range statuses {
+		if status != target {
+			return false
+		}
+	}
+	return true
+}
```
mrkaye97: maybe worth moving these out into some helper package?

internal/services/admin/v1/server.go
```diff
@@ -407,6 +407,128 @@ func (a *AdminServiceImpl) TriggerWorkflowRun(ctx context.Context, req *contract
...
+			Status:     status,
+			Error:      details.Error,
+			Output:     details.OutputPayload,
+			ReadableId: string(readableId),
+			ExternalId: details.ExternalId,
+		}
+	}
+
+	done := !a.any(statuses, "QUEUED") && !a.any(statuses, "RUNNING")
+	derivedWorkflowRunStatus := a.deriveWorkflowRunStatus(ctx, statuses)
```
mrkaye97: I was thinking about moving this out into a helper too - IIRC we're doing this in the database on the OLAP side so we don't have this defined anywhere else, but might be wrong

pkg/repository/task.go
```diff
@@ -3888,3 +3892,160 @@ func (r *TaskRepositoryImpl) FindOldestTaskInsertedAt(ctx context.Context) (*tim
...
+		return nil, fmt.Errorf("failed to list finalized workflow runs: %w", err)
+	}
+
+	if len(flat) == 0 {
+		return nil, nil
+	}
+
+	var inputRetrieveOpt RetrievePayloadOpts
+	firstTask := flat[0]
+	isDag := firstTask.DagID.Valid
```
mrkaye97, line 3944: is there a better way to do this given the info we have here? I think it'll work fine, since we only accept a single external id, so grabbing the first task from this `flat` array should be safe

pkg/repository/v1/task.go
```diff
@@ -3888,3 +3892,160 @@ func (r *TaskRepositoryImpl) FindOldestTaskInsertedAt(ctx context.Context) (*tim
...
+
+		if ok {
+			extractedInputBytes, err := json.Marshal(extractedInput)
+
+			if err != nil {
+				return nil, fmt.Errorf("failed to marshal extracted task input payload: %w", err)
+			}
+
+			input = extractedInputBytes
+		}
```
mrkaye97: wasn't sure about a better way to do this either, but it's also really janky - will look around for a better place to put this logic

pkg/repository/sqlcv1/tasks.sql
```diff
@@ -110,13 +110,12 @@ WITH lookup_rows AS (
...
-        d.external_id AS workflow_run_external_id
+        t.step_readable_id,
+        l.external_id AS workflow_run_external_id
     FROM
         lookup_rows l
     JOIN
-        v1_dag d ON d.id = l.dag_id AND d.inserted_at = l.inserted_at
-    JOIN
-        v1_dag_to_task dt ON dt.dag_id = d.id AND dt.dag_inserted_at = d.inserted_at
+        v1_dag_to_task dt ON dt.dag_id = l.dag_id AND dt.dag_inserted_at = l.inserted_at
```
mrkaye97, line 118: pretty sure this is a straightforward performance improvement - don't think we need the join to `v1_dag` here, since we don't use any of its cols, so we should be able to go straight from the lookup table to `v1_dag_to_task` afaict

internal/listutils/helpers.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+func All[T comparable](statuses []T, target T) bool {
+	for _, status := range statuses {
+		if status != target {
+			return false
+		}
+	}
+	return true
+}
```
mrkaye97, line 34: I'll probably end up adding more FP stuff here (e.g. `map`, `flatmap`, etc.) but figured I'd start with what I need for now

internal/services/admin/v1/server.go
```diff
@@ -406,6 +408,69 @@ func (a *AdminServiceImpl) TriggerWorkflowRun(ctx context.Context, req *contract
...
+	}
+
+	details, err := a.repo.Tasks().GetWorkflowRunResultDetails(ctx, tenantId, externalId.String())
+
+	if err != nil {
+		return nil, fmt.Errorf("could not get workflow run result details: %w", err)
+	}
+
+	if details == nil {
+		return &contracts.GetRunDetailsResponse{}, nil
```
abelanger5: Is it expected to return an empty object? This seems like surprising behavior instead of returning a 404

pkg/repository/task.go
abelanger5: seems fine to me

pkg/repository/task.go
```diff
@@ -3902,3 +3907,156 @@ func (r *TaskRepositoryImpl) FindOldestTaskInsertedAt(ctx context.Context) (*tim
...
+	finalizedRun := finalizedWorkflowRuns[0]
+
+	for _, event := range finalizedRun.OutputEvents {
+		outputs[event.StepReadableID] = event.Output
+		errors[event.StepReadableID] = event.ErrorMessage
+
+		status, err := statusutils.V1RunStatusFromEventType(event.EventType)
+
+		if err != nil {
+			r.l.Warn().Msgf("failed to parse event type %s: %v", event.EventType, err)
```
abelanger5: Error log here? seems like we'd want to be alerted on this.

abelanger5 (approved):

internal/services/admin/v1/server.go
mrkaye97: yeah we can 404, I'll change that

pkg/repository/task.go
mrkaye97: yeah good idea

PR #2728, opened 2025-12-29
refactor: remove v0 paths from codebase

abelanger5:
# Description

First part of engine cleanup. Removes deprecated v0 engine code paths. Migrates all tenants which are still on V0 to V1 and prevents downgrades to V1. Deprecates REST API methods which depended on v0 code paths (like the old MQ implementation). 

## Type of change


mrkaye97 (approved):
fantastic 🙌 

thoughts on renaming e.g. `ingestEventV1`, `triggerWorkflowV1,` etc. to `ingestEvent`, `triggerWorkflow`, ...?

abelanger5:
> thoughts on renaming e.g. `ingestEventV1`, `triggerWorkflowV1,` etc. to `ingestEvent`, `triggerWorkflow`, ...?

I'm trying to avoid changing anything on a v1 path at the moment, since there's still a lot of cleanup to do, but I think we could do this in a future PR!

PR #2729, opened 2025-12-30
refactor: remove v0-exclusive database queries

abelanger5:
# Description

Removes database queries exclusively used by v0 logic. 

## Type of change


internal/services/controllers/retention/controller.go
abelanger5, line 1: I think we'll be able to remove this process entirely at some point soon, we just need a good migration path for cleaning up old workers, which will come after doing some schema-level cleanup.

api/v1/server/handlers/workers/get.go
```diff
@@ -95,7 +27,7 @@ func (t *WorkerService) workerGetV1(ctx echo.Context, tenant *dbsqlc.Tenant, req
 		return nil, err
 	}
 
-	slotState, recent, err := t.config.V1.Workers().ListWorkerState(
+	slotState, err := t.config.V1.Workers().ListWorkerState(
```
abelanger5, line 30: This is potentially a breaking change, but I don't think it is. The `recent` array was reading from the old `WorkerAssignEvent` table and then joining with step runs, which will not work on v0, so this slice should always have been empty on a v1 engine.

api/v1/server/handlers/events/bulk_create.go
abelanger5, line 1: Note: we should remove these endpoints soonish, I just figure it's easier to get a 400 in response than a 404 or 500. Though modifying the populators might result in a 500 anyway.

mrkaye97 (approved):
😍

PR #2730, opened 2025-12-30
refactor: consolidate repository methods

abelanger5:
# Description

Consolidates repository methods into a single package. 

## Type of change


api/v1/server/handlers/events/create.go
abelanger5, line 1: Need to make the error better here instead of just `panic`

mrkaye97 (approved):
rubber stamping since it's a massive change, I think the compiler + the tests + clicking around the FE should be a pretty good indication here 👍

PR #2733, opened 2025-12-30
Chore: Remove v0 Docs

mrkaye97:
# Description

Removing the v0 docs now that v0 is effectively gone. I'll also remove the Vercel app once this merges

## Type of change


abelanger5 (approved):
go for it!

PR #2738, opened 2025-12-31
Proposal: Shrink task metrics chart a bunch

mrkaye97:
# Description

I think we can reduce the amount of vertical space that the run graph over time takes by doing something like this. Happy to just close it too, but thought it was worth a shot since right now this graph takes up like 1/3 of the vertical space we have

Screenshot 2025-12-30 at 9 44 01 PM

Screenshot 2025-12-30 at 9 42 41 PM

## Type of change

### DIFF
--- a/frontend/app/src/components/v1/molecules/charts/zoomable.tsx
+++ b/frontend/app/src/components/v1/molecules/charts/zoomable.tsx
@@ -397,14 +397,13 @@ function ChildAreaChart<T extends string>({
         onMouseMove={handleMouseMove}
         onMouseUp={handleMouseUp}
       >
-        <CartesianGrid vertical={false} />
         <XAxis
           dataKey="date"
           tickFormatter={formatXAxis}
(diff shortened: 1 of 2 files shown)


abelanger5 (commented):
While I am a fan of having this take up less space, I'm not a fan of the smoothing behavior here, I think it's helpful to see the discrete buckets. Is it possible to shrink it without changing the chart type?

mrkaye97:
yep! wdyt of this?

Screenshot 2025-12-31 at 8 03 33 AM

abelanger5:
> yep! wdyt of this?

Missed this! I like it, I think we can get this merged.

mrkaye97:
> > yep! wdyt of this?
> 
> Missed this! I like it, I think we can get this merged.

sounds good! think I just need a stamp here and then I'll get it merged in

abelanger5 (approved):

PR #2739, opened 2025-12-31
Fix: Make metrics in graph align with badge metrics

mrkaye97:
# Description

Been very confusing that the metrics in the graph would show something different (task event counts) than the badges, so this fixes that

## Type of change

### DIFF
--- a/api/v1/server/handlers/v1/tasks/get_point_metrics.go
+++ b/api/v1/server/handlers/v1/tasks/get_point_metrics.go
@@ -78,11 +78,11 @@ func convertToGenMetrics(metrics []*sqlcv1.GetTaskPointMetricsRow) []gen.V1TaskP
 	converted := make([]gen.V1TaskPointMetric, len(metrics))
 
 	for i, metric := range metrics {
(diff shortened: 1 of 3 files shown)


abelanger5 (commented):
I'm not sure I understand, aren't the counts based on the task's inserted_at field? Also just to clarify, this doesn't tackle aligning the failed + success counts (where retries which succeed on failure are shown in the graph), just the timestamps correct?

mrkaye97:
> I'm not sure I understand, aren't the counts based on the task's inserted_at field? Also just to clarify, this doesn't tackle aligning the failed + success counts (where retries which succeed on failure are shown in the graph), just the timestamps correct?

sorry for the confusion there! the goal here is to make it so that if you add up all of the counts of the bars in the bar chart, they sum to the total shown on the badge, since people (me included) have been confused by "why do I see all of these failed runs but then when I filter by failed it doesn't show anything?" and the reason for that was the graph showing events but the badge only counting the final statuses.

abelanger5 (approved):

PR #2744, opened 2025-12-31
Fix: Child runs counts missing filter

mrkaye97:
# Description

Fixes a bug in pagination with child runs

## Type of change


pkg/repository/sqlcv1/olap-overwrite.sql
mrkaye97, line 1: drive by deleted this - I continue to find these SQL files (the duplicates of the `*-overwrite.go` super confusing

abelanger5 (approved):

PR #2746, opened 2026-01-01
Fix: More frontend work - table cleanup, worker detail page improvements, etc.

mrkaye97:
# Description

Doing some more frontend work:

1. Added a `SimpleTable` for tables where we just need a handful of rows without pagination / react table / state management bloat to clean up a bunch of components and limit how much conditional logic we need in the `DataTable`. We were basically just using react table in a bunch of places we really didn't need to be
2. Overhauled the worker detail page (screenshot below here)
3. Slightly changed the UX of the `DataTable` (removed the border so the empty / not all the way populated state looks less weird, etc.)
4. Cleaned up the rate limits page a bit

Worker detail page:

Screenshot 2026-01-01 at 7 28 32 PM

Runs table without border:

Screenshot 2026-01-01 at 7 30 07 PM

Resource limits (example of `SimpleTable`:

Screenshot 2026-01-01 at 7 35 04 PM

Partially populated table

Screenshot 2026-01-01 at 7 36 23 PM


frontend/app/src/components/v1/molecules/simple-table/simple-table.tsx
mrkaye97, line 1: added this to handle all the cases where we need a table but don't need pagination, etc. and so we really don't need to use the `DataTable`, which lets us remove a bunch of props from that one

frontend/app/src/pages/main/v1/workers/$worker/index.tsx
```diff
@@ -213,149 +223,242 @@ export default function ExpandedWorkflowRun() {
...
                 )}
-                {worker.runtimeInfo?.os && (
-                  <div>
-                    <b>OS</b>: {worker.runtimeInfo?.os}
+              </div>
+              {maxSlots > 0 && (
+                <div className="space-y-1">
+                  <div className="h-2 w-full overflow-hidden rounded-full bg-blue-100 dark:bg-blue-900">
+                    <div
+                      className="h-full bg-red-600 transition-all dark:bg-red-500"
```
abelanger5: The red screams error for me here, can we use a different color instead?

abelanger5 (approved):
💯

PR #2750, opened 2026-01-02
refactor: set a connection-level statement timeout

abelanger5:
# Description

Sets a global statement-level timeout of 30 seconds rather than having each tx individually set a statement timeout, which causes more queries sent to Postgres on each tx. It's still possible to overwrite the connection-level timeout with a statement-level timeout using the new method `PrepareTxWithStatementTimeout`. 

Pretty much all of these tx-level values were chosen magically, either 5, 10, 15, 25, or 30 seconds, except for the intentionally very long-running statements that call `ANALYZE` or handle partitioning. There's not a good reason for the lower values, it's just important that there's some sort of statement-level timeout to avoid a class of edge cases we've seen where queries leak through to Postgres from pgx. 

This should be fairly low-risk; the biggest risk is that a query which was firing on the 5-second timeout is firing on the 30-second timeout as well, but now there might be more queries timing out at the same time which is not great for the DB. 

This was originally part of a set of performance improvements in #2258 and made a minimal impact in some benchmarking. 

## Type of change

### DIFF
--- a/pkg/config/loader/loader.go
+++ b/pkg/config/loader/loader.go
@@ -145,7 +145,9 @@ func (c *ConfigLoader) InitDataLayer() (res *database.Layer, err error) {
 
 		conn.TypeMap().RegisterType(t)
 
-		return nil
+		_, err = conn.Exec(ctx, "SET statement_timeout=30000")
+
+		return err
 	}
 
 	config, err := pgxpool.ParseConfig(databaseUrl)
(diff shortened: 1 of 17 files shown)


mrkaye97 (commented):
unless I'm missing something, I think this will break the payloads job because we have some queries that run outside of a tx but take 5-10 mins to complete - is that right?

abelanger5:
> unless I'm missing something, I think this will break the payloads job because we have some queries that run outside of a tx but take 5-10 mins to complete - is that right?

Hmm, yeah this applies to all statements - I can write a helper which doesn't use a TX but sets the statement timeout for the connection, and then unsets it afterwards. 

I'd just really like to avoid setting the statement timeout 2 times on every single tx, even a small network latency becomes more significant.

mrkaye97 (approved):

PR #2754, opened 2026-01-06
fix: typing for dependency injection

### LINKED ISSUE #2605: [FEAT] Let dependencies for dependency injection depend on other dependencies (python sdk)
Right now, dependencies cannot depend on other dependencies as it throws an `Dependency function <name> must have exactly two parameters: input and ctx`. It would be helpful if a dependency could depend on another dependency.

I'd like to be able to do something like this:
```python
class Settings(BaseModel):
    user: str
    host: str

def get_settings_dep(
    input: EmptyModel,
    ctx: Context,
) -> Settings:
    return get_settings()

def get_client(
    input: EmptyModel,
    ctx: Context,
    settings: Annotated[Settings, Depends(settings)],
) -> Client:
    return Client(settings.host)

@hatchet.task()
async def test_task(
    input: EmptyModel,
    ctx: Context,
    settings: Annotated[Settings, Depends(settings)],
    client: Annotated[Client, Depends(get_client)],
) -> EmptyModel:
    do_thing(settings.user, client)
    return EmptyModel()

```

Having to call a function in each dependency, making sure to handle caching and such myself
```python
class Settings(BaseModel):
    user: str
    host: str

def get_settings_dep(
    input: EmptyModel,
    ctx: Context,
) -> Settings:
    return get_settings()

def get_client(
    input: EmptyModel,
    ctx: Context,
) -> Client:
    return Client(get_settings().host)

@hatchet.task()
async def test_task(
    input: EmptyModel,
    ctx: Context,
    settings: Annotated[Settings, Depends(settings)],
    client: Annotated[Client, Depends(get_client)],
) -> EmptyModel:
    do_thing(settings.user, client)
    return EmptyModel()

```

**Additional context**
It seems that Hatchets implementation is made to be similar to Fastapi and Fastapi does support this.

Thanks!

StefanBRas:
# Description
Fixes #2605 (already closed yes, but the follow-up type error). See the issue for more discussion.

In short the type hint for dependency injection function requires it to take exactly two arguments, but now that they can be chained it should allow more. It would be nice to somehow type that you can ONLY give it additional args if these are `Annotated` with `Depends` but I don't think that's possible, since typecheckers completely ignore `Annotated` except for the first type. So it's typed as `..., *args: Any, **kwargs: Any`.

I had to change some `TypeVar`s to be contra/co-variant since protocols require that. I took the naming convention from [here](https://pylint.pycqa.org/en/latest/user_guide/messages/convention/invalid-name.html#predefined-naming-patterns) though the specific case is not exactly described.

Out of scope for this, but I think you should consider running Pyright as part of the linting step, maybe just on the examples.

You probably want to check this PR out and test it for yourself - I didn't really do an in-depth testing of it myself

## Type of change

## What's Changed

- [X] fixed typing of dependency injection functions


mrkaye97 (approved):
thanks for the fix! I bundled in a couple other small things, hopefully that's okay 😅

PR #2757, opened 2026-01-06
New onboarding flow

mnafees:
# Description

Introduces the new onboarding flow.

## Type of change


frontend/app/src/hooks/use-current-user.ts
```diff
@@ -0,0 +1,16 @@
...
+    retry: false,
+  });
+
+  return {
+    currentUser: currentUserQuery.data,
+    isLoading: currentUserQuery.isLoading,
+    isError: currentUserQuery.isError,
+    error: currentUserQuery.error,
+  };
+}
```
mrkaye97, line 18: this was what I was mentioning this morning - I think we should maybe have a shared provider for the organization, tenant, and user. we can probably also get rid of `useCurrentTenantId` now, since we can access the tenant id from the route params on any tenanted path, I think

frontend/app/src/pages/authenticated.tsx
```diff
@@ -197,18 +185,18 @@ export default function Authenticated() {
...
       navigate({ to: appRoutes.authLoginRoute.to, replace: true });
     }
-  }, [isAuthPage, navigate, userQuery.error]);
+  }, [isAuthPage, navigate, userError]);
 
   return (
-    <PostHogProvider user={userQuery.data}>
-      <SupportChat user={userQuery.data}>
+    <PostHogProvider user={currentUser}>
+      <SupportChat user={currentUser}>
```
mrkaye97, line 225: ideally we wouldn't need to prop drill here, and instead could use the same hook inside of these too

frontend/app/src/pages/onboarding/create-tenant/components/tenant-create-form.tsx
```diff
@@ -287,9 +241,23 @@ export function TenantCreateForm({
...
 
+        <div className="grid gap-3">
+          <Label htmlFor="name">Where did you hear about us? (optional)</Label>
+          <Input
+            id="referral_source"
+            placeholder="e.g. Twitter, LinkedIn, etc."
+            type="text"
+            autoCapitalize="none"
+            autoCorrect="off"
+          />
```
mrkaye97: we might want a dropdown here so there aren't misspellings, etc.

frontend/app/src/pages/onboarding/create-tenant/components/tenant-create-form.tsx
```diff
@@ -316,6 +284,46 @@ export function TenantCreateForm({
...
+
+        {/* Help Section */}
+        <div className="text-center text-sm text-muted-foreground">
+          Have questions?{' '}
+          <a
+            href="https://docs.hatchet.run"
+            target="_blank"
+            rel="noopener noreferrer"
+            className="text-primary hover:underline"
+          >
```
mrkaye97, line 269: I feel like we should have an `ExternalLink` component that does this - we have this duplicated too many places 😅

frontend/app/src/lib/api/generated/data-contracts.ts
```diff
@@ -1264,8 +1264,6 @@ export interface CreateTenantRequest {
   engineVersion?: TenantVersion;
   /** The environment type of the tenant. */
   environment?: TenantEnvironment;
```
mrkaye97, line 1271: I wonder if we can remove this `environment` stuff too - is this useful to us at all?

mrkaye97: upon 30 seconds of more thought and examining the git blame, I feel pretty strongly about removing this 😅 we added it in the same PR as this other stuff, and I think it doesn't provide much value and really can be quite confusing / restrictive

frontend/app/src/pages/onboarding/create-tenant/types.ts
```diff
@@ -17,8 +17,6 @@ export interface OnboardingStepProps<T = any> {
 export interface OnboardingFormData {
   name: string;
   slug: string;
-  hearAboutUs: string | string[];
-  whatBuilding: string | string[];
   environment: TenantEnvironment;
   tenantData: { name: string; environment: TenantEnvironment };
```
mrkaye97: if we remove the environment, I imagine we can remove this prop too? a bit confused about why this exists tbh since the name and environment exist elsewhere anyway

frontend/app/src/pages/onboarding/invites/index.tsx
```diff
@@ -58,11 +60,21 @@ export default function Invites() {
...
     useOrganizations();
+  const { capture } = useAnalytics();
 
   const { tenantInvites, orgInvites } = useLoaderData({
     from: appRoutes.onboardingInvitesRoute.to,
   }) as Awaited<ReturnType<typeof loader>>;
 
+  // Track invites page view
+  useEffect(() => {
+    capture('onboarding_invites_viewed', {
```
mrkaye97, line 71: flagging that we probably want to follow whatever naming convention literally wants / has established elsewhere

mrkaye97 (commented):
left some small comments, but generally looks great! just would be in favor of doing some additional refactoring here since we're working in the same places anyways

frontend/app/src/lib/api/generated/data-contracts.ts
mrkaye97: @mnafees bumping this in case it got missed!

frontend/app/src/hooks/use-current-user.ts
```diff
@@ -0,0 +1,18 @@
+import { useAppContext } from '@/providers/app-context';
+
+/**
+ * Hook to access the current user
+ *
+ * Now backed by AppContext for better performance and consistency.
+ * Maintains backward compatibility with the old API.
+ */
```
mrkaye97, line 8: nit: can we remove all these llm comments :P they're not helpful imo

mrkaye97: also, do we need this hook? or can we use `useAppContext` instead and remove this? I've noticed that Claude likes to do this sort of thing a lot, where it wants to maintain backwards compat but really doesn't need to do that

frontend/app/src/hooks/use-organizations.ts
```diff
@@ -6,14 +6,21 @@ import {
...
+import { useAppContext } from '@/providers/app-context';
 import { useQuery, useMutation } from '@tanstack/react-query';
 import { useMemo, useCallback } from 'react';
 
+/**
+ * Hook for organization data and operations
+ *
+ * Now backed by AppContext for better performance.
+ * Gets organization data from context, but keeps all mutation logic here.
+ */
```
mrkaye97, line 18: useless comment

frontend/app/src/pages/authenticated.tsx
```diff
@@ -172,20 +161,38 @@ export default function Authenticated() {
...
+            });
+          })
+          .catch(() => {
+            // On error, default to runs page
+            navigate({
+              to: appRoutes.tenantRunsRoute.to,
+              params: { tenant: targetTenant.metadata.id },
+              replace: true,
+            });
+          });
```
mrkaye97, line 195: I think this might get annoying - I feel like maybe we should redirect to the runs, but then indicate in some other way that the user should to the overview. would be curious what sebastian thinks as far as UX goes here

frontend/app/src/pages/main/v1/side-nav-items.tsx
```diff
@@ -3,16 +3,30 @@ import {
...
+//   CalendarDaysIcon,
+//   CpuChipIcon,
+//   PlayIcon,
+//   ScaleIcon,
+//   ServerStackIcon,
+//   Squares2X2Icon,
+// } from '@heroicons/react/24/outline';
+// import { ClockIcon, GearIcon } from '@radix-ui/react-icons';
+// import { Filter, SquareActivityIcon, WebhookIcon } from 'lucide-react';
+
```
mrkaye97: remove these?

frontend/app/src/providers/app-context.tsx
```diff
@@ -0,0 +1,171 @@
...
+ * Shared application context providing user, tenant, and organization data
+ *
+ * This provider consolidates the previously separate hooks:
+ * - useCurrentUser
+ * - useTenantDetails
+ * - useOrganizations
+ *
+ * Pattern inspired by PostHogProvider - provides a single source of truth
+ * for user, tenant, and organization data across the application.
+ */
```
mrkaye97, line 27: think we can remove this one too

frontend/app/src/providers/app-context.tsx
```diff
@@ -0,0 +1,171 @@
...
+  // Get tenant ID from route params (following TanStack Router best practices)
+  // This replaces the old useCurrentTenantId pattern
+  const params = useParams({ strict: false });
+  const [lastTenant, setLastTenant] = useAtom(lastTenantAtom);
+  const tenantId = params.tenant || lastTenant?.metadata.id;
+
+  // Fetch current user
+  const currentUserQuery = useQuery({
+    ...queries.user.current,
+    retry: false,
```
mrkaye97, line 71: I feel like we probably do want to retry here, just in case there's a blip

frontend/app/src/hooks/use-current-user.ts
grutt: will address in separate PR

frontend/app/src/pages/onboarding/create-tenant/components/tenant-create-form.tsx
grutt: agreed will address in future pr

frontend/app/src/lib/api/generated/data-contracts.ts
grutt: i just removed the ui affordances... its sort of interesting to see if folks are in prod but we can figure this out in other ways later

mrkaye97: yeah sounds good

frontend/app/src/hooks/use-tenant.tsx
```diff
@@ -76,17 +82,8 @@ export function useTenantDetails() {
...
-    }
-
-    return memberships?.find(
-      (membership) => membership.tenant?.metadata.id === tenantId,
-    );
-  }, [tenantId, memberships]);
-
-  const tenant = membership?.tenant;
+  // Tenant and membership now come from AppContext
+  // No need to compute them here anymore
```
mrkaye97, line 86: we need to remove these sorts of comments 😅 for some reason Claude + Co. love them, but they're so useless lol

frontend/app/src/pages/authenticated.tsx
```diff
@@ -228,3 +246,7 @@ export default function Authenticated() {
     </PostHogProvider>
   );
 }
+
+export default function Authenticated() {
+  return <AuthenticatedInner />;
+}
```
mrkaye97, line 252: do we need this wrapper if it just returns the inner component?

mrkaye97 (approved):
there's still some cleanup to do but I think we should ship it and do that next, this looks fine 👍

abelanger5 (approved):

PR #2761, opened 2026-01-08
Fix: Chunk and recursively retry too-large message sends

mrkaye97:
# Description

If we try to send a message over the internal queue that's too large, split it in chunks and retry each chunk sequentially.

## Type of change


abelanger5 (approved):
nice! very clean and simple fix 💯

PR #2780, opened 2026-01-12
Fix: Concurrency display on workflow page

mrkaye97:
# Description

Fixing the workflow version API to properly display concurrency on the workflow page, and did a bit of frontend cleanup there too to improve how that page looks:

No concurrency:
Screenshot 2026-01-12 at 3 34 59 PM

Concurrency (both workflow and task level)

Screenshot 2026-01-12 at 3 27 46 PM

## Type of change


abelanger5 (approved):
nice!

PR #2785, opened 2026-01-13
feat(cli): run hatchet locally

h0rv:
# Description

Adds `--local` flag to `hatchet server start` that runs Hatchet without Docker. Runs API and engine in-process as a single binary, auto-creates the database, and runs in headless mode (no web UI).

## Type of change

## What's Changed

- Add `--local` flag to `hatchet server start` - runs server in-process without Docker
- Auto-create database and set timezone to UTC (no manual `createdb`/`psql` needed)
- Add `--database-url`, `--api-port`, `--grpc-port`, `--healthcheck-port` flags for local mode
- Auto-detect local vs Docker mode in `hatchet server stop`
- Persist encryption keys to `~/.hatchet/local/keys.json` for reuse across restarts
- Add unit tests for identifier validation and driver initialization

## Usage

```bash
# Start with local Postgres (requires Postgres running)
hatchet server start --local

# Custom database URL
hatchet server start --local --database-url "postgresql://user:pass@localhost:5432/mydb"

# Custom ports (useful alongside Docker)
hatchet server start --local --api-port 9080 --grpc-port 9077

# Stop (auto-detects mode)
hatchet server stop

Note: Local mode is headless (no web UI). Use TUI or SDK to interact.
```

## Notes

Binary size increases from ~27MB to ~66MB due to embedding API and engine code for in-process execution. This is the tradeoff for single-binary distribution without external dependencies.

### DIFF
--- a/cmd/hatchet-cli/.goreleaser.yml
+++ b/cmd/hatchet-cli/.goreleaser.yml
@@ -26,8 +26,54 @@ builds:
       - -X github.com/hatchet-dev/hatchet/cmd/hatchet-cli/cli.Version={{.Version}}
     mod_timestamp: "{{ .CommitTimestamp }}"
 
+  - id: hatchet-api
+    dir: ../hatchet-api
+    main: ./main.go
+    binary: hatchet-api
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - darwin
+      - windows
+    goarch:
+      - amd64
+      - arm64
+    goamd64:
+      - v1
+    ignore:
+      - goos: windows
+        goarch: arm64
+    ldflags:
+      - -s -w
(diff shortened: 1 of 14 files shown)


h0rv:
due to the binary size - not sure if this should be a separate release or something

h0rv:
also for context: i want to run hatchet in a non-docker environment. i was excited to see the CLI but it didn't support running without docker.

mnafees:
Hi @h0rv, thank you for the PR. Apologies for not being able to reply sooner. I am going to close this as we have released the new [Hatchet Embedded project](https://github.com/hatchet-dev/hatchet-embedded) which pretty much does the same. We can, in the future, support it directly from the CLI as per your request.

PR #2800, opened 2026-01-15
feat: add sentinel to pydantic de/serialization

### LINKED ISSUE #2799: [FEAT] python-sdk: add sentinel value to pydantic context when deserializing
The general feature is that I would like some way to return secret values (and very large values) directly in the tasks and then have hooks to handle translating these values back and forth.

My idea for the solution is to have something like this:

```python
SecretSerializer = PlainSerializer(_encode_secret, str)

class HatchetOutput(BaseModel):
    @model_validator(mode="before")
    @classmethod
    def resolve_secret(cls, data: Any, info: ValidationInfo) -> Any:
        for field_name, field in cls.model_fields.items():
            if SecretSerializer in field.metadata:
                data[field_name] = _decode_secret(data[field_name])
        return data

class TaskReturn(HatchetOutput):
     secret: Annotated[str, SecretSerializer]
```

However I only want to run the validator in the context of de/serializing the values from the hatchet engine - I want to be able to construct the result type as normal outside it. If I could somehow determine if I was in a hatchet context inside the validator and/or serializer, this would be much easier.

In `runnables.workflow.BaseWorkflow._get_workflow_output` I would like to add the Context like so
```python
   def _get_workflow_input(self, ctx: Context) -> TWorkflowInput:
        return cast(
            TWorkflowInput,
            self.config.input_validator.validate_python(ctx.workflow_input, context=SENTINEL_VALUE),
        )
```
and similarily in `runnables.workflow.Standalone._extract_result`

Same in `runnables.workflow.BaseWorkflow._serialize_input`

Handling the whole thing manually inside each task. I would like to avoid that because of all the boilerplate

**Additional context**
If you agree with the change, I can make the PR. I think the change is not very invasive and not breaking in anyway

(Originally I wrote to pass the `Context` object, but this is not always available, so I changed that)

StefanBRas:
# Description

Basically a way to know during pydantic de/serialization if we are in a hatchet context and lets user act accordingly. 

Fixes #2799

I'm just pushing my initial implementation and wanted to hear feedback as there was multiple ways to do it.

Also I couldn't find any obvious place to test it. Do we need tests? And is there any place that would be good for that?

## Type of change

- [X] This change requires a documentation update

## What's Changed

- [X] Sentinel value added to pydantic serialization and deserialization contexts.


sdks/python/hatchet_sdk/runnables/task.py
```diff
@@ -69,6 +69,8 @@
 T_co = TypeVar("T_co", covariant=True)
 P = ParamSpec("P")
 
+HATCHET_PYDANTIC_SENTINEL = object()
```
mrkaye97: confirming we want to use a plain `object()` here vs. something like a boolean flag. e.g. maybe something like `{"is_in_hatchet_task": True}`? (not too sure on naming)

mrkaye97 (commented):
thanks for the change! I think we're missing a couple of `model_dump` calls here

sdks/python/hatchet_sdk/runnables/task.py
StefanBRas: It's just a preference thing - the nice thing with using object() for the sentinel is that you can never accidentally have another value that is equal to it. Also the id is immutable (I think, at least you have to really try).

The bad things are that it doesn't really say much if you just print the context.

But it's up to you really, an alternative I like to use is to not expose the sentinel to the use but a function so you have something like
```python
_SENTINEL = object()

def is_sentinel(obj: Any) -> bool:
    return obj is _SENTINEL
 ```

Comment (mrkaye97) [reply sdks/python/hatchet_sdk/runnables/task.py]: this seems reasonable to me - would we expose that function directly and expect it to be imported? in any case, based on the docs I think we need the context to be a dictionary?

**Comment by StefanBRas** [comment]
> thanks for the change! I think we're missing a couple of `model_dump` calls here

I think some context is missing, I can't see what "here" is a reference to but a found one place more in `step_run_callback`. At least I think so - it's not immediately clear to me how the function is used

File: sdks/python/hatchet_sdk/runnables/task.py (hunk 1 above)
Comment (StefanBRas) [reply sdks/python/hatchet_sdk/runnables/task.py]: Yes, you would then import the function into your own code so that you in serializers and validators can determine if we are in the hatchet context.

I just tested it, context can be whatever you want

Comment (StefanBRas) [reply sdks/python/hatchet_sdk/runnables/task.py]: Any preference with where the function should live then? I don't think it makes sense to have it in `runnables`

**Comment by mrkaye97** [comment]
pushed a change to fix the last place I saw! I think I'd prefer to put the sentinel into a dict with some key and a boolean value. Maybe the helper function should go in a new submodule, like `hatchet.serde` or something?

File: sdks/python/hatchet_sdk/serde/pydantic.py [hunk 2]
```diff
@@ -0,0 +1,31 @@
...
+    Determines from pydantic ValidationInfo.context whether we are in a Hatchet serialization context.
+
+    Example:
+    ```python
+
+    ```
+    """
+    if context == HATCHET_PYDANTIC_SENTINEL:
+        return True
+    if hasattr(context, "context"):
```
StefanBRas, line 26: I immediately made this error after implementing it, so made this check.

I can't think of a situation where this could happen on purpose

sdks/python/examples/serde/worker.py
StefanBRas, line 1: I don't fully understand the test/doc setup but I tried my best haha.

I verified that I could run this workflow with the expected results, but not sure if it's too verbose

sdks/python/hatchet_sdk/serde/pydantic.py
```diff
@@ -0,0 +1,31 @@
+from logging import getLogger
+from types import MappingProxyType
+from typing import Any
+
+logger = getLogger(__name__)
+
+
+HATCHET_PYDANTIC_SENTINEL = MappingProxyType({"is_hatchet_serialization_context": True})
```
StefanBRas, line 8: Took the liberty of making this `MappingProxyType` so it can't be mutated

StefanBRas (commented):
@mrkaye97 Implemented your suggestions (as I understood them) and added an example

I went freestyle some places with the `MappingProxyType` and logging the warning, so feel free to change/remove those or let me know if you want it in a different way

sdks/python/examples/serde/worker.py
mrkaye97: great stuff! thanks for writing the example. what you should be able to do is write a test in `examples/serde/test_custom_serde.py` (check out some of the other tests - basically we trigger a workflow, fetch the results, and then assert on them). Then, you'll need to register this workflow on the worker in `examples/worker.py`, and then you can run `poetry run pytest -k test_custom_serde` and it should magically work, assuming you have a token set

StefanBRas: Done!

mrkaye97:
this looks great! last thing is to bump the patch version and write a changelog entry, then I'll approve and get this merged!

StefanBRas:
@mrkaye97 Should be ready for merge now, feel free to change the wording of the changelog

mrkaye97 (approved):
🚀 great, thanks!

PR #2810, opened 2026-01-17
docs: make docs more cli-centric

abelanger5:
# Description

Make the CLI the primary entrypoint in the docs. Also simplifies the Hatchet cloud quickstart and some other files. 

## Type of change

- [X] Documentation change (pure documentation change)


mnafees (approved):
LGTM. Should we add info about the dependency on docker for `hatchet server start` to the docs?

PR #2826, opened 2026-01-20
feat: python health check improvements 

grutt:
# Description

Adds additional checks for healthy workers

## Type of change

## What's Changed

- [x] moves the health check endpoint to the listener sidecar process
- [x] checks for blocked event loop in the main thread (configurable threshold)
- [x] checks for healthy listener connection

### DIFF
--- a/sdks/python/hatchet_sdk/worker/action_listener_process.py
+++ b/sdks/python/hatchet_sdk/worker/action_listener_process.py
@@ -249,8 +249,13 @@ class WorkerActionListenerProcess:
 
     async def stop_health_server(self) -> None:
         if self._event_loop_monitor_task is not None:
-            self._event_loop_monitor_task.cancel()
+            task = self._event_loop_monitor_task
             self._event_loop_monitor_task = None
+            task.cancel()
+            try:
+                await task
+            except asyncio.CancelledError:
+                pass
 
         if self._health_runner is None:
             return
--- a/sdks/python/hatchet_sdk/worker/action_listener_process.py
+++ b/sdks/python/hatchet_sdk/worker/action_listener_process.py
@@ -149,7 +149,7 @@ class WorkerActionListenerProcess:
         if self.killing:
             return ("UNHEALTHY", False)
 
-        # If the event loop has been blocked for >5s, report unhealthy.
+        # If the event loop has been blocked longer than the configured threshold, report unhealthy.
         if (
             self._event_loop_blocked_since is not None
             and (time.time() - self._event_loop_blocked_since)
--- a/sdks/python/hatchet_sdk/worker/action_listener_process.py
+++ b/sdks/python/hatchet_sdk/worker/action_listener_process.py
@@ -1,13 +1,19 @@
 import asyncio
 import logging
+import re
 import signal
+import sys
 import time
 from dataclasses import dataclass
 from multiprocessing import Queue
 from typing import Any
 
 import grpc
+from aiohttp import web
+from aiohttp.web_request import Request
+from aiohttp.web_response import Response
 from grpc.aio import UnaryUnaryCall
+from prometheus_client import Gauge, generate_latest
 
 from hatchet_sdk.client import Client
 from hatchet_sdk.clients.dispatcher.action_listener import (
@@ -59,6 +65,8 @@ class WorkerActionListenerProcess:
         handle_kill: bool,
         debug: bool,
         labels: dict[str, str | int],
+        enable_health_server: bool = False,
+        healthcheck_port: int = 8001,
     ) -> None:
         self.name = name
         self.actions = actions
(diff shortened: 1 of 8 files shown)


sdks/python/hatchet_sdk/worker/action_listener_process.py
```diff
@@ -95,6 +115,153 @@ def __init__(
...
+
+            if lag >= self._event_loop_block_threshold_seconds:
+                if self._event_loop_blocked_since is None:
+                    self._event_loop_blocked_since = start + interval
+            else:
+                self._event_loop_blocked_since = None
+
+    def _compute_health(self) -> tuple[str, bool]:
+        if self.killing:
+            return ("UNHEALTHY", False)
```
mrkaye97: we should use an enum for this

sdks/python/hatchet_sdk/worker/action_listener_process.py
```diff
@@ -95,6 +115,153 @@ def __init__(
...
+            await asyncio.sleep(interval)
+            elapsed = time.time() - start
+            lag = max(0.0, elapsed - interval)
+            self._event_loop_last_lag_seconds = lag
+
+            if lag >= self._event_loop_block_threshold_seconds:
+                if self._event_loop_blocked_since is None:
+                    self._event_loop_blocked_since = start + interval
+            else:
+                self._event_loop_blocked_since = None
```
mrkaye97, line 160: will this work? since it's running on the listener, I'd expect the listener basically never gets blocked, and this won't detect if the action runner is blocked?

sdks/python/hatchet_sdk/worker/action_listener_process.py
```diff
@@ -95,6 +115,153 @@ def __init__(
...
+
+    def _compute_health(self) -> tuple[str, bool]:
```
mrkaye97: what's the point of the boolean here? does it map from an enum value to a yes / no if it's healthy? if yes, then I think we should have a `__bool__` method on the enum class instead that does this mapping, then we can call `bool(enum)` to get the value

sdks/python/hatchet_sdk/worker/action_listener_process.py
```diff
@@ -95,6 +115,153 @@ def __init__(
...
+            and (time.time() - self._waiting_steps_blocked_since)
+            > self._event_loop_block_threshold_seconds
+        ):
+            return ("UNHEALTHY", False)
+
+        if self.listener is None:
+            return ("STARTING", False)
+
+        # Avoid false positives before we have any listener connection attempts.
+        last_attempt = getattr(self.listener, "last_connection_attempt", 0.0) or 0.0
```
mrkaye97: we should try to avoid `getattr` since it'll break the type checker

sdks/python/hatchet_sdk/worker/action_listener_process.py
```diff
@@ -95,6 +115,153 @@ def __init__(
...
+        if last_attempt <= 0:
+            return ("STARTING", False)
+
+        listen_strategy = getattr(self.listener, "listen_strategy", "v2")
```
mrkaye97: same here

sdks/python/hatchet_sdk/worker/action_listener_process.py
```diff
@@ -95,6 +115,153 @@ def __init__(
...
+
+        if listen_strategy == "v2":
+            # Require at least one successful heartbeat.
+            time_last_hb = getattr(self.listener, "time_last_hb_succeeded", 0.0) or 0.0
```
mrkaye97: and here

sdks/python/hatchet_sdk/worker/action_listener_process.py
```diff
@@ -69,6 +76,19 @@ def __init__(
...
+        self.healthcheck_port = healthcheck_port
+
+        self._health_runner: web.AppRunner | None = None
+        self._listener_health_gauge: Gauge | None = None
+        self._event_loop_lag_gauge: Gauge | None = None
+        self._event_loop_monitor_task: asyncio.Task[None] | None = None
+        self._event_loop_last_lag_seconds: float = 0.0
+        self._event_loop_blocked_since: float | None = None
+        self._event_loop_block_threshold_seconds: float = float(
+            getattr(self.config.healthcheck, "event_loop_block_threshold_seconds", 5.0)
```
mrkaye97: another `getattr` here - should remove all of these

sdks/python/hatchet_sdk/worker/action_listener_process.py
```diff
@@ -69,6 +76,19 @@ def __init__(
...
+        )
+        self._waiting_steps_blocked_since: float | None = None
```
mrkaye97, line 100: I know we have this in existing code, but I'd prefer to use `datetime` for this

grutt:
@mrkaye97 should have addressed all things, 

tested worker exhaustion and added some casing here

mrkaye97 (approved):

PR #2828, opened 2026-01-20
Feat: Opentelemetry for TS SDK

jishnundth:
# Description

1. OTel Instrumentor for Typescript SDK (Through monkey patching)
2. OTel example for TS SDK instrumentor
3. OTel environment variables for TS SDK

## Type of change


jishnundth:
Hey @mrkaye97, 

Please take a look at this, when you have time. 

The current approach follows monkey patching files for instrumentation during load time, which is the standard pattern used in OTel TypeScript examples as well.

(References)
1. https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/instrumentation-amqplib/src/amqplib.ts
2. https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/instrumentation-graphql/src/instrumentation.ts

However, the issue with this approach is that register instrumentation must occur first, before loading any other libraries. This is because Node.js caches loaded libraries. If a library that needs to be instrumented is imported before the instrumentation is registered, the patching will not occur. The rest of the program will then use the cached version of the library, which does not contain the patched OTel code. As a result, the spans internally added in the patched module will not appear in HyperDX. 

You might already know this, just want to give a heads up on how opentelemetry works for JS.

NOTE: Also, for testing with the examples, the code has to built, packed and installed in another `node_modules` directory. Because patching follows the path of the file inside a node_module directory.

I also had to edit `handleStartStepRun` to throw the errors upstream, to capture the stacktrace in OTel.

jishnundth:
image

It's working pretty good, same as python OTel

mrkaye97:
Great, thanks! Will take a look soon :)

sdks/typescript/src/v1/client/worker/worker-internal.ts
```diff
@@ -652,12 +655,15 @@ export class V1Worker {
       });
 
       try {
-        await future.promise;
+        const taskError = await future.promise;
+        return taskError;
```
mrkaye97: nit: can we `return await future.promise` here? there should probably be a linting rule for this but 🤷

sdks/typescript/src/opentelemetry/instrumentor.ts
```diff
@@ -0,0 +1,674 @@
...
+ * - Step executions (handleStartStepRun, handleCancelStepRun)
+ *
+ * Traceparent context is automatically propagated through metadata.
+ *
+ * The instrumentor uses the global tracer/meter providers by default.
+ * Use `setTracerProvider()` and `setMeterProvider()` to configure custom providers.
+ */
+export class HatchetInstrumentor extends InstrumentationBase<HatchetInstrumentationConfig> {
+  constructor(config: Partial<HatchetInstrumentationConfig> = {}) {
+    super(INSTRUMENTOR_NAME, HATCHET_VERSION, { ...DEFAULT_CONFIG, ...config } as HatchetInstrumentationConfig);
```
mrkaye97: qq: is it possible to do this without the type assertion (`as HatchetInstrumentationConfig`) since that bypasses the type checker?

sdks/typescript/src/opentelemetry/instrumentor.ts
```diff
@@ -0,0 +1,674 @@
...
+  }
+
+  override setConfig(config: Partial<HatchetInstrumentationConfig> = {}): void {
+    super.setConfig({ ...DEFAULT_CONFIG, ...config } as HatchetInstrumentationConfig);
```
mrkaye97: same question here

mrkaye97 (commented):
left a couple small comments but basically looks good to me 🥳 

will let @grutt take a look through since he has some more context here :) great work!

sdks/typescript/src/v1/client/worker/worker-internal.ts
jishnundth: Thanks for the catch

sdks/typescript/src/opentelemetry/instrumentor.ts
jishnundth: I thought without `as`, it would trigger typing errors. Seems to be fine without it.

mrkaye97:
btw just confirming - you've tested propagating the span context 1) from trigger -> run (so the trigger workflow run and the run workflow are part of the same trace, with the run nested as a span under the trigger) and 2) from parent -> child (when a task spawns a child the child should be nested under the parent)?

jishnundth:
> btw just confirming - you've tested propagating the span context 1) from trigger -> run (so the trigger workflow run and the run workflow are part of the same trace, with the run nested as a span under the trigger) and 2) from parent -> child (when a task spawns a child the child should be nested under the parent)?

Thanks for the message. 

Case 1 is properly tested. ->  run is propagating correctly and showing proper nesting in the trace.
I missed out on Case 2. Will test and confirm shortly

jishnundth:
@mrkaye97 Both cases work seemlessly. Please check the screenshots below

## Run Workflow

This is the example in the PR. Where the workflow contains 4 tasks. 2 autoinstrumented ones (no separate spans), and 2 tasks having their own spans in the task.

workflow run


## Parent Child Workflow

A parent task spawing 3 child tasks. _parent-orchestartion_ and _spawn-child-int_ are spans inside the parent task. 
_child-compute_ is a span inside the child task. 

child_parent

Let me know, if there's issues.

Also when you're testing please remember the note I shared above: https://github.com/hatchet-dev/hatchet/pull/2828#issuecomment-3790537967

mrkaye97:
> @mrkaye97 Both cases work seemlessly. Please check the screenshots below
> 
> ## Run Workflow
> This is the example in the PR. Where the workflow contains 4 tasks. 2 autoinstrumented ones (no separate spans), and 2 tasks having their own spans in the task.
> 
> workflow run
> ## Parent Child Workflow
> A parent task spawing 3 child tasks. _parent-orchestartion_ and _spawn-child-int_ are spans inside the parent task. _child-compute_ is a span inside the child task.
> 
> child_parent
> Let me know, if there's issues.
> 
> Also when you're testing please remember the note I shared above: [#2828 (comment)](https://github.com/hatchet-dev/hatchet/pull/2828#issuecomment-3790537967)

Amazing!

mnafees:
Hey @jishnundth, sorry this is taking so long. We just merged some major changes in `main` so could you please rebase with the latest changes and I will make sure this gets shipped this week itself. Thanks so much.

jishnundth:
> Hey @jishnundth, sorry this is taking so long. We just merged some major changes in `main` so could you please rebase with the latest changes and I will make sure this gets shipped this week itself. Thanks so much.

No worries. I'm working on fixing things with the latest changes. Will keep you posted

jishnundth:
@mnafees please check now

sdks/typescript/src/opentelemetry/instrumentor.ts
```diff
@@ -0,0 +1,695 @@
...
+  InstrumentationNodeModuleDefinition,
+  InstrumentationNodeModuleFile,
+  isWrapped,
+} = otelInstrumentation;
+
+type HatchetInstrumentationConfig = OpenTelemetryConfig & InstrumentationConfig;
+type Carrier = Record<string, string>;
+
+const INSTRUMENTOR_NAME = '@hatchet-dev/typescript-sdk';
+const SUPPORTED_VERSIONS = ['>=1.16.0'];
```
grutt, line 59: can we add a FIXME to refactor this version check with the new pattern we're using in #2954

sdks/typescript/src/opentelemetry/instrumentor.ts
```diff
@@ -0,0 +1,695 @@
...
+
+function extractContext(carrier: Carrier | undefined | null): OtelContext {
+  return propagation.extract(context.active(), carrier ?? {});
+}
+
+function injectContext(carrier: Carrier): void {
+  propagation.inject(context.active(), carrier);
+}
+
+function parseAdditionalMetadata(action: Action): Record<string, any> | undefined {
```
grutt: i think we have a helper for this somewhere...

grutt: any will fail the lint check now, sorry

sdks/typescript/src/opentelemetry/instrumentor.ts
```diff
@@ -0,0 +1,695 @@
...
+    return undefined;
+  }
+}
+
+function getActionOtelAttributes(
+  action: Action,
+  excludedAttributes: string[] = [],
+  workerId?: string
+): Attributes {
+  const attributes: Attributes = {
```
grutt: can we have an exhaustiveness check on this map, if we add an attribute?

sdks/typescript/src/opentelemetry/instrumentor.ts
```diff
@@ -0,0 +1,695 @@
...
+    super(INSTRUMENTOR_NAME, HATCHET_VERSION, { ...DEFAULT_CONFIG, ...config });
+  }
+
+  override setConfig(config: Partial<HatchetInstrumentationConfig> = {}): void {
+    super.setConfig({ ...DEFAULT_CONFIG, ...config });
+  }
+
+  protected init(): InstanceType<typeof InstrumentationNodeModuleDefinition>[] {
+    const eventClientModuleFile = new InstrumentationNodeModuleFile(
+      '@hatchet-dev/typescript-sdk/clients/event/event-client.js',
```
grutt, line 149: does this break if we move this file?

sdks/typescript/src/v1/client/worker/worker-internal.ts
```diff
@@ -563,14 +564,15 @@ export class InternalWorker {
...
+      const future = new HatchetPromise<Error | undefined>(
         (async () => {
           let result: any;
           try {
             result = await run();
           } catch (e: any) {
             await failure(e);
-            return;
+            // Return error for OTel instrumentor to capture
+            return e;
```
grutt, line 576: so we're returning an error instead of re raising?

sdks/typescript/src/v1/client/worker/worker-internal.ts
```diff
@@ -603,7 +606,7 @@ export class InternalWorker {
       });
 
       try {
-        await future.promise;
+        return await future.promise;
```
grutt, line 610: why were we NOT returning this before?

grutt (changes requested):
super close! a few questions and then we'll need to resolve conficts/lint

sdks/typescript/src/opentelemetry/instrumentor.ts
jishnundth: Yes. The OTel TS references also does the same thing.

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/instrumentation-amqplib/src/amqplib.ts#L107-L126

sdks/typescript/src/v1/client/worker/worker-internal.ts
jishnundth: Task errors are by design only reported and not re-raised. Because it'll likely cause the worker to crash, if it goes upstream. 

https://github.com/hatchet-dev/hatchet/blob/main/sdks/typescript/src/v1/client/worker/worker-internal.ts#L529-L563
On failure it's reported, and task status is set as failed with logs. 

Task errors are designed to be discarded, and not flow upstream.
https://github.com/hatchet-dev/hatchet/blob/main/sdks/typescript/src/v1/client/worker/worker-internal.ts#L916

It's only returned here to capture the error on OTel as well. 
https://github.com/hatchet-dev/hatchet/pull/2828/changes#diff-ebc4ede2fd3bb6c564192fa588ac97735b916be51e599154966fbdffc2a1d313R540-R563

sdks/typescript/src/v1/client/worker/worker-internal.ts
jishnundth: The value is not used anywhere. Task runs are fire and forget only.
https://github.com/hatchet-dev/hatchet/blob/main/sdks/typescript/src/v1/client/worker/worker-internal.ts#L929-L943

With the return here, I can use it on OTel function wrapper, to validate the task execution state.

jishnundth:
PTAL, the requested changes are added

sdks/typescript/src/opentelemetry/instrumentor.ts
grutt: gotcha on python we have a comment on instrumented methods to keep them in sync. can we do something similar here?

https://github.com/hatchet-dev/hatchet/pull/3214/changes#diff-b83613a59c1e9c603402fd2679885b5a4cf05c7513aa98b0350e2f5b3dd212b7

grutt (changes requested):
one last comment to add comments on core sdk functions to prevent future breakage

jishnundth:
> one last comment to add comments on core sdk functions to prevent future breakage

Done!

mnafees (approved):

promptless-for-oss:

Updated existing suggestion: **[Document OpenTelemetry support for Python, TypeScript, and Go SDKs](https://app.gopromptless.ai/suggestions/79f91512-3bd0-45b0-a570-f6c7f0d8a4e9)**

---
_Tip: Configure how Promptless handles changelogs in [Agent Settings](https://app.gopromptless.ai/configure/settings) 📋_

PR #2831, opened 2026-01-20
feat: try to use ghostty-web for rendering log lines instead of ansiToHtml

abelanger5:
# Description

Switches log line renderer to properly render ANSI control characters using `ghostty-web` instead of ansiToHtml. 

Still to do:
- [x] Light mode
- [x] Fallback when `init` fails (due to no WASM support) - we can just dump to a div

## Type of change


frontend/app/src/components/v1/cloud/logging/logs.tsx
abelanger5, line 1: think we can remove this file but @mrkaye97 want to double check?

frontend/app/src/components/v1/cloud/logging/components/Terminal.tsx
```diff
@@ -0,0 +1,75 @@
...
+    () => ({
+      autoScroll,
+      callbacks,
+      theme,
+      onInitError: () => setInitFailed(true),
+    }),
+    [autoScroll, callbacks, theme],
+  );
+
+  useTerminal(containerRef, logs, options);
```
mrkaye97, line 34: are we calling this for side effects?

frontend/app/src/components/v1/cloud/logging/hooks/useTerminal.ts
```diff
@@ -0,0 +1,297 @@
+// @ts-ignore - ghostty-web doesn't have TypeScript declarations
```
mrkaye97: 🤮

frontend/app/src/components/v1/cloud/logging/hooks/useTerminal.ts
```diff
@@ -0,0 +1,297 @@
...
+  options?: {
+    autoScroll?: boolean;
+    callbacks?: TerminalScrollCallbacks;
+    theme?: TerminalTheme;
+    onInitError?: () => void;
+  },
+) {
+  const [terminalInitialized, setTerminalInitialized] = useState(false);
+  const terminalRef = useRef<any>(null);
+  const fitAddonRef = useRef<any>(null);
```
mrkaye97: would be great to not need to use `any` here if possible

frontend/app/src/components/v1/cloud/logging/hooks/useTerminal.ts
```diff
@@ -0,0 +1,297 @@
...
+  }
+  wasmInitPromise = init().then(() => {
+    wasmInitialized = true;
+  });
+  return wasmInitPromise;
+}
+
+export function useTerminal(
+  containerRef: RefObject<HTMLDivElement>,
+  logs: string,
```
mrkaye97, line 34: why do we pass this as a string vs a `string[]`?

frontend/app/src/components/v1/cloud/logging/components/Terminal.tsx
```diff
@@ -0,0 +1,73 @@
...
+}
+
+function Terminal({ logs, autoScroll, callbacks, className }: TerminalProps) {
+  const containerRef = useRef<HTMLDivElement>(null);
+  const [initFailed, setInitFailed] = useState(false);
+  const { theme: themeMode } = useTheme();
+
+  // Determine if dark mode
+  const isDark =
+    themeMode === 'dark' ||
```
abelanger5: should probably just leave the casing as this

frontend/app/src/components/v1/cloud/logging/log-viewer.tsx
```diff
@@ -32,82 +80,46 @@ export function LogViewer({
...
+        ? logs
+        : [
+            {
+              line: emptyMessage,
+              timestamp: new Date().toISOString(),
+              instance: 'Hatchet',
+            },
+          ];
+
+    // Sort by timestamp
```
mrkaye97: I dislike comments like this :P

frontend/app/src/components/v1/cloud/logging/logs.tsx
mrkaye97: fine with me if it's fine with the type checker

frontend/app/src/components/v1/cloud/logging/hooks/useTerminal.ts
```diff
@@ -0,0 +1,302 @@
...
+    return;
+  }
+  if (wasmInitPromise) {
+    return wasmInitPromise;
+  }
+  wasmInitPromise = init().then(() => {
+    wasmInitialized = true;
+  });
+  return wasmInitPromise;
+}
```
abelanger5, line 30: In hindsight I don't think that this is necessary, init might be safe to call multiple times. The more important thing is that we can only mount a single terminal instance at a time because there's some kind of shared memory magic going on with this wasm init.

mrkaye97 (commented):
left a couple comments but broadly looks fine, might be a pain to merge these together 😬

frontend/app/src/components/v1/cloud/logging/hooks/useTerminal.ts
```diff
@@ -0,0 +1,302 @@
...
+
+export function useTerminal(
+  containerRef: RefObject<HTMLDivElement>,
+  logs: string,
```
abelanger5, line 34: I would like to change this to an array of string, but I'm keeping it aligned with the previous impl for now

frontend/app/src/components/v1/cloud/logging/hooks/useTerminal.ts
```diff
@@ -0,0 +1,302 @@
+import { TerminalTheme } from '../terminalThemes';
+import { init, Terminal, FitAddon } from 'ghostty-web';
+import { useEffect, useRef, RefObject, useState } from 'react';
+
+export interface TerminalScrollCallbacks {
+  onTopReached?: () => void;
+  onBottomReached?: () => void;
```
abelanger5, line 7: I guess the idea is we can remove these at some point? @mrkaye97

frontend/app/src/components/v1/cloud/logging/hooks/useTerminal.ts
```diff
@@ -0,0 +1,302 @@
...
+        callbacksRef.current?.onInfiniteScroll &&
+        now - lastInfiniteScrollCallRef.current >= 100
+      ) {
+        callbacksRef.current.onInfiniteScroll({
+          scrollTop,
+          scrollHeight,
+          clientHeight,
+        });
+        lastInfiniteScrollCallRef.current = now;
+      }
```
abelanger5, line 102: Need to double check this: I did a bunch of testing on onBottomReached and onTopReached but none on `onInfiniteScroll`.

frontend/app/src/components/v1/cloud/logging/hooks/useTerminal.ts
mrkaye97: I don't think so? because this is for loading more rows as you scroll up or down

frontend/app/src/components/v1/cloud/logging/hooks/useTerminal.ts
abelanger5: this is pure hallucination, it does have TS declarations. I removed this line

mrkaye97: lmao, jesus 😂 

thanks!

frontend/app/src/components/v1/cloud/logging/hooks/useTerminal.ts
mrkaye97: if we really can only call it once, I think it probably should go in a provider at the top level

frontend/app/src/components/v1/cloud/logging/components/Terminal.tsx
abelanger5: you mean why is it a hook instead of just building a mega component? we used to have some callbacks returned but no longer.

frontend/app/src/components/v1/cloud/logging/hooks/useTerminal.ts
abelanger5: yeah will take a look

frontend/app/src/components/v1/cloud/logging/components/Terminal.tsx
mrkaye97: no I mean why are we calling this if it returns nothing 😅

mrkaye97: if we do need to call it for side effects, there must be a more correct way than this - like maybe inside of a `useMemo` or `useEffect` or something so it only is run on mount

abelanger5: well, the side effect is that it mounts the terminal on the parent component, but a child component doesn't make sense because it won't return anything. but it's complex enough that it needs multiple `useEffect` so a single `useMemo` or `useEffect` doesn't make sense to me - which leaves calling a hook that returns nothing as the last option?

mrkaye97: argh yeah, guess you're right - bummer. maybe just leave a comment for this?

frontend/app/src/components/v1/cloud/logging/hooks/useTerminal.ts
abelanger5: I thought that was the point of `onInfiniteScroll`? I thought that the caller would be able to case on loading more rows, but it sounds like I'm misunderstanding the purpose of that prop

mrkaye97: so admittedly I didn't implement this 😅  but I'm pretty sure that onInfiniteScroll might not fire if you're just sitting at the bottom of the scroll view or something - like in the case where you've scrolled all the way down, and then we want to try to fetch some more logs

mrkaye97: it also could very well not be needed though

frontend/app/src/components/v1/cloud/logging/hooks/useTerminal.ts
abelanger5: I think I'm going to leave it, because I'd like to avoid polluting other parts of the code before I'm set on this approach - but it's a good idea for down the line

mrkaye97: sounds good 👍

PR #2852, opened 2026-01-24
Fix: Invalid OTel exporter in python example, add schedule_workflow unwrap

jishnundth:
# Description

1. Fix: Invalid OTel exporter used for hyperdx grpc endpint (Python OTel example)
2. Add: Unwrap for schedule_workflow (Python OTel instrumentor)

## Type of change


mrkaye97 (approved):
thanks! mind adding a changelog entry and bumping the patch version?

jishnundth:
> thanks! mind adding a changelog entry and bumping the patch version?

There's no changelog for TS SDK. Should I start one

mrkaye97:
> > thanks! mind adding a changelog entry and bumping the patch version?
> 
> There's no changelog for TS SDK. Should I start one

sure, thanks!

sdks/typescript/src/version.ts
```diff
@@ -1 +1 @@
-export const HATCHET_VERSION = '1.9.5';
+export const HATCHET_VERSION = '1.9.6';
```
mrkaye97: I think this needs to change in the `pyproject.toml`, not here? sorry for the cofnusion!

jishnundth: My mistake. I was mostly working on the OTel for TS SDK last week. My mind is still there😑

mrkaye97: no worries, figured that :) thanks for the fix!

PR #2855, opened 2026-01-26
bugfix: allow python healthcheck server to bind on ipv6 hosts

AlfieRan:
# Description

Added a configurable "bind_address" environment variable to the python healthcheck route to allow setting to None, which allows the healthcheck web server to bind to all available hosts (ipv4 & ipv6). To do this, set `HATCHET_CLIENT_WORKER_HEALTHCHECK_BIND_ADDRESS = "None"`, `0.0.0.0` for just ipv4, or `::` for just ipv6.

Intention is to support AAAA dns lookups on Railway to healthcheck each instance of a worker deployment. See below for an example implementation:

```py
# Your railway-deployed API service 

async def get_worker_ips(hostname: str) -> list[str]:
    loop = asyncio.get_event_loop()
    results = await loop.getaddrinfo(
        hostname, None, family=socket.AF_INET6, type=socket.SOCK_STREAM
    )
    return [r[4][0]for r in results]
    
# Then inside your healthcheck:
# 1. collect the ips with the internal railway hostname=xxx.railway.internal
# 2. GET each ip via f"http://[{ip}]:{port}/health"
```

## Type of change

## What's Changed

- Added configurable host to python healthcheck endpoint 

## Footnote

You may wish to simplify this to just have a "ip mode" config? Or default to previous behaviour but allow setting all hosts? I've just validated this works on our setup via Railway but may obviously have side-effects on other services that I'm not aware of.

### DIFF
--- a/sdks/python/.gitignore
+++ b/sdks/python/.gitignore
@@ -163,6 +163,3 @@ cython_debug/
 
 openapitools.json
 .python-version
-
-# Dev-specific poetry overrides - e.g. local venvs (use pyproject.toml for project-level config)
-poetry.toml
--- a/sdks/python/.gitignore
+++ b/sdks/python/.gitignore
@@ -163,3 +163,6 @@ cython_debug/
 
 openapitools.json
 .python-version
+
+# Dev-specific poetry overrides - e.g. local venvs (use pyproject.toml for project-level config)
+poetry.toml
--- a/sdks/python/hatchet_sdk/config.py
+++ b/sdks/python/hatchet_sdk/config.py
@@ -42,6 +42,8 @@ class HealthcheckConfig(BaseSettings):
         default=timedelta(seconds=5),
         description="If the worker listener process event loop appears blocked longer than this threshold, /health returns 503. Value is interpreted as seconds.",
     )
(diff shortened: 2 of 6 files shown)


sdks/python/.gitignore
grutt, line 1: probably want to revert this

grutt (dismissed):

sdks/python/hatchet_sdk/config.py
```diff
@@ -42,6 +42,8 @@ class HealthcheckConfig(BaseSettings):
         default=timedelta(seconds=5),
         description="If the worker listener process event loop appears blocked longer than this threshold, /health returns 503. Value is interpreted as seconds.",
     )
+    # HATCHET_CLIENT_WORKER_HEALTHCHECK_BIND_ADDRESS
+    bind_address: str | None = None
```
mrkaye97: I think this needs to default to the previous hardcoded value otherwise it’s a breaking change / won’t work afaict

mrkaye97 (approved):

promptless-for-oss:

New suggestion: **[Document custom bind address configuration for Python worker healthchecks](https://app.gopromptless.ai/suggestions/7512a989-198c-4767-b19d-b7e8d1aed702)**

---

PR #2861, opened 2026-01-27
feat: durable user event log

abelanger5:
# Description

## Config changes

```sh
SERVER_ENABLE_DURABLE_USER_EVENT_LOG # OFF by default
SERVER_OPTIMISTIC_SCHEDULING_ENABLED # ON by default
SERVER_OPTIMISTIC_SCHEDULING_SLOTS # 5 by default
SERVER_GRPC_TRIGGER_WRITES_ENABLED # ON by default
SERVER_GRPC_TRIGGER_WRITE_SLOTS # 5 by default
```

## Durable event log

Adds a new table to the core database called `v1_event` which gets stored as part of the transaction which writes the triggered workflows. This will be used to make the events API more consistent for single-event getters.

## Optimistic Scheduling

(copied from #2867)

New implementation to replace #2258  

Features:
- TENTATIVE: Made some improvements to listening for a task completed event for a single-task workflow by hooking into an existing tenant message `task-completed`. We can similarly add `task-failed` and `task-cancelled` in the future. 

Drawbacks:
- Increases the complexity of scheduling as the paths for optimistic scheduling are quite different from the regular path, since we protect everything with a single transaction. Tried to minimize the amount of complexity here and made a bunch of improvements over #2258 to avoid maintenance hell. 

Limitations:
- This won't be turned on in HA mode and as `n` engines are horizontally scales the chances of optimistic scheduling reduce by `1/n` - we only use local schedulers when they have a lease on a tenant. We will need to build out a sticky load balancing strategy to take advantage of optimistic scheduling in HA setups if we'd like to tackle that. 

## Type of change


.github/workflows/test.yml
```diff
@@ -339,6 +340,7 @@ jobs:
           TESTING_MATRIX_MIGRATE: ${{ matrix.migrate-strategy }}
           TESTING_MATRIX_RABBITMQ_ENABLED: ${{ matrix.rabbitmq-enabled }}
           TESTING_MATRIX_PG_VERSION: ${{ matrix.pg-version }}
+          TESTING_MATRIX_OPTIMISTIC_SCHEDULING: ${{ matrix.optimistic-scheduling }}
```
mrkaye97, line 343: do you think we should also run the python tests w + w/o optimistic scheduling ?

mrkaye97 (approved):

promptless-for-oss:

New suggestion: **[Document optimistic scheduling and durable event log configuration options](https://app.gopromptless.ai/suggestions/f2b9eb6e-aab2-4e62-98d3-3a2153401cad)**

---

PR #2862, opened 2026-01-27
feat: db tables for user events

abelanger5:
# Description

Adds initial tables for storing user-sent events. Can turn on/off using the feature flag `SERVER_ENABLE_DURABLE_USER_EVENT_LOG`.

## Type of change

### DIFF
--- a/cmd/hatchet-migrate/migrate/migrations/20260127145041_v1_0_71.sql
+++ b/cmd/hatchet-migrate/migrate/migrations/20260127145041_v1_0_71.sql
@@ -23,7 +23,7 @@ CREATE TABLE IF NOT EXISTS v1_event_lookup_table (
     event_id BIGINT NOT NULL,
     event_seen_at TIMESTAMPTZ NOT NULL,
 
-    PRIMARY KEY (tenant_id, external_id, event_seen_at)
+    PRIMARY KEY (external_id, event_seen_at)
 ) PARTITION BY RANGE(event_seen_at);
 
 CREATE OR REPLACE FUNCTION v1_event_lookup_table_insert_function()
@@ -42,7 +42,7 @@ BEGIN
         id,
         seen_at
     FROM new_rows
-    ON CONFLICT (tenant_id, external_id, event_seen_at) DO NOTHING;
+    ON CONFLICT (external_id, event_seen_at) DO NOTHING;
 
     RETURN NULL;
 END;
@@ -68,7 +68,7 @@ CREATE TABLE IF NOT EXISTS v1_event_to_run (
 
 SELECT create_v1_range_partition('v1_event', DATE 'today');
 SELECT create_v1_weekly_range_partition('v1_event_lookup_table', DATE 'today');
-SELECT create_v1_weekly_range_partition('v1_event_to_run', DATE 'today');
+SELECT create_v1_range_partition('v1_event_to_run', DATE 'today');
 -- +goose StatementEnd
 
 -- +goose Down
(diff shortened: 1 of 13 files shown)


cmd/hatchet-migrate/migrate/migrations/20260127145041_v1_0_71.sql
```diff
@@ -0,0 +1,82 @@
...
+
+CREATE INDEX IF NOT EXISTS v1_event_key_idx ON v1_event (tenant_id, key);
+
+CREATE TABLE IF NOT EXISTS v1_event_lookup_table (
+    tenant_id UUID NOT NULL,
+    external_id UUID NOT NULL,
+    event_id BIGINT NOT NULL,
+    event_seen_at TIMESTAMPTZ NOT NULL,
+
+    PRIMARY KEY (tenant_id, external_id, event_seen_at)
```
mrkaye97: in hindsight this was a mistake (having the tenant id in the pk here), so I think maybe we should not make that same mistake twice 😅

abelanger5: why was this a mistake?

mrkaye97: because we need the tenant id to do the lookups vs. being able to populate from only an external id (the old API didn't include the tenant id in the path)

pkg/repository/sqlcv1/tasks-overwrite.go
```diff
@@ -839,3 +839,77 @@ func (q *Queries) ReleaseTasks(ctx context.Context, db DBTX, arg ReleaseTasksPar
...
+    SELECT
+        UNNEST($1::UUID[]) AS tenant_id,
+        UNNEST($2::UUID[]) AS external_id,
+        UNNEST($3::TIMESTAMPTZ[]) AS seen_at,
+        UNNEST($4::TEXT[]) AS key,
+        UNNEST($5::JSONB[]) AS additional_metadata,
+        -- Scopes are nullable
+        UNNEST($6::TEXT[]) AS scope,
+        -- Webhook names are nullable
+        UNNEST($7::TEXT[]) AS triggering_webhook_name
```
mrkaye97, line 854: nit, but I've kinda been liking having a `NULLIF()` for these, passing an empty string instead of a null, and then being able to write these as SQL instead of doing the overwrite thing

pkg/repository/sqlcv1/tasks.sql
```diff
@@ -43,6 +50,12 @@ WITH task_partitions AS (
     SELECT 'v1_log_line' AS parent_table, p::text as partition_name FROM get_v1_partitions_before_date('v1_log_line', @date::date) AS p
 ), payload_partitions AS (
     SELECT 'v1_payload' AS parent_table, p::text as partition_name FROM get_v1_partitions_before_date('v1_payload', @date::date) AS p
+), event_partitions AS (
+    SELECT 'v1_event' AS parent_table, p::text as partition_name FROM get_v1_partitions_before_date('v1_event', @date::date) AS p
+), event_lookup_table_partitions AS (
+    SELECT 'v1_event_lookup_table' AS parent_table, p::text as partition_name FROM get_v1_partitions_before_date('v1_event_lookup_table', @date::date) AS p
+), event_to_run_partitions AS (
+    SELECT 'v1_event_to_run' AS parent_table, p::text as partition_name FROM get_v1_partitions_before_date('v1_event_to_run', @date::date) AS p
```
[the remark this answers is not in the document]
mrkaye97: I think this one might be right - there's some poorly named stuff in here for working with partitions but we need to be careful to not e.g. drop the current week's partition when it's >N days old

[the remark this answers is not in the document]
abelanger5: yeah it is, twas a good catch

pkg/repository/sqlcv1/tasks-overwrite.go
abelanger5: going to keep it consistent for now, I think this might be a dedicated PR in the future

mrkaye97 (approved):

PR #2864, opened 2026-01-27
Add doc about autoscaling workers

mnafees:
# Description

Adds documentation about autoscaling workers using the task stats API endpoint.

## Type of change

### DIFF
--- a/frontend/docs/pages/home/autoscaling-workers.mdx
+++ b/frontend/docs/pages/home/autoscaling-workers.mdx
@@ -1,0 +1,149 @@
+import { Callout } from "nextra/components";
+
+# Autoscaling Workers
+
+Hatchet provides a Task Stats API that enables you to implement autoscaling for your worker pools. By querying real-time queue depths and task distribution, you can dynamically scale workers based on actual workload demand.
+
+## Task Stats API
+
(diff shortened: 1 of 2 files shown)


frontend/docs/pages/home/autoscaling-workers.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+Each task stat includes:
+
+- **total**: The total count of tasks in this state
+- **concurrency**: Distribution across concurrency groups (if concurrency limits are configured)
+
+These are available only for `queued` tasks:
+
+- **queues**: A breakdown of task counts by queue name
+- **oldest**: Timestamp of the oldest task currently enqueued (useful for detecting queue backlogs)
```
mrkaye97: I think we should have `oldest` for `running` too, but would be good to confirm that

mrkaye97 (approved):
one small comment but lgtm, thanks for putting this together!

abelanger5 (approved):
nice! the obvious question that also comes to mind is: what about scaling on worker utilization? but I think we don't have a great answer for this at the moment, so we can definitely get this merged in and tackle that later.

PR #2867, opened 2026-01-27
feat: optimistic scheduling

abelanger5:
# Description

New implementation to replace #2258  

Features:
- TENTATIVE: Made some improvements to listening for a task completed event for a single-task workflow by hooking into an existing tenant message `task-completed`. We can similarly add `task-failed` and `task-cancelled` in the future. 

Drawbacks:
- Increases the complexity of scheduling as the paths for optimistic scheduling are quite different from the regular path, since we protect everything with a single transaction. Tried to minimize the amount of complexity here and made a bunch of improvements over #2258 to avoid maintenance hell. 

Limitations:
- This won't be turned on in HA mode and as `n` engines are horizontally scales the chances of optimistic scheduling reduce by `1/n` - we only use local schedulers when they have a lease on a tenant. We will need to build out a sticky load balancing strategy to take advantage of optimistic scheduling in HA setups if we'd like to tackle that. 

## Type of change

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -8,48 +8,48 @@ on:
         required: true
 name: Release
 jobs:
-  # load:
-  #   runs-on: ubuntu-latest
-  #   timeout-minutes: 30
-  #   strategy:
-  #     matrix:
-  #       migrate-strategy: ["latest"]
-  #       rabbitmq-enabled: ["true", "false"]
-  #       pg-version: ["15-alpine", "16-alpine", "17-alpine"]
+  load:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    strategy:
+      matrix:
+        migrate-strategy: ["latest"]
+        rabbitmq-enabled: ["true", "false"]
+        pg-version: ["15-alpine", "16-alpine", "17-alpine"]
+        optimistic-scheduling: ["true", "false"]
 
-  #   steps:
-  #     - uses: actions/checkout@v6
+    steps:
+      - uses: actions/checkout@v6
 
-  #     - name: Install Task
-  #       uses: arduino/setup-task@v2
-  #       with:
-  #         repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install Task
+        uses: arduino/setup-task@v2
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
 
-  #     - name: Setup Go
-  #       uses: actions/setup-go@v6
-  #       with:
-  #         go-version: "1.25"
+      - name: Setup Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: "1.25"
 
-  #     - name: Setup pnpm
-  #       uses: pnpm/action-setup@v4
-  #       with:
-  #         version: 10.16.1
-  #         run_install: false
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: 10.16.1
+          run_install: false
 
-  #     - name: Go deps
-  #       run: go mod download
+      - name: Go deps
+        run: go mod download
 
-  #     - name: Test
-  #       run: |
-  #         go test -tags load ./... -p 5 -v -race -failfast -timeout 20m
-  #       env:
-  #         TESTING_MATRIX_MIGRATE: ${{ matrix.migrate-strategy }}
-  #         TESTING_MATRIX_RABBITMQ_ENABLED: ${{ matrix.rabbitmq-enabled }}
-  #         TESTING_MATRIX_PG_VERSION: ${{ matrix.pg-version }}
+      - name: Test
+        run: |
+          go test -tags load ./... -p 5 -v -race -failfast -timeout 20m
+        env:
+          TESTING_MATRIX_MIGRATE: ${{ matrix.migrate-strategy }}
+          TESTING_MATRIX_RABBITMQ_ENABLED: ${{ matrix.rabbitmq-enabled }}
+          TESTING_MATRIX_PG_VERSION: ${{ matrix.pg-version }}
   push-hatchet-server:
     name: Push latest
     runs-on: ubuntu-latest
-    # needs: load
     steps:
       - name: Get tag name
         id: tag_name
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -46,6 +46,8 @@ linters:
         text: "context-keys-type: should not use basic type string"
       - path: (.+)\.go$
         text: error strings should not be capitalized
+      - path: "internal/services/dispatcher/(.+).go"
+        text: "ST1016:"
       - linters:
           - staticcheck
         text: "ST1003:"
--- a/cmd/hatchet-migrate/migrate/migrations/20260127201500_v1_0_72.sql
+++ b/cmd/hatchet-migrate/migrate/migrations/20260127201500_v1_0_72.sql
@@ -1,0 +1,299 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION v1_task_insert_function()
+RETURNS TRIGGER AS $$
+DECLARE
+    rec RECORD;
+BEGIN
+    -- Only insert if there's a single task with initial_state = 'QUEUED' and concurrency_strategy_ids is not null
+    IF (SELECT COUNT(*) FROM new_table WHERE initial_state = 'QUEUED' AND concurrency_strategy_ids[1] IS NOT NULL) > 0 THEN
+        WITH new_slot_rows AS (
+            SELECT
+                id,
+                inserted_at,
+                retry_count,
+                tenant_id,
+                priority,
+                concurrency_parent_strategy_ids[1] AS parent_strategy_id,
+                CASE
+                    WHEN array_length(concurrency_parent_strategy_ids, 1) > 1 THEN concurrency_parent_strategy_ids[2:array_length(concurrency_parent_strategy_ids, 1)]
+                    ELSE '{}'::bigint[]
+                END AS next_parent_strategy_ids,
+                concurrency_strategy_ids[1] AS strategy_id,
+                external_id,
+                workflow_run_id,
+                CASE
+                    WHEN array_length(concurrency_strategy_ids, 1) > 1 THEN concurrency_strategy_ids[2:array_length(concurrency_strategy_ids, 1)]
+                    ELSE '{}'::bigint[]
+                END AS next_strategy_ids,
+                concurrency_keys[1] AS key,
+                CASE
+                    WHEN array_length(concurrency_keys, 1) > 1 THEN concurrency_keys[2:array_length(concurrency_keys, 1)]
+                    ELSE '{}'::text[]
+                END AS next_keys,
+                workflow_id,
+                workflow_version_id,
+                queue,
+                CURRENT_TIMESTAMP + convert_duration_to_interval(schedule_timeout) AS schedule_timeout_at
+            FROM new_table
+            WHERE initial_state = 'QUEUED' AND concurrency_strategy_ids[1] IS NOT NULL
+        )
+        INSERT INTO v1_concurrency_slot (
+            task_id,
+            task_inserted_at,
+            task_retry_count,
+            external_id,
+            tenant_id,
+            workflow_id,
+            workflow_version_id,
+            workflow_run_id,
+            parent_strategy_id,
+            next_parent_strategy_ids,
+            strategy_id,
+            next_strategy_ids,
+            priority,
+            key,
+            next_keys,
+            queue_to_notify,
+            schedule_timeout_at
+        )
+        SELECT
+            id,
+            inserted_at,
+            retry_count,
+            external_id,
+            tenant_id,
+            workflow_id,
+            workflow_version_id,
+            workflow_run_id,
+            parent_strategy_id,
+            next_parent_strategy_ids,
+            strategy_id,
+            next_strategy_ids,
+            COALESCE(priority, 1),
+            key,
+            next_keys,
+            queue,
+            schedule_timeout_at
+        FROM new_slot_rows;
+    END IF;
+
+    INSERT INTO v1_queue_item (
+        tenant_id,
+        queue,
+        task_id,
+        task_inserted_at,
+        external_id,
+        action_id,
+        step_id,
+        workflow_id,
+        workflow_run_id,
+        schedule_timeout_at,
+        step_timeout,
+        priority,
+        sticky,
+        desired_worker_id,
+        retry_count
+    )
+    SELECT
+        tenant_id,
+        queue,
+        id,
+        inserted_at,
+        external_id,
+        action_id,
+        step_id,
+        workflow_id,
+        workflow_run_id,
+        CURRENT_TIMESTAMP + convert_duration_to_interval(schedule_timeout),
+        step_timeout,
+        COALESCE(priority, 1),
+        sticky,
+        desired_worker_id,
+        retry_count
+    FROM new_table
+    WHERE initial_state = 'QUEUED' AND concurrency_strategy_ids[1] IS NULL;
+
+    -- Only insert into v1_dag and v1_dag_to_task if dag_id and dag_inserted_at are not null
+    IF (SELECT COUNT(*) FROM new_table WHERE dag_id IS NOT NULL AND dag_inserted_at IS NOT NULL) > 0 THEN
+        INSERT INTO v1_dag_to_task (
+            dag_id,
+            dag_inserted_at,
+            task_id,
+            task_inserted_at
+        )
+        SELECT
+            dag_id,
+            dag_inserted_at,
+            id,
+            inserted_at
+        FROM new_table
+        WHERE dag_id IS NOT NULL AND dag_inserted_at IS NOT NULL;
+    END IF;
+
+    INSERT INTO v1_lookup_table (
+        external_id,
+        tenant_id,
+        task_id,
+        inserted_at
+    )
+    SELECT
+        external_id,
+        tenant_id,
+        id,
+        inserted_at
+    FROM new_table
+    ON CONFLICT (external_id) DO NOTHING;
+
+    RETURN NULL;
+END;
+$$
+LANGUAGE plpgsql;
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION v1_task_insert_function()
+RETURNS TRIGGER AS $$
+DECLARE
+    rec RECORD;
+BEGIN
+    WITH new_slot_rows AS (
+        SELECT
+            id,
+            inserted_at,
+            retry_count,
+            tenant_id,
+            priority,
+            concurrency_parent_strategy_ids[1] AS parent_strategy_id,
+            CASE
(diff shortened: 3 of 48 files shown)


internal/services/controllers/olap/signal/signal.go
abelanger5, line 1: I had to consolidate signaling logic between the local scheduler (which creates tasks) and the tasks controller, and this was the cleanest way to do it.

cmd/hatchet-migrate/migrate/migrations/20260127201500_v1_0_72.sql
```diff
@@ -0,0 +1,299 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION v1_task_insert_function()
+RETURNS TRIGGER AS $$
+DECLARE
+    rec RECORD;
+BEGIN
+    -- Only insert if there's a single task with initial_state = 'QUEUED' and concurrency_strategy_ids is not null
+    IF (SELECT COUNT(*) FROM new_table WHERE initial_state = 'QUEUED' AND concurrency_strategy_ids[1] IS NOT NULL) > 0 THEN
```
abelanger5, line 9: Just added a couple of `IF` statements to this insert, just had to copy the whole trigger over to the migration

internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/hatchet-dev/hatchet/pkg/repository/sqlchelpers"
 	"github.com/hatchet-dev/hatchet/pkg/repository/sqlcv1"
 	"github.com/hatchet-dev/hatchet/pkg/telemetry"
+
+	schedulingv1 "github.com/hatchet-dev/hatchet/pkg/scheduling/v1"
 )
 
 func (d *DispatcherImpl) handleTaskBulkAssignedTask(ctx context.Context, msg *msgqueue.Message) error {
```
abelanger5, line 25: This is a bit of a risk: I refactored this method significantly so we could reuse some helpers in `HandleLocalAssignments`. Since this is critical path I'm confident that this didn't break the expected path but there's some risk on edge cases - would appreciate a close review here!

internal/services/scheduler/v1/scheduler.go
```diff
@@ -159,6 +162,9 @@ func New(
 	// TODO: replace with config or pull into a constant
 	tasksWithNoWorkerCache := expirable.NewLRU(10000, func(string, struct{}) {}, 5*time.Minute)
 
+	// TODO: fix propagation of replayEnabled, it's hardcoded to true here
+	signaler := signal.NewOLAPSignaler(true, opts.mq, opts.repov1, opts.l, pubBuffer)
```
abelanger5: Fix TODO before merge

pkg/repository/scheduler_queue.go
```diff
@@ -569,20 +578,34 @@ func (d *queueRepository) GetTaskRateLimits(ctx context.Context, queueItems []*s
 	return taskIdToKeyToUnits, nil
 }
 
+// TODO: it seems like there's a risk of deadlocking if we have a tx open and we call d.pool to open a new connection
```
abelanger5: Definitely need to resolve this before merge, should probably write a new method on the `Optimistic` repository to pass through a tx. Can't have nested opening of txs as we've learned the hard way.

pkg/repository/shared.go
```diff
@@ -25,19 +26,28 @@ type taskExternalIdTenantIdTuple struct {
...
-	m                         TenantLimitRepository
+	pool    *pgxpool.Pool
+	v       validator.Validator
+	l       *zerolog.Logger
+	queries *sqlcv1.Queries
+
+	// TODO: modify these caches to use expirable.LRU as well
+	queueCache               *cache.Cache
+	stepExpressionCache      *cache.Cache
+	concurrencyStrategyCache *cache.Cache
```
abelanger5: Take a look here

pkg/repository/shared.go
```diff
@@ -25,19 +26,28 @@ type taskExternalIdTenantIdTuple struct {
...
+
+	// TODO: verify whether these three caches are actually necessary
```
abelanger5: Worth a second look

pkg/scheduling/v1/queuer.go
```diff
@@ -563,3 +570,193 @@ func (q *Queuer) flushToDatabase(ctx context.Context, r *assignResults) int {
...
+	for _, qi := range qis {
+		stepIds = append(stepIds, qi.StepID)
+	}
+
+	// TODO: use tx
+	labels, err := q.repo.GetDesiredLabels(ctx, stepIds)
+
+	if err != nil {
+		return nil, nil, err
+	}
```
abelanger5: address todos

internal/services/dispatcher/dispatcher_v1.go
abelanger5, line 1: The original PR (https://github.com/hatchet-dev/hatchet/pull/2258) had some changes in the dispatcher to listen for a `task-completed` event, should consider adding that here or in a separate PR.

.golangci.yml
```diff
@@ -46,6 +46,8 @@ linters:
         text: "context-keys-type: should not use basic type string"
       - path: (.+)\.go$
         text: error strings should not be capitalized
+      - path: "internal/services/dispatcher/(.+).go"
+        text: "ST1016:"
```
abelanger5, line 50: Trying to avoid a huge diff since the linter is picking up on this error - going to leave it for now, definitely not critical:

```
internal/services/dispatcher/dispatcher_v1.go:88:26: ST1016: methods on the same type should have the same receiver name (seen 15x "d", 25x "s") (staticcheck)
func (d *DispatcherImpl) GetLocalWorkerIds() map[string]struct{} {
```

pkg/repository/trigger.go
abelanger5, line 1: Probably easier to read this file rather than view the diff - I basically pulled out the super large methods `TriggerFromEvents` and `TriggerFromWorkflowNames` into a set of helpers which we could use in the optimistic queries.

pkg/repository/scheduler_optimistic.go
```diff
@@ -0,0 +1,130 @@
...
+	return &optimisticSchedulingRepositoryImpl{
+		sharedRepository: shared,
+	}
+}
+
+func (r *optimisticSchedulingRepositoryImpl) StartTx(ctx context.Context) (*OptimisticTx, error) {
+	return r.PrepareOptimisticTx(ctx)
+}
+
+func (r *optimisticSchedulingRepositoryImpl) TriggerFromEvents(ctx context.Context, tx *OptimisticTx, tenantId string, opts []EventTriggerOpts) ([]*sqlcv1.V1QueueItem, *TriggerFromEventsResult, error) {
```
abelanger5, line 29: This and `TriggerFromWorkflowNames` are significant improvements on #2258 , there's now minimal copy-pasted code so it should be straightforward to maintain

abelanger5 (commented):
Need to address some comments, additionally we should add the optimistic scheduling paths to the testing matrix.

internal/services/controllers/olap/signal/signal.go
mrkaye97: this is basically just copy paste from the existing code right?

internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -48,267 +50,358 @@ func (d *DispatcherImpl) handleTaskBulkAssignedTask(ctx context.Context, msg *ms
...
+		return true
+	})
+
+	return workerIds
+}
+
+// Note: this is very similar to handleTaskBulkAssignedTask, with some differences in what's sync vs run in a goroutine
+// In this method, we wait until all tasks have been sent to the worker before returning
+func (d *DispatcherImpl) HandleLocalAssignments(ctx context.Context, tenantId, workerId string, tasks []*schedulingv1.AssignedItemWithTask) error {
+	ctx, span := telemetry.NewSpan(ctx, "task-local-assignmnets")
```
mrkaye97: nit: span naming

internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -48,267 +50,358 @@ func (d *DispatcherImpl) handleTaskBulkAssignedTask(ctx context.Context, msg *ms
...
+	bulkDatas, err := d.repov1.Tasks().ListTasks(ctx, tenantId, taskIds)
+
+	if err != nil {
+		for _, task := range bulkDatas {
+			requeue(task)
 		}
 
-		inputs, err := d.repov1.Payloads().Retrieve(ctx, nil, retrievePayloadOpts...)
+		d.l.Error().Err(err).Msgf("could not bulk list step run data:")
+		return nil, err
```
mrkaye97, line 156: what happens if we error here? does the error bubble up to the dispatcher and then the task just get dropped?

internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -48,267 +50,358 @@ func (d *DispatcherImpl) handleTaskBulkAssignedTask(ctx context.Context, msg *ms
...
+}
 
-				if err != nil && !errors.Is(err, ErrWorkerNotFound) {
-					return fmt.Errorf("could not get worker: %w", err)
-				}
+func (d *DispatcherImpl) sendTasksToWorker(
+	ctx context.Context,
+	requeue func(task *sqlcv1.V1Task),
+	tenantId, workerId string,
+	taskIds []int64,
```
mrkaye97, line 284: nit: seems like we maybe don't need to pass this slice of task ids here? can we loop over `taskId, task := range tasks` instead?

internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -48,267 +50,358 @@ func (d *DispatcherImpl) handleTaskBulkAssignedTask(ctx context.Context, msg *ms
...
+	return innerEg.Wait()
+}
 
-					if err != nil {
-						return fmt.Errorf("could not create failed task message: %w", err)
-					}
+func (d *DispatcherImpl) handleRetries(
+	ctx context.Context,
+	tenantId string,
+	toRetry []*sqlcv1.V1Task,
```
mrkaye97, line 366: maybe this should also be a `[]*taskWithPayload` for consistency? when I initially did this, I was mostly trying to remove the `sqlcv1.V1Task` from everywhere I could to make it less confusing what was what

internal/services/ingestor/ingestor_v1.go
```diff
@@ -49,44 +53,133 @@ func (i *IngestorImpl) ingestEventV1(ctx context.Context, tenant *sqlcv1.Tenant,
...
+				Priority:              event.EventPriority,
+				Scope:                 event.EventScope,
+				TriggeringWebhookName: event.TriggeringWebhookName,
+			})
+		}
+
+		localAssigned, schedulingErr := i.localScheduler.RunOptimisticSchedulingFromEvents(ctx, tenantId, opts, localWorkerIds)
+
+		if schedulingErr != nil {
+			if !errors.Is(schedulingErr, schedulingv1.ErrTenantNotFound) && !errors.Is(schedulingErr, schedulingv1.ErrNoOptimisticSlots) {
```
mrkaye97, line 118: comment here indicating we intentionally fall through to regular scheduling if this fails (assuming I'm reading this right)

internal/services/scheduler/v1/scheduler.go
```diff
@@ -409,7 +416,9 @@ func (s *Scheduler) scheduleStepRuns(ctx context.Context, tenantId string, res *
 				dispatcherIdToWorkerIdsToStepRuns[dispatcherId][workerId] = make([]int64, 0)
 			}
 
-			dispatcherIdToWorkerIdsToStepRuns[dispatcherId][workerId] = append(dispatcherIdToWorkerIdsToStepRuns[dispatcherId][workerId], bulkAssigned.QueueItem.TaskID)
+			if !bulkAssigned.IsAssignedLocally {
```
mrkaye97, line 418: how much risk is there here that we forget to check this flag somewhere and end up with duplicate assigns?

pkg/repository/idempotency.go
```diff
@@ -54,7 +53,7 @@ func claimIdempotencyKeys(context context.Context, queries *sqlcv1.Queries, pool
 		claimedByExternalIds[i] = claim.ClaimedByExternalId
 	}
 
-	claimResults, err := queries.ClaimIdempotencyKeys(context, pool, sqlcv1.ClaimIdempotencyKeysParams{
+	claimResults, err := queries.ClaimIdempotencyKeys(context, tx, sqlcv1.ClaimIdempotencyKeysParams{
```
mrkaye97, line 56: whoops 🤦  thanks

pkg/repository/trigger.go
mrkaye97: can we factor out even more here? e.g. we have the same logic that calls `r.prepareTriggerFromEvents(ctx, r.pool, tenantId, opts)` in both `trigger.go` and `scheduler_optimistic.go` - maybe we can have a small wrapper there to get rid of some more duplication?

pkg/repository/trigger.go
```diff
@@ -236,504 +236,186 @@ func (r *TriggerRepositoryImpl) TriggerFromEvents(ctx context.Context, tenantId
...
+	tasks, dags, err := r.triggerWorkflows(ctx, nil, tenantId, triggerOpts, createCoreEventOpts)
 
-	seenAt := time.Now().UTC() // TODO: propagate this to caller, and figure out how we should be setting this
+	if err != nil {
+		return nil, fmt.Errorf("failed to trigger workflows: %w", err)
+	}
 
-	eventKeys := make([]string, 0, len(opts))
-	uniqueEventKeys := make(map[string]struct{})
+	populateEventExternalIdToRuns(eventExternalIdToRuns, externalIdToEventIdAndFilterId, tasks, dags)
```
mrkaye97: so I suspect this is a preference of mine, but I strongly dislike this pattern of passing in a map to a function like this and mutating it, and then moving on. I'd much rather we pass in the map, do whatever we need to do, and return. IMO, calling functions like this for side effects can be hard to understand / maintain (astonishing :P )

pkg/scheduling/v1/queuer.go
```diff
@@ -563,3 +570,193 @@ func (q *Queuer) flushToDatabase(ctx context.Context, r *assignResults) int {
...
+
+func (q *Queuer) flushToDatabaseOptimistic(
+	ctx context.Context,
+	r *assignResults,
+	tx *v1.OptimisticTx,
+	localWorkerIds map[string]struct{},
+) ([]*v1.AssignedItem, *QueueResults, error) {
+	begin := time.Now()
+
+	ctx, span := telemetry.NewSpan(ctx, "flush-to-database")
```
mrkaye97: nit: span naming - maybe also include `optimistic` in here?

internal/services/controllers/olap/signal/signal.go
abelanger5: Yep, no changes to the actual methods that were copied over. I did create `NotifyEventsCreated` and `NotifyCELEvaluationFailures`.

pkg/repository/trigger.go
abelanger5: yeah, that's a good idea, I'm just not sure if it'll add too much complexity in terms of casing on the `tx.tx` and `tx` existence, but I'll give it a try

internal/services/dispatcher/dispatcher_v1.go
abelanger5: This was my attempt at keeping the logic as close as possible to the previous implementation. For `handleTaskBulkAssignedTask`, the map contains the payload lookups for all of the tasks, while the tasks slice is just the subsection of tasks that we're running on that worker. So we can't range over all the tasks in the lookup.

internal/services/dispatcher/dispatcher_v1.go
abelanger5: discussed this IRL but for posterity: tasks get requeued if we hit an error, and the caller simply continues rather than retrying

internal/services/scheduler/v1/scheduler.go
abelanger5: this should be the only place we instantiate the tasks which get sent to the dispatchers

pkg/repository/trigger.go
abelanger5: yeah, this was again a case of preserving existing chunks of the program, since there was a mutation of this map hundreds of lines below its instantiation. I refactored this to avoid the surprising behavior and I'm just returning the map out of `getEventExternalIdToRuns` now.

internal/services/scheduler/v1/scheduler.go
abelanger5: removed casing on replayEnabled since we're guarding it from the tasks controller

internal/services/dispatcher/dispatcher_v1.go
abelanger5: will do in a separate PR

pkg/repository/trigger.go
abelanger5: tried to refactor this, think it's cleaner now

internal/services/ingestor/ingestor_v1.go
abelanger5: added comments 👍

internal/services/admin/v1/server.go
```diff
@@ -482,6 +485,54 @@ func (i *AdminServiceImpl) generateExternalIds(ctx context.Context, tenantId str
...
+			for workerId, assignedItems := range localAssigned {
+				eg.Go(func() error {
+					err := i.localDispatcher.HandleLocalAssignments(ctx, tenantId, workerId, assignedItems)
+
+					if err != nil {
+						return fmt.Errorf("could not dispatch assigned items: %w", err)
+					}
+
+					return nil
+				})
```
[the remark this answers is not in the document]
abelanger5: praying for the day go 1.22+ makes it in the training data

internal/services/admin/server_v1.go
```diff
@@ -230,6 +233,54 @@ func (i *AdminServiceImpl) generateExternalIds(ctx context.Context, tenantId str
...
+			for workerId, assignedItems := range localAssigned {
+				eg.Go(func() error {
+					err := i.localDispatcher.HandleLocalAssignments(ctx, tenantId, workerId, assignedItems)
+
+					if err != nil {
+						return fmt.Errorf("could not dispatch assigned items: %w", err)
+					}
+
+					return nil
+				})
```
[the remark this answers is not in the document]
abelanger5: same as above

internal/services/ingestor/ingestor_v1.go
```diff
@@ -49,44 +53,134 @@ func (i *IngestorImpl) ingestEventV1(ctx context.Context, tenant *sqlcv1.Tenant,
...
+			for workerId, assignedItems := range localAssigned {
+				eg.Go(func() error {
+					err := i.localDispatcher.HandleLocalAssignments(ctx, tenantId, workerId, assignedItems)
+
+					if err != nil {
+						return fmt.Errorf("could not dispatch assigned items: %w", err)
+					}
+
+					return nil
+				})
```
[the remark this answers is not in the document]
abelanger5: same as above

mrkaye97 (approved):
🥳

.github/workflows/release.yaml
```diff
@@ -8,48 +8,48 @@ on:
         required: true
 name: Release
 jobs:
-  # load:
```
mrkaye97, line 11: do we want to uncomment/delete these?

mrkaye97 (approved):
lgtm, would be great to run the python tests locally with / without optimistic mode on to see how it looks + just confirm

Issue #2808, opened 2026-01-16
[FEAT] Resend or SMTP Support


jdoss:
**Is your feature request related to a problem? Please describe.**
I'd like to use another transactional email provider, like Resend, to send Email alerts.

**Describe the solution you'd like**
Better email provider support in Hatchet. Even Generic SMTP support would be great.

**Describe alternatives you've considered**
Signing up for Postmark and paying for another email provider.

abelanger5:
Hey @jdoss I think this would be great! We're happy to provide email templates/html if someone would be interested in picking this up. I will also check in with some contributors and see if we can find someone willing to contribute this feature!

gregfurman:
👋 @abelanger5 Could I be assigned this?

abelanger5:
> 👋 [@abelanger5](https://github.com/abelanger5) Could I be assigned this?

yep, done 👍

gregfurman:
Hey @jdoss. Email alerting via SMTP (with `PLAIN` auth) should now be available 🙂 Check out the [self-hosting docs](https://docs.hatchet.run/self-hosting/smtp-server) on getting this configured.

Let me know if this unblocks you!

codeafridi:
hey this is still opened? i would like to contribute

gregfurman:
@codeafridi We ended up implementing the SMTP approach with plain auth support. Think this can be closed, but I was waiting to hear back from @jdoss about whether this solved their issue.

jdoss:
Hey sorry! I will test this out soon.

PR #2868, opened 2026-01-27
add: email alert support via SMTP

gregfurman:
# Description

Currently, Hatchet only supports sending alert emails via Postmark. This PR adds an additional email provider that supports sending messages to a mail-server using SMTP.

Addresses #2808

## Type of change

## What's Changed

- Adds new `smtp.SMTPService` email provider using `github.com/wneessen/go-mail`.
- Adds mock SMTP server for testing using `github.com/emersion/go-smtp` & `github.com/emersion/go-sasl`.
- Embeds HTML email templates in `smtp` package, adapted from `postmark` to use go-template syntax instead of mustache.
- Refactors `sendEmailFromTemplateRequest` and template alias strings to the `email` package.
- Introduces new `SERVER_EMAIL_SMTP_*` config fields and updates corresponding docs.

### TODO
- [x] Add email subject support in the new `SendEmailFromTemplateRequest`

## Testing

E2E tests were run against [sj26/mailcatcher](https://github.com/sj26/mailcatcher) -- which allows us to capture emails and render their templates from the browser.

<details>
<summary><b>See E2E Testing Steps</b></summary>

1. Launch an instance of `mailcatcher` and access the UI at [`localhost:9001`](http://localhost:9001)
   ```sh
   docker run -p 9001:1080 -p 9002:1025 sj26/mailcatcher
   ```

2. Include the following environment variables:

```env
SERVER_EMAIL_KIND=smtp
SERVER_EMAIL_SMTP_ENABLED=true
SERVER_EMAIL_SMTP_SERVER_ADDR=127.0.0.1:9002 # mailcatcher smtp server
SERVER_EMAIL_SMTP_FROM_EMAIL=hatchet@example.com
SERVER_EMAIL_SMTP_FROM_NAME=Hatchet
SERVER_EMAIL_SMTP_SUPPORT_EMAIL=support@example.com
```

3. Run a failing workflow for `WorkflowRunFailed`:
   ```sh
   go run ./examples/go/on-failure/main.go
   ```

4. Create a token that will expire in 6 days (144 hours) for `ExpiringToken`:
   ```sh
   go run ./cmd/hatchet-admin token create --name dummy --tenant-id 707d0855-80ab-4e1f-a156-f1c4546cbf52 --expiresIn 144h
   ```

5. Access the UI, login as admin, and invite a new user for `TenantInvite`.

</details>

#### Screenshots

The following emails were captured, and rendered, as a result of the above e2e tests:

<details>
<summary><b>Workflow Run Failed alert email</b></summary>
image
</details>

<details>
<summary><b>Tenant Invite email</b></summary>
image
</details>

<details>
<summary><b>Expiring Token email</b></summary>
image
</details>

<details>
<summary><b>Resource Limit alert email</b></summary>

Note: This was not captured with an e2e test but instead by inspecting the body of the `TestSMTPServiceSendMail_resource_limit_alert` case in `smtp_test.go`.

image
</details>

### DIFF
--- a/frontend/docs/pages/self-hosting/configuration-options.mdx
+++ b/frontend/docs/pages/self-hosting/configuration-options.mdx
@@ -319,11 +319,19 @@ Variables marked with ⚠️ are conditionally required when specific features a
 | `SERVER_TENANT_ALERTING_SLACK_CLIENT_ID`     | Slack client ID                     |                        |
 | `SERVER_TENANT_ALERTING_SLACK_CLIENT_SECRET` | Slack client secret                 |                        |
 | `SERVER_TENANT_ALERTING_SLACK_SCOPES`        | Slack scopes                        | `["incoming-webhook"]` |
+| `SERVER_EMAIL_KIND`                          | Email integration kind              | `postmark`             |
 | `SERVER_EMAIL_POSTMARK_ENABLED`              | Enable Postmark                     |                        |
 | `SERVER_EMAIL_POSTMARK_SERVER_KEY`           | Postmark server key                 |                        |
 | `SERVER_EMAIL_POSTMARK_FROM_EMAIL`           | Postmark from email                 |                        |
 | `SERVER_EMAIL_POSTMARK_FROM_NAME`            | Postmark from name                  | `Hatchet Support`      |
 | `SERVER_EMAIL_POSTMARK_SUPPORT_EMAIL`        | Postmark support email              |                        |
+| `SERVER_EMAIL_SMTP_ENABLED`                  | Enable SMTP                         |                        |
+| `SERVER_EMAIL_SMTP_SERVER_ADDR`              | SMTP server address                 |                        |
+| `SERVER_EMAIL_SMTP_FROM_EMAIL`               | SMTP from email                     |                        |
+| `SERVER_EMAIL_SMTP_FROM_NAME`                | SMTP from name                      | `Hatchet Support`      |
+| `SERVER_EMAIL_SMTP_SUPPORT_EMAIL`            | SMTP support email                  |                        |
+| `SERVER_EMAIL_SMTP_AUTH_USERNAME`            | SMTP authentication username        |                        |
+| `SERVER_EMAIL_SMTP_AUTH_PASSWORD`            | SMTP authentication password        |                        |
 | `SERVER_MONITORING_ENABLED`                  | Enable monitoring                   | `true`                 |
 | `SERVER_MONITORING_PERMITTED_TENANTS`        | Permitted tenants for monitoring    |                        |
 | `SERVER_MONITORING_PROBE_TIMEOUT`            | Monitoring probe timeout            | `30s`                  |
--- a/pkg/config/loader/loader.go
+++ b/pkg/config/loader/loader.go
@@ -632,6 +632,8 @@ func createControllerLayer(dc *database.Layer, cf *server.ServerConfigFile, vers
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to create SMTP service: %w", err)
 		}
+	default:
+		return nil, nil, fmt.Errorf("invalid email provider of type %s, must be 'postmark' or 'smtp'", cf.Email.Kind)
 	}
 
 	additionalOAuthConfigs := make(map[string]*oauth2.Config)
--- a/pkg/config/loader/loader.go
+++ b/pkg/config/loader/loader.go
@@ -37,6 +37,7 @@ import (
 	"github.com/hatchet-dev/hatchet/pkg/errors/sentry"
 	"github.com/hatchet-dev/hatchet/pkg/integrations/email"
 	"github.com/hatchet-dev/hatchet/pkg/integrations/email/postmark"
+	"github.com/hatchet-dev/hatchet/pkg/integrations/email/smtp"
 	"github.com/hatchet-dev/hatchet/pkg/logger"
 	"github.com/hatchet-dev/hatchet/pkg/repository/cache"
 	"github.com/hatchet-dev/hatchet/pkg/repository/debugger"
@@ -604,13 +605,33 @@ func createControllerLayer(dc *database.Layer, cf *server.ServerConfigFile, vers
 
 	var emailSvc email.EmailService = &email.NoOpService{}
 
-	if cf.Email.Postmark.Enabled {
+	switch strings.ToLower(cf.Email.Kind) {
+	case "postmark":
+		if !cf.Email.Postmark.Enabled {
+			break
+		}
 		emailSvc = postmark.NewPostmarkClient(
 			cf.Email.Postmark.ServerKey,
 			cf.Email.Postmark.FromEmail,
 			cf.Email.Postmark.FromName,
 			cf.Email.Postmark.SupportEmail,
 		)
+
+	case "smtp":
+		if !cf.Email.SMTP.Enabled {
+			break
+		}
+		emailSvc, err = smtp.NewSMTPService(
+			cf.Email.SMTP.ServerAddr,
+			cf.Email.SMTP.BasicAuth.Username,
+			cf.Email.SMTP.BasicAuth.Password,
+			cf.Email.SMTP.FromEmail,
+			cf.Email.SMTP.FromName,
+			cf.Email.SMTP.SupportEmail,
+		)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to create SMTP service: %w", err)
+		}
 	}
 
 	additionalOAuthConfigs := make(map[string]*oauth2.Config)
--- a/pkg/config/server/server.go
+++ b/pkg/config/server/server.go
@@ -473,11 +473,11 @@ type RabbitMQConfigFile struct {
 }
 
 type ConfigFileEmail struct {
-	Kind string `mapstructure:"kind" json:"kind,omitempty" validate:"required,oneof=postmark smtp" default:"postmark"`
+	Kind string `mapstructure:"kind" json:"kind,omitempty"`
 
-	Postmark EmailConfig `mapstructure:"postmark" json:"postmark,omitempty"`
+	Postmark PostmarkConfigFile `mapstructure:"postmark" json:"postmark,omitempty"`
 
-	SMTP EmailConfig `mapstructure:"smtp" json:"smtp,omitempty"`
+	SMTP SMTPEmailConfig `mapstructure:"smtp" json:"smtp,omitempty"`
 }
 
 type ConfigFileMonitoring struct {
@@ -494,7 +494,16 @@ type ConfigFileMonitoring struct {
 	TLSRootCAFile string `mapstructure:"tlsRootCAFile" json:"tlsRootCAFile,omitempty"`
 }
 
-type EmailConfig struct {
+type PostmarkConfigFile struct {
+	Enabled bool `mapstructure:"enabled" json:"enabled,omitempty"`
+
+	ServerKey    string `mapstructure:"serverKey" json:"serverKey,omitempty"`
+	FromEmail    string `mapstructure:"fromEmail" json:"fromEmail,omitempty"`
+	FromName     string `mapstructure:"fromName" json:"fromName,omitempty" default:"Hatchet Support"`
+	SupportEmail string `mapstructure:"supportEmail" json:"supportEmail,omitempty"`
+}
+
+type SMTPEmailConfig struct {
 	Enabled      bool   `mapstructure:"enabled" json:"enabled,omitempty"`
 	ServerKey    string `mapstructure:"serverKey" json:"serverKey,omitempty"`
 	ServerAddr   string `mapstructure:"serverAddr" json:"serverAddr,omitempty"`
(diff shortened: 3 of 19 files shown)


pkg/config/server/server.go
```diff
@@ -473,7 +473,11 @@ type RabbitMQConfigFile struct {
 }
 
 type ConfigFileEmail struct {
-	Postmark PostmarkConfigFile `mapstructure:"postmark" json:"postmark,omitempty"`
+	Kind string `mapstructure:"kind" json:"kind,omitempty" validate:"required,oneof=postmark smtp" default:"postmark"`
```
abelanger5: note that we don't actually use these validator tags at the moment (when validating the config), so we'll need to throw an error if it's neither of these from the loader.

pkg/config/server/server.go
```diff
@@ -473,7 +473,11 @@ type RabbitMQConfigFile struct {
...
+
+	Postmark EmailConfig `mapstructure:"postmark" json:"postmark,omitempty"`
+
+	SMTP EmailConfig `mapstructure:"smtp" json:"smtp,omitempty"`
```
abelanger5: Does it make sense for both of these strategies to share a config? I'm thinking that these are likely long-term divergent and that each of these should have a separate config type.

pkg/integrations/email/smtp/templates/expiring_token.html
```diff
@@ -0,0 +1,30 @@
...
+<br>
+We are reaching out to let you know that you have a Hatchet token expiring soon:
+<br>
+<br>
+<table class="attributes" width="100%" cellpadding="0" cellspacing="0">
+  <tr>
+    <td class="attributes_content">
+      <table width="100%" cellpadding="0" cellspacing="0">
+        <tr>
+          <td class="attributes_item"><a href="{{.TokenSettings}}">{{.TokenName}}</a> expires {{.ExpiresAtRelativeDate}}</td>
```
abelanger5, line 13: Just to make extra sure, if an attacker ha somehow created a `TokenName` in the database which uses html, for example `<div>token</div>`, we properly sanitize it because `go-mail` uses `html/template` under the hood?

pkg/integrations/email/smtp/mock_smtp_test.go
abelanger5, line 1: Note that we have a pretty annoying but necessary build tag convention for testing files (you can look at some of the other testing files for reference) to avoid running unnecessary tests when certain build tags are present. For example, if I use `-tag integration` I don't want to run all of the unit tests.

pkg/integrations/email/smtp/smtp.go
```diff
@@ -0,0 +1,148 @@
...
+	if req.To != "" {
+		if err := msg.ToFromString(req.To); err != nil {
+			return err
+		}
+	}
+
+	switch data := req.TemplateModel.(type) {
+	case interface{ GetSubject() string }:
+		msg.Subject(data.GetSubject())
+	}
```
abelanger5: I'm a little confused by this - why don't we just pass in the subject as an argument? Or pass in an interface which satisfies something like:

```go
type SendRequestWithSubject interface {
  GetSubject() string
  GetEmailRequest() *email.SendEmailFromTemplateRequest
}
```

frontend/docs/pages/self-hosting/configuration-options.mdx
abelanger5, line 1: Do we think it's worth writing a tutorial/extra docs page for setting up an SMTP server? A common point of feedback for this page is that people don't know what the config options refer to and we have no guidance for these options.

pkg/integrations/email/smtp/smtp.go
gregfurman: > why don't we just pass in the subject as an argument?

This was just an easier way of extracting and constructing the email subject without changing the original `EmailProvider` interface and explicitly attaching formats/templates to each request type.

> Or pass in an interface which satisfies something like:

I was originally thinking something like this... but again I didn't want to change the structure of anything publicly exported since that'd be a breaking change for those importing `pkg/integrations/email`. Or perhaps I'm misunderstanding your suggestion 👀 

> I'm a little confused by this -

Alternatively, I could use a simpler type-switch if you think that's more explicit:

```golang
	var subject string
	switch r := req.TemplateModel.(type) {
	case email.WorkflowRunsFailedEmailData:
		subject = return fmt.Sprintf("[%s] %s", r.TenantName, r.Subject)
	case email.ExpiringTokenEmailData:
		subject = return fmt.Sprintf("[%s] %s", r.TenantName, r.Subject)
	case email.ResourceLimitAlertData:
		subject = return fmt.Sprintf("[%s] %s", r.TenantName, r.Subject)
	case email.TenantInviteEmailData:
		subject = fmt.Sprintf("%s invited you to join %s on Hatchet", r.InviteSenderName, r.TenantName)
	}
```

pkg/integrations/email/smtp/templates/expiring_token.html
gregfurman: The `SetBodyHTMLTemplate` (which uses `html/template` under-the-hood) should sanitize templates i.e  `<` becomes `&lt;` etc.

pkg/config/server/server.go
gregfurman: I was thinking that configuring an email provider is mutually exclusive: like I would either select `postmark` OR `smtp` which is why I thought they'd reside in the same config.

frontend/docs/pages/self-hosting/configuration-options.mdx
gregfurman: That seems like a good idea. A simple tutorial on how to do this for gmail (or proton mail) could be a neat!

Should we add this as a part of this PR? Else I'll open an issue and get started on a follow-up once this is in

abelanger5: I think part of this PR would be great! Thanks 💯

pkg/integrations/email/smtp/smtp.go
abelanger5: Ah, got it, I understand now. What if we modify the `templateRegistry` so that it stores the subject template as well? That will keep it consistent with Postmark where the template also stores the subject.

pkg/config/server/server.go
abelanger5: Yep they should definitely be mutually exclusive, but from a maintainability perspective it's confusing that only a subset of fields in the config are bound in `BindAllEnv` and are read in the `loader`. But I'm fine with this approach!

pkg/integrations/email/smtp/smtp.go
gregfurman: Neat idea 👆Will do that instead of my overengineered type-switching approach

frontend/docs/pages/self-hosting/configuration-options.mdx
gregfurman: Done with [frontend/docs/pages/self-hosting/smtp-server.mdx](https://github.com/hatchet-dev/hatchet/pull/2868/changes#diff-cf346f2f5fff2846b267cd36c1add9b35c19d8d13312b165d168fdbfa8f1c31f)

Lmk if you think I should add more detail here.

pkg/config/server/server.go
gregfurman: Ahh I see. Agree with your assessment! Will change.

pkg/config/loader/loader.go
```diff
@@ -604,13 +605,35 @@ func createControllerLayer(dc *database.Layer, cf *server.ServerConfigFile, vers
...
+			cf.Email.SMTP.BasicAuth.Password,
+			cf.Email.SMTP.FromEmail,
+			cf.Email.SMTP.FromName,
+			cf.Email.SMTP.SupportEmail,
+		)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to create SMTP service: %w", err)
+		}
+	default:
+		return nil, nil, fmt.Errorf("invalid email provider of type %s, must be 'postmark' or 'smtp'", cf.Email.Kind)
```
abelanger5, line 636: small fix: for backwards compatibility with existing config, I think we might want to set the kind to `postmark` if we have valid fields for `cf.Email.Postmark.ServerKey`. otherwise this is a breaking change

pkg/integrations/email/smtp/smtp.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (s *SMTPService) SendTemplateEmailBCC(ctx context.Context, bcc, templateAlias string, templateModelData any, bccSupport bool) error {
+	if bccSupport {
+		bcc = fmt.Sprintf("%s,%s", bcc, s.supportEmail)
+	}
+
+	return s.sendRequest(ctx, &email.SendEmailFromTemplateRequest{
+		From:          fmt.Sprintf("%s <%s>", s.fromName, s.fromEmail),
+		To:            "",
```
abelanger5: the `To` field should still contain the address, it's just that the additional email is bcc'd

pkg/integrations/email/smtp/smtp.go
```diff
@@ -0,0 +1,164 @@
...
+	return s.sendRequest(ctx, &email.SendEmailFromTemplateRequest{
+		From:          fmt.Sprintf("%s <%s>", s.fromName, s.fromEmail),
+		To:            to,
+		Bcc:           bcc,
+		TemplateAlias: templateAlias,
+		TemplateModel: templateModelData,
+	})
+}
+
+func (s *SMTPService) SendTemplateEmailBCC(ctx context.Context, bcc, templateAlias string, templateModelData any, bccSupport bool) error {
```
abelanger5: I just checked and we actually aren't calling `SendTemplateEmail` or `SendTemplateEmailBCC` from anywhere that's not the email packages. I'd be in favor of just removing these from the interface and just having a `sentTemplateEmail` implementation in each package which includes the `bccSupport` flag.

frontend/docs/pages/self-hosting/configuration-options.mdx
abelanger5: this looks great!

abelanger5 (changes requested):

pkg/integrations/email/smtp/smtp.go
gregfurman: Wouldn't that be a breaking change considering the interface is in `pkg` and therefore publicly exported?

abelanger5: Yes, but that's fine and we're comfortable with that - we have yet to encounter a single instance of someone importing the Hatchet engine or server into their application and even if they did, it's exceedingly unlikely that they'd import the alert package and call these methods. 

The breaking changes to avoid are on the config level or the env var level, which we want to keep backwards compatible.

pkg/integrations/email/smtp/smtp.go
gregfurman: > it's just that the additional email is bcc'd

That differs from the postmark implementation:

https://github.com/hatchet-dev/hatchet/blob/8cc22e3de10a19d572d9221ea7db6b58b61d5ae6/pkg/integrations/email/postmark/postmark.go#L96-L101

Isn't the purpose of `SendTemplateEmailBCC` to send mail where recipients are _only_ BCC'd in?

pkg/config/loader/loader.go
gregfurman: 🤦  I originally made `"postmark"` the default value for `Kind` to ensure backwards compat but mistakenly removed it when removing those `validate` tags from the config struct. From [this commit](https://github.com/gregfurman/hatchet/blob/a1c5dc823adfc22093677d9cacb5c46aec811d36/pkg/config/server/server.go#L475-L476):

```golang
type ConfigFileEmail struct {
	Kind string `mapstructure:"kind" json:"kind,omitempty" validate:"required,oneof=postmark smtp" default:"postmark"`
// ...
```

If we set the `Kind` to `"postmark"` by default, then (afaik) this shoudn't be a worry:

```golang
type ConfigFileEmail struct {
	Kind string `mapstructure:"kind" json:"kind,omitempty" default:"postmark"`
// ...
```

wdyt?

pkg/integrations/email/smtp/smtp.go
gregfurman: OK I simplified the `EmailService` interface and made a `sendRequest` method per implementation.

Think it's looking much simpler now but lmk thoughts!

> The breaking changes to avoid are on the config level or the env var level, which we want to keep backwards compatible.

Gotcha. Was unaware of this guarantee. Could be worth moving some of these packages into `internal` if they are prone to breaking changes (see [Hyrum's Law](https://www.hyrumslaw.com/) 👀)

abelanger5: I agree with this in spirit! The problem is that we import Hatchet internally to build Hatchet Cloud. If a package is in `/pkg`, we likely have extended it or imported it for this purpose. Though we can deal with breaking changes easily on our end, and if the upgrade becomes a problem it's easy for us to fix in OSS. 

Now that you mention it, it's quite likely that we're using `SendTemplateEmailBCC` in Hatchet Cloud, so I'll need to double check that. But not a blocker for this PR.

pkg/integrations/email/smtp/smtp.go
abelanger5: Ah, yep, nice catch. I believe this was intentional, if I remember correctly, so that multiple recipients are not CC'd with each other.

pkg/config/loader/loader.go
abelanger5: yep looks good!

abelanger5 (approved):
💯

abelanger5:
Load test is detecting a known `goleak`, merging and this will be in the next release!

PR #2874, opened 2026-01-28
feat: Extend webhook support for scope_expression and payload

jishnundth:
# Description
Extend webhook support for scope_expression and payload

## Type of change


api/v1/server/handlers/v1/webhooks/receive.go
```diff
@@ -197,6 +197,18 @@ func (w *V1WebhooksService) V1WebhookReceive(ctx echo.Context, request gen.V1Web
 
 	headerMap := make(map[string]string)
 
+	if len(webhook.StaticPayload) > 0 {
+		var staticPayloadMap map[string]interface{}
+		if unmarshalErr := json.Unmarshal(webhook.StaticPayload, &staticPayloadMap); unmarshalErr != nil {
+			w.config.Logger.Warn().Err(unmarshalErr).Str("webhook", webhookName).Str("tenant", tenantId).Msg("Failed to unmarshal static payload")
```
mrkaye97: I think maybe this should result in an error response?

api/v1/server/handlers/v1/webhooks/receive.go
```diff
@@ -261,7 +306,7 @@ func (w *V1WebhooksService) V1WebhookReceive(ctx echo.Context, request gen.V1Web
 		payload,
 		nil,
 		nil,
-		nil,
+		scope,
```
mrkaye97, line 315: great!

api/v1/server/oas/transformers/v1/webhooks.go
```diff
@@ -23,6 +25,19 @@ func ToV1Webhook(webhook *sqlcv1.V1IncomingWebhook) gen.V1Webhook {
...
+
+	if webhook.ScopeExpression.Valid {
+		result.ScopeExpression = &webhook.ScopeExpression.String
+	}
+
+	if len(webhook.StaticPayload) > 0 {
+		var staticPayload map[string]interface{}
+		if err := json.Unmarshal(webhook.StaticPayload, &staticPayload); err == nil {
+			result.StaticPayload = &staticPayload
+		}
```
mrkaye97: do we need to handle the error here?

mrkaye97 (commented):
This looks great! I'll play around with the UI a bit and then will approve and get it shipped - left a couple small comments

api/v1/server/oas/transformers/v1/webhooks.go
jishnundth: You're right. I'll remove it

api/v1/server/handlers/v1/webhooks/receive.go
jishnundth: This'll ideally never occur, because frontend is already handling the invalid JSON. 

~~https://github.com/hatchet-dev/hatchet/pull/2874/changes#diff-9f7b133f7583bb2aad05ef3d7dfe1d6176fb62361349cdb08c9218fd9d7f4199R137-R149~~

https://github.com/hatchet-dev/hatchet/pull/2874/changes#diff-68c0a74862d2ee036072d10825055dde0c6ad59077d2020b9201fbd0c115a85bR84-R99

I'll add it if necessary. Let me know what you think.

mrkaye97: gotcha! I think we should error if this happens anyways even though it shouldn't, murphy's law and all :)

api/v1/server/oas/transformers/v1/webhooks.go
```diff
@@ -23,6 +25,18 @@ func ToV1Webhook(webhook *sqlcv1.V1IncomingWebhook) gen.V1Webhook {
...
 		SourceName:         gen.V1WebhookSourceName(webhook.SourceName),
 	}
+
+	if webhook.ScopeExpression.Valid {
+		result.ScopeExpression = &webhook.ScopeExpression.String
+	}
+
+	if len(webhook.StaticPayload) > 0 {
+		var staticPayload map[string]interface{}
+		json.Unmarshal(webhook.StaticPayload, &staticPayload)
```
mrkaye97: whoops sorry I think I misled you - I didn't realize we couldn't return an error here. I think we should either do what you had before and log the error, or we can also have this transformer return `(*gen.V1Webhook, error)` instead and handle the error that way. Either way is okay with me, sorry about that!

jishnundth: But we have validation from Frontend -> API Handler -> Database. Should we still throw an error from here as well :)

I'll add it anyways 😁

mrkaye97 (approved):
looks great! I'll get this merged in a bit once we finish up deploying some things.

thanks for putting this in!

PR #2877, opened 2026-01-28
fix: big int alignment for cleanup function

grutt:
# Description

Fixes overflow when 2.1 billion tasks are run with concurrency... 

2026-01-28T19:22:48.062Z ERR error in pre-ack on msg task-completed: could not complete tasks: failed to release tasks: ERROR: integer out of range (SQLSTATE 22003) service=rabbitmq

## Type of change

### DIFF
--- a/cmd/hatchet-migrate/migrate/migrations/20260128120000_v1_0_71.sql
+++ b/cmd/hatchet-migrate/migrate/migrations/20260128120000_v1_0_71.sql
@@ -1,0 +1,243 @@
+-- +goose Up
+-- +goose StatementBegin
+-- Fix: v1_workflow_concurrency_slot.sort_id is BIGINT, but the cleanup function used INTEGER
+-- which can overflow once sort_id exceeds int32.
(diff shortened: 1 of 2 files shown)


cmd/hatchet-migrate/migrate/migrations/20260128120000_v1_0_71.sql
```diff
@@ -0,0 +1,243 @@
...
+-- +goose Down
+-- +goose StatementBegin
+-- Revert to previous definition (keeps BIGINT to avoid reintroducing overflow)
+CREATE OR REPLACE FUNCTION cleanup_workflow_concurrency_slots(
+    p_strategy_id BIGINT,
+    p_workflow_version_id UUID,
+    p_workflow_run_id UUID
+) RETURNS VOID AS $$
+DECLARE
+    v_sort_id BIGINT;
```
mrkaye97: think this one should be `INTEGER` for the down? 

```suggestion
    v_sort_id INTEGER;
```

mrkaye97 (approved):

PR #2883, opened 2026-01-29
Make sure we query for pending invites in the create tenant page

mnafees:
# Description

Since the create tenant page is mounted with the `rootRoute` instead of `authenticatedRoute` we were not checking for pending invites. This PR fixes that and some other logic around it.

## Type of change


grutt:
i'd love to see cypress tests for these flows

mrkaye97 (approved):

grutt (approved):

PR #2581, opened 2025-11-27
[WIP] Retire `pgtype.UUID` completely from the codebase

mnafees:
# Description

We were using two different types to represent UUIDv4 in the codebase. Not only did this lead to too many type conversions and memory allocations but led to bugs at times. This PR retires the usage of `pgtype.UUID` completely and makes `github.com/google/uuid` the de-facto UUIDv4 type across the entire codebase.

## Type of change

- [x] Chore (changes which are not directly related to any business logic)

PR #2894, opened 2026-01-30
Refactor: Attempt II at removing `pgtype.UUID` everywhere + convert string UUIDs into `uuid.UUID`

mrkaye97:
# Description

Attempt 2 at #2581 

The main goal here was to replace `pgtype.UUID` with `uuid.UUID` everywhere. This uncovered a handful of small bugs (especially in some of our queries) where we were relying on `pgtype.UUID` semantics to tell the difference between a null value and a non-null value, which breaks with a more restrictive type (e.g. not using `sqlc.narg` in queries and just implicitly allowing `@foobar::UUID` to be nullable via `pgtype.UUID`). 

I also got a little greedy, so I tried my best to replace all instances of string uuids everywhere with `uuid.UUID` too. Should help a lot with type safety everywhere, and limit the amount of type casting we're doing

I feel pretty good about this with all the tests passing and such, but it's pretty likely there are at least one or two bugs or commands or some such that are broken, so will need to track those down incrementally I think

## Type of change


api/v1/server/authz/middleware.go
```diff
@@ -78,7 +77,7 @@ func (a *AuthZ) handleCookieAuth(c echo.Context, r *middleware.RouteInfo) error
 		}
 
 		// check if the user is a member of the tenant
-		tenantMember, err := a.config.V1.Tenant().GetTenantMemberByUserID(c.Request().Context(), sqlchelpers.UUIDToStr(tenant.ID), sqlchelpers.UUIDToStr(user.ID))
+		tenantMember, err := a.config.V1.Tenant().GetTenantMemberByUserID(c.Request().Context(), tenant.ID, user.ID.String())
```
abelanger5: Why is the `userId` param a string while the `tenantId` is a UUID?

api/v1/server/handlers/api-tokens/create.go
```diff
@@ -50,7 +49,7 @@ func (a *APITokenService) ApiTokenCreate(ctx echo.Context, request gen.ApiTokenC
 
 	a.config.Analytics.Enqueue(
 		"api-token:create",
-		sqlchelpers.UUIDToStr(user.ID),
+		user.ID.String(),
```
abelanger5, line 52: Would be in favor of modifying all `userId` references to `uuid.UUID` as well

api/v1/server/handlers/slack-app/oauth_callback.go
```diff
@@ -22,7 +23,13 @@ func (g *SlackAppService) UserUpdateSlackOauthCallback(ctx echo.Context, _ gen.U
 
 	sh := authn.NewSessionHelpers(g.config)
 
-	tenantId, err := sh.GetKey(ctx, "tenant")
+	tenantIdStr, err := sh.GetKey(ctx, "tenant")
```
abelanger5: Could have a method on the session helpers called `GetKeyUUID`?

api/v1/server/oas/transformers/v1/tasks.go
```diff
@@ -320,11 +318,11 @@ func ToTask(taskWithData *v1.TaskWithPayloads, workflowRunExternalId pgtype.UUID
 		Output:                output,
 		Status:                gen.V1TaskStatus(taskWithData.Status),
 		Input:                 input,
-		TenantId:              uuid.MustParse(sqlchelpers.UUIDToStr(taskWithData.TenantID)),
-		WorkflowId:            uuid.MustParse(sqlchelpers.UUIDToStr(taskWithData.WorkflowID)),
+		TenantId:              uuid.MustParse(taskWithData.TenantID.String()),
```
abelanger5: Seeing this pattern a lot, why can't this just be `taskWithData.TenantID`?

pkg/repository/api_token.go
```diff
@@ -51,11 +52,11 @@ func newAPITokenRepository(shared *sharedRepository, cacheDuration time.Duration
 }
 
 func (a *apiTokenRepository) RevokeAPIToken(ctx context.Context, id string) error {
-	return a.queries.RevokeAPIToken(ctx, a.pool, sqlchelpers.UUIDFromStr(id))
+	return a.queries.RevokeAPIToken(ctx, a.pool, uuid.MustParse(id))
```
abelanger5: I think we should take a close look at any remaining `MustParse` calls, because we are intentionally introducing additional panic paths which could potentially crash the API servers or engine

api/v1/server/oas/transformers/v1/tasks.go
mrkaye97: whoops yep, find and replace - will fix this!

api/v1/server/oas/transformers/worker.go
```diff
@@ -28,10 +27,9 @@ func ToWorkerLabels(labels []*sqlcv1.ListWorkerLabelsRow) *[]gen.WorkerLabel {
 			value = nil
 		}
 
-		id := fmt.Sprintf("%d", labels[i].ID)
-
 		resp[i] = gen.WorkerLabel{
-			Metadata: *toAPIMetadata(id, labels[i].CreatedAt.Time, labels[i].UpdatedAt.Time),
+			// fixme: `id` needs to be a uuid
```
grutt, line 31: fixme?

mrkaye97: for later, we discussed on slack

grutt (approved):

api/v1/server/oas/transformers/worker.go
grutt: see fixme --> write fixme?

internal/queueutils/pool.go
```diff
@@ -38,7 +37,7 @@ func (p *OperationPool[T]) SetTenants(tenants []*sqlcv1.Tenant) {
 	tenantMap := make(map[string]bool)
```
abelanger5, line 37: should this just be `map[uuid.UUID]bool`?

abelanger5 (approved):

internal/queueutils/pool.go
mrkaye97: yeah maybe - I was trying not to mess with the operation pool stuff too much because of all the generics and reflection going on here, figured we could clean this up later

pkg/repository/match.go
```diff
@@ -669,7 +668,11 @@ func (m *sharedRepository) processEventMatches(ctx context.Context, tx sqlcv1.DB
 					// signals are durable, meaning they persist between retries, so a retryCount of -1 is used
 					RetryCount: -1,
 				})
-				externalIds = append(externalIds, sqlchelpers.UUIDToStr(match.SignalExternalID))
+				if match.SignalExternalID != nil {
+					externalIds = append(externalIds, *match.SignalExternalID)
+				} else {
+					externalIds = append(externalIds, uuid.Nil)
```
grutt, line 674: correct

pkg/repository/olap.go
```diff
@@ -1159,18 +1140,24 @@ func (r *OLAPRepositoryImpl) ListWorkflowRuns(ctx context.Context, tenantId stri
 				continue
 			}
 
-			outputPayload, exists := externalIdToPayload[dag.OutputEventExternalID]
+			outputEventExternalId := uuid.Nil
```
grutt, line 1143: correct

pkg/repository/olap.go
```diff
@@ -1340,20 +1327,25 @@ func (r *OLAPRepositoryImpl) ListTaskRunEvents(ctx context.Context, tenantId str
...
 	if err != nil {
 		return nil, err
 	}
 
-	externalIds := make([]pgtype.UUID, len(rows))
+	externalIds := make([]uuid.UUID, len(rows))
 
 	for i, row := range rows {
-		externalIds[i] = row.EventExternalID
+		eventExternalId := uuid.Nil
```
grutt, line 1343: incorrect

pkg/repository/olap.go
```diff
@@ -1365,7 +1357,11 @@ func (r *OLAPRepositoryImpl) ListTaskRunEventsByWorkflowRunId(ctx context.Contex
 	taskEventWithPayloads := make([]*TaskEventWithPayloads, 0, len(rows))
 
 	for _, row := range rows {
-		payload, exists := payloads[row.EventExternalID]
+		eventExternalId := uuid.Nil
```
grutt, line 1360: incorrect

pkg/repository/olap.go
```diff
@@ -1627,13 +1608,18 @@ func (r *OLAPRepositoryImpl) UpdateTaskStatuses(ctx context.Context, tenantIds [
 					eventCount = int(row.Count)
 				}
 
+				latestWorkerId := uuid.Nil
```
grutt, line 1611: incorrect

PR #2896, opened 2026-01-30
fix: only fetch finalized workflow runs

grutt:
# Description

In ListFinalizedWorkflowRuns, the code was calling listTaskOutputEvents for ALL tasks passed in, even though verifyAllTasksFinalized had already determined which workflow runs were actually finalized. This meant:
1. If 1000 workflow runs were being tracked but only 10 were finalized
2. The system would fetch large payload data for all 1000 workflow runs' tasks
3. Only the 10 finalized ones would be used; the other 990 were immediately discarded
4. Each polling iteration (every 1 second) would repeat this, causing 349MB+ of memory allocations that were never needed

The fix filters taskExternalIds to only include tasks belonging to finalized workflow runs before calling listTaskOutputEvents

## Type of change


abelanger5 (approved):

promptless-for-oss:

New suggestion: **[Add SERVER_WORKFLOW_RUN_BUFFER_SIZE configuration option](https://app.gopromptless.ai/suggestions/3eed2d85-88a4-42ed-adc9-57ad6361a3d9)**

---

PR #2902, opened 2026-02-01
[Python] Fix: Empty Output Serialization

mrkaye97:
# Description

Fixing one cause of `failed to unmarshal json` from the Python SDK sending an empty string instead of `None` when a task returns nothing

## Type of change


sdks/python/tests/worker_fixture.py
```diff
@@ -30,6 +30,8 @@ def wait_for_worker_health(healthcheck_port: int) -> bool:
             if resp.status_code == 200:
                 return True
 
+            time.sleep(1)
```
mrkaye97, line 33: drive by

abelanger5 (approved):

PR #2907, opened 2026-02-02
chore: dependabot updates

grutt:
# Description

Updates packages flagged in Security

## Type of change


mrkaye97 (approved):
thanks for dealing with this 😭

PR #2911, opened 2026-02-02
[WIP] New `Memo()` function in Go SDK for durable tasks

mnafees:
# Description

Add memoization support to the Go SDK with the `Memo` function on the `DurableContext`.

## Type of change


pkg/repository/sqlcv1/durable_event_log.sql
```diff
@@ -0,0 +1,30 @@
+-- name: GetDurableEventLog :one
+SELECT *
+FROM v1_durable_event_log
+WHERE
+    tenant_id = @tenant_id::uuid
+    AND task_id = @task_id::bigint
+    AND task_inserted_at = @task_inserted_at::timestamptz
+    AND retry_count = @retry_count::integer
```
mrkaye97: I think we should probably be memoizing across retries, so we might not want to include the retry count in this table / filtering by it

cmd/hatchet-migrate/migrate/migrations/20260203165306_v1_0_75.sql
```diff
@@ -0,0 +1,29 @@
...
+-- +goose StatementBegin
+CREATE TABLE v1_durable_event_log (
+    id BIGINT GENERATED ALWAYS AS IDENTITY,
+    tenant_id UUID NOT NULL,
+    task_id BIGINT NOT NULL,
+    task_inserted_at TIMESTAMPTZ NOT NULL,
+    retry_count INTEGER NOT NULL DEFAULT 0,
+    event_type TEXT NOT NULL,
+    key TEXT NOT NULL,
+    data JSONB,
```
mrkaye97: I'll need to think about this - I'm not sure if we want this to be json or a string 🤔

pkg/repository/sqlcv1/durable_event_log.sql
```diff
@@ -0,0 +1,30 @@
...
+) VALUES (
+    @tenant_id::uuid,
+    @task_id::bigint,
+    @task_inserted_at::timestamptz,
+    @retry_count::integer,
+    @event_type::text,
+    @key::text,
+    @data::jsonb
+)
+RETURNING *;
```
mrkaye97: I think we probably want `ON CONFLICT DO NOTHING` here, which gets us around the risk of what we were discussing before where two of the same task try to write at about the same time. If we do this, then we can just return the previously-written result instead of erroring

cmd/hatchet-migrate/migrate/migrations/20260203165306_v1_0_75.sql
mnafees: Kept it JSON since we want the memoize result to be JSON serializable like task inputs.

promptless-for-oss:

New suggestion: **[Document Memo() function for Go SDK durable tasks](https://app.gopromptless.ai/suggestions/237575c2-9ef0-4b37-8877-1c7e983a14c3)**

---
_Tip: Add or adjust Promptless's style guide in [Agent Knowledge Base](https://app.gopromptless.ai/configure/settings) ✍️_

PR #2913, opened 2026-02-02
chore:  proto naming conventions

grutt:
# Description

Aligns naming conventions for proto props. Since this only changes names, these changes are backwards compatible with existing sdks in the wild. 

## Type of change


mrkaye97:
I'm a little confused how these changes are backwards compatible with existing sdks 😅 if an SDK tries to e.g. push an event and sends `tenantId` instead of `tenant_id`, don't we lose that param?

grutt:
> I'm a little confused how these changes are backwards compatible with existing sdks 😅 if an SDK tries to e.g. push an event and sends `tenantId` instead of `tenant_id`, don't we lose that param?

sounds insane... but since we're not changing types or order information the names are only used to generate the sdk bindings. this current build is technically using the old sdks which should all pass. if they do then i'll regenerate all the sdk bindings

mrkaye97:
> > I'm a little confused how these changes are backwards compatible with existing sdks 😅 if an SDK tries to e.g. push an event and sends `tenantId` instead of `tenant_id`, don't we lose that param?
> 
> sounds insane... but since we're not changing types or order information the names are only used to generate the sdk bindings. this current build is technically using the old sdks which should all pass. if they do then i'll regenerate all the sdk bindings

interesting, TIL!

sdks/python/hatchet_sdk/contracts/compat.py
```diff
@@ -0,0 +1,198 @@
+from __future__ import annotations
+
+import warnings
+from typing import Iterable
+
+from hatchet_sdk.contracts import dispatcher_pb2, events_pb2, workflows_pb2
+from hatchet_sdk.contracts.v1 import workflows_pb2 as v1_workflows_pb2
+
+
+def _add_aliases(cls: type, aliases: Iterable[tuple[str, str]]) -> None:
```
grutt: @mrkaye97 the idea here is that if someone is somehow using one of these internals types this wont break their code... let me know if you think this isnt needed or if there's a better way of handling this

mrkaye97: I think we don't need it, but we'll just need to comb through all of the methods to make sure nothing is accepting / returning one of generated types. It looks like the event client might be the only one and it's a return, so in that case, we might be able to just write a wrapper pydantic model that matches the old schema and return that

api-contracts/v1/workflows.proto
```diff
@@ -86,11 +86,11 @@ message CreateWorkflowVersionRequest {
     repeated string cron_triggers = 5; // (optional) cron triggers for the workflow
     repeated CreateTaskOpts tasks = 6; // (required) the workflow jobs
 
-    // Deprecated: use concurrency_arr instead
+    // Deprecated: use concurrencyArr instead
```
mrkaye97: is this backwards? I think it should be `concurrency_arr`?

internal/services/admin/server.go
```diff
@@ -100,10 +100,10 @@ func (a *AdminServiceImpl) ScheduleWorkflow(ctx context.Context, req *contracts.
 	isParentTriggered := req.ParentId != nil
 
 	if isParentTriggered {
-		if req.ParentStepRunId == nil {
+		if req.ParentTaskRunId == nil {
```
mrkaye97: nit on naming: in other places we call this `ParentTaskExternalId` for disambiguation, since the `task_id` is technically its own thing (an int). if we're renaming things already, maybe we should clean this up too, it's confusing imo

internal/services/dispatcher/server.go
```diff
@@ -50,8 +50,8 @@ func (s *DispatcherImpl) Register(ctx context.Context, request *contracts.Worker
 		}
 	}
 
-	if request.MaxRuns != nil {
-		mr := int(*request.MaxRuns)
+	if request.Slots != nil {
+		mr := int(*request.Slots)
```
mrkaye97, line 54: 🙌

internal/services/dispatcher/server_v1.go
```diff
@@ -554,16 +554,16 @@ func (s *DispatcherImpl) sendStepActionEventV1(ctx context.Context, request *con
 
 	// if there's no retry count, we need to read it from the task, so we can't skip the cache
 	skipCache := request.RetryCount == nil
-	stepRunId, err := uuid.Parse(request.StepRunId)
+	taskRunId, err := uuid.Parse(request.TaskRunId)
```
mrkaye97: same comment here on naming, e.g. `taskRunExternalId` for disambiguation

sdks/python/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to Hatchet's Python SDK will be documented in this changelog
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.23.0] - 2026-02-02
+
+### Internal Only
+
+- Updated gRPC/REST contract field names to lowerCamelCase for consistency across SDKs.
```
mrkaye97:
```suggestion
- Updated gRPC/REST contract field names to snake_case for consistency across SDKs.
```

sdks/python/hatchet_sdk/clients/events.py
```diff
@@ -60,6 +60,35 @@ class BulkPushEventWithMetadata(BaseModel):
     scope: str | None = None
 
 
+class Event(BaseModel):
+    tenant_id: str
+    event_id: str
+    key: str
+    payload: str
+    event_timestamp: timestamp_pb2.Timestamp
```
mrkaye97, line 68: I think (unfortunately) these are breaking changes. this will need to be `eventTimestamp` (we can use a pydantic alias for this so we don't need multiple fields)

mrkaye97: (same for `additionalMetadata` I think)

mrkaye97: actually, using a property might be the easiest, e.g. 

```python
@property
def eventTimestamp(self) -> timestamp_pb2.Timestamp:
    return self.event_timestamp
```

api-contracts/dispatcher/dispatcher.proto
```diff
@@ -116,43 +116,43 @@ enum ActionType {
...
 
-    // the step id
-    string stepId = 7;
+    // the task id
+    string task_id = 7;
 
-    // the step run id
-    string stepRunId = 8;
+    // the task external run id
+    string task_external_id = 8;
```
abelanger5: Note: for the Go SDK, we should double check whether we're using this since it may break v0

abelanger5: Also: `task_run_external_id` is a more descriptive name

api-contracts/dispatcher/dispatcher.proto
```diff
@@ -116,43 +116,43 @@ enum ActionType {
...
+    ActionType action_type = 10;
 
     // the action payload
-    string actionPayload = 11;
+    string action_payload = 11;
 
-    // the step name
-    string stepName = 12;
+    // the task name
+    string step_name = 12;
```
abelanger5: Assuming we are not breaking anything by renaming `stepRunId` -> `task_external_id`, then should this be `task_name`?

api-contracts/dispatcher/dispatcher.proto
```diff
@@ -338,30 +338,30 @@ enum WorkflowRunEventType {
...
     optional string error = 4;
 
     optional string output = 5;
 }
 
 message OverridesData {
-    // the step run id
-    string stepRunId = 1;
+    // the task run id
+    string step_run_id = 1;
```
abelanger5: Another one: should this be `task_run_external_id`?

api-contracts/dispatcher/dispatcher.proto
```diff
@@ -306,30 +306,30 @@ enum ResourceEventType {
 
 message WorkflowEvent {
     // the id of the workflow run
-    string workflowRunId = 1;
+    string workflow_run_id = 1;
```
abelanger5, line 309: For clarity, `workflow_run_external_id`?

api-contracts/dispatcher/dispatcher.proto
```diff
@@ -370,35 +370,35 @@ message OverridesData {
...
+    google.protobuf.Timestamp heartbeat_at = 2;
 }
 
 message HeartbeatResponse {}
 
 message RefreshTimeoutRequest {
-    // the id of the step run to release
-    string stepRunId = 1;
+    // the id of the task run to release
+    string step_run_id = 1;
```
abelanger5: `task_run_external_id`?

api-contracts/dispatcher/dispatcher.proto
```diff
@@ -370,35 +370,35 @@ message OverridesData {
...
 message RefreshTimeoutResponse {
-    google.protobuf.Timestamp timeoutAt = 1;
+    google.protobuf.Timestamp timeout_at = 1;
 
 }
 message ReleaseSlotRequest {
-    // the id of the step run to release
-    string stepRunId = 1;
+    // the id of the task run to release
+    string step_run_id = 1;
```
abelanger5: same here

api-contracts/dispatcher/dispatcher.proto
abelanger5: ah, ok we can't change this without breaking the v0 Go SDK

sdks/python/hatchet_sdk/clients/events.py
grutt: @matt can you double check these

mrkaye97: should be fine I think

grutt: thank you

internal/services/shared/tasktypes/v1/event.go
```diff
@@ -34,7 +34,7 @@ func NewInternalEventMessage(tenantId uuid.UUID, timestamp time.Time, events ...
 
 type StreamEventPayload struct {
 	WorkflowRunId uuid.UUID `json:"workflow_run_id" validate:"required"`
-	StepRunId     uuid.UUID `json:"step_run_id" validate:"required"`
+	TaskRunId     uuid.UUID `json:"task_run_id" validate:"required"`
```
mnafees, line 37: in the proto files we are referring these as task run "external" IDs so wondering if we should carry on that convention here

pkg/client/event.go
```diff
@@ -41,7 +41,7 @@ type EventClient interface {
 
 	BulkPush(ctx context.Context, payloads []EventWithAdditionalMetadata, options ...BulkPushOpFunc) error
 
-	PutLog(ctx context.Context, stepRunId, msg string, level *string, taskRetryCount *int32) error
+	PutLog(ctx context.Context, taskRunId, msg string, level *string, taskRetryCount *int32) error
```
mnafees, line 44: `taskRunExternalId`?

pkg/client/event.go
```diff
@@ -190,29 +190,29 @@ func (a *eventClientImpl) BulkPush(ctx context.Context, payload []EventWithAddit
 	return nil
 }
 
-func (a *eventClientImpl) PutLog(ctx context.Context, stepRunId, msg string, level *string, taskRetryCount *int32) error {
+func (a *eventClientImpl) PutLog(ctx context.Context, taskRunId, msg string, level *string, taskRetryCount *int32) error {
```
mnafees, line 193: here too?

mnafees (approved):

PR #2914, opened 2026-02-03
wip: durable execution, new schema

abelanger5:
# Description

(currently a WIP)

## Type of change


sql/schema/v1-core.sql
```diff
@@ -2215,3 +2215,84 @@ CREATE TABLE v1_event_to_run (
...
+    -- sequence value generated from v1_durable_event_log_file.latest_entry_id
+    entry_id BIGINT NOT NULL,
+    -- Todo: Associated data for this event should be stored in the v1_payload table!
+    -- data JSONB,
+    -- The hash of the data stored in the v1_payload table to check non-determinism violations.
+    -- This can be null for event types that don't have associated data.
+    -- TODO: we can add CHECK CONSTRAINT for event types that require data_hash to be non-null.
+    data_hash BYTEA,
+    -- Can discuss: adds some flexibility for future hash algorithms
+    data_hash_alg TEXT,
```
mnafees, line 2266: Can this be an enum instead of `TEXT`?

sql/schema/v1-core.sql
```diff
@@ -2215,3 +2215,84 @@ CREATE TABLE v1_event_to_run (
...
     PRIMARY KEY (event_id, event_seen_at, run_external_id)
 ) PARTITION BY RANGE(event_seen_at);
+
+-- v1_durable_event_log represents the log file for the durable event history
+-- of a durable task. This table stores metadata like sequence values for entries.
+--
+-- Important: writers to v1_durable_event_log_entry should lock this row to increment the sequence value.
+CREATE TABLE v1_durable_event_log_file (
+    -- The id and inserted_at of the durable task which created this entry
+    durable_task_id BIGINT NOT NULL,
```
mnafees, line 2225: Are we thinking of having a separate `v1_durable_task` table given the naming used here?

abelanger5: no, this was just to make it more clear that this is the durable task and doesn't refer to any child tasks which may have been triggered in the event history

sql/schema/v1-core.sql
abelanger5: yeah, I think this is something we need to discuss - I'm not convinced that this hashing approach is the correct one to detect non-determinism, it's purely an idea right now. I agree that if we decide to take this approach we should use an enum

api-contracts/v1/dispatcher.proto
```diff
@@ -7,13 +7,85 @@ package v1;
...
+}
+
+enum DurableTaskTriggerKind {
+    DURABLE_TASK_TRIGGER_KIND_UNSPECIFIED = 0;
+    DURABLE_TASK_TRIGGER_KIND_RUN = 1;
+    DURABLE_TASK_TRIGGER_KIND_WAIT_FOR = 2;
+    DURABLE_TASK_TRIGGER_KIND_MEMO = 3;
+}
 
+message DurableTaskRequestTrigger {
```
mrkaye97, line 32: is the idea here that triggering children out of a durable task will be entirely separate from the normal trigger path?

api-contracts/v1/dispatcher.proto
```diff
@@ -7,13 +7,85 @@ package v1;
...
+    // The session_id is used to identify the worker session for this durable task, since
+    // there can be many sessions for the same durable task (e.g. retries)
+    int64 session_id = 1;
+    string durable_task_external_id = 2;
+    DurableTaskTriggerKind kind = 3;
+    optional bytes payload = 4;
+    optional DurableEventListenerConditions wait_for_conditions = 5;
+}
+
+message DurableTaskResponseTriggerAck {
```
mrkaye97, line 42: I think this name is a little confusing 😅 maybe I'm just not following, but I don't really get what it's supposed to represent

PR #2926, opened 2026-02-03
Fix: Concurrency rule violation on replays

mrkaye97:
# Description

Fixes a bug where concurrency isn't respected on a replay because of the ordering by priority first by first checking the `is_filled` in the lateral join where we find candidates to assign

## Type of change


mnafees (approved):

pkg/repository/sqlcv1/concurrency.sql
```diff
@@ -164,7 +164,7 @@ WITH eligible_slots_per_group AS (
             wcs_all.key = distinct_keys.key
             AND wcs_all.tenant_id = @tenantId::uuid
             AND wcs_all.strategy_id = @strategyId::bigint
-        ORDER BY wcs_all.priority DESC, wcs_all.sort_id ASC
+        ORDER BY wcs_all.is_filled DESC, wcs_all.priority DESC, wcs_all.sort_id ASC
```
abelanger5, line 167: I think we'll need to modify `v1_workflow_concurrency_slot_query_idx` for this to stay performant (though would be useful that this index is actually being used on large instances)

abelanger5 (changes requested):

pkg/repository/sqlcv1/concurrency.sql
mrkaye97: alright I pushed a change here, let me know if that's what you were thinking! I think we need one index for each order? maybe not though, idk

cmd/hatchet-migrate/migrate/migrations/20260317195632_v1_0_90.sql
```diff
@@ -0,0 +1,10 @@
+-- +goose Up
+-- +goose NO TRANSACTION
+CREATE INDEX CONCURRENTLY IF NOT EXISTS v1_workflow_concurrency_slot_cancel_newest_query_idx ON v1_workflow_concurrency_slot (tenant_id, strategy_id ASC, key ASC, priority DESC, is_filled DESC, sort_id ASC);
+CREATE INDEX CONCURRENTLY IF NOT EXISTS v1_workflow_concurrency_slot_cancel_in_progress_query_idx ON v1_workflow_concurrency_slot (tenant_id, strategy_id ASC, key ASC, priority DESC, is_filled ASC, sort_id DESC);
```
abelanger5, line 4: what about group round robin?

note that cancel newest and cancel in progress are expected to seq scan anyway (in the current implementation) and I don't think these indexes are going to change that, since we're doing a large join across the temp table (the only useful part of the index is `(tenant_id, strategy_id)` but we should be able to reuse the other index for that

pkg/repository/sqlcv1/concurrency.sql
abelanger5: left a comment in the migration file - I'm still not convinced?

cmd/hatchet-migrate/migrate/migrations/20260317195632_v1_0_90.sql
mrkaye97: so for GRR, seems like maybe we need `(tenant, strategy, key, is_filled desc, priority desc, sort_id asc)`, but also (oddly?) it looks like `RunParentGroupRoundRobin` orders by the `priority`, while `RunGroupRoundRobin` does not - not sure if that's intentional?

PR #2927, opened 2026-02-03
feat:  multiple slot types

grutt:
## Description

This change replaces the old durable slot model—which required two concurrent worker processes—with a scheduler‑driven approach. This change allows for multiple worker lot config rows per worker which represent different "slot pools". Initially these are "default" and "durable".

The feature is designed to extend to multiple units per task and multiple types of slots in the near future with no new schema changes.

Depends on #2913 

## Type of change

## What's Changed

- [x] Add slot type + slot capacity support across scheduler, worker APIs, and UI.
- [x] Add DB schema updates and migrations for slot types/defaults.
- [x] Update Go/Python/TS SDKs and examples.
- [x] Update worker docs and SDK docs to reflect slot configuration changes.


pkg/repository/scheduler_queue.go
```diff
@@ -639,6 +639,29 @@ func (d *queueRepository) GetDesiredLabels(ctx context.Context, tx *OptimisticTx
 	return stepIdToLabels, nil
 }
 
+func (d *queueRepository) GetStepSlotRequirements(ctx context.Context, stepIds []uuid.UUID) (map[uuid.UUID]map[string]int32, error) {
```
grutt: nit naming

pkg/repository/sqlcv1/tenant_limits.sql
```diff
@@ -97,7 +97,9 @@ AND "lastHeartbeatAt" >= NOW() - '30 seconds'::INTERVAL
 AND "isActive" = true;
 
 -- name: CountTenantWorkerSlots :one
```
grutt: are we still using this query?

pkg/repository/sqlcv1/tenants.sql
```diff
@@ -665,7 +666,7 @@ WITH active_workers AS (
 ), worker_slots AS (
     SELECT
         aw."id" AS worker_id,
-        aw."maxRuns" - (
+        (aw."maxRuns" + aw."durableMaxRuns") - (
```
grutt: probably wrong

pkg/repository/sqlcv1/workers.sql
```diff
@@ -327,6 +416,7 @@ INSERT INTO "Worker" (
     "name",
     "dispatcherId",
     "maxRuns",
```
grutt: drop unused cols

pkg/repository/worker.go
```diff
@@ -305,10 +357,24 @@ func (w *workerRepository) CreateNewWorker(ctx context.Context, tenantId uuid.UU
 		return nil, err
 	}
 
-	maxRuns := int32(100)
+	slotCapacities := opts.SlotCapacities
+	maxRuns := int32(0)
```
grutt: drop maxRun

pkg/scheduling/v1/queuer.go
```diff
@@ -207,7 +207,21 @@ func (q *Queuer) loopQueue(ctx context.Context) {
 		desiredLabelsTime := time.Since(checkpoint)
 		checkpoint = time.Now()
 
-		assignCh := q.s.tryAssign(ctx, qis, labels, rls)
+		stepRequirements, err := q.repo.GetStepSlotRequirements(ctx, stepIds)
```
grutt: naming

pkg/scheduling/v1/slot_test.go
```diff
@@ -36,8 +36,8 @@ func TestGetRankedSlots(t *testing.T) {
 				DesiredWorkerID: &stableWorkerId1,
 			},
 			slots: []*slot{
-				newSlot(&worker{ListActiveWorkersResult: &v1.ListActiveWorkersResult{ID: stableWorkerId1}}, []string{}),
-				newSlot(&worker{ListActiveWorkersResult: &v1.ListActiveWorkersResult{ID: uuid.New()}}, []string{}),
+				newSlot(&worker{ListActiveWorkersResult: &v1.ListActiveWorkersResult{ID: stableWorkerId1}}, []string{}, "default"),
+				newSlot(&worker{ListActiveWorkersResult: &v1.ListActiveWorkersResult{ID: uuid.New()}}, []string{}, "default"),
```
grutt: add test for multiple types

pkg/v1/worker/worker.go
```diff
@@ -182,38 +179,7 @@ func (w *WorkerImpl) RegisterWorkflows(workflows ...workflow.WorkflowBase) error
...
-			)
-			if err != nil {
-				return err
-			}
-			w.durableWorker = durableWorker
-		}
-
-		// Register workflow with non-durable worker if it exists
+		// Register workflow with worker if it exists
 		if w.nonDurableWorker != nil {
```
grutt: clean this up further

pkg/worker/worker.go
```diff
@@ -178,6 +182,19 @@ func WithSlots(slots int) WorkerOpt {
 	}
 }
 
+func WithDurableSlots(durableSlots int) WorkerOpt {
+	return func(opts *WorkerOpts) {
+		opts.durableSlots = &durableSlots
+	}
+}
+
+// WithSlotCapacities sets slot capacities for this worker (slot_type -> units).
```
grutt: nit commet

sdks/python/hatchet_sdk/clients/rest/models/worker.py
```diff
@@ -74,11 +74,26 @@ class Worker(BaseModel):
...
         default=None,
         description="The number of runs this worker can execute concurrently.",
         alias="availableRuns",
     )
+    durable_max_runs: Optional[StrictInt] = Field(
+        default=None,
+        description="The maximum number of durable runs this worker can execute concurrently.",
+        alias="durableMaxRuns",
+    )
+    durable_available_runs: Optional[StrictInt] = Field(
```
grutt: remove

sdks/python/hatchet_sdk/runnables/task.py
```diff
@@ -377,21 +381,27 @@ def to_proto(self, service_name: str) -> CreateTaskOpts:
...
-            retries=self.retries,
-            rate_limits=self.rate_limits,
-            worker_labels=self.desired_worker_labels,
-            backoff_factor=self.backoff_factor,
-            backoff_max_seconds=self.backoff_max_seconds,
-            concurrency=[t.to_proto() for t in concurrency],
-            conditions=self._conditions_to_proto(),
-            schedule_timeout=timedelta_to_expr(self.schedule_timeout),
-        )
+        opts: dict[str, Any] = {
```
grutt: type saftey?

sdks/typescript/src/v1/client/worker/worker-internal.ts
```diff
@@ -126,12 +135,14 @@ export class V1Worker {
     );
   }
 
+  // TODO where is this used, this doesnt make much sense
```
grutt: check

sdks/typescript/src/v1/client/worker/worker.ts
```diff
@@ -180,39 +171,116 @@ export class Worker {
...
 
   // TODO docstrings
   pause() {
-    const promises: Promise<any>[] = [];
-    if (this.nonDurable?.workerId) {
-      promises.push(this._v1.workers.pause(this.nonDurable.workerId));
-    }
-    if (this.durable?.workerId) {
-      promises.push(this._v1.workers.pause(this.durable.workerId));
+    if (!this.nonDurable?.workerId) {
```
grutt: cleanup

sdks/typescript/src/v1/client/worker/worker.ts
```diff
@@ -180,39 +171,116 @@ export class Worker {
...
+      return Promise.resolve();
     }
-    if (this.durable?.workerId) {
-      promises.push(this._v1.workers.unpause(this.durable.workerId));
+
+    return this._v1.workers.unpause(this.nonDurable.workerId);
+  }
+}
+
+function resolveWorkerOptions(options: CreateWorkerOpts) {
```
grutt: move to slot utils

sdks/typescript/src/v1/examples/simple/worker.ts
```diff
@@ -9,7 +9,8 @@ async function main() {
     // 👀 Declare the workflows that the worker can execute
     workflows: [simple, simpleWithZod, parent, child],
     // 👀 Declare the number of concurrent task runs the worker can accept
-    slots: 100,
+    // slots: 100,
```
grutt: clean up

sdks/typescript/src/v1/examples/simple/workflow.ts
```diff
@@ -6,7 +6,7 @@ export type SimpleInput = {
   Message: string;
 };
 
-export const simple = hatchet.task({
+export const simple = hatchet.durableTask({
```
grutt: revert

sql/schema/v0.sql
```diff
@@ -851,6 +852,7 @@ CREATE TABLE "Worker" (
     "name" TEXT NOT NULL,
     "dispatcherId" UUID,
     "maxRuns" INTEGER NOT NULL DEFAULT 100,
```
grutt, line 855: drop

frontend/docs/pages/sdks/python/client.mdx
```diff
@@ -73,14 +73,15 @@ Create a Hatchet worker on which to run workflows.
...
+| --------------- | --------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------- | --- |
+| `name`          | `str`                             | The name of the worker.                                                                                                                                    | _required_        |
+| `slots`         | `int`                             | Maximum number of concurrent runs (mapped to `slot_config.default`).                                                                                       | `100`             |
+| `durable_slots` | `int`                             | Maximum number of concurrent durable tasks (mapped to `slot_config.durable`).                                                                              | `1000`            |
+| `slot_config`   | `dict[str, int]`                  | Slot capacities for the worker (slot_type -> units). Defaults to `{"default": 100}`.                                                                       | `{"default":100}` |
```
grutt: we might not want to expose this right now

pkg/scheduling/v1/queuer.go
```diff
@@ -207,7 +207,21 @@ func (q *Queuer) loopQueue(ctx context.Context) {
 		desiredLabelsTime := time.Since(checkpoint)
 		checkpoint = time.Now()
 
-		assignCh := q.s.tryAssign(ctx, qis, labels, rls)
+		stepRequests, err := q.repo.GetStepSlotRequests(ctx, stepIds)
```
abelanger5: Need a cache in this query, assuming this doesn't change often and is tied to the step run id

pkg/scheduling/v1/queuer.go
```diff
@@ -594,7 +610,12 @@ func (q *Queuer) runOptimisticQueue(
 		return nil, nil, err
 	}
 
-	assignCh := q.s.tryAssign(ctx, qis, labels, rls)
+	stepRequests, err := q.repo.GetStepSlotRequests(ctx, stepIds)
```
abelanger5: This deadlocks!

pkg/scheduling/v1/scheduler.go
```diff
@@ -81,13 +81,45 @@ func (s *Scheduler) nack(ids []int) {
...
+}
+
+func (a *assignedSlots) nack() {
+	for _, slot := range a.slots {
+		slot.nack()
+	}
+	if a.rateLimitNack != nil {
+		a.rateLimitNack()
+	}
+}
```
abelanger5: nit: can we move this into `slot.go`?

pkg/scheduling/v1/scheduler.go
```diff
@@ -489,6 +606,12 @@ func (s *Scheduler) tryAssignBatch(
 		}
 	}
 
+	// Get or create the action and try to assign the batch of queue items.
+	// NOTE: if we change the position of these locks, make sure that we are still acquiring locks in the same
+	// order as the replenish() function, otherwise we may deadlock. The order is:
+	// actionsMu -> action.mu -> unackedMu
+	action := s.ensureAction(actionId)
```
abelanger5: what's the behavior if we instantiate a new action here? it seems like it's created in an empty state (no `slotsByWorker`), so when would this get populated?

pkg/scheduling/v1/scheduler.go
```diff
@@ -603,36 +723,127 @@ func (s *Scheduler) tryAssignBatch(
...
+) *assignedSlots {
+	for _, candidateSlot := range candidateSlots {
+		if !candidateSlot.active() {
+			continue
+		}
 
-	for _, slot := range candidateSlots {
-		if !slot.active() {
+		workerId := candidateSlot.getWorkerId()
+		workerSlots := action.slotsByWorker[workerId]
```
abelanger5: this screams `concurrent map read and map write` since this map isn't guarded by a mutex. this should definitely be a helper on the `action` with appropriate mutex protection

abelanger5: reading more closely, it looks like we most likely have a lock here already on the action?

pkg/scheduling/v1/queuer.go
abelanger5, line 1: Possible I'm missing something, but I'm not seeing where we're using the new `rateLimitAck` methods on the batch of slots. I'm only seeing `q.s.ack` and `q.s.nack` still.

pkg/scheduling/v1/scheduler.go
```diff
@@ -263,59 +320,105 @@ func (s *Scheduler) replenish(ctx context.Context, mustReplenish bool) error {
...
+
+		if len(workerUUIDs) == 0 {
+			continue
+		}
+
+		availableSlots, err := s.repo.ListAvailableSlotsForWorkers(ctx, s.tenantId, sqlcv1.ListAvailableSlotsForWorkersParams{
+			Tenantid:  s.tenantId,
+			Workerids: workerUUIDs,
+			Slottype:  slotType,
+		})
```
abelanger5: Why is this not just a single query instead of this `n+1` pattern?

pkg/scheduling/v1/scheduler.go
```diff
@@ -330,6 +433,7 @@ func (s *Scheduler) replenish(ctx context.Context, mustReplenish bool) error {
 		// we overwrite the slots for the action. we know that the action is in the map because we checked
 		// for it in the first pass.
 		s.actions[actionId].slots = newSlots
+		s.actions[actionId].slotsByWorker = actionsToSlotsByWorker[actionId]
```
abelanger5: I think it's worth checking that we have a lock on this action (or it's being only copied and instantiated here)

grutt: confirmed

pkg/scheduling/v1/scheduler.go
grutt: its populated in replenish(), i think a signal that bumps replenish early might be a good idea

pkg/scheduling/v1/queuer.go
grutt: queuer.go still calls q.s.ack(ackIds) / q.s.nack(nackIds) after the DB flush

these fn on the scheduler call assigned.ack()

```go
func (s *Scheduler) nack(ids []int) {
	s.unackedMu.Lock()
	defer s.unackedMu.Unlock()

	for _, id := range ids {
		if assigned, ok := s.unackedSlots[id]; ok {
			assigned.nack()
			delete(s.unackedSlots, id)
		}
	}
}
```

which handle both acks (for all slots)
```go
func (a *assignedSlots) ack() {
	for _, slot := range a.slots {
		slot.ack()
	}
	if a.rateLimitAck != nil {
		a.rateLimitAck()
	}
}
```

pkg/scheduling/v1/scheduler.go
grutt: fixed

pkg/scheduling/v1/scheduler.go
grutt: added some comments for clarity

internal/services/admin/v1/server.go
```diff
@@ -863,6 +863,12 @@ func getCreateTaskOpts(tasks []*contracts.CreateTaskOpts, kind string) ([]v1.Cre
 			TriggerConditions:   make([]v1.CreateStepMatchConditionOpt, 0),
 			RateLimits:          make([]v1.CreateWorkflowStepRateLimitOpts, 0), // Initialize to avoid nil
 			ScheduleTimeout:     stepCp.ScheduleTimeout,
+			IsDurable:           stepCp.IsDurable,
+			SlotRequests:        nil,
+		}
+
+		if stepCp.SlotRequests != nil {
```
grutt, line 918: default nil for new opts on non-durable tasks on sdk

internal/services/dispatcher/server.go
```diff
@@ -52,7 +52,16 @@ func (s *DispatcherImpl) Register(ctx context.Context, request *contracts.Worker
 
 	if request.Slots != nil {
 		mr := int(*request.Slots)
-		opts.MaxRuns = &mr
+		opts.Slots = &mr
+	}
+
+	if request.DurableSlots != nil {
```
grutt: rm

pkg/client/dispatcher.go
```diff
@@ -42,12 +42,14 @@ const (
...
-	Services   []string
-	Actions    []string
-	Slots      *int
-	Labels     map[string]interface{}
-	WebhookId  *string
+	WorkerName     string
+	Services       []string
+	Actions        []string
+	Slots          *int
+	DurableSlots   *int
```
grutt: no durable slots

pkg/repository/scheduler.go
```diff
@@ -36,12 +37,15 @@ type QueueRepository interface {
 	GetTaskRateLimits(ctx context.Context, tx *OptimisticTx, queueItems []*sqlcv1.V1QueueItem) (map[int64]map[string]int32, error)
 	RequeueRateLimitedItems(ctx context.Context, tenantId uuid.UUID, queueName string) ([]*sqlcv1.RequeueRateLimitedQueueItemsRow, error)
 	GetDesiredLabels(ctx context.Context, tx *OptimisticTx, stepIds []uuid.UUID) (map[uuid.UUID][]*sqlcv1.GetDesiredLabelsRow, error)
+	GetStepSlotRequests(ctx context.Context, tx *OptimisticTx, stepIds []uuid.UUID) (map[uuid.UUID]map[string]int32, error)
```
grutt, line 42: type in db

pkg/repository/scheduler.go
```diff
@@ -36,12 +37,15 @@ type QueueRepository interface {
...
 	Cleanup()
 }
 
 type AssignmentRepository interface {
 	ListActionsForWorkers(ctx context.Context, tenantId uuid.UUID, workerIds []uuid.UUID) ([]*sqlcv1.ListActionsForWorkersRow, error)
 	ListAvailableSlotsForWorkers(ctx context.Context, tenantId uuid.UUID, params sqlcv1.ListAvailableSlotsForWorkersParams) ([]*sqlcv1.ListAvailableSlotsForWorkersRow, error)
```
grutt, line 48: rm if unused

pkg/repository/sqlcv1/lease.sql
```diff
@@ -50,11 +50,12 @@ RETURNING l.*;
 
 -- name: ListActiveWorkers :many
 SELECT
-    w."id",
-    w."maxRuns",
+    DISTINCT w."id",
     w."name"
 FROM
     "Worker" w
+JOIN
```
grutt, line 57: go migration to iterate and insert configs

grutt: trigger to insert into this table on start of migration

grutt: add tenant filter

pkg/repository/sqlcv1/queue.sql
```diff
@@ -51,35 +51,72 @@ WHERE
     AND w."isPaused" = false;
 
 -- name: ListAvailableSlotsForWorkers :many
```
grutt: move to worker repository

pkg/repository/sqlcv1/queue.sql
```diff
@@ -51,35 +51,72 @@ WHERE
...
+-- name: ListAvailableSlotsForWorkersAndTypes :many
+WITH worker_capacities AS (
+    SELECT
+        worker_id,
+        slot_type,
+        max_units
+    FROM
+        v1_worker_slot_config
+    WHERE
+        tenant_id = @tenantId::uuid
```
grutt: tuple where (tenant_id, ....)

pkg/repository/sqlcv1/queue.sql
```diff
@@ -51,35 +51,72 @@ WHERE
...
+        AND worker_id = ANY(@workerIds::uuid[])
+        AND slot_type = ANY(@slotTypes::text[])
```
grutt: we dont need slot type filter

pkg/repository/sqlcv1/queue.sql
```diff
@@ -51,35 +51,72 @@ WHERE
...
+), worker_used_slots AS (
+    SELECT
+        worker_id,
+        slot_type,
+        SUM(units) AS used_units
+    FROM
+        v1_task_runtime_slot
+    WHERE
```
grutt: join on previous cte

pkg/repository/sqlcv1/queue.sql
```diff
@@ -280,6 +354,16 @@ FROM
...
 
+-- name: GetStepSlotRequests :many
+SELECT
+    step_id,
+    slot_type,
+    units
+FROM
+    v1_step_slot_request
+WHERE
+    step_id = ANY(@stepIds::uuid[]);
```
grutt: tenant id

pkg/repository/sqlcv1/tasks-overwrite.go
```diff
@@ -729,6 +729,13 @@ WITH input AS (
     ORDER BY
         task_id, task_inserted_at, retry_count
     FOR UPDATE
+), deleted_slots AS (
+    DELETE FROM
+        v1_task_runtime_slot
+    WHERE
+        (task_id, task_inserted_at, retry_count) IN (SELECT task_id, task_inserted_at, retry_count FROM input)
+    -- return a constant for ordering
+    RETURNING 1 AS cte_order
```
grutt: remove

grutt: from both

pkg/repository/sqlcv1/tasks.sql
```diff
@@ -912,6 +912,11 @@ WITH task AS (
     ORDER BY
         task_id, task_inserted_at, retry_count
     FOR UPDATE
+), deleted_slots AS (
+    DELETE FROM v1_task_runtime_slot
```
grutt, line 916: tripple check this is everywhere we delete

pkg/repository/sqlcv1/workers.sql
```diff
@@ -10,18 +10,77 @@ SELECT
...
+    slot_type,
+    max_units
+FROM
+    v1_worker_slot_config
+WHERE
+    tenant_id = @tenantId::uuid
+    AND worker_id = ANY(@workerIds::uuid[]);
+
+-- name: UpsertWorkerSlotConfigs :exec
+INSERT INTO v1_worker_slot_config (
```
grutt, line 25: why did i make this an upsert

pkg/repository/sqlcv1/workers.sql
```diff
@@ -10,18 +10,77 @@ SELECT
...
+    slot_type,
+    max_units,
+    created_at,
+    updated_at
+)
+SELECT
+    @tenantId::uuid,
+    @workerId::uuid,
+    unnest(@slotTypes::text[]),
+    unnest(@maxUnits::integer[]),
```
grutt, line 37: double check unnest

grutt: or cte

pkg/repository/sqlcv1/workers.sql
```diff
@@ -10,18 +10,77 @@ SELECT
...
+
 -- name: ListWorkersWithSlotCount :many
 SELECT
     sqlc.embed(workers),
     ww."url" AS "webhookUrl",
     ww."id" AS "webhookId",
-    workers."maxRuns" - (
-        SELECT COUNT(*)
-        FROM v1_task_runtime runtime
+    -- TODO do we still need this?
```
grutt: address unnest

pkg/repository/sqlcv1/workers.sql
```diff
@@ -57,13 +119,39 @@ GROUP BY
 SELECT
```
grutt, line 149: where are we using this query

pkg/repository/sqlcv1/workers.sql
```diff
@@ -10,18 +10,77 @@ SELECT
...
-        SELECT COUNT(*)
-        FROM v1_task_runtime runtime
+    -- TODO do we still need this?
+    COALESCE((
+        SELECT COALESCE(cap.max_units, 0)
+        FROM v1_worker_slot_config cap
+        WHERE
+            cap.tenant_id = workers."tenantId"
+            AND cap.worker_id = workers."id"
+            AND cap.slot_type = 'default'::text
```
grutt: hard coded

pkg/repository/sqlcv1/workflows.sql
```diff
@@ -301,9 +302,31 @@ INSERT INTO "Step" (
...
+    step_id,
+    slot_type,
+    units,
+    created_at,
+    updated_at
+)
+SELECT
+    @tenantId::uuid,
+    @stepId::uuid,
+    unnest(@slotTypes::text[]),
```
grutt, line 321: check unnest

pkg/repository/sqlcv1/workflows.sql
```diff
@@ -301,9 +302,31 @@ INSERT INTO "Step" (
     coalesce(sqlc.narg('retries')::integer, 0),
     coalesce(sqlc.narg('scheduleTimeout')::text, '5m'),
     sqlc.narg('retryBackoffFactor'),
-    sqlc.narg('retryMaxBackoff')
+    sqlc.narg('retryMaxBackoff'),
+    coalesce(sqlc.narg('isDurable')::boolean, false)
 ) RETURNING *;
 
+-- name: CreateStepSlotRequests :exec
+INSERT INTO v1_step_slot_request (
```
grutt, line 310: step migration, take out exclusive lock on "Step"

pkg/repository/worker.go
```diff
@@ -28,7 +28,13 @@ type CreateWorkerOpts struct {
 	DispatcherId uuid.UUID `validate:"required"`
 
 	// The maximum number of runs this worker can run at a time
-	MaxRuns *int `validate:"omitempty,gte=1"`
+	Slots *int `validate:"omitempty,gte=1"`
+
+	// The maximum number of durable runs this worker can run at a time
+	DurableSlots *int `validate:"omitempty,gte=0"`
```
grutt: rm

pkg/repository/worker.go
```diff
@@ -28,7 +28,13 @@ type CreateWorkerOpts struct {
 	DispatcherId uuid.UUID `validate:"required"`
 
 	// The maximum number of runs this worker can run at a time
-	MaxRuns *int `validate:"omitempty,gte=1"`
+	Slots *int `validate:"omitempty,gte=1"`
```
grutt: rm

pkg/repository/worker.go
```diff
@@ -238,6 +222,37 @@ func (w *workerRepository) ListActiveSDKsPerTenant() (map[TenantIdSDKTuple]int64
 	return tenantIdSDKTupleToCount, nil
 }
 
+func (w *workerRepository) ListTotalActiveSlotsPerTenant() (map[uuid.UUID]int64, error) {
+	rows, err := w.queries.ListTotalActiveSlotsPerTenant(context.Background(), w.pool)
```
grutt: ctx.Background()?

pkg/repository/worker.go
```diff
@@ -391,11 +448,45 @@ func (w *workerRepository) CreateNewWorker(ctx context.Context, tenantId uuid.UU
 		}
 	}
 
-	if worker == nil {
-		worker, err = w.queries.CreateWorker(ctx, tx, createParams)
+	worker, err = w.queries.CreateWorker(ctx, tx, createParams)
```
grutt: assign here rm above

pkg/repository/workflow.go
```diff
@@ -727,6 +733,10 @@ func (r *workflowRepository) createJobTx(ctx context.Context, tx sqlcv1.DBTX, te
 			Readableid:     stepOpts.ReadableId,
 			CustomUserData: customUserData,
 			Retries:        retries,
+			IsDurable: pgtype.Bool{
```
grutt: helper?

pkg/scheduling/v1/action.go
```diff
@@ -31,10 +34,18 @@ func (a *action) activeCount() int {
 	return count
 }
 
+// testHookBeforeOrderedLock exists to make lock-order assertions deterministic in
+// unit tests. It is nil in production.
+var testHookBeforeOrderedLock func(actions []*action)
```
grutt: https://github.com/sasha-s/go-deadlock

for testing only

pkg/scheduling/v1/scheduler.go
```diff
@@ -29,6 +29,8 @@ type Scheduler struct {
 	actions     map[string]*action
 	actionsMu   rwMutex
 	replenishMu mutex
+	// replenishSignal is poked to trigger an immediate replenish. It's buffered to drain signal bursts while we're already awake.
+	replenishSignal chan struct{}
```
grutt: probably remove

pkg/scheduling/v1/scheduler.go
```diff
@@ -110,6 +113,30 @@ func (s *Scheduler) getWorkers() map[uuid.UUID]*worker {
 	return s.workers
 }
 
+func (s *Scheduler) ensureAction(actionId string) *action {
```
grutt: consider removing

grutt: no writes

pkg/scheduling/v1/scheduler.go
```diff
@@ -603,41 +745,144 @@ func (s *Scheduler) tryAssignBatch(
...
+	for slotType, units := range requests {
+		if units <= 0 {
 			continue
 		}
+		normalized[slotType] = units
+	}
 
-		assignedSlot = slot
-		break
+	if len(normalized) == 0 {
```
grutt: add comment

pkg/scheduling/v1/scheduler.go
```diff
@@ -603,41 +745,144 @@ func (s *Scheduler) tryAssignBatch(
...
+	return selected, true
+}
+
+func normalizeSlotRequests(requests map[string]int32) map[string]int32 {
+	if len(requests) == 0 {
+		return map[string]int32{v1.SlotTypeDefault: 1}
+	}
+
+	normalized := make(map[string]int32, len(requests))
+	for slotType, units := range requests {
```
grutt: not needed

pkg/scheduling/v1/scheduler.go
```diff
@@ -603,41 +745,144 @@ func (s *Scheduler) tryAssignBatch(
...
+	// function. We read from action.slotsByWorker, which is rebuilt during
+	// replenish under action.mu.
+
+	for _, candidateSlot := range candidateSlots {
+		if !candidateSlot.active() {
+			continue
+		}
+
+		workerId := candidateSlot.getWorkerId()
+		workerSlots := action.slotsByWorker[workerId]
```
grutt: iterator instead

pkg/scheduling/v1/scheduler.go
```diff
@@ -670,7 +915,7 @@ func (s *Scheduler) tryAssignSingleton(
 	s.unackedSlots[res.ackId] = assignedSlot
 	s.unackedMu.Unlock()
 
-	res.workerId = assignedSlot.getWorkerId()
+	res.workerId = assignedSlot.workerId()
```
grutt, line 951: check nil

pkg/repository/sqlcv1/workers.sql
```diff
@@ -57,13 +119,39 @@ GROUP BY
...
+        FROM v1_task_runtime_slot runtime
         WHERE
             runtime.tenant_id = w."tenantId" AND
-            runtime.worker_id = w."id"
-    ) AS "remainingSlots"
+            runtime.worker_id = w."id" AND
+            runtime.slot_type = 'default'::text
+    ), 0)::int AS "remainingSlots"
+    ,
+    COALESCE((
```
grutt: look at these

pkg/repository/sqlcv1/workers.sql
grutt: list end points

cmd/hatchet-migrate/migrate/migrations/20260203170919_v1_0_76.sql
```diff
@@ -0,0 +1,110 @@
...
+-- -- +goose NO TRANSACTION
+CREATE INDEX CONCURRENTLY IF NOT EXISTS v1_task_runtime_slot_tenant_worker_type_idx
+    ON v1_task_runtime_slot (tenant_id ASC, worker_id ASC, slot_type ASC);
+
+CREATE INDEX CONCURRENTLY IF NOT EXISTS v1_step_slot_request_step_idx
+    ON v1_step_slot_request (step_id ASC);
+
+-- +goose StatementBegin
+-- TODO acquire lock
+INSERT INTO v1_worker_slot_config (tenant_id, worker_id, slot_type, max_units)
```
grutt: TODO think through data migration

pkg/scheduling/v1/scheduler.go
grutt: i don't think an iterator is the best choice here but i can be convinced otherwise. instead i've made some changes to minimize overhead

cmd/hatchet-migrate/migrate/migrations/20260211000004_v1_0_76_d_cleanup_and_old_columns.sql
```diff
@@ -0,0 +1,138 @@
+-- TODO: move this to a separate PR... this must be run after old version is destroyed and new version is deployed
```
grutt: We should really clean/formalize up this pattern....

api/v1/server/handlers/workers/slot_config.go
```diff
@@ -0,0 +1,65 @@
...
+	for _, config := range slotConfigByWorker {
+		for slotType := range config {
+			slotTypes[slotType] = struct{}{}
+		}
+	}
+
+	slotTypesArr := make([]string, 0, len(slotTypes))
+	for slotType := range slotTypes {
+		slotTypesArr = append(slotTypesArr, slotType)
+	}
```
mrkaye97: nit: can we do this in one loop?

cmd/hatchet-migrate/migrate/migrations/20260203170921_v1_0_76_c_backfill_slots.go
```diff
@@ -0,0 +1,226 @@
...
+	"github.com/pressly/goose/v3"
+)
+
+func init() {
+	goose.AddMigrationNoTxContext(up20260203170921, down20260203170921)
+}
+
+const (
+	backfillSlotsBatchSize = 10_000
+	zeroUUID               = "00000000-0000-0000-0000-000000000000"
```
mrkaye97: `uuid.Nil` here?

sql/schema/v1-core.sql
```diff
@@ -432,6 +432,48 @@ alter table v1_task_runtime set (
...
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    PRIMARY KEY (tenant_id, step_id, slot_type)
+);
+
+CREATE INDEX v1_step_slot_request_step_idx
+    ON v1_step_slot_request (step_id ASC);
+
+-- v1_task_runtime_slot stores runtime slot consumption per task.
+CREATE TABLE v1_task_runtime_slot (
```
mrkaye97, line 461: I'm a little confused by this one - what's the reason we don't put the slot info on the `v1_task_runtime` directly? is there a 1:1 mapping between this table and `v1_task_runtime`

cmd/hatchet-migrate/migrate/migrations/20260203170921_v1_0_76_c_backfill_slots.go
```diff
@@ -0,0 +1,226 @@
...
+		err := db.QueryRowContext(ctx, `
+WITH batch AS (
+	SELECT
+		"tenantId" AS tenant_id,
+		"id" AS worker_id,
+		"maxRuns" AS max_units
+	FROM "Worker"
+	WHERE "maxRuns" IS NOT NULL
+	  AND ("tenantId", "id") > ($1::uuid, $2::uuid)
+	ORDER BY "tenantId", "id"
```
mrkaye97: I think we might want to order by just the `id` here unless I'm missing something, the `tenantId` isn't part of the PK

mrkaye97: actually we definitely need this anyways (and also to change the where clause) because this will break if a new worker comes online for an already-processed tenant while this is running afaict

cmd/hatchet-migrate/migrate/migrations/20260203170921_v1_0_76_c_backfill_slots.go
```diff
@@ -0,0 +1,226 @@
...
+		worker_id,
+		'default'::text,
+		max_units
+	FROM batch
+	ON CONFLICT (tenant_id, worker_id, slot_type) DO NOTHING
+)
+SELECT
+	(SELECT COUNT(*) FROM batch) AS n,
+	(SELECT tenant_id::text FROM batch ORDER BY tenant_id DESC, worker_id DESC LIMIT 1) AS last_tenant_id,
+	(SELECT worker_id::text FROM batch ORDER BY tenant_id DESC, worker_id DESC LIMIT 1) AS last_worker_id;
```
mrkaye97: why do we cast these to text?

cmd/hatchet-migrate/migrate/migrations/20260203170921_v1_0_76_c_backfill_slots.go
```diff
@@ -0,0 +1,226 @@
...
+
+		err := db.QueryRowContext(ctx, `
+WITH batch AS (
+	SELECT
+		"tenantId" AS tenant_id,
+		"id" AS step_id,
+		"isDurable" AS is_durable
+	FROM "Step"
+	WHERE ("tenantId", "id") > ($1::uuid, $2::uuid)
+	ORDER BY "tenantId", "id"
```
mrkaye97: Same here I think, I think we need to do this by just the `id` for correctness + performance, unless I'm missing something

pkg/repository/sqlcv1/queue.sql
```diff
@@ -278,7 +284,18 @@ SELECT
 FROM
     "StepDesiredWorkerLabel"
 WHERE
-    "stepId" = ANY(@stepIds::uuid[]);
+    "stepId" = ANY($1::uuid[]);
```
mrkaye97: I think this is incorrect? probably should still be `@stepIds` unless I'm missing something

mrkaye97: I think this is incorrect? probably should still be `@stepIds` unless I'm missing something

mrkaye97: I think this is incorrect? probably should still be `@stepIds` unless I'm missing something

pkg/repository/sqlcv1/tasks-overwrite.sql
```diff
@@ -101,46 +141,48 @@ SET
     internal_retry_count = 0,
     input = CASE WHEN i.input IS NOT NULL THEN i.input ELSE v1_task.input END,
     initial_state = i.initial_state,
-    concurrency_strategy_ids = i.concurrency_strategy_ids,
```
mrkaye97, line 104: intentional to remove this?

pkg/repository/sqlcv1/tasks-overwrite.sql
```diff
@@ -101,46 +141,48 @@ SET
...
 SELECT
     i.key,
     i.task_id,
     i.task_inserted_at,
     i.value_str,
+	i.value_int,
     i.kind
 FROM
     input i
-ON CONFLICT (key, task_id, task_inserted_at, kind) DO UPDATE
```
mrkaye97, line 143: intentional?

pkg/repository/sqlcv1/tasks.sql
```diff
@@ -912,6 +912,11 @@ WITH task AS (
...
+    WHERE
+        (task_id, task_inserted_at, retry_count) IN (SELECT id, inserted_at, retry_count FROM task)
+    RETURNING task_id
```
mrkaye97, line 919: any deadlock risk here?

pkg/repository/sqlcv1/workers.sql
```diff
@@ -10,21 +10,111 @@ SELECT
...
     ww."url" AS "webhookUrl",
-    ww."id" AS "webhookId",
-    workers."maxRuns" - (
-        SELECT COUNT(*)
-        FROM v1_task_runtime runtime
-        WHERE
-            runtime.tenant_id = workers."tenantId" AND
-            runtime.worker_id = workers."id"
-    ) AS "remainingSlots"
+    ww."id" AS "webhookId"
```
mrkaye97: maybe we can just drive-by remove this? webhook workers are gone now right?

pkg/repository/sqlcv1/workflows.sql
```diff
@@ -301,9 +302,31 @@ INSERT INTO "Step" (
...
+    updated_at
+)
+SELECT
+    @tenantId::uuid,
+    @stepId::uuid,
+    unnest(@slotTypes::text[]),
+    unnest(@units::integer[]),
+    CURRENT_TIMESTAMP,
+    CURRENT_TIMESTAMP
+ON CONFLICT (tenant_id, step_id, slot_type) DO UPDATE
```
mrkaye97: what's the case that would cause a conflict here? a second insert into `v1_task_runtime`?

pkg/scheduling/v1/action.go
```diff
@@ -31,10 +40,18 @@ func (a *action) activeCount() int {
...
```
mrkaye97: I'm not too familiar here, but IMO this sort of thing is a code smell - we should be able to write this in a way such that we don't need to have testing logic inside of the code itself, either via refactoring, mocking, etc.

pkg/scheduling/v1/scheduler.go
```diff
@@ -240,82 +243,182 @@ func (s *Scheduler) replenish(ctx context.Context, mustReplenish bool) error {
...
 
-	orderedLock(actionsToReplenish)
-	unlock := orderedUnlock(actionsToReplenish)
+	orderedLock(actionsToLock)
+	unlock := orderedUnlock(actionsToLock)
 	defer unlock()
 
+	if testHookBeforeReplenishUnackedLock != nil {
+		testHookBeforeReplenishUnackedLock()
+	}
```
mrkaye97: here too re: testing

sdks/python/hatchet_sdk/runnables/task.py
```diff
@@ -149,8 +149,12 @@ def __init__(
         wait_for: list[Condition | OrGroup] | None,
         skip_if: list[Condition | OrGroup] | None,
         cancel_if: list[Condition | OrGroup] | None,
+        slot_requests: dict[str, int] | None = None,
     ) -> None:
         self.is_durable = is_durable
+        if slot_requests is None:
+            slot_requests = {"durable": 1} if is_durable else {"default": 1}
```
mrkaye97, line 156: do we only want one requested slot here by default? feels like this should be 100 or something?

sql/schema/v1-core.sql
grutt: one runtime can have many slots

sdks/python/hatchet_sdk/runnables/task.py
```diff
@@ -393,6 +397,14 @@ def to_proto(self, service_name: str) -> CreateTaskOpts:
             schedule_timeout=timedelta_to_expr(self.schedule_timeout),
         )
 
+        # Backwards/forwards compatibility across protobuf schema versions.
+        if "is_durable" in CreateTaskOpts.DESCRIPTOR.fields_by_name:
+            proto.is_durable = self.is_durable
+        if "slot_requests" in CreateTaskOpts.DESCRIPTOR.fields_by_name:
+            proto.slot_requests.update(self.slot_requests)
+
+        return proto
```
mrkaye97: I think this isn't the right order - can't we handle the conditional logic similar to how we handle the concurrency thing?

sdks/python/hatchet_sdk/utils/slots.py
```diff
@@ -0,0 +1,80 @@
...
+from hatchet_sdk.worker.slot_types import SlotType
+
+
+def normalize_slot_config(
+    slot_config: dict[SlotType | str, int],
+) -> dict[str, int]:
+    normalized: dict[str, int] = {}
+    for key, value in slot_config.items():
+        normalized_key = key.value if isinstance(key, SlotType) else key
+        normalized[normalized_key] = value
```
mrkaye97: nit, but I'd prefer to use a dict comp for this

sdks/python/hatchet_sdk/utils/slots.py
```diff
@@ -0,0 +1,80 @@
...
+    slot_config: dict[SlotType | str, int], slot_type: SlotType
+) -> bool:
+    return slot_type in slot_config or slot_type.value in slot_config
+
+
+def ensure_slot_config(
+    slot_config: dict[SlotType | str, int], slot_type: SlotType, default_value: int
+) -> None:
+    if not has_slot_config(slot_config, slot_type):
+        slot_config[slot_type] = default_value
```
mrkaye97: I really dislike this and think we shouldn't do it 😅 

IMO, we should return the new dictionary, not mutate in place

sdks/python/lint.sh
```diff
@@ -8,6 +8,8 @@ poetry run ruff check . --fix
 echo "Formatting with black"
 poetry run black . --color
 
+# echo "\nFormatting with isort"
+# poetry run isort .
```
mrkaye97: can remove these I think

sdks/python/pyproject.toml
```diff
@@ -85,6 +85,21 @@ env = ["HATCHET_CLIENT_WORKER_HEALTHCHECK_ENABLED=True"]
...
+  "loguru",
+  "protobuf",
+  "pydantic",
+  "python_dotenv",
+  "python_dateutil",
+  "pyyaml",
+  "urllib3",
+]
+extend_skip = ["hatchet_sdk/contracts/", "hatchet_sdk/v0/contracts/"]
+
```
mrkaye97: can remove this too

cmd/hatchet-migrate/migrate/migrations/20260203170921_v1_0_76_c_backfill_slots.go
grutt: good catch

cmd/hatchet-migrate/migrate/migrations/20260216000001_v1_0_79_a_multi_slot_schema_and_triggers.sql
abelanger5, line 1: Add to CI - run load test, migrate up to the current latest migration on main, then migrate keeping the old engine running, and make sure the load test completes.

cmd/hatchet-migrate/migrate/migrations/20260203170922_v1_0_76_d_cleanup_and_old_columns.sql
```diff
@@ -0,0 +1,138 @@
...
+DROP FUNCTION IF EXISTS v1_step_slot_request_insert_function();
+
+DROP TRIGGER IF EXISTS v1_task_runtime_slot_insert_trigger ON v1_task_runtime;
+DROP FUNCTION IF EXISTS v1_task_runtime_slot_insert_function();
+
+DROP TRIGGER IF EXISTS v1_task_runtime_slot_delete_trigger ON v1_task_runtime;
+DROP FUNCTION IF EXISTS v1_task_runtime_slot_delete_function();
+
+ALTER TABLE "Worker"
+    DROP COLUMN IF EXISTS "maxRuns";
```
abelanger5: Drop this, so we don't need to do a multi-step migration

pkg/repository/workflow.go
```diff
@@ -101,10 +101,10 @@ type CreateStepOpts struct {
 	RetryBackoffMaxSeconds *int `validate:"omitnil,min=1,max=86400"`
 
 	// (optional) whether this step is durable
-	IsDurable bool
+	IsDurable bool `json:"isDurable,omitempty"`
```
abelanger5, line 104: I think this should be a pointer but let's double check

pkg/scheduling/v1/scheduler.go
```diff
@@ -809,62 +803,76 @@ func findAssignableSlots(
...
+
+	selected := make([]*slot, 0, totalNeeded)
 
 	for slotType, units := range requests {
 		if units <= 0 {
 			continue
 		}
 
-		slots, ok := workerSlots[slotType]
+		byWorker, ok := slotsByType[slotType]
```
abelanger5: bySlotType might be a better name for this

pkg/scheduling/v1/slot.go
```diff
@@ -87,7 +86,11 @@ func (s *slot) getWorkerId() uuid.UUID {
 }
 
 func (s *slot) getSlotType() string {
-	return s.slotType
+	if s.meta == nil {
+		return ""
```
abelanger5: Probably throw an error?

pkg/repository/sqlcv1/tasks.sql
grutt: gc

pkg/repository/workflow.go
grutt: i actually think this is preferred since we default to false, but i can be convinced otherwise if you have strong opinions

api-contracts/dispatcher/dispatcher.proto
```diff
@@ -33,6 +33,11 @@ service Dispatcher {
     rpc ReleaseSlot(ReleaseSlotRequest) returns (ReleaseSlotResponse) {}
 
     rpc UpsertWorkerLabels(UpsertWorkerLabelsRequest) returns (UpsertWorkerLabelsResponse) {}
+
+    // GetVersion returns the dispatcher protocol version as a simple integer.
```
mrkaye97, line 37: if it's possible, shouldn't we return the tag instead?

examples/python/simple/chaos_worker.py
mrkaye97, line 1: what's this?

sdks/python/hatchet_sdk/deprecated/deprecation.py
```diff
@@ -0,0 +1,77 @@
...
+        # Phase 3: raise 1-in-5 times
+        if feature not in _already_logged:
+            logger.error(
+                f"{message} "
+                "This fallback is no longer supported and will fail intermittently."
+            )
+            _already_logged.add(feature)
+
+        if random.random() < 0.2:
+            raise DeprecationError(f"{feature}: {message}")
```
mrkaye97, line 97: I don't think we should do this? It's unfortunate that we need to deprecate and wait, but I think raising an error here is not the right solution, unfortunately. Also, realistically a very small proportion of SDKs are even using any of these durable features, so at the very least we should only be forcing upgrades on those SDKs, probably?

api-contracts/dispatcher/dispatcher.proto
grutt: ya we do... comment drift

sdks/python/hatchet_sdk/worker/worker.py
```diff
@@ -204,6 +207,35 @@ def start(self, options: WorkerStartOptions = WorkerStartOptions()) -> None:
         if self.handle_kill:
             sys.exit(0)
 
+    async def _check_engine_version(self) -> bool:
```
mrkaye97: I think we should return the version here, probably, so that we can case on it later

api-contracts/dispatcher/dispatcher.proto
mrkaye97: wait but the tag isn't an int, it needs to be a string right?

sdks/python/hatchet_sdk/deprecated/deprecation.py
grutt: this is after 6 months and it only fails 20% of the time.

mrkaye97: ah alright that's fine then, hopefully we've removed it by then anyways... I guess it should maybe be easy enough to see who's using durable tasks anyways and ask them to upgrade

examples/python/simple/chaos_worker.py
grutt: renamed this to chaos_test.py, basically just starts/stop workers and runs tasks in a loop

sdks/python/hatchet_sdk/worker/worker.py
```diff
@@ -430,6 +468,13 @@ async def _close(self) -> None:
         if self.action_runner is not None:
             self.action_runner.cleanup()
 
+        # Also clean up the durable action runner (legacy mode)
+        durable_runner: WorkerActionRunLoopManager | None = getattr(
+            self, "_legacy_durable_action_runner", None
+        )
```
mrkaye97: let's not use `getattr` (why does Claude like `getattr` so much?!?!)

sdks/python/hatchet_sdk/worker/worker.py
```diff
@@ -446,9 +491,23 @@ async def exit_gracefully(self) -> None:
             await self.action_runner.wait_for_tasks()
             await self.action_runner.exit_gracefully()
 
+        # Also clean up the durable action runner (legacy mode)
+        durable_runner: WorkerActionRunLoopManager | None = getattr(
+            self, "_legacy_durable_action_runner", None
+        )
```
mrkaye97: same here

sdks/python/hatchet_sdk/deprecated/deprecation.py
grutt: yeah i think this is fine and tbh running that large of a mismatch will likely have other issues 😅

sdks/python/hatchet_sdk/worker/worker.py
grutt: 😅

api-contracts/dispatcher/dispatcher.proto
mrkaye97: oooh nvm, I see what you mean - I was looking at this commit, not the full diff 👍

api-contracts/dispatcher/dispatcher.proto
```diff
@@ -415,5 +415,6 @@ message ReleaseSlotResponse {}
 message GetVersionRequest {}
 
 message GetVersionResponse {
-    int32 dispatcher_version = 1;
+    reserved 1;
```
mrkaye97: what's this `reserved` thing?

sdks/go/deprecation.go
```diff
@@ -105,3 +107,34 @@ func EmitDeprecationNotice(feature, message string, start time.Time, logger *zer
...
+	aMaj, aMin, aPat := ParseSemver(a)
+	bMaj, bMin, bPat := ParseSemver(b)
+	if aMaj != bMaj {
+		return aMaj < bMaj
+	}
+	if aMin != bMin {
+		return aMin < bMin
+	}
+	return aPat < bPat
+}
```
mrkaye97, line 140: there must be a library for this

sdks/python/tests/test_semver.py
```diff
@@ -0,0 +1,46 @@
+from hatchet_sdk.deprecated.deprecation import parse_semver, semver_less_than
+
+
+class TestParseSemver:
```
mrkaye97: let's not wrap these in classes to stay consistent with the existing code

sdks/python/hatchet_sdk/worker/worker.py
```diff
@@ -207,32 +207,48 @@ def start(self, options: WorkerStartOptions = WorkerStartOptions()) -> None:
...
+
     async def _check_engine_version(self) -> bool:
-        """Returns True if engine is legacy (pre-slot-config)."""
+        """Returns True if engine is legacy (pre-slot-config).
+
+        Compares the engine's semantic version against the minimum required
+        version for slot_config support.
+        """
+        from hatchet_sdk.deprecated.deprecation import semver_less_than
```
mrkaye97: does this need to be imported here or can we import it at the top level?

sdks/python/hatchet_sdk/worker/worker.py
```diff
@@ -207,32 +207,48 @@ def start(self, options: WorkerStartOptions = WorkerStartOptions()) -> None:
...
+
         try:
             from hatchet_sdk.clients.dispatcher.dispatcher import DispatcherClient
```
mrkaye97: this too

sdks/python/hatchet_sdk/worker/worker.py
```diff
@@ -207,32 +207,48 @@ def start(self, options: WorkerStartOptions = WorkerStartOptions()) -> None:
...
 
             dispatcher = DispatcherClient(self.config)
```
mrkaye97: we might be able to grab a dispatcher client from somewhere else so we don't need to make a new one, e.g. `self.client.dispatcher` or something (unsure)

sdks/go/deprecation.go
grutt: do we want to add a dep for something stupid simple tho?

mrkaye97: yeah fair enough

api-contracts/dispatcher/dispatcher.proto
grutt: ooph cursor trying to keep the proto backwards compatible in the refactor, good catch

abelanger5 (approved):

promptless-for-oss:

New suggestion: **[Add durable execution documentation page](https://app.gopromptless.ai/suggestions/e446edc3-9854-41c5-a334-c7f1a42c0708)**

---

promptless-for-oss:

New suggestion: **[Document slot types for durable task isolation](https://app.gopromptless.ai/suggestions/77409d4b-a483-4a70-a107-b6f42625aede)**

---

PR #2928, opened 2026-02-03
refactor: use typed maps

abelanger5:
# Description

Uses typed `syncx.Map` instead of `sync.Map` for more safety. This also fixes a regression in the tenant manager which didn't properly clean up tenants. 

## Type of change

### DIFF
--- a/internal/msgqueue/shared_tenant_reader.go
+++ b/internal/msgqueue/shared_tenant_reader.go
@@ -7,10 +7,12 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/hashicorp/go-multierror"
+
+	"github.com/hatchet-dev/hatchet/internal/syncx"
 )
 
 type sharedTenantSub struct {
-	fs        *sync.Map
+	fs        syncx.Map[int, AckHook]
 	counter   int
 	isRunning bool
 	mu        sync.Mutex
@@ -18,23 +20,21 @@ type sharedTenantSub struct {
 }
 
 type SharedTenantReader struct {
-	tenants *sync.Map
+	tenants *syncx.Map[uuid.UUID, *sharedTenantSub]
 	mq      MessageQueue
 }
 
 func NewSharedTenantReader(mq MessageQueue) *SharedTenantReader {
 	return &SharedTenantReader{
-		tenants: &sync.Map{},
+		tenants: &syncx.Map[uuid.UUID, *sharedTenantSub]{},
 		mq:      mq,
 	}
 }
 
 func (s *SharedTenantReader) Subscribe(tenantId uuid.UUID, postAck AckHook) (func() error, error) {
-	tenant, _ := s.tenants.LoadOrStore(tenantId, &sharedTenantSub{
-		fs: &sync.Map{},
+	t, _ := s.tenants.LoadOrStore(tenantId, &sharedTenantSub{
+		fs: syncx.Map[int, AckHook]{},
 	})
-
-	t := tenant.(*sharedTenantSub)
 
 	t.mu.Lock()
 	defer t.mu.Unlock()
@@ -59,9 +59,7 @@ func (s *SharedTenantReader) Subscribe(tenantId uuid.UUID, postAck AckHook) (fun
 		cleanupSingleSub, err := s.mq.Subscribe(q, NoOpHook, func(task *Message) error {
 			var innerErr error
 
-			t.fs.Range(func(key, value interface{}) bool {
-				f := value.(AckHook)
-
+			t.fs.Range(func(key int, f AckHook) bool {
 				if err := f(task); err != nil {
 					innerErr = multierror.Append(innerErr, err)
 				}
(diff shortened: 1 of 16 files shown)


internal/operation/pool.go
```diff
@@ -2,18 +2,20 @@ package operation
...
+
 	"github.com/hatchet-dev/hatchet/internal/services/partition"
+	"github.com/hatchet-dev/hatchet/internal/syncx"
 	v1 "github.com/hatchet-dev/hatchet/pkg/repository"
 	"github.com/hatchet-dev/hatchet/pkg/repository/sqlcv1"
 
 	"github.com/rs/zerolog"
 )
 
 type OperationPool struct {
```
abelanger5: Should probably clarify that the operation pool is a tenant-based operation pool

internal/queueutils/operation.go
```diff
@@ -7,39 +7,34 @@ import (
 	"sync"
 	"time"
 
-	"github.com/google/uuid"
 	"github.com/rs/zerolog"
 )
 
-type ID interface {
-	string | int64 | uuid.UUID
```
abelanger5, line 15: These generics were mostly used for the `int64` partitioning thing which we don't need anymore, and I just wanted to simplify the generics a bit (could really use generics easily with a nested operation because we had a syncx.Map relying on it)

internal/syncx/map.go
```diff
@@ -0,0 +1,54 @@
...
+
+// Delete deletes the value for a key.
+func (m *Map[K, V]) Delete(key K) {
+	m.m.Delete(key)
+}
+
+// LoadOrStore loads or stores the value for a key.
+func (m *Map[K, V]) LoadOrStore(key K, value V) (actual V, loaded bool) {
+	a, loaded := m.m.LoadOrStore(key, value)
+	return a.(V), loaded
```
mrkaye97, line 35: is this type assertion safe?

internal/syncx/map.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// Range calls f sequentially for each key and value present in the map.
+// If f returns false, the range stops.
+func (m *Map[K, V]) Range(f func(key K, value V) bool) {
+	m.m.Range(func(key, value any) bool {
+		return f(key.(K), value.(V))
```
mrkaye97, line 42: also this one - not sure if this is safe or not I really dont understand how these assertions work 😅

internal/syncx/map.go
abelanger5: yes, because callers can't ever access `m sync.Map` so the only way this would panic is if we wrote an incorrect value internal to this package somehow, which shouldn't be possible

internal/syncx/map.go
abelanger5: safe for the same reasons

internal/msgqueue/shared_tenant_reader.go
```diff
@@ -7,35 +7,35 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/hashicorp/go-multierror"
+
+	"github.com/hatchet-dev/hatchet/internal/syncx"
 )
 
 type sharedTenantSub struct {
-	fs        *sync.Map
+	fs        syncx.Map[int, AckHook]
```
abelanger5: should convert this into a pointer so we don't pass lock values by reference

internal/msgqueue/shared_tenant_reader.go
```diff
@@ -100,25 +98,31 @@ func (s *SharedTenantReader) Subscribe(tenantId uuid.UUID, postAck AckHook) (fun
...
+	cleanup   func() error
+	fs        syncx.Map[int, DstFunc]
+	counter   int
+	mu        sync.Mutex
+	isRunning bool
+}
+
 type SharedBufferedTenantReader struct {
-	tenants *sync.Map
+	tenants syncx.Map[uuid.UUID, *sharedBufferedTenantSub]
```
abelanger5: inconsistent pointer usage with the above

internal/operation/pool.go
```diff
@@ -100,8 +102,8 @@ func (p *OperationPool) Cleanup() {
 	p.cancel()
 
 	// stop all operations
-	p.ops.Range(func(key, value any) bool {
-		if op, ok := value.(*SerialOperation); ok {
+	p.ops.Range(func(key uuid.UUID, op *SerialOperation) bool {
+		if op != nil {
```
abelanger5: should double check this, we can probably remove (our previous implementation did not have a nil guard)

internal/operation/pool.go
```diff
@@ -122,9 +124,9 @@ func (p *OperationPool) setTenants(tenants []*sqlcv1.Tenant) {
 	}
 
 	// delete tenants that are not in the list
-	p.ops.Range(func(key, value any) bool {
-		if _, ok := tenantMap[key.(string)]; !ok {
-			if op, ok := value.(*SerialOperation); ok {
+	p.ops.Range(func(key uuid.UUID, op *SerialOperation) bool {
+		if _, ok := tenantMap[key]; !ok {
+			if op != nil {
```
abelanger5: also here

mrkaye97 (approved):

PR #2936, opened 2026-02-04
feat(py): improved cancellation signaling

grutt:
# Description

Improves Python cancelation signal handling and propagation to Hatchet internals 

## Type of change

## What's Changed

- [X] Cancellation signals will now cancel subscriptions allowing for immediate return from a task


sdks/python/hatchet_sdk/runnables/workflow.py
```diff
@@ -625,95 +632,183 @@ def run_no_wait(
         self,
         input: TWorkflowInput = cast(TWorkflowInput, EmptyModel()),
         options: TriggerWorkflowOptions = TriggerWorkflowOptions(),
+        cancellation_token: CancellationToken | None = None,
```
mrkaye97: ~I'm not totally clear on why this is on the workflow but I don't think this makes sense - won't this token be shared by all runs of this workflow if we do this?~

mrkaye97: jk read the diff wrong 🤦

sdks/python/hatchet_sdk/runnables/workflow.py
```diff
@@ -625,95 +632,183 @@ def run_no_wait(
...
+        # Use context variable if no explicit token provided
+        if cancellation_token is None:
+            cancellation_token = ctx_cancellation_token.get()
+
+        logger.debug(
+            f"Workflow.run_no_wait: triggering {self.config.name}, "
+            f"token={cancellation_token is not None}"
+        )
+
+        ref = self.client._client.admin.run_workflow(
```
mrkaye97: I think we should maybe pull some of this logic into the admin client so we can repeat it fewer places?

sdks/python/hatchet_sdk/worker/runner/utils/capture_logs.py
```diff
@@ -48,10 +54,22 @@ class ContextVarToCopyDict(BaseModel):
     value: JSONSerializableMapping | None
 
 
+class ContextVarToCopyToken(BaseModel):
+    """Special type for copying CancellationToken to threads."""
+
+    model_config = {"arbitrary_types_allowed": True}
+
+    name: Literal["ctx_cancellation_token"]
+    value: Any  # CancellationToken, but using Any to avoid Pydantic issues with custom classes
```
mrkaye97, line 63: I think this should be fine to pass the class to directly - the `model_config` might be specified wrong here though (I think it needs to be something like `ConfigDict(arbitrary_types_allowed=True)`

sdks/python/tests/test_cancellation.py
```diff
@@ -0,0 +1,399 @@
...
+import pytest
+
+from hatchet_sdk.cancellation import CancellationToken
+from hatchet_sdk.exceptions import CancellationReason, CancelledError
+from hatchet_sdk.runnables.contextvars import ctx_cancellation_token
+from hatchet_sdk.utils.cancellation import await_with_cancellation
+
+
+class TestCancellationToken:
+    """Tests for the CancellationToken class."""
```
mrkaye97: we shouldn't use class-based tests IMO, it's not very pytest-y

sdks/python/tests/test_cancellation.py
mrkaye97, line 1: IMO, these tests are not very useful. I think we need E2E tests that test the functionality against the engine, e.g. using some of the cancellation methods in durable tasks and asserting on the behavior we expect

sdks/python/hatchet_sdk/runnables/workflow.py
grutt: i was going to flag this... im actually not sure if we do want to expose this though. as currently implemented this would override the context var token, if anything it maybe should merge tokens which i dont want to support right now.

grutt: removing this.

PR #2940, opened 2026-02-04
Feat: Durable event log models

mrkaye97:
# Description

Initial migration + sqlc generation for durable event log models, so we can all start wiring up writes

## Type of change


cmd/hatchet-migrate/migrate/migrations/20260204174818_v1_0_76.sql
```diff
@@ -0,0 +1,108 @@
...
+
+CREATE TYPE v1_durable_event_log_entry_kind AS ENUM (
+    'RUN_TRIGGERED',
+    'WAIT_FOR_STARTED',
+    'MEMO_STARTED'
+);
+
+CREATE TABLE v1_durable_event_log_entry (
+    -- need an external id for consistency with the payload store logic (unfortunately)
+    external_id UUID NOT NULL,
```
mrkaye97, line 30: added this for payload consistency (still hoping to update the payload schema to not use id + inserted_at anymore)

PR #2941, opened 2026-02-04
CI: Add Python 3.14 to the Python testing matrix

mrkaye97:
# Description

Adds 3.14 (latest) to the testing matrix + makes it the default on branches

## Type of change


sdks/python/docs/generator/doc_types.py
mrkaye97, line 1: drive by for import issues

mnafees (approved):

PR #2947, opened 2026-02-05
fix:  typescript metrics clients

grutt:
# Description

Aligns with current v1 cloud implementations

## Type of change


mrkaye97 (approved):
ideally we wouldn't need the breaking change? but maybe the current endpoints being wrapped aren't working now anyways?

grutt:
correct, the breaking has already broke

PR #2950, opened 2026-02-05
Add idempotency key reclaim + tests

colonelpanic8:
## Summary
- Add optional idempotency key field to workflow trigger APIs/clients
- Implement idempotency key reclaim on terminal status with configurable TTL and recheck interval
- Track `last_denied_at` and clean up keys on terminal status updates
- Add migration for `last_denied_at` plus expanded idempotency tests (including concurrency/mixed-key cases)
- Update flake dev shell with lint tooling

## Testing
- `CGO_ENABLED=0 go test ./pkg/repository -run Idempotency -count=1`
- `CGO_ENABLED=0 go test ./... -count=1`
- `nix develop -c golangci-lint run ./... --config .golangci.yml --new-from-rev=origin/main`

## Notes
- Requires migration `20260205120000_v1_0_76.sql` before deploying server changes.


promptless-for-oss:

New suggestion: **[Document idempotency keys feature for workflow triggers](https://app.gopromptless.ai/suggestions/75efb404-df4b-4727-8770-f03676a064af)**

---

promptless-for-oss:

New suggestion: **[Document idempotency keys feature for workflow triggers](https://app.gopromptless.ai/suggestions/48e3299f-fae6-4f6a-9cae-5c36ebd6f5fa)**

---

gregfurman:
Hey @colonelpanic8 👋 Apologies for the delay here. We're currently in the middle of a release week but should have more capacity next week to give this a review

mrkaye97:
hi @colonelpanic8 - closing this as folded into #4045. feel free to take a look there and let me know if you have thoughts!

colonelpanic8:
> hi @colonelpanic8 - closing this as folded into #4045. feel free to take a look there and let me know if you have thoughts!

cool. how similar is the interface to what I had originally?

do you know when this will get merged?

mrkaye97:
> > hi @colonelpanic8 - closing this as folded into #4045. feel free to take a look there and let me know if you have thoughts!
> 
> cool. how similar is the interface to what I had originally?
> 
> do you know when this will get merged?

It's somewhat similar, but there are important differences. One big one is that idempotency is configured on the task / workflow directly via a CEL expression that uses the input or additional metadata, as opposed to being passed at runtime (as this wouldn't work for events, webhooks, etc.). Feel free to take a look through the docs and examples there! Hoping to merge in the next day or two.

colonelpanic8:
@mrkaye97 Thanks for pointing me to #4045. I took a detailed look at the current implementation, and I’m pretty disappointed that this PR was closed as “folded into” #4045, because the behavior in #4045 does not actually subsume the behavior I implemented here.

The distinction is key derivation versus key lifetime. #4045 adds a useful workflow-level mechanism for deriving a key from a CEL expression, but it implements fixed-window deduplication:

- the key is claimed with `expires_at = now + ttl`;
- an existing claim becomes claimable only when `expires_at <= now`;
- cleanup only removes expired keys;
- `IdempotencyConfig` exposes only `expression` and `ttl`;
- terminal task/DAG status updates do not release the claim;
- there is no per-trigger release, bypass, or lifetime override.

What I need is active-execution idempotency:

1. Reject another trigger with the same key while the claiming workflow run is non-terminal.
2. Release/reclaim that key when the run reaches `COMPLETED`, `FAILED`, or `CANCELLED`.
3. Keep a TTL only as a safety net for an orphaned/stuck claim or a missed terminal update.

Those semantics cannot be reproduced by choosing a different fixed TTL. For example, suppose run A claims key K at 00:00, runs for two minutes, and completes at 00:02. A trigger at 00:01 should be rejected, but a trigger at 00:03 should be allowed immediately. With a 24-hour TTL, #4045 rejects the legitimate 00:03 trigger for almost a day. With a one-minute TTL, it can admit a duplicate while A is still running. No fixed TTL provides “for the lifetime of the active execution.”

This matters concretely for Railbird. A streaming `ProcessVideo` run can intentionally finish/suspend and later needs to resume the same processing record with the same logical key. Failure and recovery paths likewise need to enqueue again as soon as the previous run is terminal, while still preventing concurrent duplicate execution. Making the application manufacture a new nonce or infer Hatchet’s run state is not equivalent: it moves lifecycle coordination out of Hatchet and introduces opportunities for both lost retries and duplicate runs.

There are things I like in #4045: workflow-level CEL expressions, broader trigger-path coverage, and returning the existing run ID on collision are all improvements. But those are orthogonal to whether a claim is fixed-TTL or released on terminal status. I don’t think it is accurate to describe #2950 as folded into #4045 while the central lifecycle behavior is absent.

After reviewing #4045’s current structure, this also does not look architecturally difficult to support as an opt-in policy. One possible design would be:

- extend the workflow idempotency config with a lifecycle policy, e.g. `FIXED_TTL` versus `UNTIL_TERMINAL_WITH_TTL_FALLBACK` (or a `release_on_terminal` flag);
- persist that policy with the claim;
- when a standalone task or DAG reaches a terminal status, delete the claim only when tenant, key, and `claimed_by_external_id` still match;
- on a collision, optionally recheck the claimant’s terminal status and reclaim atomically, covering lag or failure in asynchronous terminal cleanup;
- retain the TTL as the fallback;
- test task and DAG completion/failure/cancellation plus the race where an old run terminates after an expired key has already been claimed by a newer run.

The terminal task/DAG hooks used by this PR still exist, and #4045 already records the claimant external ID and propagates the evaluated idempotency key onto runs, so the two designs appear compatible rather than mutually exclusive. The SQL must condition deletion on the claimant ID so an old run cannot delete a newer claim after TTL-based reuse.

Would you be open to preserving this as an idempotency lifecycle option in #4045 or accepting a focused follow-up PR that layers terminal release/reclaim onto the new expression-based design? This behavior is the reason I implemented #2950, and fixed-window deduplication does not meet that requirement.

mrkaye97:
Hey @colonelpanic8 - thanks for the detailed writeup! I like the direction you'd like to go here, and we're willing to consider it, as it's not the first time we've heard this. 

But first, a clarifying question: I'm unclear on why concurrency with a `GROUP_ROUND_ROBIN` strategy that uses the idempotency key as the concurrency key doesn't work for the case you mentioned. In this case, you'd see a run "claim a key" (occupy a concurrency slot), and once it reaches a terminal status, the engine "reclaims" that key, and another run can claim it. I think I'm misunderstanding something here, as based on my read, that gives you the behavior you're looking for. What am I missing?

colonelpanic8:
@mrkaye97 Thanks—I may be misunderstanding the concurrency semantics, but I need duplicate triggers received during an active execution to be discarded, not queued. Once that execution reaches a terminal state, subsequent triggers with the same key should be accepted normally. Does `GROUP_ROUND_ROBIN` provide that behavior?

mrkaye97:
Aha! In that case, maybe `CANCEL_NEWEST`? Then Task A comes in and claims "K" (the concurrency strategy) and runs, Task B comes in concurrently and is cancelled (discarded), and then A completes, C comes in, and is able to run again?

colonelpanic8:
Ah, yes—`CANCEL_NEWEST` seems to match the execution-level behavior. The remaining distinction is the trigger contract: I need the duplicate trigger to be rejected/collapsed at admission (ideally returning the existing run ID), rather than successfully creating a new run that is subsequently marked `CANCELLED`.

I may be misunderstanding the implementation, but my reading is that `CANCEL_NEWEST` still returns the new run ID successfully and cancellation happens through scheduler state. Is that correct?

mrkaye97:
Yep, that's right - you'd need to store the run id somewhere on your side so you could look it up on a cancellation (rejection). 

I'll keep thinking about your idempotency proposal by the way - I modified the PR a bit before merging to make it a bit easier to add this going forward, so I'd say it's definitely on the table for the next few weeks

mrkaye97:
Hey @colonelpanic8 - feel free to check out [this PR](https://github.com/hatchet-dev/hatchet/pull/4460), I think it'll do what you want

PR #2953, opened 2026-02-05
feat(ts): durable slot suspension ts cancellation

grutt:
# Description

Improves cancellation signaling and propagation for the typescript sdk

## Type of change

### DIFF
--- a/sdks/typescript/src/util/abort-error.ts
+++ b/sdks/typescript/src/util/abort-error.ts
@@ -4,3 +4,89 @@ export function createAbortError(message = 'Operation aborted'): Error {
   err.code = 'ABORT_ERR';
   return err as Error;
 }
+
+export function isAbortError(err: unknown): err is Error {
+  return (
+    err instanceof Error &&
+    (err.name === 'AbortError' || (err as any).code === 'ABORT_ERR')
+  );
+}
+
+/**
+ * Helper to be used inside broad `catch` blocks so cancellation isn't accidentally swallowed.
+ *
+ * Example:
+ * ```ts
+ * try { ... } catch (e) { rethrowIfAborted(e); ... }
+ * ```
+ */
+export function rethrowIfAborted(err: unknown): void {
+  if (isAbortError(err)) {
+    throw err;
+  }
+}
+
+export type ThrowIfAbortedOpts = {
+  /**
+   * Optional: called before throwing when the signal is aborted.
+   * This lets callsites attach logging without coupling this util to a logger implementation.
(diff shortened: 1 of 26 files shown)


sdks/typescript/src/util/logger/task-run-log.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export function actionMap(action: ActionType): string {
+  switch (action) {
+    case ActionType.START_STEP_RUN:
+      return 'starting...';
+    case ActionType.CANCEL_STEP_RUN:
+      return 'cancelling...';
+    case ActionType.START_GET_GROUP_KEY:
+      return 'starting to get group key...';
+    default:
```
mrkaye97, line 11: I think we should get into the habit of having that `exhaustivenessCheck` thing in places like this - just makes it so that we can't forget a case

sdks/typescript/src/util/abort-error.ts
```diff
@@ -0,0 +1,89 @@
+export function createAbortError(message = 'Operation aborted'): Error {
+  const err: any = new Error(message);
```
mrkaye97, line 2: why do we need the type hint for `any` here?

sdks/typescript/src/util/abort-error.ts
```diff
@@ -0,0 +1,89 @@
...
+  err.name = 'AbortError';
+  err.code = 'ABORT_ERR';
+  return err as Error;
+}
+
+export function isAbortError(err: unknown): err is Error {
+  return err instanceof Error && (err.name === 'AbortError' || (err as any).code === 'ABORT_ERR');
```
mrkaye97, line 9: maybe we should factor our `AbortError` and `ABORT_ERR` into constants if we're sharing them?

sdks/typescript/src/v1/client/worker/context.ts
```diff
@@ -608,11 +640,16 @@ export class Context<T, K = {}> {
         );
       }
 
+      // `signal` must never be sent over the wire.
+      const optsWithoutSignal: Omit<ChildRunOpts, 'signal'> & { signal?: never } = { ...opts };
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      delete (optsWithoutSignal as any).signal;
```
mrkaye97, line 646: is it possible to do this without the assertion to `any`?

mrkaye97 (approved):
left a couple small comments, but broadly looks good to me 👍

PR #2954, opened 2026-02-05
Feat: Durable Execution Revamp

mrkaye97:
# Description

WIP - Long-living branch for durable execution work

## Type of change


pkg/repository/durable_events.go
```diff
@@ -0,0 +1,297 @@
...
+	DurableTaskInsertedAt pgtype.Timestamptz
+	InsertedAt            pgtype.Timestamptz
+	Kind                  string
+	Key                   string
+	NodeId                int64
+	IsSatisfied           bool
+}
+
+type DurableEventsRepository interface {
+	CreateEventLogFiles(ctx context.Context, opts []CreateEventLogFileOpts) ([]*sqlcv1.V1DurableEventLogFile, error)
```
abelanger5: Probably not an exposed method on the repository, likely just a method on `sharedRepository` which takes a `tx`, since we'll call then on an insert of a durable task?

pkg/repository/durable_events.go
```diff
@@ -0,0 +1,297 @@
...
+	GetEventLogFileForTask(ctx context.Context, durableTaskId int64, durableTaskInsertedAt pgtype.Timestamptz) (*sqlcv1.V1DurableEventLogFile, error)
+
+	CreateEventLogEntries(ctx context.Context, opts []CreateEventLogEntryOpts) ([]*sqlcv1.V1DurableEventLogEntry, error)
```
abelanger5: I suppose this could be a batched insert if a user decides to call something like `RunMany`?

pkg/repository/durable_events.go
```diff
@@ -0,0 +1,297 @@
...
+	GetEventLogEntry(ctx context.Context, durableTaskId int64, durableTaskInsertedAt pgtype.Timestamptz, nodeId int64) (*sqlcv1.V1DurableEventLogEntry, error)
+	ListEventLogEntries(ctx context.Context, durableTaskId int64, durableTaskInsertedAt pgtype.Timestamptz) ([]*sqlcv1.V1DurableEventLogEntry, error)
```
abelanger5: I imagine that this is only useful when scoped to a branch ID (or the latest branch id)

pkg/repository/durable_events.go
```diff
@@ -0,0 +1,297 @@
...
+
+	CreateEventLogCallbacks(ctx context.Context, opts []CreateEventLogCallbackOpts) ([]*sqlcv1.V1DurableEventLogCallback, error)
+	GetEventLogCallback(ctx context.Context, durableTaskId int64, durableTaskInsertedAt pgtype.Timestamptz, key string) (*sqlcv1.V1DurableEventLogCallback, error)
+	ListEventLogCallbacks(ctx context.Context, durableTaskId int64, durableTaskInsertedAt pgtype.Timestamptz) ([]*sqlcv1.V1DurableEventLogCallback, error)
+	UpdateEventLogCallbackSatisfied(ctx context.Context, durableTaskId int64, durableTaskInsertedAt pgtype.Timestamptz, key string, isSatisfied bool) (*sqlcv1.V1DurableEventLogCallback, error)
```
abelanger5: This will almost certainly have to be batched, because we update these callbacks from `v1_match` which gets updated in batches.

pkg/repository/durable_events.go
```diff
@@ -0,0 +1,297 @@
...
+
+type DurableEventsRepository interface {
+	CreateEventLogFiles(ctx context.Context, opts []CreateEventLogFileOpts) ([]*sqlcv1.V1DurableEventLogFile, error)
+	GetEventLogFileForTask(ctx context.Context, durableTaskId int64, durableTaskInsertedAt pgtype.Timestamptz) (*sqlcv1.V1DurableEventLogFile, error)
+
+	CreateEventLogEntries(ctx context.Context, opts []CreateEventLogEntryOpts) ([]*sqlcv1.V1DurableEventLogEntry, error)
+	GetEventLogEntry(ctx context.Context, durableTaskId int64, durableTaskInsertedAt pgtype.Timestamptz, nodeId int64) (*sqlcv1.V1DurableEventLogEntry, error)
+	ListEventLogEntries(ctx context.Context, durableTaskId int64, durableTaskInsertedAt pgtype.Timestamptz) ([]*sqlcv1.V1DurableEventLogEntry, error)
+
+	CreateEventLogCallbacks(ctx context.Context, opts []CreateEventLogCallbackOpts) ([]*sqlcv1.V1DurableEventLogCallback, error)
```
abelanger5: I think these fields will have to be written as part of the tx which inserts the durable event log entries, wdyt?

pkg/repository/durable_events.go
```diff
@@ -0,0 +1,297 @@
...
+	GetEventLogCallback(ctx context.Context, durableTaskId int64, durableTaskInsertedAt pgtype.Timestamptz, key string) (*sqlcv1.V1DurableEventLogCallback, error)
```
abelanger5: Looking at this further I think we'll need to store the `callback_key` field on the durable event log entry. The gRPC API is going to be using the `node_id` field to query. 

Perhaps the concept of a `key` doesn't make much sense here, and we should just be using the `node_id` throughout the system, rather than the key? And then on the `v1_match` table we just store the `callback_node_id`?

pkg/repository/durable_events.go
```diff
@@ -0,0 +1,297 @@
...
+	InsertedAt            pgtype.Timestamptz
+	Kind                  string
+	Key                   string
+	NodeId                int64
+	IsSatisfied           bool
+}
+
+type DurableEventsRepository interface {
+	CreateEventLogFiles(ctx context.Context, opts []CreateEventLogFileOpts) ([]*sqlcv1.V1DurableEventLogFile, error)
+	GetEventLogFileForTask(ctx context.Context, durableTaskId int64, durableTaskInsertedAt pgtype.Timestamptz) (*sqlcv1.V1DurableEventLogFile, error)
```
abelanger5: nit: I suspect this is implementation we'd like to keep internal to the repository. I actually don't see a use-case for anything external to be requiring this log file. I do see a use-case for querying for the latest `branch_id`.

pkg/repository/durable_events.go
abelanger5: Also, this isn't something we should expose on the repository, it will always require a tx to update to ensure tx safety.

pkg/repository/durable_events.go
```diff
@@ -0,0 +1,297 @@
...
+	DurableTaskId         int64
+	DurableTaskInsertedAt pgtype.Timestamptz
+	InsertedAt            pgtype.Timestamptz
+	Kind                  string
+	Key                   string
+	NodeId                int64
+	IsSatisfied           bool
+}
+
+type DurableEventsRepository interface {
```
abelanger5: I think we'll need a method for creating a new branch from a specific event id

pkg/repository/durable_events.go
```diff
@@ -0,0 +1,297 @@
...
+type CreateEventLogEntryOpts struct {
+	TenantId              uuid.UUID
+	ExternalId            uuid.UUID
+	DurableTaskId         int64
+	DurableTaskInsertedAt pgtype.Timestamptz
+	InsertedAt            pgtype.Timestamptz
+	Kind                  string
+	NodeId                int64
+	ParentNodeId          int64
+	BranchId              int64
```
abelanger5: These are sequence values generated internally (or joined through the log file table), so they shouldn't be settable by the user

pkg/repository/durable_events.go
```diff
@@ -0,0 +1,297 @@
...
+
+type CreateEventLogCallbackOpts struct {
+	DurableTaskId         int64
+	DurableTaskInsertedAt pgtype.Timestamptz
+	InsertedAt            pgtype.Timestamptz
+	Kind                  string
+	Key                   string
+	NodeId                int64
+	IsSatisfied           bool
+}
```
abelanger5: see note below: I don't think these should be exposed

pkg/repository/durable_events.go
```diff
@@ -0,0 +1,297 @@
...
+
+func (r *durableEventsRepository) GetEventLogFileForTask(ctx context.Context, durableTaskId int64, durableTaskInsertedAt pgtype.Timestamptz) (*sqlcv1.V1DurableEventLogFile, error) {
+	return r.queries.GetDurableEventLogFileForTask(ctx, r.pool, sqlcv1.GetDurableEventLogFileForTaskParams{
+		Durabletaskid:         durableTaskId,
+		Durabletaskinsertedat: durableTaskInsertedAt,
+	})
+}
+
+func (r *durableEventsRepository) CreateEventLogEntries(ctx context.Context, opts []CreateEventLogEntryOpts) ([]*sqlcv1.V1DurableEventLogEntry, error) {
```
abelanger5: Biggest concern here is that we need to lock the file row in an exclusive mode and we need serial updates based on the latest read. In other words we can't just use `latest_node_id + 1` to generate these log entries because concurrent writes will read the same `latest_node_id`. 

These cases should be exceedingly rare though, so perhaps we're ok with an advisory `TryLock` and a simple retry? If we're getting a lot of concurrent writes to these tables we're doing something wrong (should be serialized by the `Durable` listener anyway)

promptless-for-oss:

New suggestion: **[Document NonDeterminismError exception and reset_durable_task methods](https://app.gopromptless.ai/suggestions/48aec471-52b2-4ea5-b411-5e402a482483)**

---
_Tip: Tell your friends working on non-commercial open-source projects to apply for free Promptless access at [promptless.ai/oss](https://promptless.ai/oss) ❤️_

promptless-for-oss:

Updated existing suggestion: **[Document durable execution with fork and replay capabilities](https://app.gopromptless.ai/suggestions/e446edc3-9854-41c5-a334-c7f1a42c0708)**

---
_Tip: Add or adjust Promptless's style guide in [Agent Knowledge Base](https://app.gopromptless.ai/configure/settings) ✍️_

grutt (approved):
🪨

promptless-for-oss:

New suggestion: **[Document NonDeterminismError exception and reset_durable_task methods](https://app.gopromptless.ai/suggestions/8722bfff-7718-422c-89ae-1c13e578f6fa)**

---

PR #2956, opened 2026-02-05
Feat: Durable event log wiring

mrkaye97:
# Description

First pass at the work on wiring up the new durable event backend. Basically got the event sends / acks and the callback registrations / acks / results wired up so that they match the current functionality. And then in the next couple PRs (smaller) I can work on adding new features, e.g. branching logic, nondeterminism detection, etc.

Left a couple comments on things, but the basic idea is that for processing the events, we just do everything on the dispatcher, but for callbacks we do the waiting on the controllers when necessary and try to optimistically pub messages out to the dispatcher when things happen, like matches being satisfied. I don't think it's perfect but the tests seem to be passing so I figured this was a good place to leave this one

## Type of change


internal/services/controllers/olap/controller.go
```diff
@@ -435,19 +435,19 @@ func (tc *OLAPControllerImpl) handleBufferedMsgs(tenantId uuid.UUID, msgId strin
 	}()
 
 	switch msgId {
-	case "created-task":
+	case msgqueue.MsgIDCreatedTask:
```
mrkaye97: drive by fixed some of these

pkg/repository/trigger.go
```diff
@@ -2166,3 +2173,67 @@ func (r *sharedRepository) prepareTriggerFromWorkflowNames(ctx context.Context,
 
 	return triggerOpts, nil
 }
+
+func (r *sharedRepository) NewTriggerOpt(
```
mrkaye97: factored this out so we could share it in a few places

pkg/repository/durable_events.go
```diff
@@ -110,21 +134,27 @@ func (r *durableEventsRepository) CreateEventLogFiles(ctx context.Context, opts
...
-		Durabletaskinsertedat: durableTaskInsertedAt,
+func (r *durableEventsRepository) GetOrCreateEventLogFileForTask(ctx context.Context, tenantId uuid.UUID, durableTaskId int64, durableTaskInsertedAt pgtype.Timestamptz) (*sqlcv1.GetOrCreateEventLogFileForTaskRow, error) {
+	return r.queries.GetOrCreateEventLogFileForTask(ctx, r.pool, sqlcv1.GetOrCreateEventLogFileForTaskParams{
+		Tenantid:                      tenantId,
+		Durabletaskid:                 durableTaskId,
+		Durabletaskinsertedat:         durableTaskInsertedAt,
+		Latestinsertedat:              sqlchelpers.TimestamptzFromTime(time.Now().UTC()),
+		Latestnodeid:                  0,
+		Latestbranchid:                1,
+		Latestbranchfirstparentnodeid: 0,
```
mrkaye97: plan here will be to figure out how to deal with these in the next PR or two, but hard-coding them for now

pkg/repository/durable_events.go
```diff
@@ -176,13 +202,38 @@ func (r *durableEventsRepository) CreateEventLogEntries(ctx context.Context, opt
...
+	for i, entry := range entries {
+		opt, ok := externalIdToOpts[entry.ExternalID]
+
+		if !ok {
+			continue
+		}
+
+		hash := sha256.Sum256(opt.Data)
+		dataHashes[i] = hash[:]
+		dataHashAlgs[i] = "sha256"
```
mrkaye97: will figure out how to throw a nondeterminism error out of here somehow (or maybe factor this out)

pkg/repository/match.go
```diff
@@ -117,6 +127,9 @@ type EventMatchResults struct {
 
 	// The list of tasks which were replayed from the matches
 	ReplayedTasks []*V1TaskWithPayload
+
+	// The list of satisfied durable callbacks from matches
+	SatisfiedCallbacks []SatisfiedCallback
```
mrkaye97, line 134: lets us pub back to the dispatcher out of here

cmd/hatchet-engine/engine/run.go
```diff
@@ -377,12 +377,15 @@ func runV0Config(ctx context.Context, sc *server.ServerConfig) ([]Teardown, erro
...
 			dispatcherv1.WithMessageQueue(sc.MessageQueueV1),
 			dispatcherv1.WithLogger(sc.Logger),
+			dispatcherv1.WithDispatcherId(d.DispatcherId()),
 		)
 
 		if err != nil {
 			return nil, fmt.Errorf("could not create dispatcher (v1): %w", err)
 		}
 
+		d.SetDurableCallbackHandler(dv1.(*dispatcherv1.DispatcherServiceImpl))
```
abelanger5: Discuss this

sql/schema/v1-core.sql
```diff
@@ -2316,6 +2329,8 @@ CREATE TABLE v1_durable_event_log_callback (
     -- Whether this callback has been seen by the engine or not. Note that is_satisfied _may_ change multiple
     -- times through the lifecycle of a callback, and readers should not assume that once it's true it will always be true.
     is_satisfied BOOLEAN NOT NULL DEFAULT FALSE,
+    -- The dispatcher ID that registered this callback, used to route completion messages
+    dispatcher_id UUID,
```
abelanger5: to discuss

pkg/repository/sqlcv1/durable_event_log.sql
```diff
@@ -28,18 +31,57 @@ FROM
 RETURNING *
 ;
 
--- name: GetDurableEventLogFileForTask :one
+-- name: UpdateLatestNodeId :one
```
abelanger5: discuss concurrent writes

internal/services/controllers/olap/controller.go
abelanger5: please can this be a separate PR

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (tc *TasksControllerImpl) processSingleSatisfiedCallback(ctx context.Context, tenantId uuid.UUID, cb v1.SatisfiedCallback) error {
+	if cb.DispatcherId == nil {
+		return fmt.Errorf("callback %s has no dispatcher_id set", cb.CallbackKey)
+	}
+
+	dispatcherId := *cb.DispatcherId
+
+	callbackKey, err := tc.repov1.DurableEvents().ParseCallbackKey(ctx, cb.CallbackKey)
```
abelanger5: to discuss - probably not n+1 but should double check

internal/services/dispatcher/v1/server.go
```diff
@@ -318,3 +320,624 @@ func waitFor(wg *sync.WaitGroup, timeout time.Duration, l *zerolog.Logger) {
...
+		return status.Errorf(codes.InvalidArgument, "invalid durable_task_external_id: %v", err)
+	}
+
+	task, err := d.repo.Tasks().GetTaskByExternalId(ctx, invocation.tenantId, taskExternalId, false)
+	if err != nil {
+		return status.Errorf(codes.NotFound, "task not found: %v", err)
+	}
+
+	nodeId := invocation.getNextNodeId(req.DurableTaskExternalId, req.InvocationCount)
+	existingEntry, err := d.repo.DurableEvents().GetEventLogEntry(ctx, invocation.tenantId, task.ID, task.InsertedAt, nodeId)
```
abelanger5: discuss transactional safety

internal/services/dispatcher/v1/server.go
```diff
@@ -318,3 +320,624 @@ func waitFor(wg *sync.WaitGroup, timeout time.Duration, l *zerolog.Logger) {
...
+	for {
+		select {
+		case <-ticker.C:
+			signalEvents := []v1.TaskIdInsertedAtSignalKey{{
+				Id:         task.ID,
+				InsertedAt: task.InsertedAt,
+				SignalKey:  signalKey,
+			}}
+
+			completedEvents, err := d.repo.Tasks().ListSignalCompletedEvents(pollCtx, invocation.tenantId, signalEvents)
```
abelanger5: we should probably not be using this method

sql/schema/v1-core.sql
abelanger5: Are we using this somewhere?

pkg/repository/trigger.go
```diff
@@ -325,14 +332,18 @@ func getEventExternalIdToRuns(opts []EventTriggerOpts, externalIdToEventIdAndFil
...
+func (s *sharedRepository) triggerFromWorkflowNames(ctx context.Context, tx sqlcv1.DBTX, tenantId uuid.UUID, opts []*WorkflowNameTriggerOpts) ([]*V1TaskWithPayload, []*DAGWithData, error) {
+	triggerOpts, err := s.prepareTriggerFromWorkflowNames(ctx, tx, tenantId, opts)
 
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to prepare trigger from workflow names: %w", err)
 	}
 
-	return r.triggerWorkflows(ctx, nil, tenantId, triggerOpts, nil)
+	return s.triggerWorkflows(ctx, nil, tenantId, triggerOpts, nil)
+}
```
abelanger5: This looks like a deadlock. We have a tx open, but we're passing in `nil`, which means we likely start a tx inside of `triggerWorkflows`

abelanger5: obviously not a deadlock if we pass in `r.pool` as the `tx` parameter, but is a deadlock in any other case.

pkg/repository/trigger.go
```diff
@@ -2166,3 +2177,67 @@ func (r *sharedRepository) prepareTriggerFromWorkflowNames(ctx context.Context,
...
 	return triggerOpts, nil
 }
+
+func (r *sharedRepository) NewTriggerOpt(
+	ctx context.Context,
+	tenantId uuid.UUID,
+	req *v1contracts.TriggerWorkflowRequest,
+	parentTask *sqlcv1.FlattenExternalIdsRow,
+) (*WorkflowNameTriggerOpts, error) {
+	ctx, span := telemetry.NewSpan(ctx, "admin_service.new_trigger_opt")
```
abelanger5: Nit: span naming - this shouldn't be `admin_service`?

pkg/repository/trigger.go
```diff
@@ -11,11 +11,16 @@ import (
 	"github.com/google/uuid"
 	"github.com/jackc/pgx/v5/pgtype"
 	"github.com/rs/zerolog"
+	"go.opentelemetry.io/otel/attribute"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/hatchet-dev/hatchet/internal/cel"
+	v1contracts "github.com/hatchet-dev/hatchet/internal/services/shared/proto/v1"
```
abelanger5, line 17: I feel a little strange drilling `v1contracts` down to the repository level, it feels like a SoC violation. I also saw that we're directly returning grpc codes out of this repo - I think this is a little sketchy, since these get sent to the user directly. I'd still prefer that we have a clear boundary between the repository and API layers, particularly when it comes to types and errors.

pkg/repository/trigger.go
```diff
@@ -2166,3 +2177,67 @@ func (r *sharedRepository) prepareTriggerFromWorkflowNames(ctx context.Context,
...
+		t.ParentExternalId = &parentExternalId
+		t.ParentTaskId = &parentTask.ID
+		t.ParentTaskInsertedAt = &parentTask.InsertedAt.Time
+		t.ChildIndex = &childIndex
+		t.ChildKey = req.ChildKey
+	}
+
+	return &WorkflowNameTriggerOpts{
+		TriggerTaskData: t,
+		ExternalId:      uuid.New(),
```
abelanger5: Just confirming that this is always generated, and shouldn't be overwritten in any case? Just thinking about the previous implementation where external id sometimes aligned with the previously generated child workflow id.

pkg/repository/match.go
```diff
@@ -140,6 +157,16 @@ type GroupMatchCondition struct {
 	Data []byte
 }
 
+type SatisfiedCallback struct {
+	// need to figure out if we want the id + inserted at, external id, or both here - seems like we might need both
```
abelanger5: nit: update comment?

pkg/repository/match.go
```diff
@@ -653,6 +684,60 @@ func (m *sharedRepository) processEventMatches(ctx context.Context, tx sqlcv1.DB
...
+			durableTaskId := match.DurableEventLogCallbackDurableTaskID.Int64
+			cb := SatisfiedCallback{
+				DurableTaskExternalId: *match.DurableEventLogCallbackDurableTaskExternalID,
+				DurableTaskId:         &durableTaskId,
+				DurableTaskInsertedAt: match.DurableEventLogCallbackDurableTaskInsertedAt,
+				NodeId:                match.DurableEventLogCallbackNodeID.Int64,
+				Data:                  match.McAggregatedData,
+			}
+
+			callback, err := m.queries.UpdateDurableEventLogCallbackSatisfied(ctx, tx, sqlcv1.UpdateDurableEventLogCallbackSatisfiedParams{
```
abelanger5: I still think writes should be batched, while reads perhaps shouldn't be (in some cases). I'm becoming less excited about the `UNNEST` pattern because it's annoying to debug and use, despite it being the best performance-wise. I've started to enjoy `batch.Queue` -- if you look at the implementation of `ReleaseTasks` you'll see what I'm talking about. Might be a good candidate here?

pkg/repository/match.go
```diff
@@ -653,6 +684,60 @@ func (m *sharedRepository) processEventMatches(ctx context.Context, tx sqlcv1.DB
...
+			if callback.Kind.Valid && callback.Kind.V1DurableEventLogCallbackKind == sqlcv1.V1DurableEventLogCallbackKindRUNCOMPLETED {
+				if extracted, extractErr := ExtractOutputFromMatchData(cb.Data); extractErr != nil {
+					m.l.Error().Err(extractErr).Msgf("failed to extract output from RUN_COMPLETED match data for callback %d", cb.NodeId)
+				} else {
+					cb.Data = extracted
+				}
+			}
+
+			if len(cb.Data) > 0 {
+				err = m.payloadStore.Store(ctx, tx, StorePayloadOpts{
```
abelanger5: Since `Store` is variadic, we should be able to easily append and store in a single call, regardless of what we do above?

pkg/repository/durable_events.go
```diff
@@ -2,16 +2,21 @@ package repository
...
-	"crypto/sha256"
+	"errors"
+	"fmt"
+	"time"
 
 	"github.com/google/uuid"
+	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgtype"
 
+	v1 "github.com/hatchet-dev/hatchet/internal/services/shared/proto/v1"
```
abelanger5: See note in the `trigger.go` file about this - I'm still not sure about this

pkg/repository/durable_events.go
```diff
@@ -2,16 +2,21 @@ package repository
...
 	"github.com/hatchet-dev/hatchet/pkg/repository/sqlchelpers"
 	"github.com/hatchet-dev/hatchet/pkg/repository/sqlcv1"
 )
 
 type CreateEventLogFileOpts struct {
```
abelanger5: We need validators for all these types -- I think it's critical here

pkg/repository/durable_events.go
```diff
@@ -21,40 +26,80 @@ type CreateEventLogFileOpts struct {
 }
 
 type CreateEventLogEntryOpts struct {
```
abelanger5: Lots of these seem unused?

pkg/repository/durable_events.go
```diff
@@ -67,231 +112,502 @@ func newDurableEventsRepository(shared *sharedRepository) DurableEventsRepositor
...
+			Branchid:              params.Branchid,
+			Datahash:              params.Datahash,
+			Datahashalg:           params.Datahashalg,
+		})
+
+		if err != nil {
+			return nil, err
+		}
+
+		entry = &sqlcv1.V1DurableEventLogEntry{
```
abelanger5: Is `newEntry` not of type `*sqlcv1.V1DurableEventLogEntry`? why are we doing this?

pkg/repository/durable_events.go
```diff
@@ -67,231 +112,502 @@ func newDurableEventsRepository(shared *sharedRepository) DurableEventsRepositor
...
+				Payload:    payload,
+				TenantId:   tenantId,
+			})
+
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		callback = &sqlcv1.V1DurableEventLogCallback{
```
abelanger5: same question here

pkg/repository/sqlcv1/durable_event_log.sql
```diff
@@ -64,98 +63,77 @@ INSERT INTO v1_durable_event_log_entry (
...
+SELECT cb.*, t.external_id AS task_external_id
+FROM v1_durable_event_log_callback cb
+JOIN v1_task t ON t.id = cb.durable_task_id
+    AND t.inserted_at = cb.durable_task_inserted_at
+    AND t.tenant_id = @tenantId::UUID
+WHERE (t.external_id, cb.node_id) IN (
+    SELECT
+        unnest(@taskExternalIds::uuid[]),
+        unnest(@nodeIds::bigint[])
+)
```
abelanger5: This where clause concerns me...can we lookup the task in a CTE to simplify the structure a bit? I'd be concerned about the query planner grabbing the `node_id` as the first thing it looks for, which would obviously lead to a seq scan

pkg/repository/durable_events.go
```diff
@@ -67,231 +112,502 @@ func newDurableEventsRepository(shared *sharedRepository) DurableEventsRepositor
...
+		payload, err := r.payloadStore.RetrieveSingle(ctx, r.pool, RetrievePayloadOpts{
+			Id:         row.ID,
+			InsertedAt: row.InsertedAt,
+			Type:       sqlcv1.V1PayloadTypeDURABLEEVENTLOGCALLBACKRESULTDATA,
+			TenantId:   tenantId,
+		})
 		if err != nil {
-			return nil, err
+			r.l.Warn().Err(err).Msgf("failed to retrieve payload for callback %d", row.NodeID)
+			payload = nil
```
abelanger5, line 235: Is this what we want to be doing? Seems like a bunch of things upstream will break

pkg/repository/sqlcv1/durable_event_log.sql
```diff
@@ -1,58 +1,57 @@
...
+FROM v1_durable_event_log_file
+WHERE durable_task_id = @durableTaskId::BIGINT
+    AND durable_task_inserted_at = @durableTaskInsertedAt::TIMESTAMPTZ
+FOR UPDATE;
+
+-- name: UpdateLogFileNodeIdInvocationCount :one
+UPDATE v1_durable_event_log_file
+SET
+    latest_node_id = COALESCE(sqlc.narg('nodeId')::BIGINT, v1_durable_event_log_file.latest_node_id),
+    latest_invocation_count = COALESCE(sqlc.narg('invocationCount')::BIGINT, v1_durable_event_log_file.latest_invocation_count)
```
abelanger5, line 12: Are these correct? I thought we'd be incrementing the `latest_node_id` and `latest_invocation_count` here

abelanger5: actually, I see that we have some casing later on to set/increment these, but I'm still a bit confused

pkg/repository/durable_events.go
```diff
@@ -67,231 +112,502 @@ func newDurableEventsRepository(shared *sharedRepository) DurableEventsRepositor
...
-		kinds[i] = opt.Kind
-		keys[i] = opt.Key
-		nodeIds[i] = opt.NodeId
-		isSatisfieds[i] = opt.IsSatisfied
+	// take a lock of the log file so nothing else can concurrently write to it and e.g. increment the node id or branch
+	// id while this tx is running
+	logFile, err := r.queries.GetAndLockLogFile(ctx, tx, sqlcv1.GetAndLockLogFileParams{
+		Durabletaskid:         task.ID,
+		Durabletaskinsertedat: task.InsertedAt,
+	})
```
abelanger5, line 272: This isn't strict enough of a lock from what I can tell - I think we need to do the equivalent of a `TryLock` and error or retry if we don't acquire it. The reason is that the `UpdateLogFileNodeIdInvocationCount` will only see rows from before the tx started, so we won't catch the update to the log file. We might get lucky with a unique constraint violation later but this feels fragile.

pkg/repository/durable_events.go
```diff
@@ -67,231 +112,502 @@ func newDurableEventsRepository(shared *sharedRepository) DurableEventsRepositor
...
+	if logFile.LatestInvocationCount < opts.InvocationCount {
+		isNewInvocation = true
 	}
 
-	return callbacks, nil
-}
+	var nodeId int64
+	if isNewInvocation {
+		newNode, err := r.queries.UpdateLogFileNodeIdInvocationCount(ctx, tx, sqlcv1.UpdateLogFileNodeIdInvocationCountParams{
+			NodeId:                sqlchelpers.ToBigInt(1),
```
abelanger5, line 298: I think this might come back to bite us - I think having a unique node id across invocations of the same durable task will be necessary. I know the focus of this isn't branching, but is the thinking that a node is actually identified by its `(invocation, node_id)` tuple within a task?

pkg/repository/sqlcv1/durable_event_log.sql
abelanger5: ok, I do see now that we could call this up to 3 different times in the same tx - but we should be able to do better than that with just a single update call?

pkg/repository/durable_events.go
```diff
@@ -67,231 +112,502 @@ func newDurableEventsRepository(shared *sharedRepository) DurableEventsRepositor
...
-		Durabletaskid:         durableTaskId,
-		Durabletaskinsertedat: durableTaskInsertedAt,
-		Key:                   key,
-		Issatisfied:           isSatisfied,
+	var spawnedTasks []*V1TaskWithPayload
+	var spawnedDAGs []*DAGWithData
+
+	if !logEntry.AlreadyExisted {
+		switch opts.Kind {
+		case sqlcv1.V1DurableEventLogEntryKindWAITFORSTARTED:
```
abelanger5: nit: each of these switch cases should be pulled into their own method, this is really hard to read

pkg/repository/durable_events.go
```diff
@@ -67,231 +112,502 @@ func newDurableEventsRepository(shared *sharedRepository) DurableEventsRepositor
...
+
+			if err != nil {
+				return nil, fmt.Errorf("failed to trigger workflows: %w", err)
+			}
+
+			taskId := task.ID
+			taskExternalId := task.ExternalID
+			callbackNodeId := callbackResult.Callback.NodeID
+
+			for _, childTask := range createdTasks {
```
abelanger5: It appears that we're registering each task as its own match condition, I don't think this is what we want? Let's say we're invoking a child workflow with tasks `a` and `b`, I think our match condition would look something like:

```
( a-completed || a-failed || a-cancelled) && (b-completed || b-failed || b-cancelled)
```

As in the finalization events all share an or group and are anded together with other tasks in the workflow.

pkg/repository/trigger.go
mrkaye97: whoops yeah, good catch

pkg/repository/trigger.go
mrkaye97: haha thanks, copy paste 👍

pkg/repository/durable_events.go
mrkaye97: ah yeah you're right, this was leftover from before - for a little bit I had it so it returned the entry + a boolean flag for if it's new, so had to do this. I can remove this now!

pkg/repository/durable_events.go
mrkaye97: thanks yeah, leftover from the original skeleton

pkg/repository/durable_events.go
mrkaye97: yeah, exactly - I was thinking that on replay, we want to increment the invocation count but use the existing `node_id`s, which makes the lookups for existing callbacks really easy

sql/schema/v1-core.sql
mrkaye97: yeah, we use it in `processSingleSatisfiedCallback` which we call when matches are satisfied. I wasn't sure how else to get around this

pkg/repository/trigger.go
mrkaye97: I have mixed feelings on the contracts thing in this case - in general I agree, but the method using this is basically a transformer from proto -> `TriggerWorkflowRequest`, so it seems silly to me to have another struct that's basically just a copy of that original proto in the middle there for SoC reasons

Agree on the grpc codes though, will fix those (another copy paste thing)

pkg/repository/durable_events.go
abelanger5: Ah, so `isNewInvocation` means that this is a completely new invocation, not just a replay?

mrkaye97: Oh so I was thinking of a replay as an invocation, but I think this is going to need to change when branching starts to get implemented because that’ll invalidate some of the assumptions here

pkg/repository/sqlcv1/durable_event_log.sql
mrkaye97: going to leave this for the branching logic changes - I think all of this will need to change

pkg/repository/durable_events.go
mrkaye97: think I'll figure this out with the branching stuff too - agreed here though

pkg/repository/durable_events.go
mrkaye97: Just to make sure I understand, this is only relevant in the case where you spawn multiple children in bulk, right? I wasn't sure if we'd want one match or multiple (if I understand right, I think multiple would let you have separate callbacks with their own outputs and such for each child?)

abelanger5: No, I'm talking about a case where you spawn a child workflow which has multiple tasks. We are not registering a separate callback for each task in that workflow, right? However, we are registering each task as its own match condition? This is the inconsistency/bug I'm flagging. Very possible I'm misunderstanding though!

pkg/repository/durable_events.go
abelanger5: I think the safest bet is to treat replays as new invocations, so as long as `isNewInvocation` means the first time that the durable task is triggered, this makes sense to me. I just want to avoid ever resetting the nodeId to 1.

pkg/repository/trigger.go
mrkaye97: thanks, this was fine but confusing - I refactored a bit to just return the `TriggerTaskData` instead

pkg/repository/durable_events.go
mrkaye97: I think this is a bit of the same trap as always, which is we don't have a guarantee that we find the payload in the db (since I'm not sure we're e.g. writing empty maps as rows vs. not writing them). I think the best fix long term is to make sure we always have a row written, but since we can't easily differentiate between empty payload and didn't find a payload, having this seems like the best we can do right now (at least to me). 

I'll need to think more about how to make this less brittle, because it's a mess

pkg/repository/durable_events.go
mrkaye97: gotcha, I _think_ I implemented this right this time but would be good to get 👀 here

internal/services/shared/tasktypes/v1/task.go
```diff
@@ -214,3 +214,26 @@ type CandidateFinalizedPayload struct {
 	// (required) the workflow run id (can either be a workflow run id or single task)
 	WorkflowRunId uuid.UUID `validate:"required"`
 }
+
+type DurableCallbackCompletedPayload struct {
+	TaskExternalId  uuid.UUID
+	NodeId          int64
+	InvocationCount int64
+	Payload         []byte
+}
```
abelanger5, line 223: Do we want to include the destination worker ID which we used to discover the dispatcher, so that the dispatcher can reject the message in the case of an incorrect worker ID?

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+	return nil
+}
+
+func (tc *TasksControllerImpl) processSingleSatisfiedCallback(ctx context.Context, tenantId uuid.UUID, cb v1.SatisfiedCallback) error {
+	if cb.DispatcherId == nil {
+		return fmt.Errorf("callback has no dispatcher_id set")
+	}
+
+	dispatcherId := *cb.DispatcherId
```
abelanger5: Could you help me trace out where we're getting this dispatcher ID from? I am assuming that when we write the callback we are storing the dispatcher ID. It seems exceedingly likely that we will send the callback to the wrong dispatcher ID since it can be reassigned many times over the course of a task (and due to hatchet engine deployments).

Prior art is that we store the information in the database about which worker is running which durable task, and then query right before we send to the dispatcher, to avoid as many false positives. 

Why don't we store the dispatcher id for each worker listener which is currently active and look it up that way, using the `v1_task_runtime` to figure out the worker ID?

mrkaye97: we're sending it through `getOrCreateEventLogCallback` via `IngestDurableTaskEvent` which we call from the dispatcher (where we have the id).

sounds like a good idea to me to store the worker instead, I can fix that 👍

abelanger5 (approved):

PR #2959, opened 2026-02-05
feat: add python and typescript webhook client

jishnundth:
# Description
1. Add python and TS webhook clients, to directly access them through hatchet client. 
2. EventKeyExpression optional for webhook update. (DB update retains the old value in case input is null)

## Type of change


sdks/python/hatchet_sdk/features/webhooks.py
```diff
@@ -0,0 +1,290 @@
...
+                        authType="API_KEY",
+                        auth=V1WebhookAPIKeyAuth(
+                            headerName=header_name,
+                            apiKey=api_key,
+                        ),
+                    )
+                ),
+            )
+
+    async def aio_create_with_api_key(
```
mrkaye97: I'm a little early here, but one note! we needed to do it this way in Go because Go's type system doesn't cleanly support discriminated unions (🤮), but Python and TS do! do I think instead of separate methods, we should be able to have a single `create` / `aio_create` that takes a Pydantic model called `auth` that has a discriminator for the type (e.g. basic, api key, hmac), and then has the correct params for each case.

https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions-with-str-discriminators

Edit: It might be even simpler to allow one of a union of three models, like `Basic | ApiKey | HMAC` without the discriminator - whichever works!

mrkaye97: (TS will be even easier)

jishnundth: Hey, Thank you for this early review. I'll check and refactor it.

sdks/python/hatchet_sdk/features/webhooks.py
```diff
@@ -0,0 +1,178 @@
...
+            )
+            return v1_webhook_get(
+                tenant=self.tenant_id,
+                v1_webhook=webhook_name,
+            )
+
+    async def aio_get(self, webhook_name: str) -> V1Webhook:
+        return await asyncio.to_thread(self.get, webhook_name)
+
+    def create(self, request: CreateWebhookRequest) -> V1Webhook:
```
mrkaye97: sorry for not being clear enough here! I think we should try to keep consistent with the other feature clients and use separate params instead of a single pydantic model. I edited my comment from before, but it might end up looking like:

```python
def create(
    self,
    name: str,
    auth: BasicAuth | APIKeyAuth | HMACAuth,
    ...
) -> ...
    ...
```

sdks/python/examples/webhooks/test_webhooks.py
```diff
@@ -14,19 +14,6 @@
...
-)
-from hatchet_sdk.clients.rest.models.v1_create_webhook_request_api_key import (
-    V1CreateWebhookRequestAPIKey,
-)
-from hatchet_sdk.clients.rest.models.v1_create_webhook_request_basic_auth import (
-    V1CreateWebhookRequestBasicAuth,
-)
-from hatchet_sdk.clients.rest.models.v1_create_webhook_request_hmac import (
-    V1CreateWebhookRequestHMAC,
-)
```
mrkaye97, line 29: 🙌

mrkaye97 (commented):
left one small comment on the signature of the create method, but otherwise this looks great to me!

would you mind writing changelog entries for TS and Python? And also, you'll need to bump the minor version of the TS package (in `package.json`) the same way as in the `pyproject.toml`

jishnundth:
> left one small comment on the signature of the create method, but otherwise this looks great to me!
> 
> would you mind writing changelog entries for TS and Python? And also, you'll need to bump the minor version of the TS package (in `package.json`) the same way as in the `pyproject.toml`

Done. There's some tests failing, should pass with retries I think.

mrkaye97:
> > left one small comment on the signature of the create method, but otherwise this looks great to me!
> > would you mind writing changelog entries for TS and Python? And also, you'll need to bump the minor version of the TS package (in `package.json`) the same way as in the `pyproject.toml`
> 
> Done. There's some tests failing, should pass with retries I think.

Thanks! I think we might've broken the generation one, I'll look into that

sdks/python/hatchet_sdk/features/webhooks.py
```diff
@@ -0,0 +1,214 @@
...
+    async def aio_get(self, webhook_name: str) -> V1Webhook:
+        return await asyncio.to_thread(self.get, webhook_name)
+
+    def create(
+        self,
+        source_name: V1WebhookSourceName,
+        name: str,
+        event_key_expression: str,
+        auth_type: Literal["BASIC", "API_KEY", "HMAC"],
+        auth: V1WebhookBasicAuth | V1WebhookAPIKeyAuth | V1WebhookHMACAuth,
```
mrkaye97: I thought I commented here but maybe it got lost, but I think one more tiny thing is we don't need the `auth_type` param here, since we can infer it from the `auth` one. e.g. `V1WebhookBasicAuth` -> `BASIC`, etc.

jishnundth: Done. Right now go and python SDK are consistent because the auth_type does not need to be parsed. But it's not the case for the TS SDK. ~Should I remove it for TS as well ?~

I just added the logic for TS as well to make it consistent.

mrkaye97: sure, thanks!

jishnundth: auth type magic strings are removed. 🙌🏻

mrkaye97 (approved):
looks great, nice work! I'll get this shipped once CI passes 🥳

promptless-for-oss:

New suggestion: **[Document webhooks client for Python SDK](https://app.gopromptless.ai/suggestions/432d67dd-1ee7-4fb8-9e2a-c8e6c73ffdfb)**

---

promptless-for-oss:

New suggestion: **[Document webhooks client for Python and TypeScript SDKs](https://app.gopromptless.ai/suggestions/c9dfc3ac-f2ad-480c-aacc-698f3081f486)**

---

PR #2960, opened 2026-02-05
chore: deprecate  ts v0

grutt:
# Description

Removes v0 typescript code

## Type of change

- [x] Refactor (non-breaking changes to code which doesn't change any behavior)


promptless-for-oss:

New suggestion: **[Update migration guides to reflect V0 SDK deprecation](https://app.gopromptless.ai/suggestions/b3dc45e0-3f4f-4eb1-8589-938a76a96139)**

---

grutt:
I've restored the sdks/typescript/src/examples/ directory from main and confirmed no type errors and workflows can continue to run

sdks/typescript/src/clients/hatchet-client/hatchet-client.ts
mrkaye97, line 23: I'm a little confused on the state of the world so looking for clarification here - if this is `Legacy`, should we be removing it? and if not, can we make it into `HatchetClient`? I feel like one of the things we need to fix here is importing `LegacyFoobar`, or `FoobarV0`, etc. in different places but ideally if we're getting rid of all the V0 stuff, we could also consolidate those

WDYT?

grutt: good call out... unfortunately this is for the multi-slot durable fallback so it'll have stick around.

i think we should just be importing the LegacyWorkflow type in a variety of places which we'll need to keep around as well because this gets transformed internally

ill consider how to rename these to be a bit clearer

grutt: actually i think i might have been confused here.. i think i can remove this

mrkaye97: Cool, yeah sounds good, consolidating things like `LegacyWorkflow` to work more similar to Python's `BaseWorkflow` would be great IMO

sdks/typescript/.npmrc
```diff
@@ -0,0 +1 @@
+minimumReleaseAge=1440
```
mrkaye97: I think we're using the `pnpm-workspace.yaml` for this nowadays

mrkaye97 (approved):

sdks/typescript/src/clients/hatchet-client/hatchet-client.ts
grutt: Ok, couldn't remove it completely but the only refs to it should be the other legacy things and moved over all core services to the main client.

promptless-for-oss:

New suggestion: **[Add TypeScript V0 SDK deprecation notice and migration guide updates](https://app.gopromptless.ai/suggestions/12f2dcaf-1492-4772-a4b6-894fa7fb8377)**
Updated existing suggestion: **[Add TypeScript V0 SDK deprecation notice and migration details](https://app.gopromptless.ai/suggestions/4bebc838-b8be-4767-9970-f944f183d460)**

---

PR #2965, opened 2026-02-06
Feat  improved py cancellation

grutt:
# Description

Improves python cancellation signaling

## Type of change


frontend/docs/pages/home/cancellation.mdx
```diff
@@ -22,6 +22,48 @@ When a task is canceled, Hatchet sends a cancellation signal to the task. The ta
...
+```python
+from hatchet_sdk import CancelledError, CancellationReason
+
+@hatchet.task()
+def my_task(input: MyInput, ctx: Context) -> dict:
+    try:
+        result = child_workflow.run(input)
+    except CancelledError as e:
+        # Handle cancellation - i.e. perform cleanup, then re-raise
+        print(f"Task cancelled: {e.reason}")
```
mrkaye97: is this catching when the _child_ is cancelled, or when the _parent_ is cancelled?

frontend/docs/pages/home/cancellation.mdx
```diff
@@ -22,6 +22,48 @@ When a task is canceled, Hatchet sends a cancellation signal to the task. The ta
...
+        # Always re-raise CancelledError so Hatchet can properly handle the cancellation
+        raise
+    except Exception as e:
+        # This will NOT catch CancelledError
+        print(f"Other error: {e}")
+        raise
```
mrkaye97: nit: if we end up re-raising, we should remove this

frontend/docs/pages/home/cancellation.mdx
```diff
@@ -22,6 +22,48 @@ When a task is canceled, Hatchet sends a cancellation signal to the task. The ta
...
+  handlers. This is intentional and mirrors the behavior of Python's
+  `asyncio.CancelledError`.
+</Callout>
+
+```python
+from hatchet_sdk import CancelledError, CancellationReason
+
+@hatchet.task()
+def my_task(input: MyInput, ctx: Context) -> dict:
+    try:
```
mrkaye97: let's make this into a generated example instead of inline python, since this won't lint as-is :)

sdks/python/examples/simple/worker.py
```diff
@@ -1,17 +1,21 @@
...
 
-from hatchet_sdk import Context, EmptyModel, Hatchet
+from hatchet_sdk import Context, DurableContext, EmptyModel, Hatchet
 
 hatchet = Hatchet(debug=True)
 
 
 @hatchet.task()
 def simple(input: EmptyModel, ctx: Context) -> dict[str, str]:
+    time.sleep(50)
```
mrkaye97: can we remove this? this is the example I always run to test stuff locally and don't want it to sleep for a minute 😅

sdks/python/examples/simple/worker.py
```diff
@@ -1,17 +1,21 @@
...
     return {"result": "Hello, world!"}
 
 
 @hatchet.durable_task()
-def simple_durable(input: EmptyModel, ctx: Context) -> dict[str, str]:
+async def simple_durable(input: EmptyModel, ctx: DurableContext) -> dict[str, str]:
+    res = await simple.aio_run(input)
```
mrkaye97: once my changes merge (tbd since I haven't thought through this that much) this will probably be `ctx.spawn_child` instead

sdks/python/hatchet_sdk/clients/listeners/pooled_listener.py
```diff
@@ -221,8 +237,42 @@ async def subscribe(self, id: str) -> T:
             if not self.listener_task or self.listener_task.done():
                 self.listener_task = asyncio.create_task(self._init_producer())
 
-            return await self.events[subscription_id].get()
+            logger.debug(
+                f"PooledListener.subscribe: waiting for event on id={id}, "
+                f"subscription_id={subscription_id}, token={cancellation_token is not None}"
+            )
```
mrkaye97, line 243: claude really likes adding these sorts of debug lines but I think we should be careful about adding these, since I think a lot of people might have `debug=True` set, and these are really internal logs that aren't all that helpful to the user

sdks/python/hatchet_sdk/clients/listeners/pooled_listener.py
```diff
@@ -221,8 +237,42 @@ async def subscribe(self, id: str) -> T:
...
+                done, pending = await asyncio.wait(
+                    [cancel_task, result_task],
+                    return_when=asyncio.FIRST_COMPLETED,
+                )
+
+                # Cancel pending tasks
+                for task in pending:
+                    task.cancel()
+                    with contextlib.suppress(asyncio.CancelledError):
+                        await task
```
mrkaye97: what happens here if someone handles `asyncio.CancelledError` in their task? could this hang until that finishes?

sdks/python/hatchet_sdk/clients/listeners/pooled_listener.py
```diff
@@ -221,8 +237,42 @@ async def subscribe(self, id: str) -> T:
...
+                    )
+                    raise asyncio.CancelledError(
+                        "Operation cancelled by cancellation token"
+                    )
+
+                logger.debug(f"PooledListener.subscribe: received event for id={id}")
+                return result_task.result()
+            result = await self.events[subscription_id].get()
+            logger.debug(f"PooledListener.subscribe: received event for id={id}")
+            return result
```
mrkaye97:
```suggestion
            return await self.events[subscription_id].get()
```

sdks/python/hatchet_sdk/clients/listeners/pooled_listener.py
```diff
@@ -221,8 +237,42 @@ async def subscribe(self, id: str) -> T:
...
         except asyncio.CancelledError:
+            logger.debug(f"PooledListener.subscribe: externally cancelled for id={id}")
```
mrkaye97, line 250: I'd remove this one too

sdks/python/hatchet_sdk/context/context.py
```diff
@@ -169,36 +196,53 @@ def workflow_run_id(self) -> str:
...
+        self.cancellation_token.cancel(reason)
 
     def cancel(self) -> None:
         """
         Cancel the current task run. This will call the Hatchet API to cancel the step run and set the exit flag to True.
 
         :return: None
         """
-        logger.debug("cancelling step...")
+        logger.debug(f"Context: cancel() called for step_run_id={self.step_run_id}")
```
mrkaye97: nit: `task` here instead of `step`?

sdks/python/hatchet_sdk/context/context.py
```diff
@@ -169,36 +196,53 @@ def workflow_run_id(self) -> str:
...
+        self._set_cancellation_flag(CancellationReason.USER_REQUESTED)
 
     async def aio_cancel(self) -> None:
         """
         Cancel the current task run. This will call the Hatchet API to cancel the step run and set the exit flag to True.
 
         :return: None
         """
-        logger.debug("cancelling step...")
+        logger.debug(f"Context: aio_cancel() called for step_run_id={self.step_run_id}")
```
mrkaye97: here too re: task / step

sdks/python/hatchet_sdk/context/context.py
```diff
@@ -169,36 +196,53 @@ def workflow_run_id(self) -> str:
...
+        self._set_cancellation_flag(CancellationReason.USER_REQUESTED)
 
     def done(self) -> bool:
         """
         Check if the current task run has been cancelled.
 
         :return: True if the task run has been cancelled, False otherwise.
         """
-        return self.exit_flag
+        return self.cancellation_token.is_cancelled
```
mrkaye97: I think we should revert this change? no reason to have the duplication here imo (in 2.0.0 I'll remove this entirely)

sdks/python/hatchet_sdk/context/context.py
```diff
@@ -502,20 +553,30 @@ async def aio_wait_for(
...
         self.durable_event_listener.register_durable_event(request)
 
-        return await self.durable_event_listener.result(
-            task_id,
-            signal_key,
+        # Use await_with_cancellation to respect the cancellation token
+        return await await_with_cancellation(
+            self.durable_event_listener.result(task_id, signal_key),
+            self.cancellation_token,
         )
```
mrkaye97, line 564: just a note that this will be moot once my change merges

sdks/python/hatchet_sdk/runnables/workflow.py
```diff
@@ -621,6 +628,44 @@ def greet(input, ctx):
     and can be arranged into complex dependency patterns.
     """
 
+    def _resolve_check_cancellation_token(self) -> CancellationToken | None:
```
mrkaye97, line 635: I'm not convinced that the workflow is the right place for these methods, but also might not be completely understanding them (can circle back to this)

sdks/python/hatchet_sdk/runnables/workflow.py
```diff
@@ -621,6 +628,44 @@ def greet(input, ctx):
...
+    def _register_child_with_token(
+        self,
+        cancellation_token: CancellationToken | None,
+        workflow_run_id: str,
+        log_prefix: str,
+    ) -> None:
+        if not cancellation_token:
+            return
+
+        logger.debug(f"{log_prefix}: registered child {workflow_run_id} with token")
```
mrkaye97: I think we should remove this and remove the `log_prefix` arg that we pass around

sdks/python/hatchet_sdk/runnables/workflow.py
mrkaye97: also, regardless of that ^, `_resolve_check_cancellation_token` is an awkward name imo

sdks/python/hatchet_sdk/runnables/workflow.py
```diff
@@ -742,15 +860,58 @@ def run_many(
...
         :returns: A list of results for each workflow run.
+        :raises CancelledError: If the cancellation token is triggered (and return_exceptions is False).
+        :raises Exception: If a workflow run fails (and return_exceptions is False).
         """
+        cancellation_token = self._resolve_check_cancellation_token()
+
+        logger.debug(
+            f"Workflow.run_many: triggering {len(workflows)} workflows, "
+            f"token={cancellation_token is not None}"
+        )
```
mrkaye97: would remove this too - tbh I'd just remove basically all of these debug lines

sdks/python/hatchet_sdk/runnables/workflow.py
```diff
@@ -742,15 +860,58 @@ def run_many(
...
+                        )
+                    )
+                    break
+                raise
+            except Exception as e:
+                if return_exceptions:
+                    results.append(e)
+                else:
+                    raise
+        return results
```
mrkaye97, line 902: I'm not totally sure this logic is correct - I don't think we should ever `break` out of this loop (the list might just end up having both exceptions and results, which is okay)

mrkaye97: I also am a bit unclear on why we replaced `_get_result` with `result()` here

sdks/python/hatchet_sdk/utils/cancellation.py
```diff
@@ -0,0 +1,123 @@
...
+) -> T:
+    """
+    Await an awaitable with cancellation support.
+
+    This function races the given awaitable against a cancellation token. If the
+    token is cancelled before the awaitable completes, the awaitable is cancelled
+    and an asyncio.CancelledError is raised.
+
+    Args:
+        coro: The awaitable to await (coroutine, Future, or Task).
```
mrkaye97: nit: might be good to clarify this is `asyncio.Task`

sdks/python/hatchet_sdk/utils/cancellation.py
```diff
@@ -0,0 +1,123 @@
...
+        ```
+    """
+
+    async def _invoke_cancel_callback() -> None:
+        if not cancel_callback:
+            return
+
+        result = cancel_callback()
+        if inspect.isawaitable(result):
+            await result
```
mrkaye97: Not sure I follow the reason for this approach - we know from the type hint that the result is awaitable, so we should be able to simplify this to `await cancel_callback()` afaict

sdks/python/hatchet_sdk/utils/cancellation.py
```diff
@@ -0,0 +1,123 @@
...
+        result = await await_with_cancellation(
+            long_running_task(),
+            token,
+            cancel_callback=cleanup,
+        )
+        ```
+    """
+
+    async def _invoke_cancel_callback() -> None:
+        if not cancel_callback:
```
mrkaye97: can we factor this out so we don't have the closure here? I don't think we really need this to be a closure unless I'm missing something

sdks/python/hatchet_sdk/utils/cancellation.py
```diff
@@ -0,0 +1,123 @@
...
+
+    logger.debug("await_with_cancellation: starting with cancellation token")
+
+    # Check if already cancelled
+    if token.is_cancelled:
+        logger.debug("await_with_cancellation: token already cancelled")
+        if cancel_callback:
+            logger.debug("await_with_cancellation: invoking cancel callback")
+            await _invoke_cancel_callback()
+        raise asyncio.CancelledError("Operation cancelled by cancellation token")
```
mrkaye97, line 134: do we want to raise an asyncio cancellederror here, or our own cancellederror?

sdks/python/hatchet_sdk/exceptions.py
```diff
@@ -171,3 +172,57 @@ class IllegalTaskOutputError(Exception):
...
+    """The workflow run was cancelled."""
+
+    TOKEN_CANCELLED = "token_cancelled"
+    """The cancellation token was cancelled."""
+
+    UNKNOWN = "unknown"
+    """Unknown or unspecified cancellation reason."""
+
+
+class CancelledError(BaseException):
```
mrkaye97, line 195: I wonder if this should inherit from `asyncio.CancelledError` instead, but I'm not sure 🤔

sdks/python/hatchet_sdk/utils/cancellation.py
```diff
@@ -0,0 +1,123 @@
...
+
+        if cancel_task in done:
+            logger.debug("await_with_cancellation: cancelled before completion")
+            if cancel_callback:
+                logger.debug("await_with_cancellation: invoking cancel callback")
+                await _invoke_cancel_callback()
+            raise asyncio.CancelledError("Operation cancelled by cancellation token")
+
+        logger.debug("await_with_cancellation: completed successfully")
+        return main_task.result()
```
mrkaye97: this logic looks familiar from before - can we factor it out into a helper?

sdks/python/hatchet_sdk/worker/runner/utils/capture_logs.py
```diff
@@ -24,6 +27,9 @@
     LogLevel,
 )
 
+if TYPE_CHECKING:
+    pass
```
mrkaye97: lol
```suggestion
```

sdks/python/hatchet_sdk/worker/runner/utils/capture_logs.py
```diff
@@ -48,10 +54,22 @@ class ContextVarToCopyDict(BaseModel):
     value: JSONSerializableMapping | None
 
 
+class ContextVarToCopyToken(BaseModel):
+    """Special type for copying CancellationToken to threads."""
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    name: Literal["ctx_cancellation_token"]
+    value: Any  # CancellationToken, but using Any to avoid Pydantic issues with custom classes
```
mrkaye97: I think Pydantic should be fine with this?

sdks/python/hatchet_sdk/cancellation.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def get_child_run_ids(self) -> list[str]:
+        """
+        Get a copy of the registered child run IDs.
+
+        Returns:
+            A list of child workflow run IDs.
+        """
+        with self._lock:
+            return self._child_run_ids.copy()
```
mrkaye97: can this be a property that returns `self._child_run_ids`? also, why do we need the lock and copy here?

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -480,28 +491,100 @@ def force_kill_thread(self, thread: Thread) -> None:
...
+            # Wait until warning threshold
+            await asyncio.sleep(warning_threshold)
+            elapsed = time.monotonic() - start_time
+            elapsed_ms = round(elapsed * 1000)
 
-                if self.config.enable_force_kill_sync_threads:
-                    self.force_kill_thread(thread)
-                    await asyncio.sleep(1)
+            # Check if task is still running after warning threshold
+            task_was_running = key in self.tasks and not self.tasks[key].done()
```
mrkaye97: we should probably check the exit flag here instead, the naming is so confusing - this is checking if there's a current task and the task isn't cancelled, but it reads as checking if the task isn't "done"

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -480,28 +491,100 @@ def force_kill_thread(self, thread: Thread) -> None:
...
+            if task_was_running:
                 logger.warning(
-                    f"thread {self.threads[key].ident} with key {key} is still running after cancellation. This could cause the thread pool to get blocked and prevent new tasks from running."
+                    f"Cancellation: task {action.action_id} has not cancelled after "
+                    f"{elapsed_ms}ms (warning threshold {warning_threshold_ms}ms). "
+                    f"Consider checking for blocking operations. "
+                    f"See https://docs.hatchet.run/home/cancellation"
                 )
+
+                # Continue waiting until grace period only if task is still running
```
mrkaye97: can we remove some of these comments? I don't think they're helpful and they add a lot of clutter

sdks/python/hatchet_sdk/cancellation.py
```diff
@@ -0,0 +1,195 @@
...
+    """
+
+    def __init__(self) -> None:
+        self._cancelled = False
+        self._reason: CancellationReason | None = None
+        self._async_event: asyncio.Event | None = None
+        self._sync_event = threading.Event()
+        self._child_run_ids: list[str] = []
+        self._callbacks: list[Callable[[], None]] = []
+        self._lock = threading.Lock()
```
mrkaye97, line 51: how does the async code play with `threading.Lock`?

sdks/python/hatchet_sdk/cancellation.py
```diff
@@ -0,0 +1,195 @@
...
+
+            self._cancelled = True
+            self._reason = reason
+
+            # Signal both event types
+            if self._async_event is not None:
+                self._async_event.set()
+            self._sync_event.set()
+
+            # Invoke callbacks
```
mrkaye97: I'd remove this one too (and a bunch of the others) - not helpful IMO, very obvious that this code that calls `callback()` for each callback is in fact invoking the callbacks 🤦

sdks/python/hatchet_sdk/exceptions.py
```diff
@@ -171,3 +172,57 @@ class IllegalTaskOutputError(Exception):
...
+    """The parent workflow or task was cancelled."""
+
+    WORKFLOW_CANCELLED = "workflow_cancelled"
+    """The workflow run was cancelled."""
+
+    TOKEN_CANCELLED = "token_cancelled"
+    """The cancellation token was cancelled."""
+
+    UNKNOWN = "unknown"
+    """Unknown or unspecified cancellation reason."""
```
mrkaye97: should we allow this? maybe it's a better idea to just assume `USER_REQUESTED` unless there's some other reason provided?

sdks/python/tests/test_cancellation.py
mrkaye97, line 1: same comment as before on these tests 😅 I think it's fine to write these sorts of unit tests, but we really need some end to end tests that test these code paths in a real task setting (can double as examples)

frontend/docs/pages/home/cancellation.mdx
grutt: its when the parent is cancelled, i'll clarify in the example

sdks/python/hatchet_sdk/clients/listeners/pooled_listener.py
grutt: i actually had claude add these because it makes tracing these cancellation stacks much easier.. i think they are helpful fwiw but maybe we can consider namespacing debug lines in the future or something...

i would like to keep them

sdks/python/tests/test_cancellation.py
grutt: yeah i really want to just test all of this with durable execution examples once all that is sorted tbh

sdks/python/hatchet_sdk/worker/runner/runner.py
grutt: the intent her is to ensure we have indeed completed the task so i think it needs to use aio's done. but agreed on naming.

sdks/python/hatchet_sdk/exceptions.py
grutt: i think no, because im not sure how it would potentially interact with asyncio

grutt: also this is intended for only sync code

sdks/python/hatchet_sdk/utils/cancellation.py
grutt: yes asyncio since this is async

sdks/python/hatchet_sdk/utils/cancellation.py
grutt: good call

sdks/python/hatchet_sdk/cancellation.py
grutt: this should be fine afict i refactored some of the callback stuff to minimize any contention

sdks/python/hatchet_sdk/cancellation.py
grutt: good call, cleaning this up

mrkaye97 (approved):
lgtm pending tests + a minor version bump and changelog entry

promptless-for-oss:

New suggestion: **[Document CancelledError exception and CancellationReason enum for Python SDK](https://app.gopromptless.ai/suggestions/9892add2-56e9-423f-ae35-9dfe33e23c3b)**

---

promptless-for-oss:

New suggestion: **[Document Python SDK cancellation improvements (CancelledError, CancellationReason, config options)](https://app.gopromptless.ai/suggestions/79f6468a-f675-4167-a0e1-3b692cfc51af)**

---

PR #2968, opened 2026-02-07
test(rest): add typed REST transport exceptions + diagnostics tests

BloggerBust:
Adds typed REST transport exceptions for urllib3/network failures and unit tests verifying translation, diagnostics, and backward compatibility. Updates apply_patches.py to patch generated REST client code and exceptions.

Follow-up to: #2872

# Description

This PR introduces typed REST transport exceptions for the Python SDK to replace the generic `ApiException` previously raised for urllib3/network-level failures.

Specifically:
- Adds a `RestTransportError` base class and concrete subtypes for TLS, timeout, connection, and protocol errors
- Updates the generated REST client error handling to translate urllib3 exceptions into the appropriate typed exception
- Preserves backward compatibility: all new exceptions inherit from `ApiException`
- Improves diagnostic messages by consistently including method, URL, and timeout
- Adds comprehensive unit tests covering translation, diagnostics, and backward compatibility
- Updates `apply_patches.py` so regenerated clients receive the new imports and handlers idempotently

No retry behavior is changed in this PR.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Test changes (add, refactor, improve or change a test)

## What's Changed

- Add typed REST transport exception hierarchy (`RestTransportError`, `RestTLSError`, `RestTimeoutError`, `RestConnectionError`, `RestProtocolError`)
- Translate urllib3 exceptions into typed REST exceptions in the generated REST client
- Ensure exception handler ordering is correct for urllib3 inheritance behavior
- Add unit tests covering:
  - Exception translation
  - Diagnostic message contents
  - Backward compatibility with `ApiException`
- Update `apply_patches.py` to patch generated REST clients and exception imports idempotently

Issue #2872, opened 2026-01-28
[FEAT] Improve Python SDK reliability, retries, and failure diagnostics


TheEleventhAvatar:
The Python SDK can currently surface generic or low-context errors when encountering
common failure scenarios such as server unavailability, timeouts, or partial workflow
registration failures. In both local development and production environments, this
can make it difficult for users to quickly determine whether an issue is caused by
connectivity, configuration problems, or transient infrastructure issues.

I’d like to improve the reliability and debuggability of the Python SDK by making
these failure modes more explicit and easier to reason about. This would include
centralizing request error handling, introducing clearer and more actionable
exceptions for common scenarios, adding configurable timeout and retry behavior
where appropriate, and improving healthcheck robustness while preserving backward
compatibility with existing SDK usage.

While documenting common failure cases is an alternative, improving behavior directly
in the SDK provides better defaults and faster feedback for users running Hatchet
workflows in production. I’m happy to approach this incrementally and align on scope
before implementation, and wanted to confirm that this is an area where contributions
and potential longer-term ownership would be welcome.

mrkaye97:
hey @TheEleventhAvatar, would love to see a detailed sketch of your vision here before implementation to get a sense for what you’re thinking and make sure we’re aligned, but definitely agree with the broader goal!

TheEleventhAvatar:
That sounds good, happy to sketch it out.

At a high level, my thinking is to keep this incremental and non-invasive, focusing
first on making existing failure modes easier to understand rather than changing
behavior.

The initial scope I have in mind would be:
- Introduce a small, explicit exception hierarchy for the Python SDK (e.g. connection
  errors, timeouts, unexpected server responses) to replace generic exceptions while
  preserving existing control flow.
- Centralize request execution and error translation in a single place, so retries,
  timeouts, and diagnostics aren’t scattered across the SDK.
- Add conservative, opt-in timeout and retry defaults that can be configured, but do
  not change behavior for existing users unless explicitly enabled.
- Improve healthcheck failure reporting to clearly distinguish between connectivity
  issues and server-side readiness.

I’d plan to tackle this in small PRs (starting with exception normalization and
centralized request handling), and avoid any breaking changes. If helpful, I can write
up a more concrete proposal or pseudo-code for the first step before starting.

Let me know if this direction aligns with how you’re thinking about it, or if there’s
a specific area you’d prefer to prioritize first.

mrkaye97:
[...]

PR #2942, opened 2026-02-04
feat(sdk): improve python REST client transport error diagnostics

BloggerBust:
Refs #2872

# Description

This PR improves failure diagnostics in the Python SDK REST client by expanding
the set of transport level exceptions captured and enriching the resulting
`ApiException` message with request context.

Specifically, when low level `urllib3` transport errors occur, such as timeouts,
connection failures, protocol errors, the raised exception now includes:

- the underlying exception type
- the original error message
- HTTP method
- request URL
- configured request timeout

This makes timeout and connectivity failures significantly easier to diagnose
without changing retry behavior or public APIs.

This addresses part of the diagnostics concerns raised in #2872.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## What's Changed

- Expand the set of `urllib3` transport exceptions handled by the REST client
- Improve error messages for transport failures with request context
- Apply the change via the existing Python SDK patching mechanism

PR #2971, opened 2026-02-08
test(sdk): cover python REST and gRPC retry predicate behavior

BloggerBust:
**Follow-up to:** #2872 ( test coverage of current behavior)
**stacked on PR:** #2968 

# Description

This PR adds explicit, parameterized unit tests documenting the current retry behavior of the Python SDK’s tenacity retry predicate for both REST and gRPC calls.

The tests make retry semantics clear and intentional without changing any runtime behavior. In particular, they codify which exception types and gRPC status codes are retried vs not retried today.

This PR strictly adds tests documenting current behavior and does not modify retry logic.

### Context / Motivation

Retry behavior is a sensitive area with implications for idempotency, load amplification, and user expectations. Before changing retry semantics (as discussed in issue #2872), it’s important to have clear test coverage that documents existing behavior.

These tests provide that baseline and make future policy changes deliberate and reviewable.

### Relationship to other work

This PR is stacked on top of the earlier Python SDK PR: #2968 

It is intended to be reviewed either:
- after #2968, or
- independently, as it only adds tests.

It also provides groundwork for follow-up discussion and potential changes related to:
- https://github.com/hatchet-dev/hatchet/issues/2872
---

## Type of change

---

## What's Changed

- [x] Added parameterized unit tests for the Python SDK tenacity retry predicate
- [x] Covered REST exception retry behavior (5xx, 404, other 4xx, transport errors)
- [x] Covered generic exception behavior (no retries)
- [x] Covered gRPC retry behavior across retried vs non-retried status codes
- [x] Documented current retry semantics without changing implementation

### DIFF
--- a/sdks/python/CHANGELOG.md
+++ b/sdks/python/CHANGELOG.md
@@ -6,19 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
-## [1.23.4] - 2026-02-12
-
-### Changed
-
-- Adds parameterized unit tests documenting current retry behavior of the Python SDK’s tenacity retry predicate for REST and gRPC errors.
-
-
 ## [1.23.3] - 2026-02-12
 
 ### Added
 
 - Adds type-hinted `Standalone.output_validator` and `Standalone.output_validator_type` properties to support easier type-safety and match the `input_validator` property pattern on `BaseWorkflow`.
 - Adds type-hinted `Task.output_validator` and `Task.output_validator_type` properties to support easier type-safety and match the patterns on `BaseWorkflow/Standalone`.
+- Adds parameterized unit tests documenting current retry behavior of the Python SDK’s tenacity retry predicate for REST and gRPC errors.
 
 
 ## [1.23.2] - 2026-02-11
--- a/sdks/python/CHANGELOG.md
+++ b/sdks/python/CHANGELOG.md
@@ -5,12 +5,21 @@ All notable changes to Hatchet's Python SDK will be documented in this changelog
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [1.23.4] - 2026-02-12
+
+### Changed
+
+- Adds parameterized unit tests documenting current retry behavior of the Python SDK’s tenacity retry predicate for REST and gRPC errors.
+
+
 ## [1.23.3] - 2026-02-12
 
 ### Added
 
 - Adds type-hinted `Standalone.output_validator` and `Standalone.output_validator_type` properties to support easier type-safety and match the `input_validator` property pattern on `BaseWorkflow`.
 - Adds type-hinted `Task.output_validator` and `Task.output_validator_type` properties to support easier type-safety and match the patterns on `BaseWorkflow/Standalone`.
+
 
 ## [1.23.2] - 2026-02-11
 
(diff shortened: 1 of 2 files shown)


sdks/python/hatchet_sdk/clients/rest/rest.py
```diff
@@ -239,14 +246,47 @@ def request(
...
+            msg = "\n".join(
+                [
+                    type(e).__name__,
+                    str(e),
+                    f"method={method}",
+                    f"url={url}",
+                    f"timeout={_request_timeout}",
+                ]
+            )
+            raise RestTLSError(status=0, reason=msg) from e
```
mrkaye97, line 259: qq: I suspect it's very unlikely that this matters much in practice, but this is a breaking change right? since we're re-raising a different class of exception here, and hypothetically someone could have been catching `SSLError` explicitly (again, highly unlikely, but I think important to ask)

BloggerBust: Good question, this shouldn’t be a breaking change in practice.

We already wrapped urllib3.exceptions.SSLError into an SDK exception in #2942 (raising `ApiException(status=0, ...))`, so user code couldn’t reliably catch the raw `SSLError` before this change either.

In #2968 we’re keeping that behavior (still no raw `SSLError` being raised), but making it more specific by raising `RestTLSError` (which subclasses `ApiException`). We also preserve the original exception via `raise ... from e` so it’s accessible as `exc.__cause__`.

If you’d like, I can add a short note to the changelog calling out the new typed transport exceptions, but existing except `ApiException`: handlers will continue to work unchanged.

mrkaye97: Sorry for just circling back! This is fine by me, adding a changelog entry and leaving this how you have it here sounds good 👍

BloggerBust: @mrkaye97 I added a changelog entry for the next patch version.

sdks/python/tests/test_tenacity_utils.py
```diff
@@ -0,0 +1,237 @@
...
+    RestTimeoutError,
+    RestTLSError,
+    RestTransportError,
+    ServiceException,
+    UnauthorizedException,
+)
+from hatchet_sdk.clients.rest.tenacity_utils import tenacity_should_retry
+
+
+class TestRestExceptionRetryPredicate:
```
mrkaye97: I think I made this comment before (maybe I forgot to submit it 😅) - I think we shouldn't wrap these in classes, but just for consistency with other parts of the codebase for now

BloggerBust: Good catch and thanks for the reminder. I’ve unwrapped these into module-level tests for consistency with the rest of the SDK. I also updated the changelog version and moved the entry to the top.

sdks/python/CHANGELOG.md
```diff
@@ -5,13 +5,22 @@ All notable changes to Hatchet's Python SDK will be documented in this changelog
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [1.23.4] - 2026-02-12
+
+### Changed
+
+- Adds parameterized unit tests documenting current retry behavior of the Python SDK’s tenacity retry predicate for REST and gRPC errors.
```
mrkaye97: I think maybe for this one we should just bundle this with the previous patch (CI won't release a new one) since it's just test changes

BloggerBust: Good call! I have bundled the test only changes into 1.23.3 to avoid triggering a patch release. Thanks!

mrkaye97 (approved):

promptless-for-oss:

Updated existing suggestion: **[Document REST transport exceptions for Python SDK](https://app.gopromptless.ai/suggestions/9808640a-7872-4803-86df-f8777ec1e938)**

---
_Tip: Worried about broken links? Ask Promptless to find and fix them automatically 🔗_

PR #2972, opened 2026-02-08
chore(docs): Improvements to CONTRIBUTING docs and Taskfile

gregfurman:
# Description

This PR improves upon the existing `CONTRIBUTING.md` by adding a section on testing, clarifying guidelines for contributors, as well as fixing up outdated/unclear instructions for getting started with running locally.

In addition, the `Taskfile` was modified to include Go testing & formatting, in addition to some tasks being reworked (or omitted) that were deemed unnecessary to getting setup, or otherwise unused.

## Type of change

## What's Changed

- Run `go fmt` (via `task fmt-go`) and commit formatting changes
- Improve `CONTRIBUTING.md` structure and clarity
- Modifications to `Taskfile.yaml`:
   - Add `fmt-go`, `fmt-app`, and `fmt-docs` tasks for code formatting (aliased to `task fmt`)
   - Add `test` and `test-integration` tasks for Go testing
   - Remove unused `generate-go` task
   - Add `dotenv` support to seed tasks instead of `./hack/dev/run-go-with-env.sh` usage
   - Improve task aliases, and simplify `setup` task dependencies


CONTRIBUTING.md
```diff
@@ -1,128 +1,121 @@
...
-SERVER_ENCRYPTION_CLOUDKMS_KEY_URI=gcp-kms://projects/<PROJECT>/locations/global/keyRings/development/cryptoKeys/development
+You should see the following logs if the workflow was started against your local instance successfully:
+```log
+{"level":"debug","service":"client","message":"connecting to 127.0.0.1:7070 without TLS"}
+{"level":"info","service":"client","message":"gzip compression enabled for gRPC client"}
+{"level":"debug","service":"worker","message":"worker simple-worker is listening for actions: [process-message:process-message]"}
+{"level":"debug","service":"client","message":"No compute configs found, skipping cloud registration and running all actions locally."}
+{"level":"debug","service":"client","message":"Registered worker with id: c47cc839-8c3b-4b0f-a904-00e37f164b7d"}
+{"level":"debug","service":"client","message":"Starting to listen for actions"}
+{"level":"debug","service":"client","message":"updating worker c47cc839-8c3b-4b0f-a904-00e37f164b7d heartbeat"}
```
abelanger5, line 115: Small note on this: all of us have the following env vars set and these should probably be generated as part of the quickstart:

```
SERVER_LOGGER_LEVEL=warn
SERVER_LOGGER_FORMAT=console
DATABASE_LOGGER_LEVEL=warn
DATABASE_LOGGER_FORMAT=console
```

It's nearly impossible to reason about the JSON debug logs. The console-level `warn` logs seem like a good middle ground/default for contributors.

CONTRIBUTING.md
abelanger5, line 1: Small note here: while it probably is a good idea to generate _encryption keys_ as part of the setup steps, I think we shouldn't require generating _certificates_ since it's possible to run the engine in "insecure" mode without TLS enabled, and the encryption keys can be a headache to manage. It would also simplify the environment generation, etc. What are your thoughts?

cmd/hatchet-cli/cli/tui/dag/graph_test.go
abelanger5, line 1: Small nit: can we create a separate PR with the `go fmt` changes?

docs/development/updating-configuration.md
abelanger5, line 1: nice, this is definitely a useful page!

hack/dev/init-dev-token-and-env.sh
```diff
@@ -1,8 +1,15 @@
...
+
+get_token() {
+  go run ./cmd/hatchet-admin token create --name local --tenant-id 707d0855-80ab-4e1f-a156-f1c4546cbf52 2>/dev/null
+}
+
+cat <<EOF
 HATCHET_CLIENT_TENANT_ID=707d0855-80ab-4e1f-a156-f1c4546cbf52
 HATCHET_CLIENT_TLS_ROOT_CA_FILE=$PWD/hack/dev/certs/ca.cert
 HATCHET_CLIENT_TLS_SERVER_NAME=cluster
+HATCHET_CLIENT_TLS_STRATEGY=none
```
abelanger5, line 12: Note that this is incongruous with the env vars directly about this (`HATCHET_CLIENT_TLS_ROOT_CA_FILE` and `HATCHET_CLIENT_TLS_SERVER_NAME`) - see note in contributing guide doc

abelanger5 (changes requested):
Nice!

examples/go/migration-guides/mergent.go
```diff
@@ -45,7 +45,6 @@ func ProcessImageMergent(req MergentRequest) (*MergentResponse, error) {
 	}, nil
 }
 
-
```
mnafees, line 48: I think these line breaks are introduced as part of an automated generation script. Can we make sure we would not be getting a duplicate change to reintroduce these in a follow up PR?

Taskfile.yaml
```diff
@@ -56,16 +58,22 @@ tasks:
         SERVER_INTERNAL_CLIENT_BASE_STRATEGY=none
         SERVER_INTERNAL_CLIENT_BASE_INHERIT_BASE=false
         EOF
+  fmt-go:
+    cmd: gofmt -s -w .
```
mnafees, line 57: I wonder if this will work as expected along with `task pre-commit-run` which depends on `pre-commit run` and `golangci-lint`.

CONTRIBUTING.md
```diff
@@ -1,128 +1,121 @@
...
 
-```sh
-task setup
-```
+Next, ensure all changes are:
 
-**_Note: You might need to run this as `sudo` so it can install certificates._**
+- Unit tested with `task test`
+- Linted with `task lint`
+- Formatted with `task fmt`
```
mnafees, line 45: Personally, I think we should make `task pre-commit-run` as the command to depend on for linting and formatting.

Taskfile.yaml
```diff
@@ -1,18 +1,20 @@
 version: "3"
 
+vars:
+  PRE_COMMIT_CMD: '{{.PRE_COMMIT_CMD | default "pre-commit"}}'
```
mnafees: What is the idea behind making this as a global var?

gregfurman: Originally I planned to replace `pre-commit` with  `prek` (https://prek.j178.dev/) which is a rust alternative for pre-commit. That way we wouldn't need an entire python `venv` for a single dependency.

On my local, this command allows me to run:

```sh
task pre PRE_COMMIT_CMD=prek
```

We could localize the variable to the `pre` task instead, if you think that is better?

cmd/hatchet-cli/cli/tui/dag/graph_test.go
gregfurman: Good idea. The diff is pretty large atm so that'll definitely help.

hack/dev/init-dev-token-and-env.sh
gregfurman: Gotcha. So shall I just remove the `HATCHET_CLIENT_TLS_ROOT_CA_FILE` and `HATCHET_CLIENT_TLS_SERVER_NAME` then?

CONTRIBUTING.md
gregfurman: That definitely sounds ideal! The certificate generation seems like overkill (plus it means we don't need sudo for the setup steps 👀 ). Where in the docs could I find how to run hatchet in "insecure" mode?

gregfurman: Related: I also think the dependency on `caddy` could be removed in favour of setting `SERVER_AUTH_COOKIE_DOMAIN` from `app.dev.hatchet-tools.com` to `localhost`, but this requires modifying those those `hack/dev` scripts or implementing the `go run` commands purely in `Taskfile` tasks (which IMO I am in favour of),

The PR was already getting quite large so I can do this in a followup if we think it's a good idea!

CONTRIBUTING.md
gregfurman: There is already a `pre` command for this. Are you suggesting we rename or alias this to also work with the `pre-commit-run`?

IMO I'm personally a fan of providing seperate commands for formatting and linting since it doesn't obfuscate what will be run in the same way a single compound command will.

CONTRIBUTING.md
gregfurman: Gotcha! Will change this my default in the `.env` file

Taskfile.yaml
gregfurman: I think `go fmt` and `golangci-lint` work well together. Do you think the pre-commit hook will conflict with the outcome of `go fmt`?

CONTRIBUTING.md
mnafees: There is already a `task pre-commit-run` command that is meant to be run for lint fixing so that our GitHub Actions workflows do not fail.

In terms of separate commands, I think we can keep the existing separate commands but for the contributing guide itself, it might make sense to at least point to the `task pre-commit-run` command's existence for pull requests.

Taskfile.yaml
mnafees: Got it. I'd be open to using `prek`.

gregfurman: So do you think we can leave the envar approach as-is or would you prefer fully replacing the `pre-commit` usage?

CONTRIBUTING.md
gregfurman: FYI These logs were from the hatchet workflow, not the server. I noticed that the hatchet worker logger is actually not ever configured, unless I've missed something:

See ref to `HATCHET_CLIENT_LOG_LEVEL` here:
https://github.com/hatchet-dev/hatchet/blob/871d63d4296895b9c53f2914d104648aea504c88/frontend/docs/pages/self-hosting/worker-configuration-options.mdx?plain=1#L47

But it's never bound (or even present in the ClientConfig) here:
https://github.com/hatchet-dev/hatchet/blob/871d63d4296895b9c53f2914d104648aea504c88/pkg/config/client/client.go#L66

Should I support setting a client's log level via `HATCHET_CLIENT_LOG_LEVEL` in a follow-up?

CONTRIBUTING.md
gregfurman: > while it probably is a good idea to generate encryption keys as part of the setup steps, I think we shouldn't require generating certificates [...]

Done!

CONTRIBUTING.md
gregfurman: Ahh I see what you're saying. Unsure if instructing a contributor to run this is necessary since isn't the purpose of the `pre-commit` hook that we don't expect anyone to explicitly run it -- instead it should be configured + installed during setup to auto run?

Lmk if you agree with the above. I'm happy to mention (perhaps as a footnote) that this should be configured.

hack/dev/init-dev-token-and-env.sh
gregfurman: Removed these ☝️

cmd/hatchet-cli/cli/tui/dag/graph_test.go
gregfurman: https://github.com/hatchet-dev/hatchet/pull/2984

Taskfile.yaml
mnafees: Let us leave the env var for now but still depend on `pre-commit`. Migration away from `pre-commit` should be another PR perhaps.

CONTRIBUTING.md
gregfurman: OK have updated the Taskfile to make installing `pre-commit` easier in a `venv`. Also, have mentioned this in the contributing guide.



The `fmt` job should now be running `pre-commit-run`. In addition, `pre-commit-run` depends on `pre-commit-install` which will automate the `venv` creation + installation 👇 

```yaml
  venv:
    status: [test -d .venv]
    cmd: python3 -m venv .venv
  pre-commit-install:
    status: [source .venv/bin/activate && pre-commit -V]
    deps: [venv]
    cmds:
      - 'source .venv/bin/activate && pip install pre-commit'
      - 'source .venv/bin/activate && pre-commit install'
  pre-commit-run:
    deps: [pre-commit-install]
    cmd: 'source .venv/bin/activate && pre-commit run --all-files || pre-commit run --all-files'
```

CONTRIBUTING.md
abelanger5: > Related: I also think the dependency on caddy could be removed in favour of setting SERVER_AUTH_COOKIE_DOMAIN from app.dev.hatchet-tools.com to localhost, but this requires modifying those those hack/dev scripts or implementing the go run commands purely in Taskfile tasks (which IMO I am in favour of),

Yeah I think I agree, but would also want @mrkaye97 and @grutt to chime in on that - let's perhaps discuss in the context of a separate PR!

CONTRIBUTING.md
```diff
@@ -1,128 +1,121 @@
...
+Then, before contributing, check out the following sections:
 
-```
-DATABASE_URL='postgresql://hatchet:hatchet@127.0.0.1:5431/hatchet'
-
-SERVER_ENCRYPTION_MASTER_KEYSET_FILE=./hack/dev/encryption-keys/master.key
-SERVER_ENCRYPTION_JWT_PRIVATE_KEYSET_FILE=./hack/dev/encryption-keys/private_ec256.key
-SERVER_ENCRYPTION_JWT_PUBLIC_KEYSET_FILE=./hack/dev/encryption-keys/public_ec256.key
+- [Development Environment Setup](#development-environment-setup)
+- [Contribution Guidelines](#contribution-guidelines)
```
abelanger5: This tag doesn't lead anywhere

CONTRIBUTING.md
```diff
@@ -1,128 +1,121 @@
...
+- [Go 1.25+](https://go.dev/doc/install)
+- [Node.js v18+](https://nodejs.org/en/download)
+   - We recommend using [nvm](https://github.com/nvm-sh/nvm) for managing node versions to match the version defined in [`.nvmrc`](.nvmrc)
+- [pnpm](https://pnpm.io/installation) installed globally (`npm i -g pnpm`)
+- [Docker](https://docs.docker.com/engine/install/)
+- [task](https://taskfile.dev/docs/installation)
+- [protoc](https://grpc.io/docs/protoc-installation/)
+- [Caddy](https://caddyserver.com/docs/install)
+- [goose](https://pressly.github.io/goose/installation/)
+- [atlas](https://atlasgo.io/)
```
abelanger5: I don't think we need atlas anymore? We don't use it anywhere, perhaps we have a task in the taskfile that references it, in which case we should remove it. We use goose for migrations and write them by hand now.

Taskfile.yaml
```diff
@@ -30,13 +26,14 @@ tasks:
         SERVER_ENCRYPTION_MASTER_KEYSET_FILE=./hack/dev/encryption-keys/master.key
         SERVER_ENCRYPTION_JWT_PRIVATE_KEYSET_FILE=./hack/dev/encryption-keys/private_ec256.key
         SERVER_ENCRYPTION_JWT_PUBLIC_KEYSET_FILE=./hack/dev/encryption-keys/public_ec256.key
```
abelanger5, line 29: nit: we should remove these right?

Taskfile.yaml
```diff
@@ -30,13 +26,14 @@ tasks:
...
+        SERVER_TLS_STRATEGY=none
 
         SERVER_PORT=8080
         SERVER_URL=http://localhost:8080
 
         SERVER_AUTH_COOKIE_SECRETS="$(randstring 16) $(randstring 16)"
-        SERVER_AUTH_COOKIE_DOMAIN=app.dev.hatchet-tools.com
-        SERVER_AUTH_COOKIE_INSECURE=false
+        SERVER_AUTH_COOKIE_DOMAIN=localhost
```
abelanger5: I thought this might break instances? Does it not?

abelanger5: The more I think about it, the more in favor I am of removing Caddy and using the `./hatchet-dashboard` binary to serve the API and serving the engine directly over localhost. That'll also let us access it over localhost instead of the aliased hostname. What are your thoughts?

CONTRIBUTING.md
mrkaye97: I'd be in favor of removing caddy and using localhost!

CONTRIBUTING.md
gregfurman: Ahh meant to link it to `Pull Requests`. Thanks!

Taskfile.yaml
gregfurman: OK cool! Am very happy with that. Will look at that `hatchet-dashboard` implementation for this

Taskfile.yaml
gregfurman: AFAIK we need them to load the Encryption service

For master keyset:
https://github.com/hatchet-dev/hatchet/blob/4fd7b94751db4e34697fe4b6da3a756b936203e6/pkg/config/loader/loader.go#L746-L754

For JWT:
https://github.com/hatchet-dev/hatchet/blob/4fd7b94751db4e34697fe4b6da3a756b936203e6/pkg/config/loader/loader.go#L760-L765

Otherwise we end up with:

```
engine failure: could not load server config: could not load encryption service: encryption is required
```

or

```
error starting API: error loading server config: could not load encryption service: jwt encryption is required
```

We could instead generate these on demand instead of relying on static files? i.e

```env
SERVER_ENCRYPTION_MASTER_KEYSET="<base64-encoded-keyset>"
SERVER_ENCRYPTION_JWT_PUBLIC_KEYSET="<base64-encoded-jwt-public>"
SERVER_ENCRYPTION_JWT_PRIVATE_KEYSET="<base64-encoded-jwt-private>"
```

CONTRIBUTING.md
gregfurman: OK cool! So should remove the `caddy start` commands from the `dev` scripts then? Or do we want to rather have this done in a follow-up PR?

edit: nvm I see your latest comment on using the hatchet dashboard approach https://github.com/hatchet-dev/hatchet/pull/2972#discussion_r2791218764

Taskfile.yaml
abelanger5: whoops sorry, I misread these as certificates, not encryption keys -- you're right, we should keep these

Taskfile.yaml
gregfurman: > I thought this might break instances? Does it not?

@abelanger5 Seems to run fine for me with `localhost`. 

> The more I think about it, the more in favor I am of removing Caddy and using the ./hatchet-dashboard binary to serve the API and serving the engine directly over localhost. That'll also let us access it over localhost instead of the aliased hostname. What are your thoughts?

This sounds a lot like using `hatchet-lite` instead launching the engine + server in seperate processes. Any suggestions for starting points for the `hatchet-dashboard` approach? Else, should we perhaps do this in a follow-up PR?

abelanger5: Ah, for some reason I thought our `dashboard.Dockerfile` was using a single Go binary, but I see it's still using an NGINX image and a Go binary. A separate PR would make sense to me then.

gregfurman: Opened https://github.com/hatchet-dev/hatchet/issues/3005

CONTRIBUTING.md
gregfurman: > Personally, I think we should make task pre-commit-run as the command to depend on for linting and formatting.

@mnafees FYI running the `task fmt` will trigger the `pre-commit-run`. 

```yaml
  pre:
    aliases: [fmt]
    cmds:
      - task: fmt-go
      - task: fmt-app
      - task: fmt-docs
      - task: pre-commit-run
```

Should we update the `.pre-commit-config.yaml` to include the prettier hooks?

CONTRIBUTING.md
```diff
@@ -1,128 +1,120 @@
...
 ```
 
-### OpenTelemetry
-
-You can set the following to enable distributed tracing:
+**Note:** You might need to run this with `sudo` to install certificates.
 
+3. Start the Hatchet engine, API server, and frontend:
+```sh
+task start-dev # or task start-dev-tmux if you want to use tmux panes
```
abelanger5, line 89: After this runs, we should make it very clear to the user which URL they should go to in order to see the Hatchet UI. People always navigate to the Vite URL which is not the right one (`http://localhost:5173/`).

CONTRIBUTING.md
```diff
@@ -1,128 +1,120 @@
...
-SERVER_LOGGER_LEVEL=debug
-
-# json or console
-SERVER_LOGGER_FORMAT=json
 
-DATABASE_LOGGER_LEVEL=debug
-DATABASE_LOGGER_FORMAT=console
+2. Install Go & Node.js dependencies, run migrations, generate encryption keys, and seed the database:
+```sh
+task setup
```
abelanger5, line 84: The setup script nearly worked but not quite. I think it's missing `pnpm run docs:gen` in the `frontend/app` folder. Otherwise I hit the error:

```
10:51:50 AM [vite] Internal server error: Failed to resolve import "@/lib/generated/docs" from "src/pages/main/v1/workflow-runs-v1/components/runs-table.tsx". Does the file exist?
```

CONTRIBUTING.md
```diff
@@ -1,128 +1,120 @@
...
 ```
 
-### OpenTelemetry
-
-You can set the following to enable distributed tracing:
+**Note:** You might need to run this with `sudo` to install certificates.
```
abelanger5: Can probably remove this line?

Taskfile.yaml
```diff
@@ -46,26 +42,37 @@ tasks:
...
         EOF
+  fmt-go:
+    cmd: gofmt -s -w .
+  fmt-app:
+    dir: frontend/app
+    cmd: pnpm run prettier:fix && pnpm run prettier:check
+  fmt-docs:
+    dir: frontend/docs
+    cmd: pnpm run prettier:fix && pnpm run prettier:check
   pre:
```
abelanger5, line 64: This task created some diffs against the latest main - we should open that in a separate PR so contributors aren't adding random files to their PRs

abelanger5 (changes requested):

CONTRIBUTING.md
gregfurman: Weird. Which task was this failing at?

abelanger5: It didn't fail while running the task itself, this was thrown later when I accessed the frontend after setup.

CONTRIBUTING.md
gregfurman: Are you referring to `http://localhost:8081`? We're not launching the `hatchet-staticfileserver` so am unsure where the UI should be accessible from other than the Vite URL at `http://localhost:5173/`

abelanger5: It should be accessible via `https://app.dev.hatchet-tools.com`. But I often modify my Caddyfile to use localhost instead:

```
http://localhost:8888 {
	handle /api/* {
		reverse_proxy localhost:8080
	}

	handle /* {
		reverse_proxy localhost:5173
	}
}
```

Accessing the UI at `http://localhost:5173/` shouldn't work? Because it needs to be able to reach the API over the same domain, but all calls to `http://localhost:5173/api...` will fail.

CONTRIBUTING.md
gregfurman: That `src/pages/main/v1/workflow-runs-v1/components/runs-table.tsx` definitely exists. Which page in the UI are you navigating to? I'm not seeing any issues on my end.

abelanger5: Yep the `runs-table.tsx` file exists, the error is that `@/lib/generated/docs` doesn't exist until it's generated with `pnpm run docs:gen` in the `frontend/app` folder. If you run the contributing guide setup from a completely fresh codebase I think you'll see this error? Unless I missed a step while testing somewhere

Taskfile.yaml
gregfurman: Looks like some of those the generated go examples in `examples/go` are not properly formatted when generated. I've changed the `generate-docs` command to include a `fmt` step which should prevent this in future.

Here's the PR updating these: https://github.com/hatchet-dev/hatchet/pull/3009

CONTRIBUTING.md
gregfurman: Ahh I see the issue. I was running `generate.py` from my root dir, which meant those files were being generated in the wrong place.

Have added a `task generate-docs` which should do this correctly as a part of setup steps.

abelanger5 (approved):

promptless-for-oss:

New suggestion: **[Update contributing docs to align with CONTRIBUTING.md improvements](https://app.gopromptless.ai/suggestions/bf34a080-34d6-4e6a-8268-586eba258bbb)**

---
_Tip: Filter the [Dashboard](https://app.gopromptless.ai) by labels or assignees to focus on what matters to you 🔎_

PR #2975, opened 2026-02-09
Feat: durable slot suspension e2e eviction

grutt:
# Description

Adds support for client side eviction decisions and restore decisions on engine.

## Type of change


examples/python/simple/worker.py
```diff
@@ -1,24 +1,43 @@
 # > Simple
+from datetime import timedelta
```
grutt: sorry matt... yet another simple worker offense

internal/services/controllers/task/controller.go
```diff
@@ -437,6 +437,9 @@ func (tc *TasksControllerImpl) handleBufferedMsgs(tenantId uuid.UUID, msgId stri
 		return tc.handleProcessInternalEvents(context.Background(), tenantId, payloads)
 	case msgqueue.MsgIDTaskTrigger:
 		return tc.handleProcessTaskTrigger(context.Background(), tenantId, payloads)
+	case msgqueue.MsgIDDurableRestoreTask:
+		// TODO: should we be using context.Background() here?
```
grutt: general question for this file

api-contracts/dispatcher/dispatcher.proto
```diff
@@ -238,6 +243,10 @@ enum StepActionEventType {
     STEP_EVENT_TYPE_COMPLETED = 2;
     STEP_EVENT_TYPE_FAILED = 3;
     STEP_EVENT_TYPE_ACKNOWLEDGED = 4;
+    // Monitoring-only events (used for OLAP observability)
+    STEP_EVENT_TYPE_CANCELLING = 9;
+    STEP_EVENT_TYPE_CANCELLED_CONFIRMED = 5;
+    STEP_EVENT_TYPE_CANCELLATION_FAILED = 6;
```
mrkaye97, line 243: weird ordering here?

api-contracts/openapi/paths/v1/tasks/tasks.yaml
```diff
@@ -111,6 +111,60 @@ getTask:
     tags:
       - Task
 
+restoreTask:
+  post:
+    x-resources: ["tenant", "task"]
+    description: TEMP. Restores an evicted durable task by requeueing it at highest priority.
```
mrkaye97: are we intending to remove this?

cmd/hatchet-migrate/migrate/migrations/20260206120000_v1_0_76.sql
mrkaye97, line 1: we'll need to update these migration numbers

cmd/hatchet-migrate/migrate/migrations/20260206120000_v1_0_76.sql
```diff
@@ -0,0 +1,52 @@
...
+ALTER TABLE v1_task_events_olap_tmp
+    ALTER COLUMN event_type TYPE v1_event_type_olap_old
+    USING event_type::text::v1_event_type_olap_old;
+
+ALTER TABLE v1_task_events_olap
+    ALTER COLUMN event_type TYPE v1_event_type_olap_old
+    USING event_type::text::v1_event_type_olap_old;
+
+DROP TYPE v1_event_type_olap;
+ALTER TYPE v1_event_type_olap_old RENAME TO v1_event_type_olap;
```
mrkaye97, line 50: this is not a good idea 😅 

I think on the down migration, we should just not do anything and have the type exist and be unused, and the up migration should check `IF NOT EXISTS`. if we run this down migration as-is, it'll probably take down the db

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/step-run-detail.tsx
```diff
@@ -136,6 +149,61 @@ export const TaskRunDetail = ({
...
+      if (isTerminalState(taskRun?.status)) {
+        return 5000;
+      }
+      return 1000;
+    },
+  });
+
+  const isEvicted = (() => {
+    const rows = taskRunEventsQuery.data?.rows ?? [];
+    const sorted = [...rows].sort(
```
mrkaye97, line 168: using the `[...rows]` here seems silly - can't we do `rows.sort`?

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/step-run-detail.tsx
```diff
@@ -136,6 +149,61 @@ export const TaskRunDetail = ({
...
+    enabled: !!tenantId && !!taskRunId,
+    refetchInterval: () => {
+      if (isTerminalState(taskRun?.status)) {
+        return 5000;
+      }
+      return 1000;
+    },
+  });
+
+  const isEvicted = (() => {
```
mrkaye97, line 166: this should probably be wrapped in `useMemo` - also written a bit oddly (as a function we immediately call)

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/step-run-detail.tsx
```diff
@@ -136,6 +149,61 @@ export const TaskRunDetail = ({
...
+    );
+
+    let evicted = false;
+    for (const e of sorted) {
+      if (e.eventType === V1TaskEventType.DURABLE_EVICTED) {
+        evicted = true;
+      } else if (e.eventType === V1TaskEventType.DURABLE_RESTORING) {
+        evicted = false;
+      }
+    }
```
mrkaye97, line 180: I feel like maybe instead of this iterative thing, we can sort the list, and then filter for the first instance of `DURABLE_EVICTED` or `DURABLE_RESTORING`, and then check which it is. that'd be the more functional, JS-ish way of doing this IMO and would also be more performant

internal/services/controllers/olap/controller.go
```diff
@@ -708,7 +709,19 @@ func (tc *OLAPControllerImpl) handleCreateMonitoringEvent(ctx context.Context, t
 		case sqlcv1.V1EventTypeOlapFAILED:
 			readableStatuses = append(readableStatuses, sqlcv1.V1ReadableStatusOlapFAILED)
 		case sqlcv1.V1EventTypeOlapCANCELLED:
+			// Backwards compatibility (older clients).
```
mrkaye97, line 711: rm comment?

internal/services/controllers/olap/controller.go
```diff
@@ -708,7 +709,19 @@ func (tc *OLAPControllerImpl) handleCreateMonitoringEvent(ctx context.Context, t
...
 			readableStatuses = append(readableStatuses, sqlcv1.V1ReadableStatusOlapCANCELLED)
+		case sqlcv1.V1EventTypeOlapCANCELLING:
+			readableStatuses = append(readableStatuses, sqlcv1.V1ReadableStatusOlapCANCELLED)
+		case sqlcv1.V1EventTypeOlapCANCELLEDCONFIRMED:
+			readableStatuses = append(readableStatuses, sqlcv1.V1ReadableStatusOlapCANCELLED)
+		case sqlcv1.V1EventTypeOlapCANCELLATIONFAILED:
+			readableStatuses = append(readableStatuses, sqlcv1.V1ReadableStatusOlapCANCELLED)
+		case sqlcv1.V1EventTypeOlapDURABLEEVICTED:
+			// TODO-DURABLE: i'm not sure what we want to do here for status...
```
mrkaye97, line 720: we probably are going to need to add a new state, realistically, but it's such a pain that I think leaving this as `RUNNING` is reasonable

internal/services/controllers/task/controller.go
mrkaye97: I think so? unless you mean to add a `WithTimeout` or something

internal/services/controllers/task/controller.go
```diff
@@ -483,6 +486,11 @@ func (tc *TasksControllerImpl) handleTaskCompleted(ctx context.Context, tenantId
 
 	tc.notifyQueuesOnCompletion(ctx, tenantId, res.ReleasedTasks)
 
+	tc.l.Warn().Msgf("handleTaskCompleted: %d internal events from %d completed tasks", len(res.InternalEvents), len(msgs))
+	for i, ev := range res.InternalEvents {
+		tc.l.Warn().Msgf("  internal event[%d]: taskExternalId=%s type=%s key=%s", i, ev.TaskExternalID, ev.EventType, ev.EventKey)
+	}
```
mrkaye97: should we do this? it's going to produce a _**ton**_ of logs

internal/services/controllers/task/controller.go
```diff
@@ -1061,6 +1070,8 @@ func (tc *TasksControllerImpl) handleProcessInternalEvents(ctx context.Context,
 
 	msgs := msgqueue.JSONConvert[v1.InternalTaskEvent](payloads)
 
+	tc.l.Warn().Msgf("handleProcessInternalEvents: received %d internal events", len(msgs))
```
mrkaye97: same here - this will produce billions of logs a day across our instances

internal/services/controllers/task/controller.go
```diff
@@ -1127,6 +1138,9 @@ func (tc *TasksControllerImpl) processInternalEvents(ctx context.Context, tenant
 		return fmt.Errorf("could not process internal event matches: %w", err)
 	}
 
+	tc.l.Warn().Msgf("processInternalEvents: %d candidates -> %d created, %d replayed, %d satisfied durable entries",
+		len(candidateMatches), len(matchResult.CreatedTasks), len(matchResult.ReplayedTasks), len(matchResult.SatisfiedDurableEventLogEntries))
+
```
mrkaye97: and here

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -3,22 +3,27 @@ package task
...
 	v1 "github.com/hatchet-dev/hatchet/pkg/repository"
+	"github.com/hatchet-dev/hatchet/pkg/repository/sqlcv1"
 )
 
 func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context, tenantId uuid.UUID, callbacks []v1.SatisfiedEntry) error {
 	if len(callbacks) == 0 {
 		return nil
 	}
 
+	tc.l.Warn().Msgf("processSatisfiedEventLogEntry: processing %d satisfied callbacks", len(callbacks))
```
mrkaye97: this too - not sure this really adds much info unless I'm missing something (I'd really just remove all of these cursor log adds. In general, I feel like Cursor is way too willing to add useless logs everywhere, and these really don't help us that much afaict, and will just be super spammy)

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -3,22 +3,27 @@ package task
...
+
 	idInsertedAtTuples := make([]v1.IdInsertedAt, 0)
 
 	for _, cb := range callbacks {
+		tc.l.Warn().Msgf("  satisfied entry: durableTaskExternalId=%s durableTaskId=%d nodeId=%d", cb.DurableTaskExternalId, cb.DurableTaskId, cb.NodeId)
```
mrkaye97: here too

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -31,9 +36,31 @@ func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context
 		return fmt.Errorf("could not list dispatcher ids for tasks: %w", err)
 	}
 
+	tc.l.Warn().Msgf("  found %d dispatcher mappings for %d tasks", len(idInsertedAtToDispatcherId), len(idInsertedAtTuples))
```
mrkaye97: and here

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -31,9 +36,31 @@ func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context
...
+
+		// TODO-DURABLE: we probably shouldn't do this lookup here
+		task, err := tc.repov1.Tasks().GetTaskByExternalId(ctx, tenantId, cb.DurableTaskExternalId, true)
+		if err != nil {
+			tc.l.Error().Err(err).Msgf("failed to look up task %s for invocation count", cb.DurableTaskExternalId)
+			continue
+		}
+
+		// TODO-DURABLE: use the latest invocation count, this is not correct
+		taskInvocationCounts[cb.DurableTaskExternalId] = task.RetryCount + 1 + task.DurableInvocationCount
```
mrkaye97: 👍

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -70,3 +109,105 @@ func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context
 
 	return nil
 }
+
+func (tc *TasksControllerImpl) handleDurableRestoreTask(ctx context.Context, tenantId uuid.UUID, payloads [][]byte) error {
+	msgs := msgqueue.JSONConvert[tasktypes.DurableRestoreTaskPayload](payloads)
+
+	tc.l.Warn().Msgf("handleDurableRestoreTask: received %d restore messages", len(msgs))
```
mrkaye97: another not-so-helpful log imo - there are a bunch in this function

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -70,3 +109,105 @@ func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context
...
+
+	for _, msg := range msgs {
+		tc.l.Warn().Msgf("  restoring task externalId=%s invocationCount=%d reason=%s", msg.TaskExternalId, msg.InvocationCount, msg.Reason)
+
+		task, err := tc.repov1.Tasks().GetTaskByExternalId(ctx, tenantId, msg.TaskExternalId, true)
```
mrkaye97: let's add a todo here too, we shouldn't do this lookup in a loop

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -70,3 +109,105 @@ func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context
...
+			tc.l.Warn().Msgf("  rejecting stale restore for %s: msg invocation %d < current %d",
+				msg.TaskExternalId, msg.InvocationCount, currentInvocationCount)
+			continue
+		}
+
+		requeued, queue, err := tc.repov1.Tasks().RestoreEvictedTask(ctx, tenantId, v1.TaskIdInsertedAtRetryCount{
+			Id:         task.ID,
+			InsertedAt: task.InsertedAt,
+			RetryCount: task.RetryCount,
+		})
```
mrkaye97: we should probably batch this

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -70,3 +109,105 @@ func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context
...
+
+		if queue != "" {
+			tc.notifySchedulerQueue(ctx, tenantId, queue)
+		}
+	}
+
+	return nil
+}
+
+func (tc *TasksControllerImpl) notifySchedulerQueue(ctx context.Context, tenantId uuid.UUID, queue string) {
```
mrkaye97: is this relevant to this PR? maybe I'm missing something

internal/services/dispatcher/v1/server.go
```diff
@@ -606,9 +610,50 @@ func (d *DispatcherServiceImpl) handleEvictInvocation(
...
+			EventTimestamp: time.Now(),
+			EventType:      sqlcv1.V1EventTypeOlapDURABLEEVICTED,
+			EventPayload:   "",
+			EventMessage:   "Evicted via DurableTask stream",
+		},
+	)
+	if err != nil {
+		d.l.Warn().Err(err).Msg("failed to build DURABLE_EVICTED monitoring message")
+	} else if err := d.pubBuffer.Pub(ctx, msgqueue.OLAP_QUEUE, msg, false); err != nil {
+		d.l.Warn().Err(err).Msg("failed to publish DURABLE_EVICTED to OLAP")
```
mrkaye97: should we error here instead of returning an ack? I'm not too sure since this is olap-side

internal/services/dispatcher/v1/server.go
```diff
@@ -631,6 +677,7 @@ func (d *DispatcherServiceImpl) handleWorkerStatus(
 			TaskExternalId: taskExternalId,
 			NodeId:         cb.NodeId,
 		})
+		invocationCounts[fmt.Sprintf("%s/%d", taskExternalId, cb.NodeId)] = cb.InvocationCount
```
mrkaye97: I've been preferring to use a struct as the key for things like this instead of a formatted string like this - I think it's easier to reason about, and is much more type safe

internal/services/dispatcher/dispatcher.go
```diff
@@ -406,17 +406,30 @@ func (d *DispatcherImpl) handleDurableCallbackCompleted(ctx context.Context, tas
...
 
 		if err != nil {
-			d.l.Error().Err(err).Msgf("failed to deliver callback completion for task %s", payload.TaskExternalId)
+			d.l.Warn().Err(err).Msgf("failed to deliver callback completion for task %s, publishing restore", payload.TaskExternalId)
+
+			restoreMsg, msgErr := tasktypes.DurableRestoreTaskMessage(tenantId, payload.TaskExternalId, payload.InvocationCount, "callback delivery failed")
+			if msgErr != nil {
+				d.l.Error().Err(msgErr).Msgf("failed to create restore message for task %s", payload.TaskExternalId)
+				continue
+			}
```
mrkaye97: Unless I'm misunderstanding, I'm not sure this is correct - we could error here for any number of reasons, and I don't think we want to try to restore in all of those cases?

internal/services/dispatcher/server.go
```diff
@@ -634,6 +634,12 @@ func (d *DispatcherImpl) RefreshTimeout(ctx context.Context, request *contracts.
 	return d.refreshTimeoutV1(ctx, tenant, request)
 }
 
+func (d *DispatcherImpl) RestoreEvictedTask(ctx context.Context, request *contracts.RestoreEvictedTaskRequest) (*contracts.RestoreEvictedTaskResponse, error) {
+	tenant := ctx.Value("tenant").(*sqlcv1.Tenant)
+
+	return d.restoreEvictedTaskV1(ctx, tenant, request)
```
mrkaye97, line 608: nit: naming - we don't need the `V1` suffix here, seems like maybe this was added for some faux-backwards compat or something, but we should only have `restoreEvictedTask` to prevent confusion

internal/services/scheduler/v1/scheduler.go
```diff
@@ -688,15 +688,15 @@ func (s *Scheduler) notifyAfterConcurrency(ctx context.Context, tenantId uuid.UU
 
 	// handle cancellations
 	for _, cancelled := range res.Cancelled {
-		eventType := sqlcv1.V1EventTypeOlapCANCELLED
+		eventType := sqlcv1.V1EventTypeOlapCANCELLING
```
mrkaye97, line 654: I'm a little nervous this will have weird side effects with status updates but I'm not sure, just a thing to check

internal/services/shared/tasktypes/v1/olap.go
```diff
@@ -137,6 +160,32 @@ func MonitoringEventMessageFromActionEvent(tenantId uuid.UUID, taskId int64, ret
...
+
+	// TODO-DURABLE: This is a hack to get the durable eviction event type to work. We should remove this once we have a proper durable eviction event types.
+	// Python SDK uses CancellationReason.EVICTED.value ("evicted").
+	// Be defensive in case other SDKs/versions use slightly different strings.
+	switch reason {
+	case "evicted", "durable_evicted", "durable-evicted":
+		return true
+	default:
+		return false
+	}
```
mrkaye97: IMO, we should absolutely not do this 😅 - we should try to be more exact with the types we're checking against. i.e. all the SDKs should be sending the same value, and ideally that value would be defined in e.g. an openapi spec, a protobuf, etc. as an enum, so there's no need for string matching. To me, this is very brittle and will absolutely break in the future and be hard to debug

pkg/repository/sqlcv1/tasks-overwrite.go
```diff
@@ -729,11 +729,6 @@ WITH input AS (
     ORDER BY
         task_id, task_inserted_at, retry_count
     FOR UPDATE
-), deleted_slots AS (
-    DELETE FROM
-        v1_task_runtime_slot
-    WHERE
-        (task_id, task_inserted_at, retry_count) IN (SELECT task_id, task_inserted_at, retry_count FROM input)
```
mrkaye97, line 742: intentional to remove this?

sdks/python/examples/simple/worker.py
```diff
@@ -1,24 +1,43 @@
...
+from hatchet_sdk.runnables.eviction import EvictionPolicy
+from hatchet_sdk.worker.durable_eviction.manager import DurableEvictionConfig
 
 hatchet = Hatchet(debug=True)
 
 
 @hatchet.task()
-def simple(input: EmptyModel, ctx: Context) -> dict[str, str]:
+def simple5(input: EmptyModel, ctx: Context) -> dict[str, str]:
+    time.sleep(10)
```
mrkaye97: need to revert this as usual 😅

mrkaye97 (commented):
Left some comments on the internals - haven't looked through the SDK impl yet

PR #2976, opened 2026-02-09
Do not replay invalid tasks

mnafees:
# Description

We need to ignore invalid tasks before replaying them.

## Type of change


pkg/repository/task.go
```diff
@@ -272,6 +272,8 @@ type TaskRepository interface {
 
 	// run "details" getter, used for retrieving payloads and status of a run for external consumption without going through the REST API
 	GetWorkflowRunResultDetails(ctx context.Context, tenantId uuid.UUID, externalId uuid.UUID) (*WorkflowRunDetails, error)
+
+	FilterValidTasks(ctx context.Context, taskIds []int64, taskInsertedAts []pgtype.Timestamptz, taskRetryCounts []int32) (map[int64]struct{}, error)
```
abelanger5: for consistency, the second argument should be the `tenantId` (passed into the query for safety) and the third argument should be a struct type which contains `taskId, taskInsertedAt, and taskRetryCount`. otherwise there's a risk that the caller uses arrays of different lengths which would be very unsafe.

pkg/repository/sqlcv1/tasks.sql
```diff
@@ -1224,3 +1224,22 @@ FROM
     input
 RETURNING
     *;
+
+-- name: FilterValidTasks :many
```
abelanger5, line 1228: this query should use the `tenantId` (see comment below)

abelanger5 (changes requested):

pkg/repository/task.go
```diff
@@ -272,6 +278,8 @@ type TaskRepository interface {
 
 	// run "details" getter, used for retrieving payloads and status of a run for external consumption without going through the REST API
 	GetWorkflowRunResultDetails(ctx context.Context, tenantId uuid.UUID, externalId uuid.UUID) (*WorkflowRunDetails, error)
+
+	FilterValidTasks(ctx context.Context, tenantId uuid.UUID, opts *FilterValidTasksOpts) (map[int64]struct{}, error)
```
abelanger5: I think the last argument should be `[]*FilterValidTasksTuple`, or we actually might have a tuple called something like `[]*TaskIdInsertedAtRetryCount` which we can use.

abelanger5 (changes requested):

abelanger5 (approved):

PR #2983, opened 2026-02-10
[hotfix] Fix Docker frontend build issue

mnafees:
# Description

Fixes frontend Docker build from failing.

## Type of change


mrkaye97 (commented):
I think if we use `useCurrentTenantId()` instead, this can never be `undefined` or `null`?

mrkaye97 (approved):

PR #2999, opened 2026-02-11
fix: Don't skip errors with no message

### LINKED ISSUE #2997: [BUG] Python SDK: Exception with no or falsy first arg gets swallowed.
**Describe the issue**

If an exception is raised that has no first arg or that first arg is falsy, then the whole exception is swallowed and there's no error log at all.

I think raising the bare class is bad behavior, but it should still be possible to get the stacktrace and name of the error.

**Update**: I've located the issue to happen here: https://github.com/hatchet-dev/hatchet/blob/64a85498f5b42a0d61888865fa9b795faa930a98/sdks/python/hatchet_sdk/exceptions.py#L40-L43

And made a PR to fix it #2999 

**Environment**
- SDK: Python 1.23.1
- Engine: Selfhosted, latest

**Expected behavior**
The error should be clearly shown.

Run this and see that only the last error is shown.
```python
import pytest
from hatchet_sdk import Context, EmptyModel, Hatchet, TriggerWorkflowOptions

hatchet = Hatchet()

test_workflow = hatchet.workflow(name="test_workflow_pytest")


@test_workflow.task()
async def test_class(input: EmptyModel, ctx: Context):
    raise ValueError


@test_workflow.task()
async def test_instance_no_input(input: EmptyModel, ctx: Context):
    raise ValueError()


@test_workflow.task()
async def test_instance_empty_input(input: EmptyModel, ctx: Context):
    raise ValueError("")


@test_workflow.task()
async def test_instance_non_empty_input(input: EmptyModel, ctx: Context):
    raise ValueError("Truthy")


@pytest.mark.asyncio
async def test_worker():
    await test_workflow.aio_run(EmptyModel(), TriggerWorkflowOptions())
```

**Additional context**
It specifically happened to me because `cryptography` can raise `cryptography.fernet.InvalidToken` as just the class so it's not just some theoretical gotcha

StefanBRas:
# Description

Fixes #2997 

I deleted both checks, because honestly I can't see how you could reach this point without actually having the trace or anything.
If you somehow managed to create the TaskRunError with no exc, exc_type OR trace then maybe it's okay with just returning and empty string with the metadata? I have no hard opinions about how this is handled, as long as normal errors aren't hidden

## Type of change

## What's Changed

- [X] Correctly handles python task errors that have no message


mrkaye97:
hey @StefanBRas - thanks for the improvement! I'm a little nervous about the empty string case, mostly for formatting. I'm not sure how that'd actually happen but there must be a reason I had it that way initially 😅 Maybe we should update the serialization logic to check if those strings are empty so the message doesn't look awkward if one / both are?

StefanBRas:
> hey @StefanBRas - thanks for the improvement! I'm a little nervous about the empty string case, mostly for formatting. I'm not sure how that'd actually happen but there must be a reason I had it that way initially 😅 Maybe we should update the serialization logic to check if those strings are empty so the message doesn't look awkward if one / both are?

Sure, makes sense. Will do!

StefanBRas:
@mrkaye97 I have done the changes now! Some things to note:

- We now only create the metadata string if we need it
- We still return an empty string if we get an empty input EVEN if include_metadata is set. That mimics behavior from before
- I changed the error message for when a workflow is missing tasks to also include the workflow name. Hope that's okay.
- I couldn't find an obvious way to test that the output is as expected, without doing some kind of snapshot testing. If you want to add that, I suggest https://pypi.org/project/inline-snapshot/


<details>

<summary>As proof of here's some screenshots:</summary>

image

image


</details>

mrkaye97 (approved):
looks great, nice work! we'll need to bump the patch version + write a changelog entry for this one too, then I'll ship it

StefanBRas:
Done!

PR #3023, opened 2026-02-14
fix(ci/generate): Run format-friendly generate task

gregfurman:
# Description

The newly introduced Taskfile changes automatically format generated files. This PR ensures that the (currently broken)  `gen-examples.yml::generate` CI job uses this task approach, similarly to `test.yml::generate` i.e

https://github.com/hatchet-dev/hatchet/blob/06e0da50fa81b48004135d908a5e1d0b14eb3c73/.github/workflows/test.yml#L25-L39

Once the examples are regenerated, this should fixe the broken `generate` job: https://github.com/hatchet-dev/hatchet/actions/runs/21994591182/job/63550922252

Depends on revert PR https://github.com/hatchet-dev/hatchet/pull/3024

## Type of change

## What's Changed

- Aligns the `gen-examples.yml::generate` job to use the same taskfile approach as `test.yml:::generate`

### DIFF
--- a/.github/workflows/gen-examples.yml
+++ b/.github/workflows/gen-examples.yml
@@ -11,12 +11,27 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v6
 
+    - name: Install Task
+      uses: arduino/setup-task@v2
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Setup Go
+      uses: actions/setup-go@v6
+      with:
+        go-version: "1.25"
+
+    - name: Setup pnpm
+      uses: pnpm/action-setup@v4
+      with:
+        version: 10.16.1
+        run_install: false
+
     - uses: actions/setup-python@v6
       with:
         python-version: '3.14'
     - name: Generate snippets
-      working-directory: frontend/snippets
-      run: python3 generate.py
+      run: task install-dependencies pre-commit-install generate-docs -v
 
     - name: Check for changes in examples directory
       id: verify-changed-files


gregfurman:
I ran some examples against my personal fork to illustrate the CI working:

- See original run picking up the incorrectly formatted Go files: https://github.com/gregfurman/hatchet/actions/runs/22018400928/job/63623532681#step:9:28 (this fails due to non-existent labels when trying to open PR)
- See how committing the formatted changes allows the CI to pass without issue: https://github.com/gregfurman/hatchet/actions/runs/22018524619/job/63623837145

mnafees (approved):

.github/workflows/gen-examples.yml
```diff
@@ -11,12 +11,27 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v6
 
+    - name: Install Task
+      uses: arduino/setup-task@v2
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
```
mnafees, line 17: Do we need this? Just a bit nervous giving authZ to a third party Action like this.

mnafees (changes requested):

.github/workflows/gen-examples.yml
gregfurman: We need the action since we are now running `task` commands instead of directly running `python3 generate.py`. Unsure about this repo token https://github.com/arduino/setup-task?tab=readme-ov-file#repo-token

mnafees (approved):

PR #3028, opened 2026-02-15
chore:  suppress some expected errors

grutt:
# Description

Some errors are expected behavior and should be suppresed

## Type of change

### DIFF
--- a/pkg/repository/ticker.go
+++ b/pkg/repository/ticker.go
@@ -2,9 +2,12 @@ package repository
 
 import (
 	"context"
+	"errors"
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/jackc/pgx/v5"
+
 	"github.com/hatchet-dev/hatchet/pkg/repository/sqlchelpers"
 	"github.com/hatchet-dev/hatchet/pkg/repository/sqlcv1"
 )
@@ -65,6 +68,10 @@ func (t *tickerRepository) IsTenantAlertActive(ctx context.Context, tenantId uui
 	res, err := t.queries.IsTenantAlertActive(ctx, t.pool, tenantId)
 
 	if err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			return false, time.Time{}, nil
+		}
+
 		return false, time.Now(), err
 	}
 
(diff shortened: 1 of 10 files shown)


pkg/repository/trigger.go
```diff
@@ -201,6 +202,8 @@ func (r *sharedRepository) makeTriggerDecisions(ctx context.Context, filters []*
 					Source:       sqlcv1.V1CelEvaluationFailureSourceFILTER,
 					ErrorMessage: err.Error(),
 				})
+
+				continue
```
grutt, line 206: afict we were inserting decision twice here for the same filterId

pkg/repository/trigger.go
```diff
@@ -201,6 +202,8 @@ func (r *sharedRepository) makeTriggerDecisions(ctx context.Context, filters []*
...
 			}
```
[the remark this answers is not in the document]
grutt: i dont think its a functional bug... i think we probably dedupe this somewhere higher, but i did not trace this

pkg/repository/ticker.go
```diff
@@ -65,6 +68,10 @@ func (t *tickerRepository) IsTenantAlertActive(ctx context.Context, tenantId uui
 	res, err := t.queries.IsTenantAlertActive(ctx, t.pool, tenantId)
 
 	if err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			return false, time.Time{}, nil
+		}
+
 		return false, time.Now(), err
 	}
```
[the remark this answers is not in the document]
grutt: yep

pkg/repository/trigger.go
```diff
@@ -184,7 +184,8 @@ func (r *sharedRepository) makeTriggerDecisions(ctx context.Context, filters []*
 			shouldTrigger, err := r.processWorkflowExpression(ctx, filter.Expression, opt, filter.Payload)
 
 			if err != nil {
-				r.l.Error().
+				// This is notified to the user via an olap event.
```
mrkaye97, line 187: it is? I'm pretty sure we write this to the db but don't show it on the frontend or anything, at least right now

pkg/repository/trigger.go
mrkaye97: good catch, this looks like a bug to me 👍 thanks

PR #3040, opened 2026-02-16
Refactor: Consolidate Dispatchers

mrkaye97:
# Description

Refactor to consolidate the two dispatchers (with a thin wrapper for backwards compat) to simplify some of the durable execution implementation and clean up a bit

## Type of change


mrkaye97:
gonna do this later, turns out it's a big surface area b/c of the SDKs

PR #3041, opened 2026-02-16
Feat: Non-determinism errors

mrkaye97:
# Description

First pass at a non-determinism check in the durable execution path. The basic idea here is when we get a new event log entry and we found an existing one in the db, we hash some of its data (the payload, the kind, the conditions if they're present), and we check the current hash against the existing one. If they don't agree, we send a `NonDeterminismError` back to the worker, and it raises.

A couple notes here:
1. I think this decision about what's included in the hash made sense (didn't think it'd make sense to include other options like metadata, priority, etc.) but could be convinced otherwise
2. I removed the `hash_alg` and `data_hash` columns in favor of a single `idempotency_key` because I figured YAGNI - didn't seem to add much value to have separate columns or support multiple hashing strategies (could be convinced otherwise here too)
3. I think, somewhat unfortunately, it makes the most sense to send the error to the worker and have it raise instead of marking the task failed on the engine, just so we duplicate less state. So that's what I'm doing here

## Type of change


api-contracts/v1/dispatcher.proto
```diff
@@ -84,11 +84,19 @@ message DurableTaskRequest {
     }
 }
 
+message DurableTaskErrorResponse {
+    string durable_task_external_id = 1;
+    int64 invocation_count = 2;
+    int64 node_id = 3;
+    string error_message = 4;
```
grutt: thinking ahead here do we maybe want an error_type enum, and error_details string?

mrkaye97: yeah, good call

pkg/repository/durable_events.go
```diff
@@ -211,6 +266,71 @@ func getDurableTaskSignalKey(taskExternalId uuid.UUID, nodeId int64) string {
 	return fmt.Sprintf("durable:%s:%d", taskExternalId.String(), nodeId)
 }
 
+func (r *durableEventsRepository) createIdempotencyKey(ctx context.Context, opts IngestDurableTaskEventOpts) ([]byte, error) {
```
grutt: perhaps we should cover this with unit tests?

sdks/python/hatchet_sdk/runnables/workflow.py
```diff
@@ -686,11 +685,6 @@ def run_no_wait(
         """
         cancellation_token = self._resolve_check_cancellation_token()
 
-        logger.debug(
```
grutt, line 689: you hate my log lines 😢

sdks/python/hatchet_sdk/utils/cancellation.py
```diff
@@ -120,30 +118,21 @@ async def long_running_task():
...
-            logger.debug("await_with_cancellation: invoking cancel callback")
             await _invoke_cancel_callback(cancel_callback)
+
         raise asyncio.CancelledError("Operation cancelled by cancellation token")
 
     main_task = asyncio.ensure_future(coro)
 
     try:
-        result = await race_against_token(main_task, token)
-        logger.debug("await_with_cancellation: completed successfully")
```
grutt, line 140: i really think we should keep at least the cancellation logs for a while since this is critical for us to be certain we're freeing resources

sdks/python/hatchet_sdk/exceptions.py
```diff
@@ -4,6 +4,20 @@
...
+    def __init__(
+        self, task_external_id: str, invocation_count: int, message: str, node_id: int
+    ) -> None:
+        self.task_external_id = task_external_id
+        self.invocation_count = invocation_count
+        self.message = message
+        self.node_id = node_id
+
+        super().__init__(
+            f"Non-determinism detected in task {task_external_id} on invocation {invocation_count} at node {node_id}.\nCheck out our documentation for more details on expectations of durable tasks: https://docs.hatchet.run/home/durable-best-practices"
```
grutt, line 17: we're 100% going to forget these links exist and we're going to move paths around... maybe we should flatten these into a docs dictionary or something

pkg/repository/durable_events.go
mrkaye97: yeah, seems reasonable. I'm hesitant to do it until we implement the branching logic b/c I suspect this might change

sdks/python/hatchet_sdk/exceptions.py
mrkaye97: that's a good idea - we have a lot of these, unfortunately... we probably can make them some enum or something

sdks/python/hatchet_sdk/runnables/workflow.py
mrkaye97: lol, I don't think most of these are helpful at all 😂 cursor loves spamming logs, I do not

sdks/python/hatchet_sdk/utils/cancellation.py
mrkaye97: yeah fair enough

pkg/repository/durable_events.go
grutt: yeah but at least if we cover and they break we will remember to fix them instead of forgetting to add them. i think this is the only critical change on this

grutt (changes requested):
only need tests on idempotency key

grutt (approved):

sql/schema/v1-core.sql
```diff
@@ -2318,14 +2318,9 @@ CREATE TABLE v1_durable_event_log_entry (
...
-    -- data JSONB,
-    -- The hash of the data stored in the v1_payload table to check non-determinism violations.
-    -- This can be null for event types that don't have associated data.
-    -- TODO: we can add CHECK CONSTRAINT for event types that require data_hash to be non-null.
-    data_hash BYTEA,
-    -- Can discuss: adds some flexibility for future hash algorithms
-    data_hash_alg TEXT,
+    -- An idempotency key generated from the incoming data (using the type of event + wait for conditions or the trigger event payload + options)
+    -- to determine whether or not there's been a non-determinism error
+    idempotency_key BYTEA,
```
abelanger5: can or should this ever be null?

pkg/repository/durable_events.go
```diff
@@ -211,6 +266,71 @@ func getDurableTaskSignalKey(taskExternalId uuid.UUID, nodeId int64) string {
...
+
+			dataToHash = append(dataToHash, []byte(toHash)...)
+		}
+	}
+
+	h := sha256.New()
+	h.Write(dataToHash)
+	hashBytes := h.Sum(nil)
+	idempotencyKey := make([]byte, hex.EncodedLen(len(hashBytes)))
+	hex.Encode(idempotencyKey, hashBytes)
```
abelanger5, line 329: This is a fair amount of data to be storing per entry -- just wondering if we've thought about this.

abelanger5 (approved):
nice!

sql/schema/v1-core.sql
mrkaye97: good call 👍

pkg/repository/durable_events.go
mrkaye97: you mean just in terms of the number of bytes we're storing in the table right? that's definitely fair 🤔 we could try and either compress or use some more compact hash, I guess the tradeoff is more collision risk or something. WDYT?

mrkaye97: we could also put this in the payloads table and have it offload

mrkaye97: (gonna merge and circle back)

abelanger5: I'm not up to date on hash functions - I could see the payloads table making sense if we're doing a read anyway I guess?

promptless-for-oss:

Updated existing suggestion: **[Document NonDeterminismError exception and reset_durable_task methods](https://app.gopromptless.ai/suggestions/48aec471-52b2-4ea5-b411-5e402a482483)**

---

promptless-for-oss:

Updated existing suggestion: **[Document durable execution with NonDeterminismError and best practices](https://app.gopromptless.ai/suggestions/e446edc3-9854-41c5-a334-c7f1a42c0708)**

---

promptless-for-oss:

New suggestion: **[Document NonDeterminismError and determinism requirements for durable tasks](https://app.gopromptless.ai/suggestions/d5ec2d05-a40d-40d1-8535-82d5461df10e)**

---
_Tip: Connect Jira, Linear, Confluence, and more in [Integrations](https://app.gopromptless.ai/integrations) to enrich suggestion quality 🔗_

PR #3045, opened 2026-02-17
Refactor: Remove separate callback table

mrkaye97:
# Description

Removes the separate callbacks table in favor of just storing the `is_satisfied` directly on the event log entry

## Type of change


cmd/hatchet-migrate/migrate/migrations/20260216174818_v1_0_80.sql
```diff
@@ -1,5 +1,43 @@
 -- +goose Up
 -- +goose StatementBegin
+DROP FUNCTION IF EXISTS create_v1_range_partition(text, date);
+CREATE OR REPLACE FUNCTION create_v1_range_partition(
+    targetTableName text,
+    targetDate date,
+    fillfactor integer DEFAULT 100
```
mrkaye97, line 7: this is the fillfactor change - easier to see in the schema file

cmd/hatchet-migrate/migrate/migrations/20260216174818_v1_0_80.sql
```diff
@@ -68,74 +106,76 @@ CREATE TABLE v1_durable_event_log_entry (
...
-    -- Definite: we'll query directly for the key when a worker is checking if a callback is satisfied
-    -- Definite: we'll query directly for the key when a v1_match has been satisfied and we need to mark the callback as satisfied
-    CONSTRAINT v1_durable_event_log_callback_pkey PRIMARY KEY (durable_task_id, durable_task_inserted_at, node_id)
+    CONSTRAINT v1_durable_event_log_entry_pkey PRIMARY KEY (durable_task_id, durable_task_inserted_at, node_id)
 ) PARTITION BY RANGE(durable_task_inserted_at);
 
-SELECT create_v1_range_partition('v1_durable_event_log_callback', NOW()::DATE);
-SELECT create_v1_range_partition('v1_durable_event_log_callback', (NOW() + INTERVAL '1 day')::DATE);
+SELECT create_v1_range_partition('v1_durable_event_log_entry', NOW()::DATE, 80);
+SELECT create_v1_range_partition('v1_durable_event_log_entry', (NOW() + INTERVAL '1 day')::DATE, 80);
```
mrkaye97, line 117: I just picked `80` as the fill factor for these, but we can make it whatever. my understanding of this is basically that 80 means that "for every 4 tuples that are inserted into a page, we leave space for one tuple empty". So I think the idea would be that if we e.g. set it to 50, we'd expect great performance if every row were updated once. I suspect every row in this table will be updated less than once (some events won't end up being satisfied), so I set it to a higher number 🤷

grutt (approved):

PR #3047, opened 2026-02-17
feat: automatic doc generator for TS SDK

jishnundth:
# Description

1. Automatic TS SDK mdx generator for Nextra framework with typedoc. 

## Type of change


sdks/typescript/docs/generate.ts
mrkaye97, line 1: I sort of wonder if we could reuse the python doc generation script for this (and the doc_types, e.g.) - do you have a sense for how much of this is shared? if it's pretty different from the python one, I'd say we're good to leave this, but thought I'd ask!

mrkaye97:
looks like we need to fix some conflicts here, but in general this looks great to me

sdks/typescript/docs/generate.ts
jishnundth: It’ll be very hard to do that because they’re incompatible at the tool level. Apart from the meta.js generation, everything is different between the Python and TypeScript generators, including the directory structure.

mrkaye97: Alright, sounds good to me!

jishnundth:
> looks like we need to fix some conflicts here, but in general this looks great to me

Done!

mrkaye97:
This looks really good! A couple small bits of feedback (not sure how much we can realistically do about these, so feel free to say it's not easy and move on 😅)

 
Some of the type definitions look a little odd (the inline type, the list, there might be more):

Screenshot 2026-03-02 at 3 32 05 PM

It's a little hard for me to differentiate between different things that should be headers here, I think (api, crons, ...)

Screenshot 2026-03-02 at 3 32 36 PM

Similar thing here

Screenshot 2026-03-02 at 3 33 10 PM

jishnundth:
> This looks really good! A couple small bits of feedback (not sure how much we can realistically do about these, so feel free to say it's not easy and move on 😅)
> 
> Some of the type definitions look a little odd (the inline type, the list, there might be more):
> 
> Screenshot 2026-03-02 at 3 32 05 PM
> It's a little hard for me to differentiate between different things that should be headers here, I think (api, crons, ...)
> 
> Screenshot 2026-03-02 at 3 32 36 PM
> Similar thing here
> 
> Screenshot 2026-03-02 at 3 33 10 PM

The header size and style can be controlled. But, controlling the type expansion of variables requires manual intervention. The only field I can control as of now is, https://typedoc.org/documents/Options.Input.html#maxtypeconversiondepth

What you see in the screenshot you've shared is maxtypeconversiondepth of 3. If I make it 2 it'll look like this

Also in the below screenshot you can see that the parameters header is not bold anymore. So, this'll help the user to clearly distinguish the method headers. 

image

jishnundth:
I've pushed the changes, please check now. Only method headers are highlighted now. But, controlling type expansion is difficult.

mrkaye97 (approved):
let's ship it 🚀

PR #3048, opened 2026-02-18
Fix--durable-slot-acquisition

grutt:
# Description

Fixes double slot acquistion and python registration

## Type of change

### DIFF
--- a/cmd/hatchet-migrate/migrate/migrations/20260216000001_v1_0_79_a_multi_slot_schema_and_triggers.sql
+++ b/cmd/hatchet-migrate/migrate/migrations/20260216000001_v1_0_79_a_multi_slot_schema_and_triggers.sql
@@ -113,8 +113,14 @@ BEGIN
         worker_id,
         'default'::text,
         1
-    FROM new_rows
-    WHERE worker_id IS NOT NULL
+    FROM new_rows nr
+    WHERE nr.worker_id IS NOT NULL
+    AND NOT EXISTS (
+        SELECT 1 FROM v1_task_runtime_slot s
+        WHERE s.task_id = nr.task_id
+          AND s.task_inserted_at = nr.task_inserted_at
+          AND s.retry_count = nr.retry_count
+    )
     ON CONFLICT (task_id, task_inserted_at, retry_count, slot_type) DO NOTHING;
 
     RETURN NULL;
--- a/cmd/hatchet-migrate/migrate/migrations/20260216000004_v1_0_80_fix_slot_trigger.sql
+++ b/cmd/hatchet-migrate/migrate/migrations/20260216000004_v1_0_80_fix_slot_trigger.sql
@@ -1,0 +1,106 @@
+-- +goose Up
+-- +goose StatementBegin
+
+-- The original trigger always inserted a 'default' slot into v1_task_runtime_slot
+-- when a row was inserted into v1_task_runtime. This caused durable tasks to also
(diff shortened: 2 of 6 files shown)


abelanger5 (approved):

cmd/hatchet-migrate/migrate/migrations/20260216000001_v1_0_79_a_multi_slot_schema_and_triggers.sql
mrkaye97, line 1: do we want to modify an existing migration?

cmd/hatchet-migrate/migrate/migrations/20260216000004_v1_0_80_fix_slot_trigger.sql
```diff
@@ -0,0 +1,79 @@
+-- +goose Up
+-- +goose StatementBegin
+
+-- The original trigger always inserted a 'default' slot into v1_task_runtime_slot
+-- when a row was inserted into v1_task_runtime. This caused durable tasks to also
+-- acquire a default slot because the new code path (assigned_slots CTE) already
+-- inserts the correct slot type, but the trigger would fire afterwards and add
+-- an extra 'default' row. The NOT EXISTS guard skips the trigger insert when the
+-- new code path has already written slots for the task.
+CREATE OR REPLACE FUNCTION v1_task_runtime_slot_insert_function()
```
mrkaye97, line 10: do we need an update to the schema for these changes?

mrkaye97 (changes requested):
going to block this for those two comments just to be safe since alexander approved already and I wasn't sure

cmd/hatchet-migrate/migrate/migrations/20260216000004_v1_0_80_fix_slot_trigger.sql
grutt: no, these are only for the migration and will be dropped in #3020

cmd/hatchet-migrate/migrate/migrations/20260216000001_v1_0_79_a_multi_slot_schema_and_triggers.sql
grutt: we've done this before for incorrect migrations to avoid confusion, i dont have a strong opinion here

cmd/hatchet-migrate/migrate/migrations/20260216000004_v1_0_80_fix_slot_trigger.sql
mrkaye97: ah okay nice!

cmd/hatchet-migrate/migrate/migrations/20260216000001_v1_0_79_a_multi_slot_schema_and_triggers.sql
mrkaye97: gotcha, my concern was if this has already been applied some places, but if that's not an issue then fine by me

mrkaye97 (approved):

cmd/hatchet-migrate/migrate/migrations/20260216000001_v1_0_79_a_multi_slot_schema_and_triggers.sql
grutt: yeah good call out, goose doesnt hash migrations like some migration managers so as long as the timestamp is the same it will not apply twice.

PR #3055, opened 2026-02-18
Feat: Replay as new (or from a node)

mrkaye97:
# Description

Initial implementation of replay from a specific node (including node 1, which would constitute "as new"). The basic idea here is:

1. Added a route on the admin service which basically calls `handleReset` from `durable_events.go`
2. Proxies over the REST API to that endpoint so that the frontend / SDKs can run this
3. `HandleReset` basically just sets the `branch_first_parent_node_id` and increments the latest node and latest branch so that we know when to start following a new path down the tree. Note that I'm not 100% sure this is the right approach and I haven't figured out yet how this generalizes to branching off branches, but it does let us do the branching without needing to reconstruct the tree on every event, which seemed like a win
4. Once we increment, we pub a replay message to replay the task, and then it ends up going down the new branch.

For instance, if we run this task:

```python
@hatchet.durable_task(name="DurableSleepTask")
async def durable_sleep_task(input: EmptyModel, ctx: DurableContext) -> dict:
    await ctx.aio_sleep_for(timedelta(seconds=3))
    await ctx.aio_sleep_for(timedelta(seconds=3))
    await ctx.aio_sleep_for(timedelta(seconds=3))
```

and then replay it from `node_id=2`, we'll get this in the db in the durable event log:

```
hatchet=# select id, kind, parent_branch_id, branch_id, parent_node_id, node_id from v1_durable_event_log_entry where durable_task_id = 1956 order by 4, 

 id |   kind   | parent_branch_id | branch_id | parent_node_id | node_id 
----+----------+------------------+-----------+----------------+---------
 41 | WAIT_FOR |                1 |         1 |                |       1
 42 | WAIT_FOR |                1 |         1 |              1 |       2
 43 | WAIT_FOR |                1 |         1 |              2 |       3
 44 | WAIT_FOR |                1 |         1 |                |       1
 45 | WAIT_FOR |                1 |         2 |              1 |       2
 46 | WAIT_FOR |                2 |         2 |              2 |       3
(6 rows)
```

(on the second run, the parent branch id for the second node is (branch = 1, node = 1), and the third node has parent branch = 2, parent node = 2

## Type of change


pkg/repository/durable_events.go
```diff
@@ -587,3 +652,47 @@ func (r *durableEventsRepository) handleTriggerRuns(ctx context.Context, tx *Opt
...
+		NodeId:                  sqlchelpers.ToBigInt(&zero),
+		InvocationCount:         sqlchelpers.ToBigInt(&newInvocationCount),
+		BranchFirstParentNodeId: sqlchelpers.ToBigInt(&lastFastForwardedNode),
+		Durabletaskid:           task.ID,
+		Durabletaskinsertedat:   task.InsertedAt,
+	})
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to update log file for reset: %w", err)
+	}
```
mrkaye97, line 725: this is the meat of it - basically, when we call reset, we update the `BranchFirstParentNodeId` and the latest branch and node so that when we reach this point in the event log next time, we know to create a new branch

api-contracts/openapi/components/schemas/v1/workflow_run.yaml
```diff
@@ -144,3 +144,40 @@ V1TriggerWorkflowRunRequest:
   required:
     - workflowName
     - input
+
+V1ResetDurableTaskRequest:
```
grutt: nit: is reset there right verb here? maybe fork?

mrkaye97: yeah I dunno, I think reset is what temporal uses 🤷

pkg/repository/durable_events.go
```diff
@@ -587,3 +652,49 @@ func (r *durableEventsRepository) handleTriggerRuns(ctx context.Context, tx *Opt
...
+
+	logFile, err := r.getAndLockLogFile(ctx, tx, task.ID, task.InsertedAt)
+
+	if err != nil || logFile == nil {
+		return nil, fmt.Errorf("failed to lock log file: %w", err)
+	}
+
+	newBranchId := logFile.LatestBranchID + 1
+
+	// TODO-DURABLE: we probably don't want to do this here - probably should only be incremented on the dispatcher
```
grutt: agreed.

one question though... what happens if a run is reset again before it is assigned/run?

do we need to guard here for this?

mrkaye97: advisory lock should handle this

grutt: similar what happens if we reset a running run

mrkaye97: re: running - that's a good question, I guess we'll increment the invocation count and the old one will be evicted, at least that's what probably should happen

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
```diff
@@ -294,7 +307,11 @@ async def _handle_response(self, response: DurableTaskResponse) -> None:
                 if not error_pending_ack_future.done():
                     error_pending_ack_future.set_exception(exc)
 
-            callback_key = (error.durable_task_external_id, error.node_id)
+            callback_key = (
+                error.durable_task_external_id,
+                error.branch_id,
```
grutt, line 312: i have a similar q on my branch but should this just be durable task + invocation id tuple for simplicity

grutt: if not... do we even need the invocation id?

mrkaye97: this one definitely needs to be (external_id, branch_id, node_id), since that's what uniquely identifies a node in the tree. the invocation count shouldn't be included, since it doesn't matter here (and will cause issues with replays if it is included)

grutt (commented):
looking good! a few clarifying questions

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
mrkaye97: also one thing is having `node_id` is sort of confusing, since it doesn't uniquely identify an actual "node", which is really a `(branch_id, node_id)` pair 🤦 naming is hard

pkg/repository/durable_events.go
```diff
@@ -331,6 +352,29 @@ func (r *durableEventsRepository) createIdempotencyKey(opts IngestDurableTaskEve
 	return idempotencyKey, nil
 }
 
+func (r *durableEventsRepository) getAndLockLogFile(ctx context.Context, tx sqlcv1.DBTX, durableTaskId int64, durableTaskInsertedAt pgtype.Timestamptz) (*sqlcv1.V1DurableEventLogFile, error) {
+	lockKey := fmt.Sprintf("durable_log_file:%d:%s", durableTaskId, durableTaskInsertedAt.Time.String())
+
+	lockAcquired, err := r.queries.TryAdvisoryLock(ctx, tx, hash(lockKey))
```
abelanger5: I wonder if we should think about namespacing locks via `int32, int32` instead. We're increasing the likelihood of collisions pretty dramatically (we haven't used advisory locks on a path with this number of writes). Easy change in the future though, just figured I'd flag that it's an option.

pkg/repository/durable_events.go
```diff
@@ -331,6 +352,29 @@ func (r *durableEventsRepository) createIdempotencyKey(opts IngestDurableTaskEve
...
+
+	if err != nil || !lockAcquired {
+		return nil, fmt.Errorf("failed to acquire advisory lock: %w", err)
+	}
+
+	logFile, err := r.queries.GetLogFile(ctx, tx, sqlcv1.GetLogFileParams{
+		Durabletaskid:         durableTaskId,
+		Durabletaskinsertedat: durableTaskInsertedAt,
+	})
```
abelanger5: I'm curious about the case with two transactions, where tx 1 commits its changes in between tx 2 starting its tx and acquiring the advisory lock. we'll successfully acquire the lock but we won't read new data on the log file. 

There's probably not a way to solve this without `SERIALIZABLE`, which we definitely don't want. 

The other thing that would solve this is if we can guarantee a unique constraint violation after commit, since these are implicitly serialized via internal pg locks as well.

abelanger5: storing a revision history on the log file, for example, with a uniqueness constraint.

mrkaye97: let me think about this - agreed it's a good callout though

sdks/python/hatchet_sdk/clients/rest/models/bulk_create_event_response.py
mrkaye97, line 1: removed all these, bunch of dead code

cmd/hatchet-migrate/migrate/migrations/20260216174818_v1_0_80.sql
```diff
@@ -121,6 +124,9 @@ ALTER TABLE v1_match
     ADD COLUMN durable_event_log_entry_branch_id BIGINT
 ;
 
+-- needs to be nullable so we don't have to backfill
```
grutt, line 127: todo?

internal/services/dispatcher/v1/server.go
```diff
@@ -587,12 +587,13 @@ func (d *DispatcherServiceImpl) handleDurableTaskEvent(
 		return status.Errorf(codes.Internal, "failed to send trigger ack: %v", err)
 	}
 
-	if ingestionResult.EventLogEntry.Entry.IsSatisfied {
+	if ingestionResult.IsSatisfied {
 		err := d.DeliverDurableEventLogEntryCompletion(
 			taskExternalId,
+			ingestionResult.InvocationCount,
```
grutt, line 593: doesnt have to block this merge but id be in favor of renaming this to DurableInvocationCount since we've decided to scope it to the feature

internal/services/dispatcher/dispatcher_v1.go
```diff
@@ -119,12 +119,41 @@ func (d *DispatcherImpl) HandleLocalAssignments(ctx context.Context, tenantId, w
...
+				InsertedAt: assigned.Task.InsertedAt,
+			})
+		}
+	}
+
+	if len(getDurableInvocationCountOpts) > 0 {
+		invocationCounts, err := d.repov1.DurableEvents().GetDurableTaskInvocationCounts(ctx, tenantId, getDurableInvocationCountOpts)
+
+		if err != nil {
+			d.l.Error().Err(err).Msgf("could not get durable task invocation counts for %d tasks", len(getDurableInvocationCountOpts))
```
grutt, line 145: is this case safe to log and continue?

grutt (approved):

promptless-for-oss:

Updated existing suggestion: **[Document durable execution with NonDeterminismError and best practices](https://app.gopromptless.ai/suggestions/e446edc3-9854-41c5-a334-c7f1a42c0708)**

---
_Tip: Adjust how proactive or focused Promptless is in [Agent Settings](https://app.gopromptless.ai/configure/settings) ⚙️_

promptless-for-oss:

New suggestion: **[Document reset_durable_task methods for durable task replay](https://app.gopromptless.ai/suggestions/339b8ef2-7423-4fae-aa2a-10e1a7fc7b26)**

---

PR #3067, opened 2026-02-20
Fix: Move event log to a tab on the task run detail

mrkaye97:
# Description

Moved the event log to its own tab instead of having this weird scrolling below every other component

Screenshot 2026-02-20 at 1 43 29 PM

Also fixed a bug where the output payload was displaying the input payload initially while the task was queued / running, and fixed a couple more UUID type bugs

## Type of change


pkg/repository/sqlcv1/dags.sql
```diff
@@ -45,7 +45,7 @@ SELECT
     i.display_name,
     i.workflow_id,
     i.workflow_version_id,
-    i.parent_task_external_id
+    NULLIF(i.parent_task_external_id, '00000000-0000-0000-0000-000000000000'::uuid)
```
mrkaye97, line 48: hack to not need an overwrite for this

mnafees (approved):

PR #3068, opened 2026-02-20
New "create organization" and "create tenant" interfaces

TehShrike:
# Description

This adds two new interfaces: one that creates a new organization and a tenant in that organization, and an interface that creates a new tenant.

The former is only relevant in the cloud environment.  It shows up in an onboarding screen in the cloud environment if you belong to no organizations.  It also shows up in its own screen if you click "create organization".

The latter shows up in an onboarding screen if you belong to no tenants.  It also shows up in a modal that can be summoned from anywhere in the app.  In the cloud environment, there is a dropdown to select the associated organization.

Both onboarding screens are very minimal and don't have all the words that were on the old onboarding create-tenant UI – let me know if you want anything in particular added back in.  Gabriel made it sound like the "create organization" onboarding screen was probably going to be expanded at some point in the future.

These changes depend on an as-of-yet-unopened PR against the cloud codebase to prevent a first organization (but not tenant) from being implicitly created when a new user is created.

## Type of change

## Other changes

- added an editorconfig so I could touch both the frontend and the go files
- added the ability to write unit tests for frontend code
- moved a pnpm config out of the npmrc files so that npm wouldn't complain at me when I reflexively use `npm run`


frontend/app/src/components/forms/new-organization-input-form.tsx
```diff
@@ -0,0 +1,69 @@
+import { Button } from '@/components/v1/ui/button';
+import { Input } from '@/components/v1/ui/input';
+import { Label } from '@/components/v1/ui/label';
+import { useState } from 'react';
+
+type NewOrganizationInputFormProps = {
+  defaultOrganizationName?: string;
+  defaultTenantName?: string;
+  isSaving?: boolean;
```
mrkaye97: any particular reason `isSaving` is optional?

frontend/app/src/components/forms/new-organization-input-form.tsx
```diff
@@ -0,0 +1,69 @@
...
+}: NewOrganizationInputFormProps) {
+  const [organizationName, setOrganizationName] = useState(
+    defaultOrganizationName,
+  );
+  const [tenantName, setTenantName] = useState(defaultTenantName);
+
+  const handleSubmit = (e: React.FormEvent) => {
+    e.preventDefault();
+    onSubmit({ organizationName, tenantName });
+  };
```
mrkaye97: not sure if the linter will yell about this, but if it does, you might need to wrap this in `useCallback`

frontend/app/src/components/forms/new-organization-input-form.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  return (
+    <form onSubmit={handleSubmit} className="grid gap-4 max-w-lg w-full">
```
mrkaye97, line 33: I'm not 100% sure, but we might want to use a shadcn form here (via React Hook Form or TanStack Form, which I'd somewhat prefer) - I haven't really used them before and don't know what prior art we have in the codebase, but I know they exist at least

frontend/app/src/components/forms/new-organization-saver-form.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  const { data: tenant } = await cloudApi.organizationCreateTenant(
+    organization.metadata.id,
+    {
+      name: tenantName,
+      slug: generateTenantSlug(tenantName),
+    },
+  );
+
+  await Promise.all([refetchTenantMemberships(), refetchOrganizations()]);
```
mrkaye97: We probably want to use react query for this (`useMutation`) instead of calling these directly and needing to worry about errors, resolving promises, etc.

frontend/app/src/components/forms/new-organization-saver-form.tsx
```diff
@@ -0,0 +1,98 @@
...
+  afterSave,
+}: NewOrganizationSaverFormProps) {
+  const [isSaving, setIsSaving] = useState(false);
+  const { isCloudEnabled, refetchTenantMemberships, refetchOrganizations } =
+    useAppContext();
+  const { capture } = useAnalytics();
+
+  if (!isCloudEnabled) {
+    // I feel like there should be an assert here instead
+    return <div>Cloud is not enabled</div>;
```
mrkaye97: Yeah... is it even possible that this renders if cloud is not enabled? Or would something farther upstream cause it to be hidden? I'm assuming something upstream would cause it to be hidden (unsure), in which case I think an assertion makes sense.

One thing we have some places is this `invariant` from `tiny-invariant`, which just throws if something isn't truthy (I think). Might be worth checking that out

frontend/app/src/components/forms/new-organization-saver-form.tsx
```diff
@@ -0,0 +1,98 @@
...
+    })
+      .then((result) => {
+        capture('onboarding_tenant_created', {
+          tenant_type: 'cloud',
+          is_cloud: true,
+        });
+        afterSave(result);
+      })
+      .catch(handleApiError)
+      .finally(() => setIsSaving(false));
```
mrkaye97: re: above comment - `useMutation` will help here!

frontend/app/src/components/forms/new-tenant-input-form.tsx
```diff
@@ -0,0 +1,109 @@
...
+  defaultTenantName = '',
+  defaultOrganizationId = '',
+  isSaving = false,
+  isCloudEnabled,
+  organizations = [],
+  onSubmit,
+}: NewTenantInputFormProps) {
+  const [tenantName, setTenantName] = useState(defaultTenantName);
+  const [organizationId, setOrganizationId] = useState(
+    defaultOrganizationId || undefined,
```
mrkaye97, line 86: extreme nit: if this is going to set `""` to `undefined`, should we also make the default `undefined`?

frontend/app/src/components/forms/new-tenant-saver-form.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+    return { type: 'cloud', tenant, organizationId };
+  } else {
+    const { data: tenant } = await api.tenantCreate({
+      name: tenantName,
+      slug,
+    });
+
+    await refetchTenantMemberships();
+
```
mrkaye97: same here with `useMutation` - should make this a lot easier. We have a lot of examples of that elsewhere! I'd recommend making some hook like `useNewTenant` (could maybe even put a lot of things into a big `useTenant`, I don't feel too strongly), and then define a mutation with these api calls that has callbacks like `onSuccess`, etc., is disabled is cloud isn't enabled, etc. and wrap it in a callback for better parameter handling

frontend/app/src/lib/global-emitter.ts
```diff
@@ -0,0 +1,5 @@
+import makeEmitter from 'better-emitter';
+
+export const globalEmitter = makeEmitter<{
+  'new-tenant': { defaultOrganizationId?: string };
+}>();
```
mrkaye97, line 5: neat! I haven't seen this library before - are the keys type safe?

frontend/app/src/lib/assert.ts
```diff
@@ -0,0 +1,7 @@
+function assert(condition: any, message?: string): asserts condition {
+  if (!condition) {
+    throw new Error(message ?? 'Assertion failed');
+  }
+}
+
+export default assert;
```
mrkaye97: I think we can replace this with `tiny-invariant`, most likely

frontend/app/src/pages/onboarding/create-tenant/index.tsx
```diff
@@ -1,330 +1,38 @@
...
-        referralSource: value.referralSource,
-      },
-      name: value.name,
-      environment: value.environment as TenantEnvironment,
-      referralSource: value.referralSource,
-    });
-  };
+  const defaultOrganizationId =
+    organizations && organizations.length > 0
+      ? organizations[0].metadata.id
```
mrkaye97, line 13: possible bug 😅 - is `organizations` sorted here? I think it might be sorted in the hook where it comes from, but I don't remember. if it's not, it probably should be for this to work how we want

frontend/app/src/pages/onboarding/create-tenant/index.tsx
```diff
@@ -1,330 +1,38 @@
...
+        <h1 className="text-2xl font-bold text-center">Create a new tenant</h1>
+
+        <div className="flex justify-center">
+          <NewTenantSaverForm
+            defaultOrganizationId={defaultOrganizationId}
+            afterSave={(result) => {
+              const tenantId =
+                result.type === 'cloud'
+                  ? result.tenant.id
+                  : result.tenant.metadata.id;
```
mrkaye97, line 28: this is a bit confusing to me, and makes me nervous. we have this convention of using a `metadata` object on everything on the API which includes its id and some timestamps. I don't necessarily love this convention, but I think it's going to be hard to follow what the difference between `tenant.id` and `tenant.metadata.id` is. at the very least, we should probably have a comment here to disambiguate, but honestly I'm not sure even reading this that I understand why these are different 😅

frontend/app/src/pages/authenticated.tsx
```diff
@@ -220,6 +263,13 @@ function AuthenticatedInner() {
     }
   }, [isAuthPage, navigate, userError]);
 
+  useEffect(() =>
+    globalEmitter.on('new-tenant', ({ defaultOrganizationId }) => {
+      setDefaultOrganizationId(defaultOrganizationId);
+      setNewTenantModalOpen(true);
+    }),
+  );
```
mrkaye97, line 264: this probably needs a dependency array

frontend/app/src/providers/app-context.tsx
```diff
@@ -133,11 +176,10 @@ export function AppContextProvider({ children }: AppContextProviderProps) {
       tenantId,
       isTenantLoading: membershipsQuery.isLoading,
       membership: membership?.role,
+      refetchTenantMemberships: membershipsQuery.refetch,
```
mrkaye97: One note here (been trying to leave this comment on other FE PRs too and make incremental progress, to mixed degrees of success): I think using `queryClient.invalidateQueries()` is preferred to `refetch`ing directly, at least according to my somewhat limited understanding of react query. So I'd prefer to start making more use of `invalidateQueries`, or be proven wrong there

mrkaye97 (commented):
Left some small comments, but largely this looks great! One request I've been making on frontend changes is to post some screenshots - that'd be great here too

promptless-for-oss:

New suggestion: **[Update cloud quickstart to reflect organization + tenant onboarding flow](https://app.gopromptless.ai/suggestions/e9ddadd4-138a-477b-b8c4-e743b14f9382)**

---

frontend/app/src/components/forms/new-organization-input-form.tsx
TehShrike: nope, don't think so, I'll change that

frontend/app/src/components/forms/new-organization-input-form.tsx
TehShrike: ah yup I think you're right

frontend/app/src/components/forms/new-organization-input-form.tsx
TehShrike: So I added a `form` for
1. letting people hit enter in an input and have the form automatically submit
2. leaning on native browser form validation to make sure that users had filled the values out before they could submit

It looks like the tanstack form is mostly about fancier types of validation that we don't need here?  I think it would just result in more code getting executed and more DOM elements in this form without any change in behavior, unless I'm missing something.

frontend/app/src/components/forms/new-organization-saver-form.tsx
TehShrike: The idea is that something upstream should always prevent this from showing up.  This form is/should only be displayed in the new organization onboarding screen, and the dedicated new organization screen, and we shouldn't be sending the user to those routes if they're on the cloud interface.

I had originally started building out a unified form to be used in both contexts, but the saving flow and state started feeling divergent enough that I thought that having an org-creation-specific UI would be a lot simpler overall.

I'll add an assertion here.

frontend/app/src/components/forms/new-tenant-input-form.tsx
TehShrike: You're right – I changed the state to default to undefined once I figured out it was necessary for to trigger `required` on the `Select`, and now that prop default doesn't make sense, I'll drop it

frontend/app/src/lib/global-emitter.ts
TehShrike: Yeah, keys (and arguments) are all type-checked.  Also I find the "listener methods return unsubscribe methods" pattern so much more reasonable in component lifecycles where you use them inside of something like `useEffect`.

frontend/app/src/lib/assert.ts
TehShrike: ah yeah, I'd looked a bit to see if there was an assertion function in the codebase already but had missed tiny-invariant.

Though I suppose I wouldn't have thought to make their main design decision (besides allowing the message to be a function): https://github.com/alexreardon/tiny-invariant/blob/master/src/tiny-invariant.ts

If, God forbid, an assertion does get thrown in production, I'd rather see the message show up in Sentry than not.  But I suppose hopefully we have functioning sourcemaps in production and the stack trace should be about as good in theory.

But that's not a serious concern, I can switch the new code over to tiny-invariant

frontend/app/src/pages/onboarding/create-tenant/index.tsx
TehShrike: I looked around to try to find any "identify the appropriate default organization" code and wasn't able to find anything, so this is very much me throwing up my hands and randomly grabbing one from the pile.

Is there any knowable logic for picking a default organization from the list?  I'd much rather stick a `getDefaultOrganization(organizations)` function somewhere reusable if so.

frontend/app/src/pages/onboarding/create-tenant/index.tsx
TehShrike: I agree.  I saw that the pattern was being used [in the old create-tenant component](https://github.com/hatchet-dev/hatchet/pull/3068/changes#diff-475619b18d5efcdcfd7c3dd78d03b2a28be427fdcccb92e017a42825ab5390f2L148-L156), so I mostly just covered my eyes and duplicated it.

The bones of the issue as far as I've looked into it is that the OSS API endpoint that creates a tenant returns this shape: https://github.com/hatchet-dev/hatchet/blob/44069cef89bc155483347cddf7b2985e511d16f5/frontend/app/src/lib/api/generated/data-contracts.ts#L757-L771

and the Cloud API returns this shape: https://github.com/hatchet-dev/hatchet/blob/44069cef89bc155483347cddf7b2985e511d16f5/frontend/app/src/lib/api/generated/cloud/data-contracts.ts#L686-L699

which triggered my "there are fundamental differences between the APIs that I probably shouldn't think about this week" senses and here we are.  What's the story between the two APIs?

It would be great if both APIs returned about the same shape, then we could e.g. pass in a different "create tenant" function into a component depending on the environment, and the component's saving flow could be simpler.

frontend/app/src/pages/authenticated.tsx
TehShrike: oh snap, yeah it does, I'd forgotten that no argument !== empty array

frontend/app/src/providers/app-context.tsx
TehShrike: I do like that idea.

In testing, I ran into some runtime errors after creating an organization+tenant because there was no guarantee that the organizations would finish fetching before the tenants did, and there was code that assumed if a tenant referenced an organization id, that the organization must exist in the organizations list.  So I used the refetch function and had the save function wait on everything being refreshed before redirecting the user.

One solution would be to put both organizations and tenants behind the same query that could be invalidated/fetched together.

Probably worth calling out, I get nervous every time I look at useOrganizations, and even the small amount of global cached state in appContext was feeling pretty rickety to me.  And I felt like I was adding a ton of "make sure the orgs/tenants are done fetching before we continue on" boilerplate everywhere that I'd rather not have to mess with.

---

I'm starting to think that tanstack queries may not be the appropriate interface here, at least for regular pages.  Dependency data like organizations/tenants should generally be fetched in the route's loading function, which means that we want the underlying promise, not `isSuccess` + `data`.

An interface like this would probably be preferable:
```
type QueryThingy<T> = ({
	data: T
	dataIsLoaded: true
} | {
	data: null
	dataIsLoaded: false
}) & {
	getLatestResultAndStartNewRequestIfNecessary: () => Promise<T>
	invalidate: () => void
}
```

---

Anyways, in the short term fetching organizations and tenants in the same query in the app context is probably the best way to move in the direction you were thinking of, let me know if you have any other thoughts.

frontend/app/src/components/forms/new-organization-saver-form.tsx
TehShrike: I looked at existing uses of useMutation and read the docs, and couldn't figure out what it was actually adding – it gives you properties named `onSuccess` and `onError` that look like functional replacements for the promise lifecycle methods, and you get an `isPending` property.  Is there something else?  It looked like it usually took up a few more lines of code and I couldn't figure out what the advantage would be for this code, since it's all promises either way..  Am I missing something?

frontend/app/src/components/forms/new-tenant-saver-form.tsx
TehShrike: huh, I didn't see any way that useMutation would improve this flow, but maybe that's unfamiliarity/a failure of imagination.  I can give it a go.

TehShrike: Were you thinking something like ba5d5f7 or something else?

The refetch calls will go away once I move both tenant+organization queries in the app context into the a singular query that can be invalidated in onSuccess/when the save promise resolves.

frontend/app/src/components/forms/new-organization-input-form.tsx
mrkaye97: Sounds good to me!

frontend/app/src/components/forms/new-organization-saver-form.tsx
mrkaye97: Yeah I think there are a bunch of things - I do think the `onSuccess`, `onMutate`, `onError`, etc. callbacks are quite useful but agree that promise methods are helpful in some of those cases (maybe not on `onMutate`). There's also built in retry logic which I appreciate but of course you could also just build that with plain fetch / promises.

For me, the biggest thing (true of `useQuery` too) is the handling of different states becomes trivial (fetching, is fetched, is pending, etc.) so you can use the return values from the `useMutation` hook instead of needing to wrangle state separately (presumably in the promise lifecycle methods)

The last thing I think is just consistency with other places in the codebase (mutations) and also how this bit of react-query code will play with e.g. uses of `useQuery` in other places (potentially things updating the query cache, etc.)

frontend/app/src/lib/assert.ts
mrkaye97: Great, thanks! Yeah this is more of an appeal for consistency than anything else. I don't have strong feelings about the tooling here (and agree on the point on Sentry), so as long as we're consistent, anything would work for me

frontend/app/src/pages/onboarding/create-tenant/index.tsx
mrkaye97: nope 😂 I think some other places we sort alphabetically and pick the first, but I do think the logic at least should be stable

frontend/app/src/pages/onboarding/create-tenant/index.tsx
mrkaye97: ugh, yeah, you're totally right - going to cc @mnafees here because I'm not too sure, but I agree. it'd be great if they returned something in the same shape

frontend/app/src/providers/app-context.tsx
mrkaye97: Yeah, combining the fetches sounds good to me! And yeah would be curious about the app context thing 👍

frontend/app/src/providers/user-universe.tsx
```diff
@@ -0,0 +1,214 @@
...
+        };
+  },
+  enabled: isCloudLoaded,
+});
+
+export const invalidateUserUniverse = () => {
+  queryClient.invalidateQueries({
+    queryKey: ['user-universe'],
+  });
+};
```
mrkaye97: I _think_ it's generally recommended to use `useQueryClient` instead of importing it directly, but I might be misremembering that - easy enough to put this inside the provider and return it as a callback though

frontend/app/src/providers/user-universe.tsx
```diff
@@ -0,0 +1,214 @@
...
+  children,
+}: {
+  children: React.ReactNode;
+}) {
+  const { isCloudEnabled, isCloudLoaded } = useCloud();
+  const tenantMembershipAndOrganizationsQuery = useQuery(
+    useMemo(
+      () => userUniverseQuery({ isCloudEnabled, isCloudLoaded }),
+      [isCloudEnabled, isCloudLoaded],
+    ),
```
mrkaye97: I think we probably don't need to memoize here? might be my own misunderstanding too, but I don't think I've seen a memo inside of a `useQuery` like this before, although that's not to say it's necessarily wrong or anything

frontend/app/src/providers/user-universe.tsx
```diff
@@ -0,0 +1,214 @@
...
+  );
+
+  const invalidate = useCallback(() => {
+    queryClient.resetQueries({
+      queryKey: ['user-universe'],
+    });
+  }, []);
```
mrkaye97: ah nice, seems like this is duped from the invalidate function that's exported above? I'd think we'd keep this one and remove that one, assuming I understand right

frontend/app/src/providers/user-universe.tsx
```diff
@@ -0,0 +1,214 @@
...
+  const value = useMemo<UserUniverse>(() => {
+    const isLoaded = tenantMembershipAndOrganizationsQuery.isSuccess;
+    if (isCloudEnabled) {
+      const getWithOrganizations = get as () => Promise<{
+        organizations: OrganizationForUserList['rows'];
+        tenantMemberships: TenantMember[];
+      }>;
+
+      if (isLoaded) {
+        invariant(tenantMembershipAndOrganizationsQuery.data.organizations);
```
mrkaye97, line 142: I'd leave a comment here explaining how we know that this is truthy

frontend/app/src/providers/user-universe.tsx
```diff
@@ -0,0 +1,214 @@
...
+      }
+
+      return isLoaded
+        ? {
+            isCloudEnabled,
+            isLoaded,
+            organizations: tenantMembershipAndOrganizationsQuery.data
+              .organizations as OrganizationForUserList['rows'],
```
mrkaye97: is it possible to not have the type assertion here?

frontend/app/src/components/forms/new-organization-saver-form.tsx
```diff
@@ -0,0 +1,105 @@
...
+    isCloudEnabled,
+  } = useUserUniverse();
+
+  const saveOrganizationMutation = useSaveOrganization({ afterSave });
+
+  if (!isUserUniverseLoaded) {
+    return <></>;
+  }
+
+  invariant(isCloudEnabled, 'Cloud must be enabled');
```
mrkaye97: comment here too probably - I imagine this is safe because we'd never load this component if orgs aren't enabled, which implies we're on cloud?

frontend/app/src/providers/user-universe.tsx
TehShrike: I think the consensus is that you're supposed to be right?  I couldn't find the docs saying explicitly, but based on [`useBaseQuery.ts`](https://github.com/TanStack/query/blob/main/packages/react-query/src/useBaseQuery.ts) and [`queryObserver.ts`](https://github.com/TanStack/query/blob/main/packages/query-core/src/queryObserver.ts) it looks like they assume that queries are unique based on the key.

So I think each call will produce a new observer object, but should leave the previous query object, though in some cases a call can overwrite the options on the previous query.

In any case, I'm guessing the idiomatic way here is to just let all that code run on every render, but move `isCloudEnabled` into the key, so I'll do that.

frontend/app/src/providers/user-universe.tsx
TehShrike: turns out yes, though I have to drop my beautiful ternary 😭

frontend/app/src/providers/user-universe.tsx
TehShrike: Theoretically it's explained by the line right above it, except that variable name is too generic to be obvious – I'll try changing it

frontend/app/src/components/forms/new-organization-saver-form.tsx
TehShrike: Good call, I improved the message

mrkaye97 (approved):
🚀

promptless-for-oss:

New suggestion: **[Update documentation for new organization and tenant creation flow](https://app.gopromptless.ai/suggestions/fbcd69a6-801f-436d-8d26-248fc655dadc)**

---

PR #3075, opened 2026-02-22
Feat:  durable eviction take 2

grutt:
# Description

Simplified implementation of durable eviction

## Type of change

## To Dos

- [x] Additional tests
- [ ] Early exit on new callback
- [x] Investigate OLAP events


internal/services/dispatcher/v1/server.go
```diff
@@ -610,9 +612,50 @@ func (d *DispatcherServiceImpl) handleEvictInvocation(
...
 	req *contracts.DurableTaskEvictInvocationRequest,
 ) error {
-	// TODO-DURABLE: implement eviction here
+	taskExternalId, err := uuid.Parse(req.DurableTaskExternalId)
+	if err != nil {
+		return status.Errorf(codes.InvalidArgument, "invalid durable_task_external_id: %v", err)
+	}
 
-	return nil
+	// TODO-DURABLE: should we be doing a lookup here?
```
grutt: should we be confirming if this is the latest here?

promptless-for-oss:

Updated existing suggestion: **[Document durable task eviction for Go and Python SDKs](https://app.gopromptless.ai/suggestions/b0f0cbdd-a099-4c1d-b03f-e32cb8f89e60)**

---

pkg/repository/sqlcv1/tasks.sql
```diff
@@ -395,6 +395,8 @@ WITH tasks_on_inactive_workers AS (
     WHERE
         w."tenantId" = @tenantId::uuid
         AND w."lastHeartbeatAt" < NOW() - INTERVAL '30 seconds'
+        -- evicted tasks are not eligible for re-assignment
```
grutt, line 398: we should never be in this case since eviction sets worker id to null

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -379,11 +419,13 @@ def cleanup_run_id(self, key: ActionKey) -> None:
             del self.threads[key]
 
         if key in self.contexts:
-            if self.contexts[key].exit_flag:
+            if self.contexts[key].exit_flag:  # noqa: SIM102
```
grutt: i'm not sure why this is now required

pkg/repository/sqlcv1/tasks.sql
```diff
@@ -954,6 +956,113 @@ WHERE
...
+    RETURNING 1
+)
+SELECT
+    COALESCE((SELECT 1 FROM updated_runtime LIMIT 1), 0)::int AS "evicted";
+
+-- name: RestoreEvictedTask :one
+-- Restores an evicted task by inserting it directly into the assignment queue.
+-- The evicted runtime row stays (evicted_at set); when the queue item is assigned,
+-- the ON CONFLICT in UpdateTasksToAssigned clears evicted_at and re-creates slots.
+-- TODO-DURABLE: check if invocation has increased and if yes, do nothing
```
grutt: i dont actually think this is required

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -42,7 +44,17 @@ func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context
 		dispatcherId, ok := idInsertedAtToDispatcherId[key]
 
 		if !ok {
-			tc.l.Warn().Msgf("could not find dispatcher id for task %d inserted at %s, skipping callback", cb.DurableTaskId, cb.DurableTaskInsertedAt.Time)
+			tc.l.Debug().Msgf("no dispatcher for task %d (evicted or worker gone), publishing restore message", cb.DurableTaskId)
```
mrkaye97: I'm not 100% sure this is the right thing to do here, need to think about it a bit. Is there a risk that there's some other way that no dispatcher is found, but we _don't_ need to restore? e.g. if there's a reassign or something, and we're racing the reassign and the restore? Not sure if that's possible, just wondering

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -72,3 +84,84 @@ func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context
 
 	return nil
 }
+
+func (tc *TasksControllerImpl) handleDurableRestoreTask(ctx context.Context, tenantId uuid.UUID, payloads [][]byte) error {
+	msgs := msgqueue.JSONConvert[tasktypes.DurableRestoreTaskPayload](payloads)
+
+	for _, msg := range msgs {
+		task, err := tc.repov1.Tasks().GetTaskByExternalId(ctx, tenantId, msg.TaskExternalId, true)
```
mrkaye97: N+1 query here - let's look these up in a batch?

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -72,3 +84,84 @@ func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context
...
+		if err != nil {
+			tc.l.Error().Err(err).Msgf("failed to look up task %s for restore", msg.TaskExternalId)
+			continue
+		}
+
+		requeued, queue, err := tc.repov1.Tasks().RestoreEvictedTask(ctx, tenantId, v1.TaskIdInsertedAtRetryCount{
+			Id:         task.ID,
+			InsertedAt: task.InsertedAt,
+			RetryCount: task.RetryCount,
+		})
```
mrkaye97: Can we batch this too?

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -72,3 +84,84 @@ func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context
...
+		if err != nil {
+			tc.l.Error().Err(err).Msgf("failed to restore task %s", msg.TaskExternalId)
+			continue
+		}
```
mrkaye97: if we get an error, should we return an error here or continue?

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -72,3 +84,84 @@ func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context
...
+				EventMessage:   fmt.Sprintf("Restoring evicted task: %s", msg.Reason),
+			},
+		)
+		if err == nil {
+			if pubErr := tc.pubBuffer.Pub(ctx, msgqueue.OLAP_QUEUE, olapMsg, false); pubErr != nil {
+				tc.l.Warn().Err(pubErr).Msg("failed to publish DURABLE_RESTORING to OLAP")
+			}
+		}
+
+		tc.notifySchedulerQueue(ctx, tenantId, queue)
```
mrkaye97: this also results in an N+1 query afaict? looking up the tenant

internal/services/dispatcher/v1/server.go
mrkaye97: Confirming the task being evicted is the latest invocation?

pkg/repository/sqlcv1/tasks.sql
mrkaye97: Not too sure I'm following here, but for my own understanding mostly - if the invocation count has increased, we definitely do need to do _something_ to eagarly evict any other invocations of that task that are running

pkg/repository/task.go
```diff
@@ -1568,6 +1572,60 @@ func (r *TaskRepositoryImpl) ReleaseSlot(ctx context.Context, tenantId, external
 	return resp, nil
 }
 
+func (r *TaskRepositoryImpl) EvictTask(ctx context.Context, tenantId uuid.UUID, task TaskIdInsertedAtRetryCount) (bool, error) {
```
mrkaye97: something I've been liking that I feel we should do more, is making a type for this bool like

```go
type WasEvicted bool
```

and then having `EvictTask` return `(WasEvicted, error)`

IMO, it's a lot easier to read the code that way, and harder to mix up what `true` / `false` mean

pkg/repository/task.go
```diff
@@ -1568,6 +1572,60 @@ func (r *TaskRepositoryImpl) ReleaseSlot(ctx context.Context, tenantId, external
...
+	})
+	if err != nil {
+		return false, "", err
+	}
+
+	if err := commit(ctx); err != nil {
+		return false, "", err
+	}
+
+	queueName, _ := row.Queue.(string)
```
mrkaye97: this seems odd - what type is this if not a string?

sdks/python/examples/durable_eviction/test_durable_eviction.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+async def _poll_for_event(
+    hatchet: Hatchet,
+    workflow_run_id: str,
+    event_type: V1TaskEventType,
+) -> set[V1TaskEventType]:
+    """Poll run details until *event_type* appears (or timeout)."""
+    for _ in range(MAX_POLLS):
+        details = await hatchet.runs.aio_get(workflow_run_id)
```
mrkaye97: it's probably better to use `aio_get_details` or whatever I called that since it goes over grpc - will be less racy that way

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
```diff
@@ -63,6 +64,7 @@ def from_proto(cls, proto: DurableTaskEventLogEntryCompletedResponse) -> Self:
 
 PendingCallback = tuple[TaskExternalId, BranchId, NodeId]
```
mrkaye97: just fyi, we might (I'm not 100% sure, honestly) still need to make this tuple include the invocation count, but it might be fine to leave it as-is. I'll think about it

sdks/python/hatchet_sdk/context/context.py
```diff
@@ -535,12 +542,11 @@ async def aio_sleep_for(self, duration: Duration) -> dict[str, Any]:
             SleepCondition(duration=duration),
         )
 
-    ## todo: instrumentor for this
```
mrkaye97, line 538: need to leave this comment so I don't forget 😅

mrkaye97 (commented):
looks great, left a couple questions / comments on the engine. will go through the SDK more carefully tuesday morning most likely

pkg/repository/task.go
grutt: miscast in old query, good catch

pkg/repository/task.go
grutt: i like this

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
grutt: so i was going to re-raise this but i think we need this. currently i think there's a potential race if a callback completes before we cleanup but after we ack the eviction

grutt: afict adding invocation count will eliminate this because we wont "reuse" this callback

internal/services/controllers/task/durable_callbacks.go
grutt: i think it was okay, but this could be more clear

internal/services/dispatcher/v1/server.go
grutt: yeah basically if we get an old eviction request but there's already a new invocation we should probably exit, right?

api/v1/server/handlers/v1/workflow-runs/list.go
```diff
@@ -15,6 +15,30 @@ import (
 	transformers "github.com/hatchet-dev/hatchet/api/v1/server/oas/transformers/v1"
 )
 
+// TODO-DURABLE: Hack until eviction is modeled in v1_readable_status_olap.
+func normalizeWorkflowRunStatuses(statuses []gen.V1TaskStatus) []sqlcv1.V1ReadableStatusOlap {
+	normalized := make([]sqlcv1.V1ReadableStatusOlap, 0, len(statuses))
+	seen := make(map[sqlcv1.V1ReadableStatusOlap]struct{}, len(statuses))
+
+	for _, status := range statuses {
+		mapped := sqlcv1.V1ReadableStatusOlap(status)
```
mrkaye97: how does this compile if the enums aren't the same......

I hate Go

cmd/hatchet-migrate/migrate/migrations/20260222120000_v1_0_82_durable_eviction.sql
```diff
@@ -0,0 +1,9 @@
+-- +goose Up
+
+ALTER TABLE v1_task_runtime ADD COLUMN IF NOT EXISTS evicted_at TIMESTAMPTZ DEFAULT NULL;
```
mrkaye97, line 3: if something is evicted multiple times, is the idea that we'll go from this being null to non-null to null again? that's fine with me, but just checking

examples/python/durable_eviction/worker.py
```diff
@@ -0,0 +1,67 @@
...
+
+@hatchet.durable_task(
+    execution_timeout=timedelta(minutes=5),
+    eviction=EvictionPolicy(
+        ttl=timedelta(seconds=EVICTION_TTL_SECONDS),
+        allow_capacity_eviction=True,
+        priority=0,
+    ),
+)
+async def evictable_sleep(input: EmptyModel, ctx: DurableContext) -> dict[str, object]:
```
mrkaye97: nit: awkward return type

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -39,10 +41,30 @@ func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context
 			InsertedAt: cb.DurableTaskInsertedAt,
 		}
 
-		dispatcherId, ok := idInsertedAtToDispatcherId[key]
+		dispatcherLookup, ok := idInsertedAtToDispatcherId[key]
```
mrkaye97, line 44: I sort of think this name was fine? 😅

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -39,10 +41,30 @@ func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context
...
 
 		if !ok {
-			tc.l.Warn().Msgf("could not find dispatcher id for task %d inserted at %s, skipping callback", cb.DurableTaskId, cb.DurableTaskInsertedAt.Time)
+			tc.l.Warn().Msgf("no runtime/dispatcher lookup row for task %d, skipping callback delivery", cb.DurableTaskId)
+			continue
+		}
+
+		if dispatcherLookup.IsEvicted {
```
mrkaye97, line 51: maybe I'm confused - isn't `dispatcherLookup` (previously `dispatcherId`) here a uuid?

mrkaye97: oh I see, we changed the return type of that getter 👍 

I think this is a bit confusing though - why is `IsEvicted` on the dispatcher?

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -59,7 +81,7 @@ func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context
 			continue
 		}
 
-		dispatcherToMsgs[dispatcherId] = append(dispatcherToMsgs[dispatcherId], msg)
+		dispatcherToMsgs[*dispatcherLookup.DispatcherId] = append(dispatcherToMsgs[*dispatcherLookup.DispatcherId], msg)
```
mrkaye97: nit: let's factor out `dispatcherLookup.DispatcherId` into a variable `dispatcherId` on line 65 before we do the nil check for readability?

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -72,3 +94,116 @@ func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context
...
+		reasonByExternalId[msg.TaskExternalId] = msg.Reason
+	}
+
+	flatTasks, err := tc.repov1.Tasks().FlattenExternalIds(ctx, tenantId, externalIds)
+	if err != nil {
+		return fmt.Errorf("failed to batch-lookup tasks for restore: %w", err)
+	}
+
+	if len(flatTasks) == 0 {
+		tc.l.Warn().Msgf("no tasks found for %d restore messages", len(msgs))
```
mrkaye97: probably don't need this log? this is a fine state to be in IMO

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -72,3 +94,116 @@ func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context
...
+				EventMessage:   fmt.Sprintf("Restoring evicted task: %s", reason),
+			},
+		)
+		if err == nil {
+			if pubErr := tc.pubBuffer.Pub(ctx, msgqueue.OLAP_QUEUE, olapMsg, false); pubErr != nil {
+				tc.l.Warn().Err(pubErr).Msg("failed to publish DURABLE_RESTORING to OLAP")
+			}
+		}
+
+		if restored.Queue != "" {
```
mrkaye97, line 164: if this can be an empty string, should we make it nullable instead?

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -72,3 +94,116 @@ func (tc *TasksControllerImpl) processSatisfiedEventLogEntry(ctx context.Context
...
+	}
+
+	if len(queues) > 0 {
+		tc.notifySchedulerQueues(ctx, tenantId, queues)
+	}
+
+	return nil
+}
+
+func (tc *TasksControllerImpl) notifySchedulerQueues(ctx context.Context, tenantId uuid.UUID, queues map[string]struct{}) {
```
mrkaye97: seems like this should return `error`?

internal/services/dispatcher/v1/server.go
mrkaye97: hmm... I think the footgun here is we need to make sure that _somehow_ all old invocations are evicted, so as long as we can guarantee that, then sure, I think exiting here is fine. But I'm nervous if we exit here then we'll fail to evict old invocations somehow

pkg/repository/sqlcv1/tasks-overwrite.go
```diff
@@ -850,6 +850,114 @@ func (q *Queries) ReleaseTasks(ctx context.Context, db DBTX, arg ReleaseTasksPar
...
+WITH input AS (
+    SELECT
+        task_id, task_inserted_at, retry_count
+    FROM
+        (
+            SELECT
+                unnest($1::bigint[]) AS task_id,
+                unnest($2::timestamptz[]) AS task_inserted_at,
+                unnest($3::integer[]) AS retry_count
+        ) AS subquery
```
mrkaye97: I know we have this all over, but we don't need it (would prefer to remove in the future 😅). We can do this:

```sql
WITH input AS (
      SELECT
          unnest($1::bigint[]) AS task_id,
          unnest($2::timestamptz[]) AS task_inserted_at,
          unnest($3::integer[]) AS retry_count
)
```

pkg/repository/sqlcv1/tasks-overwrite.go
```diff
@@ -850,6 +850,114 @@ func (q *Queries) ReleaseTasks(ctx context.Context, db DBTX, arg ReleaseTasksPar
...
+        AND r.evicted_at IS NOT NULL
+    ORDER BY r.task_id, r.task_inserted_at, r.retry_count
+    FOR UPDATE
+), selected_tasks AS (
+    SELECT
+        t.*
+    FROM
+        v1_task t
+    JOIN
+        evicted_runtimes er ON t.id = er.task_id AND t.inserted_at = er.task_inserted_at
```
mrkaye97, line 836: do we need to join on the retry count here too?

pkg/repository/sqlcv1/tasks-overwrite.go
```diff
@@ -850,6 +850,114 @@ func (q *Queries) ReleaseTasks(ctx context.Context, db DBTX, arg ReleaseTasksPar
...
+    FROM
+        selected_tasks t
+    ON CONFLICT DO NOTHING
+    RETURNING task_id, task_inserted_at
+)
+SELECT
+    st.id AS task_id,
+    st.inserted_at AS task_inserted_at,
+    st.retry_count,
+    CASE WHEN iq.task_id IS NOT NULL THEN 1 ELSE 0 END::int AS queued,
```
mrkaye97: should this be a boolean instead of a 0 or 1?

pkg/repository/sqlcv1/tasks-overwrite.go
```diff
@@ -850,6 +850,114 @@ func (q *Queries) ReleaseTasks(ctx context.Context, db DBTX, arg ReleaseTasksPar
...
+type RestoreEvictedTasksParams struct {
+	Taskids         []int64              `json:"taskids"`
+	Taskinsertedats []pgtype.Timestamptz `json:"taskinsertedats"`
+	Retrycounts     []int32              `json:"retrycounts"`
+	Tenantid        uuid.UUID            `json:"tenantid"`
+}
+
+type RestoreEvictedTasksRow struct {
+	TaskInsertedAt pgtype.Timestamptz `json:"task_inserted_at"`
+	Queue          string             `json:"queue"`
```
mrkaye97, line 874: maybe this was that variable from before that could've been a pointer?

pkg/repository/sqlcv1/workers.sql
```diff
@@ -502,7 +502,7 @@ SELECT
     rt.*,
     w."durableTaskDispatcherId"
 FROM v1_task_runtime rt
-JOIN "Worker" w ON rt.worker_id = w.id
+LEFT JOIN "Worker" w ON rt.worker_id = w.id
```
mrkaye97, line 505: intentional to change this with no other changes to this query?

pkg/repository/task.go
```diff
@@ -4125,16 +4192,21 @@ func (r *TaskRepositoryImpl) GetWorkflowRunResultDetails(ctx context.Context, te
 	}
 
 	externalIdToIsRunning := make(map[string]bool)
+	externalIdToIsEvicted := make(map[string]bool)
```
mrkaye97: these should probably be `map[uuid.UUID]bool` instead

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
```diff
@@ -382,10 +401,63 @@ async def wait_for_callback(
...
+    def cleanup_task_state(self, durable_task_external_id: str) -> None:
+        """Remove all pending callbacks, acks, and buffered completions for a task."""
+        stale_cb_keys = [
+            k for k in self._pending_callbacks if k[0] == durable_task_external_id
+        ]
+        for k in stale_cb_keys:
+            fut = self._pending_callbacks.pop(k)
+            if not fut.done():
+                fut.cancel()
+
```
mrkaye97, line 441: shouldn't we only do this for old invocations?

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
```diff
@@ -382,10 +401,63 @@ async def wait_for_callback(
...
+        stale_ack_keys = [
+            ak for ak in self._pending_event_acks if ak[0] == durable_task_external_id
+        ]
+        for ak in stale_ack_keys:
+            ack_fut = self._pending_event_acks.pop(ak)
+            if not ack_fut.done():
+                ack_fut.cancel()
```
mrkaye97, line 450: same here

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
```diff
@@ -83,9 +85,13 @@ def __init__(self, config: ClientConfig, admin_client: AdminClient):
         self._pending_event_acks: dict[
             PendingEventAck, asyncio.Future[DurableTaskEventAck]
         ] = {}
+        self._pending_eviction_acks: dict[PendingEvictionAck, asyncio.Future[None]] = {}
         self._pending_callbacks: dict[
             PendingCallback, asyncio.Future[DurableTaskEventLogEntryResult]
         ] = {}
+        self._early_completions: dict[
+            PendingCallback, DurableTaskEventLogEntryResult
+        ] = {}
```
mrkaye97, line 98: what's an early completion here?

sdks/python/hatchet_sdk/context/context.py
mrkaye97: bumping this

sdks/python/hatchet_sdk/runnables/eviction.py
```diff
@@ -0,0 +1,38 @@
...
+    model_config = ConfigDict(frozen=True)
+
+    ttl: timedelta | None
+    """Maximum continuous waiting duration before TTL-eligible eviction."""
+
+    allow_capacity_eviction: bool = True
+    """Whether this task may be evicted under durable-slot pressure."""
+
+    priority: int = 0
+    """Lower values are evicted first when multiple candidates exist."""
```
mrkaye97: let's follow the existing docs patterns we have for things like this (might be helpful to look at e.g. the `ConcurrencyExpression`)

sdks/python/hatchet_sdk/worker/durable_eviction/cache.py
```diff
@@ -0,0 +1,204 @@
...
+
+from pydantic import BaseModel, ConfigDict
+
+from hatchet_sdk.logger import logger
+from hatchet_sdk.runnables.action import ActionKey
+from hatchet_sdk.runnables.eviction import EvictionPolicy
+
+
+class DurableRunRecord(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
mrkaye97: do we need `arbitrary_types_allowed` here? Not too sure, but I feel like maybe this will just work

sdks/python/hatchet_sdk/worker/durable_eviction/cache.py
```diff
@@ -0,0 +1,204 @@
...
+    step_run_id: str
+    eviction: EvictionPolicy | None
+    registered_at: datetime
+
+    # Waiting state
+    waiting_since: datetime | None = None
+    wait_kind: str | None = None
+    wait_resource_id: str | None = None
+
+    def is_waiting(self) -> bool:
```
mrkaye97, line 24: this should probably be a `property`

sdks/python/hatchet_sdk/worker/durable_eviction/cache.py
```diff
@@ -0,0 +1,204 @@
...
+        return self.waiting_since is not None
+
+
+class DurableEvictionCache(Protocol):
```
mrkaye97: I haven't read through this at all, but I find it very unlikely that we need a protocol here

sdks/python/hatchet_sdk/worker/durable_eviction/cache.py
```diff
@@ -0,0 +1,204 @@
...
+
+class InMemoryDurableEvictionCache(DurableEvictionCache):
+    def __init__(self) -> None:
+        self._runs: dict[ActionKey, DurableRunRecord] = {}
+
+    def register_run(
+        self,
+        key: ActionKey,
+        step_run_id: str,
+        *,
```
mrkaye97: let's not do this - seems not super worth it to force positional args for the key and id and not allow them for `now` and the policy?

sdks/python/hatchet_sdk/worker/durable_eviction/cache.py
```diff
@@ -0,0 +1,204 @@
...
+    def get(self, key: ActionKey) -> DurableRunRecord | None:
+        return self._runs.get(key)
+
+    def get_all_waiting(self) -> list[DurableRunRecord]:
+        return [r for r in self._runs.values() if r.is_waiting()]
+
+    def mark_waiting(
+        self,
+        key: ActionKey,
+        *,
```
mrkaye97: here too

sdks/python/hatchet_sdk/worker/durable_eviction/cache.py
```diff
@@ -0,0 +1,204 @@
...
+    ) -> None:
+        rec = self._runs.get(key)
+        if not rec:
+            return
+
+        rec.waiting_since = now
+        rec.wait_kind = wait_kind
+        rec.wait_resource_id = resource_id
+
+    def mark_active(self, key: ActionKey, *, now: datetime) -> None:
```
mrkaye97: and here

sdks/python/hatchet_sdk/worker/durable_eviction/cache.py
```diff
@@ -0,0 +1,204 @@
...
+            return None
+
+        # Prefer TTL-eligible candidates first.
+        ttl_eligible: list[DurableRunRecord] = []
+        for r in waiting:
+            ttl = r.eviction.ttl if r.eviction else None
+            if ttl is None or r.waiting_since is None:
+                continue
+            if (now - r.waiting_since) >= ttl:
+                ttl_eligible.append(r)
```
mrkaye97: let's use a list comp for this

sdks/python/hatchet_sdk/worker/durable_eviction/cache.py
```diff
@@ -0,0 +1,204 @@
...
+
+        capacity_candidates: list[DurableRunRecord] = []
+        for r in waiting:
+            if not r.eviction or not r.eviction.allow_capacity_eviction:
+                continue
+            if r.waiting_since is None:
+                continue
+            if (now - r.waiting_since) < min_wait_for_capacity_eviction:
+                continue
+            capacity_candidates.append(r)
```
mrkaye97: list comp

sdks/python/hatchet_sdk/worker/durable_eviction/instrumentation.py
```diff
@@ -0,0 +1,53 @@
...
+    mgr = ctx_durable_eviction_manager.get()
+
+    if not key or mgr is None or not ctx_is_durable.get():
+        return None, None
+
+    return key, mgr
+
+
+@contextmanager
+def durable_eviction_wait(wait_kind: str, resource_id: str) -> Iterator[None]:
```
mrkaye97: do we need a sync version of this?

sdks/python/hatchet_sdk/worker/durable_eviction/instrumentation.py
```diff
@@ -0,0 +1,53 @@
...
+
+@asynccontextmanager
+async def aio_durable_eviction_wait(
+    wait_kind: str, resource_id: str
+) -> AsyncIterator[None]:
+    """
+    Async variant of `durable_eviction_wait`.
+    """
+
+    with durable_eviction_wait(wait_kind, resource_id):
```
mrkaye97: I'm nervous about this blocking

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -171,6 +184,29 @@ def run(self, action: Action) -> None:
             self.running_tasks.add(t)
             t.add_done_callback(lambda task: self.running_tasks.discard(task))
 
+    def _eviction_cancel_callback(self, key: ActionKey) -> None:
+        """Called from CancellationToken when the eviction manager evicts a run."""
```
mrkaye97: old comment?

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -171,6 +184,29 @@ def run(self, action: Action) -> None:
...
+        if key in self.contexts:
+            ctx = self.contexts[key]
+            ctx._set_cancellation_flag()
+            if self.durable_event_listener is not None and hasattr(ctx, "step_run_id"):
```
mrkaye97: let's not use `hasattr`

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -171,6 +184,29 @@ def run(self, action: Action) -> None:
...
+    def _make_eviction_request(
+        listener: DurableEventListener,
+    ) -> Callable[[ActionKey, DurableRunRecord], Awaitable[None]]:
+        async def _request(key: ActionKey, rec: DurableRunRecord) -> None:
+            await listener.send_evict_invocation(
+                durable_task_external_id=rec.step_run_id,
+                invocation_count=0,
+            )
+
+        return _request
```
mrkaye97: this seems a little odd to need to do (having a closure here with the async function) - not sure what the right way is off the top of my head, but just seemed unusual

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -270,6 +306,10 @@ async def async_wrapped_action_func(
         ctx_durable_context.set(
             ctx if isinstance(ctx, DurableContext) and task.is_durable else None
         )
+        ctx_is_durable.set(task.is_durable)
+        ctx_durable_eviction_manager.set(
+            self.durable_eviction_manager if task.is_durable else None
+        )
```
mrkaye97: I wonder if we should use a contextvar for this, or if we should prop drill it down to the `DurableContext` and then pass it into those methods instead? I think I'd prefer the second - IMO we should use contextvars as little as possible (basically only for when we really need to) and it seems like here we probably don't need to

sdks/python/hatchet_sdk/hatchet.py
```diff
@@ -551,6 +555,7 @@ def durable_task(
         backoff_max_seconds: int | None = None,
         default_filters: list[DefaultFilter] | None = None,
         default_additional_metadata: JSONSerializableMapping | None = None,
+        eviction: EvictionPolicy | None = DEFAULT_DURABLE_TASK_EVICTION_POLICY,
```
mrkaye97: nit: I might name this `eviction_policy` instead

cmd/hatchet-migrate/migrate/migrations/20260222120000_v1_0_82_durable_eviction.sql
grutt: yes

pkg/repository/sqlcv1/tasks-overwrite.go
grutt: agreed this is more readable... i'll make this change

pkg/repository/sqlcv1/tasks-overwrite.go
grutt: yes

pkg/repository/sqlcv1/tasks-overwrite.go
grutt: good catch, i was counting in some iteration of this query for no good reason

sdks/python/hatchet_sdk/hatchet.py
grutt: yeah was thinking the same actually...

sdks/python/hatchet_sdk/runnables/eviction.py
grutt: oh right, i was going to flag this... probably my ide setup but the current docstring pattern doesn't register in intellisense for me, but this format does...

grutt: this is on the actual param, not on the object itself

sdks/python/hatchet_sdk/context/context.py
grutt: are you on the latest commit? this was added back in...

sdks/python/hatchet_sdk/worker/durable_eviction/instrumentation.py
grutt: no, we don't. this was copied over from the before times

api/v1/server/handlers/v1/workflow-runs/list.go
grutt: good catch. strings are strings i guess...

internal/services/dispatcher/v1/server.go
grutt: lets talk through this sync. tracing this was helpful though, i wasn't properly handling the error case here which would be problematic

pkg/repository/sqlcv1/workers.sql
grutt: yes, since worker id is reset to null on eviction this needs to be left joined or else we wont list these tasks

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
grutt: i was seeing some races where the completion event was received before wait for callback is called... maybe this is checked at the start of the wait but maybe this is a separate issue we want to address differently

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
grutt: good catch

sdks/python/hatchet_sdk/worker/runner/runner.py
grutt: yeah seems reasonable

sdks/python/hatchet_sdk/worker/durable_eviction/cache.py
grutt: i think we can just simplify this to not have a base class for now since we only have one implementation

pkg/repository/sqlcv1/tasks-overwrite.go
mrkaye97: wait should it be a pointer? or should it not be?

grutt: not a pointer. it was incorrectly cast before. we aways have a queue

mrkaye97: ah nice!

api/v1/server/handlers/v1/workflow-runs/list.go
```diff
@@ -34,10 +66,7 @@ func (t *V1WorkflowRunsService) WithDags(ctx context.Context, request gen.V1Work
 
 	if request.Params.Statuses != nil {
 		if len(*request.Params.Statuses) > 0 {
-			statuses = []sqlcv1.V1ReadableStatusOlap{}
```
mrkaye97, line 37: one thing I'm curious about (which isn't just a thing for here, but is relevant since this is repeated) is that we have these nested checks and _then_ we call the helper function. I wonder if instead we should have the helper function do these nil and length checks too, especially since we ultimately assign back to this empty list of `statuses`. WDYT? (relevant in the blocks below too)

sdks/python/examples/durable_eviction/test_durable_eviction.py
```diff
@@ -0,0 +1,274 @@
...
+
+@pytest.mark.asyncio(loop_scope="session")
+async def test_non_evictable_task_completes(hatchet: Hatchet) -> None:
+    """A durable task with eviction disabled should finish normally."""
+    result = await non_evictable_sleep.aio_run()
+
+    assert result["status"] == "completed"
+    runtime = result["runtime"]
+    assert isinstance(runtime, (int, float))
+    assert runtime >= 10
```
mrkaye97: we should return a pydantic model from the task instead of a dict so we don't need to assert on types. I think I did this with a dict somewhere else, but shouldn't have

sdks/python/examples/durable_eviction/test_durable_eviction.py
```diff
@@ -0,0 +1,274 @@
...
+    ref = evictable_sleep.run_no_wait()
+
+    await _poll_until_status(hatchet, ref.workflow_run_id, V1TaskStatus.RUNNING)
+    details = await _poll_until_status(
+        hatchet, ref.workflow_run_id, V1TaskStatus.EVICTED
+    )
+    task_id = _get_task_id(details)
+
+    with hatchet.runs.client() as client:
+        TaskApi(client).v1_task_restore(task=task_id)
```
mrkaye97, line 111: do we need to do this directly because the restore method isn't meant to be exported? (this is fine, just checking)

sdks/python/examples/durable_eviction/test_durable_eviction.py
```diff
@@ -0,0 +1,274 @@
...
+
+    details = await _poll_until_status(
+        hatchet, ref.workflow_run_id, V1TaskStatus.RUNNING
+    )
+    statuses = {t.status for t in details.task_runs.values()}
+
+    assert (
+        V1TaskStatus.RUNNING in statuses
+    ), f"Expected RUNNING after restore, got: {statuses}"
```
mrkaye97, line 120: if a task was running, then evicted, but does _not_ resume, will this test not still pass since it was previously running?

sdks/python/examples/durable_eviction/test_durable_eviction.py
```diff
@@ -0,0 +1,274 @@
...
+    statuses = {t.status for t in details.task_runs.values()}
+
+    assert (
+        V1TaskStatus.COMPLETED in statuses
+    ), f"Expected COMPLETED after restore + event, got: {statuses}"
+
+
+# ---------------------------------------------------------------------------
+# evictable_child_spawn
+# ---------------------------------------------------------------------------
```
mrkaye97: can we remove these comment blocks? they're basically just the test names 🤦

sdks/python/examples/durable_eviction/test_durable_eviction.py
```diff
@@ -0,0 +1,274 @@
...
+
+
+# ---------------------------------------------------------------------------
+# multiple_eviction  (sleep -> evict -> restore -> sleep -> evict -> restore)
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.asyncio(loop_scope="session")
+async def test_multiple_eviction_cycle(hatchet: Hatchet) -> None:
+    """The task should survive two eviction+restore cycles."""
```
mrkaye97, line 243: this test too, just want to make sure that a previous running or evicted state doesn't make the assertions trivially pass

sdks/python/examples/durable_eviction/trigger.py
```diff
@@ -0,0 +1,6 @@
+"""Quick manual trigger for the evictable_sleep task."""
```
mrkaye97: rm comment

sdks/python/examples/durable_eviction/worker.py
```diff
@@ -0,0 +1,134 @@
...
+    """Has eviction disabled -- should never be evicted."""
+    start = time.time()
+    await ctx.aio_sleep_for(timedelta(seconds=10))
+    return {"runtime": time.time() - start, "status": "completed"}
+
+
+def main() -> None:
+    worker = hatchet.worker(
+        "eviction-worker",
+        slots=2,
```
mrkaye97: so I'm forgetting where we're at on this now - will `slots=2` do what we want here?

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
mrkaye97: huh... I'm not too sure but this feels a little confusing to me. I also am nervous about leaking memory here - is there any risk of that?

sdks/python/hatchet_sdk/context/context.py
mrkaye97: still looks removed to me?

sdks/python/hatchet_sdk/runnables/eviction.py
mrkaye97: I think this is pretty likely to be specifically a classes issue? e.g. for concurrency expression on the durable exec branch we have:

```python
class ConcurrencyExpression(BaseModel):
    """
    Defines concurrency limits for a workflow using a CEL expression.

    :ivar expression: CEL expression to determine concurrency grouping. (i.e. "input.user_id")
    :ivar max_runs: Maximum number of concurrent workflow runs.
    :ivar limit_strategy: Strategy for handling limit violations.


    **Example**
    ```python
    ConcurrencyExpression(
        "input.user_id",
        5,
        ConcurrencyLimitStrategy.CANCEL_IN_PROGRESS
    )
    ```
    """
```

mrkaye97 (approved):

sdks/python/examples/durable_eviction/test_durable_eviction.py
grutt: yeah, this is helpful for us and but i'm not sure if we want to make this public... tbh i'm not 100% we even want this endpoint at all. i dont know what the usecase would be for a manual restore. its a simple proxy endpoint so not much to maintain, but worth a thought before we merge everything

sdks/python/examples/durable_eviction/worker.py
grutt: i think thats just from the example i copied, ill remove it

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
grutt: adding a todo to look deeper at this

api/v1/server/handlers/v1/workflow-runs/list.go
grutt: i think we'll be rewriting this today to natively support these olap events, this is all too hacky... i have a todo-durable here

internal/services/controllers/task/durable_callbacks.go
grutt: this is a safety check.. we should never have an empty queue... adding a warning just in case here...

promptless-for-oss:

Updated existing suggestion: **[Document durable task eviction for Go and Python SDKs](https://app.gopromptless.ai/suggestions/b0f0cbdd-a099-4c1d-b03f-e32cb8f89e60)**

---
_Tip: Adjust how proactive or focused Promptless is in [Agent Settings](https://app.gopromptless.ai/configure/settings) ⚙️_

promptless-for-oss:

---
_Tip: Tell your friends working on non-commercial open-source projects to apply for free Promptless access at [promptless.ai/oss](https://promptless.ai/oss) ❤️_

promptless-for-oss:

New suggestion: **[Document configurable eviction policies for durable tasks](https://app.gopromptless.ai/suggestions/f27e6f6a-3b1c-480d-a134-70e40fa510b4)**

---
_Tip: Tell your friends working on non-commercial open-source projects to apply for free Promptless access at [promptless.ai/oss](https://promptless.ai/oss) ❤️_

PR #3088, opened 2026-02-23
streaming: document ~8KB limit and guard Python put_stream payload size

### LINKED ISSUE #3087: [DEVEX] Document put_stream payload size limit (~8KB) from pg_notify
## Summary
`put_stream()` appears to have an undocumented payload size constraint due to PostgreSQL `pg_notify` limits (~8KB).

When stream payloads get larger (in my case, JSON progress payloads with many active transfers), streaming fails and the SDK surfaces a generic internal error.

## Observed behavior
- Small payloads work normally
- Larger payloads fail with generic gRPC internal error in SDK
- Engine logs show Postgres error:
  - `ERROR: payload string too long (SQLSTATE 22023)`

## Suspected root cause
The engine sends stream payloads through Postgres NOTIFY:

`select pg_notify($1,$2)`

Postgres NOTIFY payload size is capped (roughly 8KB minus overhead), so oversized stream events fail.

## Why this is hard to debug
- `put_stream()` docs/docstring do not mention size limits
- Error shown to SDK user is generic (`An internal error occurred`)
- Event is effectively dropped

## Suggested improvements
1. **Docs:** add a note to streaming docs that `put_stream` payloads must stay under ~8KB due to `pg_notify`
2. **SDK guardrail (optional):** add a preflight size check/warning in `put_stream()` (similar spirit to log truncation guardrails)
3. **Error clarity (optional):** surface a more actionable error when payload is too large

## Workaround currently used
Truncate large arrays in stream payloads and include counts separately to stay under the limit.

## Environment
- Python SDK: `hatchet-sdk==1.25.2`
- Engine: self-hosted (Docker)
- PostgreSQL: 17.x

nigelcrabwell:
## Summary
Adds guardrails for oversized stream payloads and documents the practical size constraint.

## Why
`put_stream` payloads flow through PostgreSQL `NOTIFY` (effective ~8KB payload ceiling). Without guardrails, oversized payloads can fail with hard-to-debug internal errors.

## Changes
- **Docs:** add warning callout in `frontend/docs/pages/home/streaming.mdx` about the ~8KB limit
- **Python SDK:** add stream payload guard in `sdks/python/hatchet_sdk/clients/events.py`
  - If payload > 8,000 bytes, log warning and truncate to 8,000 bytes
  - Pattern intentionally mirrors existing `log()` truncation behavior

Closes #3087

### DIFF
--- a/frontend/docs/pages/home/streaming.mdx
+++ b/frontend/docs/pages/home/streaming.mdx
@@ -11,6 +11,13 @@ Hatchet tasks can stream data back to a consumer in real-time. This has a number
 
 You can stream data out of a task run by using the `put_stream` (or equivalent) method on the `Context`.
 
+<Callout type="warning" emoji="📏">
+  Stream event payloads are sent via PostgreSQL `NOTIFY`, which has a payload
+  size limit of roughly 8KB (minus envelope overhead). Keep stream payloads
+  small (for example: summaries, deltas, or a capped list) to avoid dropped
+  events or internal errors when publishing oversized messages.
+</Callout>
+
 <UniversalTabs items={["Python", "Typescript", "Go", "Ruby"]}>
   <Tabs.Tab title="Python">
 <Snippet src={snippets.python.streaming.worker.streaming} />
--- a/sdks/python/hatchet_sdk/clients/events.py
+++ b/sdks/python/hatchet_sdk/clients/events.py
@@ -255,6 +255,12 @@ class EventClient(BaseRestClient):
         else:
             raise ValueError("Invalid data type. Expected str, bytes, or file.")
 
+        if len(data_bytes) > 8_000:
+            logger.warning(
+                "truncating stream payload to 8,000 bytes to avoid PostgreSQL NOTIFY payload limits"
+            )
+            data_bytes = data_bytes[:8_000]
+
         request = PutStreamEventRequest(
             task_run_external_id=step_run_id,
             created_at=proto_timestamp_now(),


nigelcrabwell:
Clean test summary:

Ran targeted Python SDK tests for the `events.py` streaming/event changes against local Hatchet using:
- HATCHET_CLIENT_TLS_STRATEGY=none
- local API token
- HATCHET_CLIENT_HOST_PORT=localhost:7077
- HATCHET_CLIENT_SERVER_URL=http://localhost:8080

Results:
- examples/streaming/test_streaming.py ✅
- examples/events/test_event.py ✅
- examples/serde/test_serde.py ✅

Summary: 18 passed in 117.13s.

Note: teardown logs included:
`RuntimeError: aclose(): asynchronous generator is already running`
but all targeted tests passed and pytest exited 0.

sdks/python/hatchet_sdk/clients/events.py
```diff
@@ -255,6 +255,12 @@ def stream(self, data: str | bytes, step_run_id: str, index: int) -> None:
         else:
             raise ValueError("Invalid data type. Expected str, bytes, or file.")
 
+        if len(data_bytes) > 8_000:
+            logger.warning(
+                "truncating stream payload to 8,000 bytes to avoid PostgreSQL NOTIFY payload limits"
+            )
+            data_bytes = data_bytes[:8_000]
+
```
mrkaye97, line 263: Unfortunately, I don't think this makes sense to do on the SDK. We only use listen / notify in PG when we're running PG as the internal message queue instead of Rabbit, which we generally don't recommend for any HA deployments and don't do for our managed cloud offering. Because of that, in most cases this limit will not be hit, so there will be a warning but no actual issue to be warning about

frontend/docs/pages/home/streaming.mdx
```diff
@@ -11,6 +11,13 @@ Hatchet tasks can stream data back to a consumer in real-time. This has a number
 
 You can stream data out of a task run by using the `put_stream` (or equivalent) method on the `Context`.
 
+<Callout type="warning" emoji="📏">
+  Stream event payloads are sent via PostgreSQL `NOTIFY`, which has a payload
+  size limit of roughly 8KB (minus envelope overhead). Keep stream payloads
+  small (for example: summaries, deltas, or a capped list) to avoid dropped
+  events or internal errors when publishing oversized messages.
+</Callout>
```
mrkaye97, line 19: same comment as below - I believe this is only true when we're using postgres as the internal message queue. I'm unsure if we have this limit documented anywhere in our docs (we should!) but it affects more than streaming - it'll affect every message we pass over the queue.

nigelcrabwell:
Closing this since the issue was resolved in our environment after updating Hatchet engine to latest; local reproduction was on older v0.77.19.

PR #3089, opened 2026-02-23
Fix cross-strategy slot contamination in chained concurrency gates

mnafees:
# Description

Fixes a bug where chained concurrency strategies (e.g., CANCEL_NEWEST -> GROUP_ROUND_ROBIN) bypass the second gate entirely.

The `updated_slots` CTE in `RunChildCancelNewest` and `RunChildCancelInProgress` matched slots by `(task_id, task_inserted_at, task_retry_count, key)` but was missing `strategy_id` in the WHERE clause. When the first strategy's query ran again after its own slots were already filled, it matched the second strategy's unfilled slots - which share the same `(task_id, key)` - and filled them directly, bypassing the second strategy's scheduling logic entirely.

Added `strategy_id` and `tenant_id` to the `updated_slots` WHERE clause so each strategy can only fill its own slots.

## Type of change

### DIFF
--- a/pkg/repository/scheduler_concurrency.go
+++ b/pkg/repository/scheduler_concurrency.go
@@ -545,6 +545,8 @@ func (c *ConcurrencyRepositoryImpl) runCancelNewest(
 		}
 
 		if !parentAcquired {
+			// Log the event when the parent advisory lock is not acquired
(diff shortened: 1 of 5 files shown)


abelanger5 (approved):
nice find!

yedidyakfir:
Thx

PR #3090, opened 2026-02-23
Make sure we create older partitions for payload tables to match data retention policy

mnafees:
# Description

We do not backfill `v1_payload` and `v1_payloads_olap` with older partitions that respect the current data retention policy leading to possible `no partition of relation` SQL errors.

## Type of change


mrkaye97 (commented):
I wonder if there's an easier way here 🤔 like we might be able to call `create_v1_range_partition` in a `SELECT` statement, e.g.:

```sql
WITH dates_to_create AS (
    somehow get a column of all the dates to create partitions for
)
SELECT create_v1_range_partition('v1_payload', dt)
FROM dates_to_create
;
```

I haven't tested it, but I think it might work and would be simpler than looping

mnafees:
Closing this because it is quite stale at this point

PR #3102, opened 2026-02-25
Fix type syntax for dependencies (and tests on main)

### LINKED ISSUE #3101: [BUG] Python sdk release broken
**Describe the issue**
See https://github.com/hatchet-dev/hatchet/actions/runs/22372826448 - apparently my tests breaks on some versions of python. I can recreate it locally with python 3.13. Sorry about that.

It means that my fix only works for python 3.14, so as such the PR didn't introduce any bugs, but it just only fixed the particular bug on the version

Apparently it's not very easy to determine if something is a TypeAliasType.
I saw that `fastapi` uses `i[s_typealiastype](https://pydantic.github.io/typing-inspection/dev/api/typing_objects/#typing_inspection.typing_objects.is_typealiastype)` from `typing_inspection` to do it. Maybe that's better then? The library is maintained by Pydantic so it should be safe. Only dependency is `typing_extensions` which is already a dependency.

**Environment**
- SDK: python 1.25.3
- Engine: [Cloud or Self-hosted v1.5.0]

**Expected behavior**
Tests should not fail and the release should happen.
Ideally the `type X = ...` syntax will also work python versions other than 3.14

See test in action

**Additional context**

We could also do a `if getattr(obj, "__value__")` but this fails for other objects with that specific field

@mrkaye97 Pinging you since it's related to the PR we communicated on

StefanBRas:
# Description

Fixes broken tests on main introduced by #2951.
Also makes it correctly work for python version 3.12 and 3.13 now, which would fail before.

Adds a dependency on `typing-inspection` to handle all the cases with `TypeAliasType`.
I did a minor bump because it's adding a dependency. Also, I just removed the change log intro for the earlier version, since it was never released.

Fixes #3101

## Type of change

## What's Changed

- [ ] Fixes broken tests on main introduced by and makes `type X = ...` syntax correctly work for dependencies for python version 3.12 and 3.13

### DIFF
--- a/sdks/python/CHANGELOG.md
+++ b/sdks/python/CHANGELOG.md
@@ -11,12 +11,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixes dependencies not working when using `type Dependency = Annotated[..., ...]` syntax for annotations on python version 3.12 and 3.13. Adds `typing-inspection` as a dependency.
 
-## [1.25.3] - 2026-02-24
-
-### Fixed
-
-- Fixes dependencies not working when using `type Dependency = Annotated[..., ...]` syntax for annotations.
-
 ### Changed
 
 - Changes one function in the python SDK to use `inspect.iscoroutinefunction` instead of `asyncio.iscoroutinefunction` which is deprecated.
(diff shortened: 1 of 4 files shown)


StefanBRas:
@mrkaye97 this is ready for review now. There's a test fail, but it looks unrelated to these changes

mrkaye97 (approved):

PR #3103, opened 2026-02-25
Pool gzip writers to reduce RabbitMQ message compression allocations

mnafees:
# Description

- Pool `gzip.Writer` instances via `sync.Pool` in RabbitMQ message compression instead of allocating a new writer per payload
- Add benchmarks and roundtrip tests for `compressPayloads`
- Add a go-benchmarks GitHub Actions workflow that runs `benchstat` comparisons on PRs

## Benchmark results

| Benchmark | Before (ns/op) | After (ns/op) | Before (B/op) | After (B/op) | Before (allocs) | After (allocs) |
|---|---|---|---|---|---|---|
| 1x10KiB | 135,764 | 71,999 | 824,943 | 11,187 | 22 | 5 |
| 10x10KiB | 1,288,930 | 730,845 | 8,248,852 | 111,186 | 202 | 32 |
| 10x100KiB | 9,549,054 | 9,125,271 | 11,191,558 | 3,058,775 | 232 | 63 |
| Concurrent 5x10KiB | 155,614 | 33,664 | 4,124,444 | 55,669 | 102 | 17 |

- **~74x fewer bytes allocated** for small payloads (825 KiB → 11 KiB)
- **~3.7x fewer bytes** for large payloads (10.7 MiB → 2.9 MiB)
- **~1.9x faster** single-goroutine, **~4.6x faster** concurrent
- **~4-6x fewer allocations** across all cases

## Type of change


internal/msgqueue/rabbitmq/gzip.go
```diff
@@ -53,17 +60,23 @@ func (t *MessageQueueImpl) compressPayloads(payloads [][]byte) (*CompressionResu
...
 
-		if _, err := gzipWriter.Write(payload); err != nil {
-			gzipWriter.Close()
+		w := gzipWriterPool.Get().(*gzip.Writer)
+		w.Reset(&buf)
+
+		if _, err := w.Write(payload); err != nil {
+			w.Close()
+			gzipWriterPool.Put(w)
 			return nil, fmt.Errorf("failed to write to gzip writer: %w", err)
```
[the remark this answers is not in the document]
mnafees: This is safe to do since after getting a writer from the pool, we always call `Reset()` which makes sure to reinitialize the gzip writer internal state.

internal/msgqueue/rabbitmq/gzip.go
```diff
@@ -53,17 +60,23 @@ func (t *MessageQueueImpl) compressPayloads(payloads [][]byte) (*CompressionResu
...
 		}
 
-		if err := gzipWriter.Close(); err != nil {
+		if err := w.Close(); err != nil {
+			gzipWriterPool.Put(w)
 			return nil, fmt.Errorf("failed to close gzip writer: %w", err)
```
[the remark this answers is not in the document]
mnafees: This is safe to do since after getting a writer from the pool, we always call Reset() which makes sure to reinitialize the gzip writer internal state.

internal/msgqueue/rabbitmq/gzip.go
```diff
@@ -53,17 +64,21 @@ func (t *MessageQueueImpl) compressPayloads(payloads [][]byte) (*CompressionResu
 
 	for i, payload := range payloads {
 		var buf bytes.Buffer
-		gzipWriter := gzip.NewWriter(&buf)
 
-		if _, err := gzipWriter.Write(payload); err != nil {
-			gzipWriter.Close()
+		w := gzipWriterPool.Get().(*gzip.Writer)
```
mrkaye97, line 70: I know this is yet another "Go generics don't work" thing, but needing the type assertion here is so. stupid.

internal/msgqueue/rabbitmq/gzip_test.go
```diff
@@ -0,0 +1,132 @@
...
+	mq := newMQ()
+	payloads := generatePayloads(5, 10*1024)
+
+	result, err := mq.compressPayloads(payloads)
+	if err != nil {
+		t.Fatalf("compressPayloads: %v", err)
+	}
+
+	if !result.WasCompressed {
+		t.Fatal("expected WasCompressed to be true")
```
mrkaye97: is there no equivalent of `assert` or `expect`? like `expect(result.WasCompressed, true)`? this seems kinda odd, but maybe it's just how `testing` works? idk

mrkaye97 (approved):

internal/msgqueue/rabbitmq/gzip.go
```diff
@@ -5,6 +5,11 @@ import (
 	"compress/gzip"
 	"fmt"
 	"io"
+
+	// we need to do this aliasing to avoid the load-deadlock GitHub action from
+	// failing which depends on replacing the standard "sync" package with
+	// github.com/sasha-s/go-deadlock
```
abelanger5: this feels fragile and any usage of the `sync` package seems like it'll bite us in the future. can we modify that github action so that it greps for `sync.Mutex` and `sync.RWMutex` and writes them to `deadlocksync.Mutex` and `deadlocksync.RWMutex` instead? There will be some casing involving in the imports, but that feels like a better approach to me?

internal/msgqueue/rabbitmq/gzip.go
abelanger5: To add to this -- it should be pretty trivial to wrap `stdsync.Pool` to implement a generic interface? not critical but just a thought

internal/msgqueue/rabbitmq/gzip.go
```diff
@@ -17,6 +22,12 @@ type CompressionResult struct {
 	CompressionRatio float64
 }
 
+var gzipWriterPool = stdsync.Pool{
```
abelanger5: Do we want to limit the size of the pool at all? This seems like it could grow unbounded, though practically that should never be an issue. Don't think this is critical, just flagging

abelanger5 (approved):

PR #3112, opened 2026-02-26
Feat: Durable Memoization

mrkaye97:
# Description

Fills in the `MEMO` functionality on the engine / Python SDK

todo: test nondeterminism checks work here

## Type of change


promptless-for-oss:

Updated existing suggestion: **[Document Memo() function for durable tasks across Python, TypeScript, and Go SDKs](https://app.gopromptless.ai/suggestions/237575c2-9ef0-4b37-8877-1c7e983a14c3)**

---
_Tip: Whenever you leave a comment tagged `@Promptless` on a Promptless PR, Promptless will remember it for future suggestions 🧠_

api-contracts/v1/dispatcher.proto
```diff
@@ -9,12 +9,23 @@ import "v1/shared/trigger.proto";
...
     rpc DurableTask(stream DurableTaskRequest) returns (stream DurableTaskResponse) {}
+    rpc LookUpCachedDurableMemoEntry(LookUpCachedDurableMemoEntryRequest) returns (LookUpCachedDurableMemoEntryResponse) {}
 
     // NOTE: deprecated after DurableEventLog is implemented
     rpc RegisterDurableEvent(RegisterDurableEventRequest) returns (RegisterDurableEventResponse) {}
     rpc ListenForDurableEvent(stream ListenForDurableEventRequest) returns (stream DurableEvent) {}
 }
 
+message LookUpCachedDurableMemoEntryRequest {
+    string task_run_external_id = 1;
```
grutt: does this need to be scoped to a branch node or invocation?

examples/python/durable/worker.py
```diff
@@ -275,6 +275,60 @@ async def durable_replay_reset(
...
+    await asyncio.sleep(SLEEP_TIME)
+
+    return SleepResult(message=message, duration=SLEEP_TIME)
+
+
+@hatchet.durable_task(input_validator=MemoInput)
+async def memo_task(input: MemoInput, ctx: DurableContext) -> SleepResult:
+    start = time.time()
+    res = await ctx.aio_memo(
+        fn=lambda: expensive_computation(input.message),
```
grutt: hmm... we don't use lambdas anywhere else is this the pattern we want to use?

perhaps with block would be more ergonomic here?
```
with ctx.memo(...): 
```

api-contracts/v1/dispatcher.proto
mrkaye97: hmm, that's a good question - maybe we do need to scope to a node, lemme see

examples/python/durable/worker.py
grutt: i guess this wont work since we need the result

sdks/python/hatchet_sdk/context/context.py
```diff
@@ -36,13 +46,22 @@
...
     from hatchet_sdk.runnables.task import Task
     from hatchet_sdk.runnables.workflow import (
         BaseWorkflow,
     )
 
 
+def _compute_memo_key(task_run_external_id: str, deps: list[Any]) -> bytes:
+    h = hashlib.sha256()
+    h.update(task_run_external_id.encode())
+    h.update(json.dumps(deps, default=str, sort_keys=True).encode())
```
grutt: are we enforcing that deps need to be serializable in some way?

grutt (commented):
Looks great! A few open questions to hash out

examples/python/durable/worker.py
mrkaye97: the lambda is actually a bad solution too because I think this will block - we're going to need to pass a function and then its arguments separately, I guess

sdks/python/hatchet_sdk/context/context.py
mrkaye97: I think the best we can do is have this raise if they're not

examples/python/durable/worker.py
```diff
@@ -294,41 +294,14 @@ async def expensive_computation(message: str) -> SleepResult:
 async def memo_task(input: MemoInput, ctx: DurableContext) -> SleepResult:
     start = time.time()
     res = await ctx.aio_memo(
-        fn=lambda: expensive_computation(input.message),
-        deps=[input.message],
-        result_validator=SleepResult,
+        expensive_computation,
+        SleepResult,
+        input.message,
```
grutt, line 299: can this only take one dep now?

grutt: i see.... you just add n dependencies inline here.. i'm not sure i prefer that over deps=[]

grutt: which is more python-native?

grutt (commented):
looking great! one last thing to agree upon

examples/python/durable/worker.py
mrkaye97: We can’t do the deps thing that useMemo does because 1) the type checking doesn’t work and 2) it’s misleading because the dependencies really are only sort of the cache key (other stuff is added + the node id matters) so IMO this is the only way (same for TS most likely)

grutt: 😅 i'm not sure i follow... lets sync on this tomorrow morning... want to add a todo to revisit and get this merged though?

grutt (approved):

examples/python/durable/worker.py
mrkaye97: For type checking, the problem is we can’t use a lambda since it’ll block, which means we need to pass the deps as parameters to the function, which is different than how useMemo works. But then, if they’re being passed and the function is e.g. (int, str) -> bool, the deps would need to be an array of int | str, which doesn’t work by position (it can’t know that the first item is int and the second is string). We could fix this by making deps a tuple instead, but IMO this is a big variation from prior art. This is much closer to other more common uses of partial application (this is how it’s done basically everywhere else, e.g. asyncio)

for the cache key, the thing is that if the deps change on the second run, it’s pretty likely the result will change. In that case, should we raise a non determinism error? Or should we allow it and persist the new result? And otherwise, the key obviously also includes the run id and the node id and branch id (implicitly) so it’s a little misleading because to the user, deps seems like the only thing included there when it’s not

happy to discuss more tomorrow, not sure if that helps at all (also on mobile so excuse typos and such, not that I avoid them on desktop anyways)

promptless-for-oss:

Updated existing suggestion: **[Document Memo() function for durable tasks across Python, TypeScript, and Go SDKs](https://app.gopromptless.ai/suggestions/237575c2-9ef0-4b37-8877-1c7e983a14c3)**

---

promptless-for-oss:

Updated existing suggestion: **[Document aio_memo() method for durable task memoization in Python SDK](https://app.gopromptless.ai/suggestions/b8893ab0-6bf7-4fba-89b7-8e69df03a8f3)**

---

PR #3115, opened 2026-02-27
Feat  durable olap refactor

grutt:
# Description

Adds evicted state to all the OLAP things!
Fixes de-dupe

## Type of change


promptless-for-oss:

Updated existing suggestion: **[Document durable task eviction for Go and Python SDKs](https://app.gopromptless.ai/suggestions/b0f0cbdd-a099-4c1d-b03f-e32cb8f89e60)**

---

cmd/hatchet-migrate/migrate/migrations/20260226120001_v1_0_83_b_olap_evicted_status.sql
```diff
@@ -0,0 +1,61 @@
...
+        FOR date_partition IN
+            SELECT inhrelid::regclass::text AS partition_name
+            FROM pg_inherits
+            WHERE inhparent = parent_table::regclass
+        LOOP
+            PERFORM create_v1_partition_with_status(date_partition.partition_name, 'EVICTED');
+        END LOOP;
+    END LOOP;
+END;
+$$;
```
mrkaye97, line 92: I mentioned the same thing to Nafees on a different PR - I wonder if we can do this without a loop, with something like:

```sql
WITH partitions AS (
      SELECT inhrelid::regclass::text AS partition_name
      FROM pg_inherits
      WHERE inhparent = parent_table::regclass
)

SELECT create_v1_partition_with_status(partition_name, 'EVICTED')
FROM partitions
```

I haven't tested in this specific case, but this worked for me on local:

```
hatchet=# with dts as (select distinct inserted_at::date as iat from v1_task limit 5) select create_v1_range_partition('v1_task', iat) from dts; 
 create_v1_range_partition 
---------------------------
                         0
                         0
                         0
                         0
                         0
(5 rows)
```

cmd/hatchet-migrate/migrate/migrations/20260226120000_v1_0_83_olap_evicted_status.sql
```diff
@@ -0,0 +1,61 @@
...
+            FROM pg_inherits
+            WHERE inhparent = parent_table::regclass
+        LOOP
+            PERFORM create_v1_partition_with_status(date_partition.partition_name, 'EVICTED');
+        END LOOP;
+    END LOOP;
+END;
+$$;
+-- +goose StatementEnd
+
```
mrkaye97: we probably need to run analyze at the end of this migration

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/index.tsx
```diff
@@ -60,6 +60,8 @@ function statusToBadgeVariant(status: V1TaskStatus) {
       return 'cancelled';
     case V1TaskStatus.QUEUED:
       return 'queued';
+    case V1TaskStatus.EVICTED:
+      return 'evicted';
```
mrkaye97, line 66: can we remove the default here and add an exhaustive check (likely gonna leave this comment a bunch of places...)

frontend/app/src/pages/main/v1/workflow-runs-v1/components/run-statuses.tsx
```diff
@@ -14,6 +14,8 @@ function createV2IndicatorVariant(eventType: V1TaskStatus | undefined) {
       return 'border-transparent rounded-full bg-slate-500';
     case V1TaskStatus.COMPLETED:
       return 'border-transparent rounded-full bg-green-500';
+    case V1TaskStatus.EVICTED:
+      return 'border-transparent rounded-full bg-indigo-500';
     default:
```
mrkaye97: exhaustive check

frontend/app/src/pages/main/v1/workflow-runs/components/run-statuses.tsx
```diff
@@ -62,6 +62,8 @@ function createV1RunStatusVariant(status: V1TaskStatus): RunStatusVariant {
       return { text: 'Running', variant: 'inProgress' };
     case V1TaskStatus.QUEUED:
       return { text: 'Queued', variant: 'queued' };
+    case V1TaskStatus.EVICTED:
+      return { text: 'Evicted', variant: 'evicted' };
```
mrkaye97, line 66: exhaustive check

internal/services/controllers/task/durable_callbacks.go
```diff
@@ -134,6 +134,17 @@ func (tc *TasksControllerImpl) handleDurableRestoreTask(ctx context.Context, ten
...
+	invCountOpts := make([]v1.IdInsertedAt, 0, len(flatTasks))
+	for _, t := range flatTasks {
+		invCountOpts = append(invCountOpts, v1.IdInsertedAt{ID: t.ID, InsertedAt: t.InsertedAt})
+	}
+
+	invocationCounts, err := tc.repov1.DurableEvents().GetDurableTaskInvocationCounts(ctx, tenantId, invCountOpts)
+	if err != nil {
+		tc.l.Warn().Err(err).Msg("could not get durable task invocation counts for restoring tasks")
+		invocationCounts = make(map[v1.IdInsertedAt]*int32)
+	}
```
mrkaye97, line 145: I feel like this should return the error instead?

internal/services/dispatcher/server_v1.go
```diff
@@ -592,22 +602,23 @@ func (s *DispatcherImpl) sendStepActionEventV1(ctx context.Context, request *con
...
 	case contracts.StepActionEventType_STEP_EVENT_TYPE_FAILED:
-		return s.handleTaskFailed(ctx, task, retryCount, request)
+		return s.handleTaskFailed(ctx, task, retryCount, durableInvCount, request)
 	}
 
 	return nil, status.Errorf(codes.InvalidArgument, "invalid task external run id %s", request.TaskRunExternalId)
 }
 
-func (s *DispatcherImpl) handleTaskStarted(inputCtx context.Context, task *sqlcv1.FlattenExternalIdsRow, retryCount int32, request *contracts.StepActionEvent) (*contracts.ActionEventResponse, error) {
+func (s *DispatcherImpl) handleTaskStarted(inputCtx context.Context, task *sqlcv1.FlattenExternalIdsRow, retryCount int32, durableInvocationCount int32, request *contracts.StepActionEvent) (*contracts.ActionEventResponse, error) {
```
mrkaye97: extreme Go nit

```suggestion
func (s *DispatcherImpl) handleTaskStarted(inputCtx context.Context, task *sqlcv1.FlattenExternalIdsRow, retryCount, durableInvocationCount int32, request *contracts.StepActionEvent) (*contracts.ActionEventResponse, error) {
```

internal/services/scheduler/v1/scheduler.go
```diff
@@ -430,6 +430,20 @@ func (s *Scheduler) scheduleStepRuns(ctx context.Context, tenantId uuid.UUID, re
...
+			invCountOpts = append(invCountOpts, repov1.IdInsertedAt{
+				ID:         a.QueueItem.TaskID,
+				InsertedAt: a.QueueItem.TaskInsertedAt,
+			})
+		}
+
+		invocationCounts, invCountErr := s.repov1.DurableEvents().GetDurableTaskInvocationCounts(ctx, tenantId, invCountOpts)
+		if invCountErr != nil {
+			s.l.Warn().Err(invCountErr).Msg("could not get durable task invocation counts for assigned tasks")
+			invocationCounts = make(map[repov1.IdInsertedAt]*int32)
```
mrkaye97: this should probably error too

sdks/python/hatchet_sdk/worker/durable_eviction/cache.py
```diff
@@ -157,8 +172,32 @@ def select_eviction_candidate(
...
+    match cause:
+        case EvictionCause.TTL_EXCEEDED:
+            ttl_str = f" ({ttl})" if ttl else ""
+            return f"Wait TTL{ttl_str} exceeded while waiting on {wait_desc}"
+        case EvictionCause.CAPACITY_PRESSURE:
+            return f"Worker at capacity while waiting on {wait_desc}"
+        case EvictionCause.WORKER_SHUTDOWN:
+            return f"Worker shutdown while waiting on {wait_desc}"
+        case _ as unreachable:
+            assert_never(unreachable)
```
mrkaye97, line 203: unfortunately this won't work 😭 `assert_never` was introduced in 3.11. there are some workarounds here: https://typing.python.org/en/latest/guides/unreachable.html#assert-never-and-exhaustiveness-checking

mrkaye97 (commented):
left a couple comments but this looks really good to me!

cmd/hatchet-migrate/migrate/migrations/20260226120001_v1_0_83_b_olap_evicted_status.sql
grutt: neat, i like this and i think it will work

cmd/hatchet-migrate/migrate/migrations/20260226120000_v1_0_83_olap_evicted_status.sql
grutt: So, I'm thinking is there any downside to running a global analyze after all migrations complete in the goose fixture since this is a fairly common pattern and we dont want to run multiple analyzes within the migration itself.

I suppose this will add some time for migrations to complete on larger DBs

grutt: Curious @abelanger5 's thoughts here as well.

sdks/python/hatchet_sdk/worker/durable_eviction/cache.py
grutt: oooph, maybe we should add a linter rule for this?

cmd/hatchet-migrate/migrate/run.go
```diff
@@ -263,6 +263,11 @@ func RunMigrations(ctx context.Context, opts ...RunMigrationsOpt) {
 			log.Fatalf("goose: failed to apply migrations: %v", err)
 		}
 	}
+
+	_, err = db.ExecContext(ctx, "ANALYZE")
```
grutt: @mrkaye97 @abelanger5 this relates to the other comment regarding analyze. can remove here if we think it introduces risk.

cmd/hatchet-migrate/migrate/migrations/20260226120000_v1_0_83_olap_evicted_status.sql
```diff
@@ -40,13 +40,70 @@ $$;
...
+ALTER TABLE v1_task_events_olap ADD COLUMN IF NOT EXISTS durable_invocation_count INT NOT NULL DEFAULT 0;
+
+ANALYZE v1_tasks_olap;
+ANALYZE v1_dags_olap;
+ANALYZE v1_runs_olap;
+ANALYZE v1_task_events_olap_tmp;
+ANALYZE v1_task_events_olap;
+ANALYZE v1_incoming_webhook_validation_failures_olap;
+ANALYZE v1_payloads_olap;
```
mrkaye97: I feel like we don't need to analyze a bunch of these if they don't have a `readable_status` column that was changed? maybe I'm wrong there though

cmd/hatchet-migrate/migrate/migrations/20260226120000_v1_0_83_olap_evicted_status.sql
```diff
@@ -40,13 +40,70 @@ $$;
...
-DO $$
+WITH partitions AS (
+      SELECT inhrelid::regclass::text AS partition_name
+      FROM pg_inherits
+      WHERE inhparent = parent_table::regclass
+)
+
+SELECT create_v1_partition_with_status(partition_name, 'EVICTED')
+FROM partitions
+-- +goose StatementEnd
```
mrkaye97: I don't think we need multiple statements here anymore if there's no `DO` block, and I think we can also run this in a transaction now

sdks/python/hatchet_sdk/worker/durable_eviction/cache.py
mrkaye97: yeah maybe easier said than done, not too sure - if we run this code path on main in the tests, they'll fail since our version matrix includes 3.10. but on branches we're only running 3.14 so it gets missed

cmd/hatchet-migrate/migrate/migrations/20260226120000_v1_0_83_olap_evicted_status.sql
grutt: i guess only v1_tasks_olap, v1_dags_olap, v1_runs_olap?

cmd/hatchet-migrate/migrate/migrations/20260226120000_v1_0_83_olap_evicted_status.sql
grutt: still need it for the create fn afict

mrkaye97: don't think so!

grutt: can run in a txn, just need one statement wrapping the create fn

cmd/hatchet-migrate/migrate/migrations/20260227174818_v1_0_83.sql
```diff
@@ -66,13 +66,16 @@ CREATE TABLE v1_durable_event_log_file (
...
-    'MEMO'
-);
+DO $$ BEGIN
+    CREATE TYPE v1_durable_event_log_kind AS ENUM (
+        'RUN',
+        'WAIT_FOR',
+        'MEMO'
+    );
+EXCEPTION WHEN duplicate_object THEN NULL;
+END $$;
```
mrkaye97: feels like we shouldn't need this? how could this type already exist?

cmd/hatchet-migrate/migrate/migrations/20260227174818_v1_0_83.sql
```diff
@@ -66,13 +66,16 @@ CREATE TABLE v1_durable_event_log_file (
...
 
-CREATE TABLE v1_durable_event_log_entry (
+CREATE TABLE IF NOT EXISTS v1_durable_event_log_entry (
```
mrkaye97: just as a general principle on things like this, I much prefer not having `IF NOT EXISTS` in migrations, because having it masks an underlying problem. if it's the case that we're trying to run this migration but this table already exists, we should fail, since that state should not be possible if the migrations have been applied correctly. and if they haven't, we need to fix the migration state

mrkaye97: I suspect this is also the problem in the comment above here btw - I think it'd be much better to revert these changes and then if the migration is failing on someone's local, help them fix that failure instead of having workarounds like this

grutt: i go back and forth on if migrations should be idempotent but can see the benefit on create tables since obviously the cols might not be correct...

cmd/hatchet-migrate/migrate/migrations/20260227174818_v1_0_83.sql
grutt: i suspect i missed a down migration at some point in life

cmd/hatchet-migrate/migrate/migrations/20260227174818_v1_0_83.sql
mrkaye97: yeah, in general, I think it's almost always better for them to be idempotent if they can be, especially when the hiccup is a local dev issue (imo this sort of thing should be sorted out locally and not make it to main)

cmd/hatchet-migrate/migrate/migrations/20260226120001_v1_0_83_b_olap_evicted_status.sql
```diff
@@ -0,0 +1,92 @@
...
+ALTER TABLE v1_task_events_olap ADD COLUMN IF NOT EXISTS durable_invocation_count INT NOT NULL DEFAULT 0;
+
+ANALYZE v1_tasks_olap;
+ANALYZE v1_dags_olap;
+ANALYZE v1_runs_olap;
+-- +goose StatementEnd
+
+-- +goose Down
+
+ALTER TABLE v1_task_events_olap DROP COLUMN IF EXISTS durable_invocation_count;
```
mrkaye97, line 58: doesn't this need to be inside of the `StatementBegin`?

sdks/python/hatchet_sdk/worker/durable_eviction/cache.py
mrkaye97: unresolving this b/c I think we need to remove it 😅

grutt: im importing from typing_extensions which is a dep of pydanic (per your link)

mrkaye97: ah nice! okay yeah I think this should work, thanks!

mrkaye97 (approved):

promptless-for-oss:

---
_Tip: Create additional [Docs Collections](https://app.gopromptless.ai/doc-collections) to keep multiple repositories in sync 📚_

PR #3144, opened 2026-03-03
tests: compat testing

grutt:
# Description

Adds compatibility testing to the growing number of test cases

## Type of change


sdks/python/hatchet_sdk/context/context.py
```diff
@@ -513,7 +530,6 @@ def _increment_wait_index(self) -> int:
 
         return index
 
-    # todo: instrumentor for this
```
mrkaye97, line 516: let's not remove this, part 10 :P

why does cursor want to remove this so badly

sdks/python/hatchet_sdk/context/context.py
```diff
@@ -531,6 +547,11 @@ async def aio_wait_for(
         if self.durable_event_listener is None:
             raise ValueError("Durable task client is not available")
 
+        if not self._supports_durable_eviction:
+            return await aio_wait_for_pre_eviction(self, signal_key, *conditions)
+
+        assert isinstance(self.durable_event_listener, DurableEventListener)
```
mrkaye97: we shouldn't need an assert here, we should use a typeguard instead

sdks/python/hatchet_sdk/context/context.py
```diff
@@ -574,15 +595,13 @@ async def aio_sleep_for(self, duration: Duration) -> dict[str, Any]:
             SleepCondition(duration=duration),
         )
 
-    # todo: instrumentor for this
```
mrkaye97, line 577: anotha one 

Image

sdks/python/hatchet_sdk/context/context.py
```diff
@@ -574,15 +595,13 @@ async def aio_sleep_for(self, duration: Duration) -> dict[str, Any]:
...
     async def _spawn_child(
         self,
         workflow: BaseWorkflow[TWorkflowInput],
         input: TWorkflowInput = cast(Any, EmptyModel()),
         options: TriggerWorkflowOptions | None = None,
     ) -> dict[str, Any]:
-        if self.durable_event_listener is None:
-            raise ValueError("Durable task client is not available")
```
mrkaye97, line 585: I think we need this check?

sdks/python/hatchet_sdk/context/context.py
```diff
@@ -574,15 +595,13 @@ async def aio_sleep_for(self, duration: Duration) -> dict[str, Any]:
...
+        assert isinstance(self.durable_event_listener, DurableEventListener)
```
mrkaye97: typeguard

sdks/python/hatchet_sdk/context/context.py
```diff
@@ -640,11 +659,17 @@ async def aio_memo(
...
+                "Engine does not support memoization (requires >= %s). "
+                "aio_memo will execute the function but results will not be "
+                "persisted across replays. Upgrade your engine to enable durable memoization.",
+                MIN_DURABLE_EVICTION_VERSION,
+            )
+            return await fn(*args, **kwargs)
 
-        if self.durable_event_listener is None:
-            raise ValueError("Durable event listener is not available")
+        assert isinstance(self.durable_event_listener, DurableEventListener)
```
mrkaye97: typeguard

sdks/python/hatchet_sdk/context/pre_eviction.py
```diff
@@ -0,0 +1,47 @@
...
+if TYPE_CHECKING:
+    from hatchet_sdk.context.context import DurableContext
+
+
+async def aio_wait_for_pre_eviction(
+    ctx: DurableContext,
+    signal_key: str,
+    *conditions: SleepCondition | UserEventCondition | OrGroup,
+) -> dict[str, Any]:
+    assert isinstance(ctx.durable_event_listener, PreEvictionDurableEventListener)
```
mrkaye97: type guard I think? idk what should happen here though tbh, maybe an error

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -127,11 +133,25 @@ def __init__(
...
+            if self._supports_durable_eviction:
+                self.durable_event_listener = DurableEventListener(
+                    self.config, admin_client=self.admin_client
+                )
+            else:
+                self.durable_event_listener = PreEvictionDurableEventListener(
+                    self.config
+                )
+        else:
+            self.durable_event_listener = None
```
mrkaye97: nit but a preference of mine - can we get rid of the nested ifs instead of if / else if / else?

sdks/python/hatchet_sdk/worker/worker.py
```diff
@@ -227,6 +230,39 @@ def _emit_legacy_deprecation(self) -> None:
...
+            task
+            for task in self.action_registry.values()
+            if task.durable_eviction is not None
+        ]
+        if not tasks_with_eviction:
+            return
+
+        from datetime import datetime, timezone
+
+        from hatchet_sdk.deprecated.deprecation import emit_deprecation_notice
```
mrkaye97: shouldn't have imports inside here if we can avoid it

sdks/python/hatchet_sdk/worker/worker.py
```diff
@@ -227,6 +230,39 @@ def _emit_legacy_deprecation(self) -> None:
...
+                f"(requires >= {MIN_DURABLE_EVICTION_VERSION}). "
+                f"Eviction policies will be ignored for tasks: {names}. "
+                "Please upgrade your Hatchet engine."
+            ),
+            start=datetime(2026, 3, 3, tzinfo=timezone.utc),
+            error_days=180,
+        )
+
+        for task in tasks_with_eviction:
+            task.durable_eviction = None
```
mrkaye97: so this might be tricky, but I would strongly prefer to not mutate the task. in generally, I'm strongly in favor of treating everything we can as immutable (even though python doesn't enforce this) just because this sort of mutation makes it so much more complicated to debug when things are going wrong

sdks/python/hatchet_sdk/exceptions.py
```diff
@@ -184,3 +184,19 @@ class IllegalTaskOutputError(Exception):
 
 class LifespanSetupError(Exception):
     pass
+
+
+MIN_SLOT_CONFIG_VERSION = "v0.78.23"
+MIN_DURABLE_EVICTION_VERSION = "v0.80.0"
+
```
mrkaye97: let's pull these into their own file, there will probably be more of them. might also be good to make this an enum like `MinimumEngineVersionForFeature(str, Enum):` that maps from the feature to the version

mrkaye97: (this enum could also have a helper method that does the `semver_less_than` thing if we want)

sdks/python/hatchet_sdk/hatchet.py
```diff
@@ -189,6 +189,16 @@ def namespace(self) -> str:
...
 
+    async def aio_engine_version(self) -> str | None:
+        """Fetch the engine version via the dispatcher's GetVersion RPC.
+
+        :return: The engine version string, or None if the engine is too old to support GetVersion.
+        """
+        try:
+            return await self._client.dispatcher.get_version()
+        except Exception:
+            return None
```
mrkaye97: We might want a timeout here (either via grpc or via asyncio), I'm not sure how long this could hang for, but it'd be odd to the user if it took a very long time.

but also, there's some risk here if this fails flakily, right? I think we'll need to retry this too (failure case here would be we get a 500, swallow it, and then the SDK thinks that it can't use any of the new features and everything starts erroring)

mrkaye97: might want to actually kill the worker here if we get 5xx back

.github/workflows/sdk-python.yml
```diff
@@ -160,6 +160,280 @@ jobs:
...
+        run: |
+          docker run --rm --network host \
+            -e DATABASE_URL="postgresql://hatchet:hatchet@127.0.0.1:5431/hatchet?sslmode=disable" \
+            ghcr.io/hatchet-dev/hatchet/hatchet-migrate:${{ env.LATEST_TAG }}
+
+      - name: Setup config and seed database
+        working-directory: .
+        run: |
+          mkdir -p generated
+          docker run --rm --network host \
```
mrkaye97: we sure we want to be using `--network host` for all of these? seems like we shouldn't need that if we're using compose (which it appears we're not)? maybe we should use compose instead of manually orchestrating the containers like this?

.github/workflows/sdk-python.yml
```diff
@@ -160,6 +160,280 @@ jobs:
...
+      - name: Set HATCHET_CLIENT_NAMESPACE
+        run: |
+          SHORT_SHA=$(git rev-parse --short HEAD)
+          echo "HATCHET_CLIENT_NAMESPACE=old-engine-new-sdk-${SHORT_SHA}" >> $GITHUB_ENV
+
+      - name: Run pytest
+        env:
+          HATCHET_CLIENT_TLS_STRATEGY: none
+          HATCHET_CLIENT_HOST_PORT: localhost:7077
+          HATCHET_CLIENT_SERVER_URL: http://localhost:8080
```
mrkaye97: shouldn't need to set these

.github/workflows/sdk-python.yml
```diff
@@ -160,6 +160,280 @@ jobs:
...
+        run: poetry install --no-interaction --all-extras
+
+      - name: Set HATCHET_CLIENT_NAMESPACE
+        run: |
+          SHORT_SHA=$(git rev-parse --short HEAD)
+          echo "HATCHET_CLIENT_NAMESPACE=old-engine-new-sdk-${SHORT_SHA}" >> $GITHUB_ENV
+
+      - name: Run pytest
+        env:
+          HATCHET_CLIENT_TLS_STRATEGY: none
```
mrkaye97, line 242: may or may not need to set this with compose, I don't remember

.github/workflows/sdk-ruby.yml
```diff
@@ -224,6 +224,350 @@ jobs:
...
+        run: |
+          docker run --rm --network host \
+            -e DATABASE_URL="postgresql://hatchet:hatchet@127.0.0.1:5431/hatchet?sslmode=disable" \
+            ghcr.io/hatchet-dev/hatchet/hatchet-migrate:${{ env.LATEST_TAG }}
+
+      - name: Setup config and seed database
+        working-directory: .
+        run: |
+          mkdir -p generated
+          docker run --rm --network host \
```
mrkaye97: same here, probably a lot simpler with compose

sdks/python/hatchet_sdk/context/context.py
```diff
@@ -502,6 +513,26 @@ def __init__(
...
+    def _get_durable_listener(self) -> DurableEventListener:
+        if self.durable_event_listener is None:
+            raise ValueError("Durable task client is not available")
+
+        if not isinstance(self.durable_event_listener, DurableEventListener):
+            raise TypeError(
+                "Expected DurableEventListener, got "
+                f"{type(self.durable_event_listener).__name__}"
+            )
+        return self.durable_event_listener
```
mrkaye97: this can probably be a property instead of a method

sdks/python/hatchet_sdk/context/pre_eviction.py
mrkaye97: bumping this, same comment here

sdks/python/hatchet_sdk/context/pre_eviction.py
```diff
@@ -0,0 +1,47 @@
...
+        conditions=flatten_conditions(list(conditions)),
+        config=ctx.runs_client.client_config,
+    )
+
+    ctx.durable_event_listener.register_durable_event(request)
+
+    result: dict[str, Any] = await ctx.durable_event_listener.result(
+        task_id, signal_key
+    )
+    return result
```
mrkaye97: I don't think we should need this type hint? 
```suggestion
    return await ctx.durable_event_listener.result(
        task_id, signal_key
    )
```

sdks/python/hatchet_sdk/hatchet.py
```diff
@@ -189,6 +189,14 @@ def namespace(self) -> str:
         """
         return self._client.config.namespace
 
+    async def aio_engine_version(self) -> str | None:
```
mrkaye97: naming nit:
```suggestion
    async def aio_get_engine_version(self) -> str | None:
```

.github/workflows/sdk-python.yml
grutt: we do need this

mrkaye97 (approved):
😍

PR #3147, opened 2026-03-03
Feat: Event getter on core db when enabled

mrkaye97:
# Description

Updates the event getter to use the core db when event writes on the core db are enabled for improved consistency over the API

## Type of change


mrkaye97:
going to close this since we don't need it (yet)

PR #3148, opened 2026-03-03
Fix: Add parent node + branch FKs

mrkaye97:
# Description

Adds a self-referencing FK for parent node and parent branch id

## Type of change


mrkaye97:
turns out we maybe don't need these cols at all!

PR #3136, opened 2026-03-02
Add seconds granularity to cron jobs

juliusgeo:
# Description

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation change (pure documentation change)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking changes to code which doesn't change any behaviour)
- [ ] CI (any automation pipeline changes)
- [ ] Chore (changes which are not directly related to any business logic)
- [ ] Test changes (add, refactor, improve or change a test)
- [ ] This change requires a documentation update

## What's Changed

- [ ] Add a list of tasks or features here...

PR #3149, opened 2026-03-03
Add queue to update scheduled cron triggers on-demand

juliusgeo:
# Description

A previous [PR](https://github.com/hatchet-dev/hatchet/pull/3136/changes) added crons with seconds granularity. However, the code that updated the active crons only ran every 15 seconds. This PR adds a new queue that is pub'ed to by both the GRPC and REST API whenever a new cron is added so that it will begin triggering instantaneously, rather than after 15 seconds.

Fixes # (issue)

## Type of change

## What's Changed

### DIFF
--- a/api/v1/server/handlers/workflows/create_cron.go
+++ b/api/v1/server/handlers/workflows/create_cron.go
@@ -73,7 +73,7 @@ func (t *WorkflowService) CronWorkflowTriggerCreate(ctx echo.Context, request ge
 		return gen.CronWorkflowTriggerCreate400JSONResponse(apierrors.NewAPIErrors("error creating cron trigger")), nil
 	}
 	msg := tasktypes.NewCronUpdateMessage(request.Tenant, msgqueue.MsgIDCronCreate)
-	err = t.config.MessageQueueV1.SendMessage(ctx.Request().Context(), msgqueue.CRON_TRIGGER_UPDATE_QUEUE, msg)
+	err = t.config.MessageQueueV1.SendMessage(ctx.Request().Context(), msgqueue.TICKER_UPDATE_QUEUE, msg)
 	if err != nil {
 		t.config.Logger.Err(err).Msg("could not send cron trigger update message")
 	}
--- a/api/v1/server/handlers/workflows/create_cron.go
+++ b/api/v1/server/handlers/workflows/create_cron.go
@@ -10,6 +10,7 @@ import (
 	"github.com/hatchet-dev/hatchet/api/v1/server/oas/gen"
 	"github.com/hatchet-dev/hatchet/api/v1/server/oas/transformers"
 	"github.com/hatchet-dev/hatchet/internal/msgqueue"
+	tasktypes "github.com/hatchet-dev/hatchet/internal/services/shared/tasktypes/v1"
 	v1 "github.com/hatchet-dev/hatchet/pkg/repository"
 	"github.com/hatchet-dev/hatchet/pkg/repository/sqlcv1"
 )
@@ -71,7 +72,7 @@ func (t *WorkflowService) CronWorkflowTriggerCreate(ctx echo.Context, request ge
 
 		return gen.CronWorkflowTriggerCreate400JSONResponse(apierrors.NewAPIErrors("error creating cron trigger")), nil
 	}
-	msg := msgqueue.NewCronUpdateMessage(request.Tenant, "create-cron")
+	msg := tasktypes.NewCronUpdateMessage(request.Tenant, "create-cron")
 	err = t.config.MessageQueueV1.SendMessage(ctx.Request().Context(), msgqueue.CRON_TRIGGER_UPDATE_QUEUE, msg)
 	if err != nil {
 		t.config.Logger.Err(err).Msg("could not send cron trigger update message")
--- a/api/v1/server/handlers/workflows/create_cron.go
+++ b/api/v1/server/handlers/workflows/create_cron.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hatchet-dev/hatchet/api/v1/server/oas/apierrors"
 	"github.com/hatchet-dev/hatchet/api/v1/server/oas/gen"
 	"github.com/hatchet-dev/hatchet/api/v1/server/oas/transformers"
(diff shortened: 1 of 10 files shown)


grutt (approved):

internal/msgqueue/msgqueue.go
```diff
@@ -50,6 +50,7 @@ const (
 	TASK_PROCESSING_QUEUE        staticQueue = "task_processing_queue_v2"
 	OLAP_QUEUE                   staticQueue = "olap_queue_v2"
 	DISPATCHER_DEAD_LETTER_QUEUE staticQueue = "dispatcher_dlq_v2"
+	CRON_TRIGGER_UPDATE_QUEUE    staticQueue = "cron_trigger_update_queue"
```
abelanger5: Does a staticQueue make sense here? The RabbitMQ characteristics of this queue type are that it's durable, non-exclusive, and non-auto-deleted. Do these characteristics match the use-case?

internal/msgqueue/msg.go
```diff
@@ -59,6 +59,19 @@ func NewTenantMessage[T any](tenantId uuid.UUID, id string, immediatelyExpire, p
 	}, nil
 }
 
+func NewCronUpdateMessage(tenantId uuid.UUID) *Message {
```
abelanger5: I believe we usually place this in the `tasktypes` file, and I think the `ID` usually has a semantic meaning, like `cron-created`

internal/services/ticker/ticker.go
```diff
@@ -228,6 +244,10 @@ func (t *TickerImpl) Start() (func() error, error) {
 
 		cancel()
 
+		if queueCleanupFunc() != nil {
+			t.l.Log().Err(err).Msg("Could not cleanup cron trigger update queue")
```
abelanger5: This doesn't quite look consistent with how we typically call the logger, but could be wrong about that

abelanger5 (changes requested):
nice! I have some questions here

internal/msgqueue/msgqueue.go
juliusgeo: I am not sure about this. I think it makes sense for the queue to stick around considering its only function is to send notifications to refresh the crons, but it could also be set up as a tenant queue of some kind. If I'm understanding the queue setup correctly, then on the staticQueue it would be shared between tenants and would continue to exist across queue restarts? That does seem to fit this use case, especially with immediatelyExpire set to true, but I'm not sure.

internal/msgqueue/msg.go
juliusgeo: Fixed so each instance where we send this uses a different id to keep track of what kind of events are triggering the cron refresh. Also moved to tasktypes.

internal/services/ticker/ticker.go
juliusgeo: Fixed this, I believe. It matches the log statement line 288.

internal/msgqueue/msgqueue.go
abelanger5: nice, yep! I think I agree. not a gotcha question, just making sure we think about this. the queue continues to exist across restarts, which makes sense. but the "durable" boolean in RabbitMQ also refers to whether the messages in the queue are persisted to disk or not. Admittedly the `immediatelyExpire` makes this more confusing. 

From the RabbitMQ guide:

> In environments and use cases where durability is important, applications must use durable queues and make sure that [publishers](https://www.rabbitmq.com/docs/publishers) mark published messages as persisted.

> Transient queues will be deleted on node boot. They therefore will not survive a node restart, by design. Messages in transient queues will also be discarded.

> Durable queues will be recovered on node boot, including messages in them published as persistent. Messages published as transient will be discarded during recovery, even if they were stored in durable queues.

In this case, I'd argue we don't really care about the messages being persisted to disk, but since we have `immediatelyExpire` I think we're safe.

juliusgeo: Noice, yeah I definitely wanted immediatelyExpire to avoid causing a ton of refreshes right after boot up of the engine. Makes sense with the persistence to disk, shouldn't happen if the messages are being dropped immediately when non-consumable.

api/v1/server/handlers/workflows/create_cron.go
```diff
@@ -71,7 +72,7 @@ func (t *WorkflowService) CronWorkflowTriggerCreate(ctx echo.Context, request ge
 
 		return gen.CronWorkflowTriggerCreate400JSONResponse(apierrors.NewAPIErrors("error creating cron trigger")), nil
 	}
-	msg := msgqueue.NewCronUpdateMessage(request.Tenant)
+	msg := tasktypes.NewCronUpdateMessage(request.Tenant, "create-cron")
```
abelanger5: There's a file called `message_ids.go` that we can store these in

api/v1/server/handlers/workflows/delete_cron.go
```diff
@@ -26,7 +27,7 @@ func (t *WorkflowService) WorkflowCronDelete(ctx echo.Context, request gen.Workf
 		return nil, err
 	}
 
-	msg := msgqueue.NewCronUpdateMessage(request.Tenant)
+	msg := tasktypes.NewCronUpdateMessage(request.Tenant, "cron-delete")
```
abelanger5: nit: should try to keep the naming consistent, so these message ids should probably be `cron-create`, `cron-delete` and `cron-update`

internal/services/ticker/ticker.go
abelanger5, line 146: Let's say that 100s of cron schedules get created per second. Have we confirmed that:
1. `refreshCronSchedules` is concurrency safe? 
2. If it is concurrency safe, that we have some kind of backpressure mechanism so we don't spawn 100 of the cron polling queries at the same time (which can get expensive)?

abelanger5 (changes requested):

internal/services/ticker/ticker.go
juliusgeo: Added another lock that prevents updates from happening simultaneously. Tested it locally and it seems to work.

abelanger5 (approved):

PR #3150, opened 2026-03-03
Feat: Branching off branches

mrkaye97:
# Description

Implements branching off branches, which actually wasn't too bad! I added a `v1_durable_event_log_branch_point` table which lets us reconstruct arbitrary branch given a start branch id and following it down through each branch decision, which lets us "construct the tree" without needing any recursion, while only storing O(N) rows where N scales with the number of branches instead of the number of nodes, which I think is a win

We also don't need to store `parent_node_id` or `parent_branch_id` on the event log entry anymore since that other table handles that.

## Type of change


promptless-for-oss:

---

grutt (approved):

PR #3158, opened 2026-03-04
feat: backwards compatibility of max runs

grutt:
# Description

Multislot changed an api signature for workers, this reverts that change but marks these fields as deprecated

## Type of change


api/v1/server/oas/transformers/v1/worker.go
```diff
@@ -79,31 +79,52 @@ func ToWorkerSqlc(worker *sqlcv1.Worker, slotConfig map[string]gen.WorkerSlotCon
...
+					availableRuns += *v.Available
+				} else {
+					allNonDurableAvailableKnown = false
+				}
+			}
 		}
 		slotConfigInt = &tmp
+		if allNonDurableAvailableKnown {
+			availableRunsPtr = &availableRuns
+		}
```
mrkaye97: I'm a little confused by this `allNonDurableAvailableKnown` thing - what's this doing? I'd think we could increment `availableRunsPtr` like we do `maxRuns`

api/v1/server/oas/transformers/worker.go
```diff
@@ -75,24 +75,45 @@ func ToWorkerSqlc(worker *sqlcv1.Worker, slotConfig map[string]gen.WorkerSlotCon
 		status = gen.INACTIVE
 	}
 
+	const slotTypeDurable = "durable"
```
mrkaye97, line 78: do we ever call this transformer at this point? maybe we can remove this instead

mrkaye97 (approved):

PR #3173, opened 2026-03-05
feat: durable-bulk-spawn

grutt:
# Description

Durable bulk child spawning


promptless-for-oss:

---
_Tip: Tag @Promptless in GitHub PR comments to guide documentation changes during code review 🐙_

pkg/repository/match.go
```diff
@@ -155,6 +155,45 @@ type GroupMatchCondition struct {
 	Data []byte
 }
 
+func ChildTerminalMatchConditions(childExternalIdHints []string, readableDataKey string) []GroupMatchCondition {
```
grutt, line 158: @mrkaye97 let me know if there's a reason we cant share this that i'm missing

sdks/python/examples/concurrency_cancel_in_progress/test_concurrency_cancel_in_progress.py
```diff
@@ -14,7 +14,7 @@
 @pytest.mark.asyncio(loop_scope="session")
 async def test_run(hatchet: Hatchet) -> None:
     test_run_id = str(uuid4())
-    refs: list[WorkflowRunRef] = []
+    refs = []
```
grutt: fix types here

sdks/python/hatchet_sdk/context/context.py
```diff
@@ -612,40 +615,43 @@ async def aio_sleep_for(self, duration: Duration) -> dict[str, Any]:
         )
 
     # todo: instrumentor for this
```
grutt, line 617: look, i'm learning :)

pkg/repository/durable_events.go
```diff
@@ -626,6 +642,438 @@ func (r *durableEventsRepository) IngestDurableTaskEvent(ctx context.Context, op
...
+	h := sha256.New()
+	h.Write(dataToHash)
+	hashBytes := h.Sum(nil)
+	idempotencyKey := make([]byte, hex.EncodedLen(len(hashBytes)))
+	hex.Encode(idempotencyKey, hashBytes)
+
+	return idempotencyKey, nil
+}
+
+func (r *durableEventsRepository) IngestBulkDurableTaskRunEvents(
```
mrkaye97: so, unfortunately, this is what I was hoping to avoid 😭 

If we can avoid it, I'd strongly prefer to not have split code paths (this being separate from the regular ingest event function), and instead allow IngestEvent to handle bulk spawning. IMO, the amount of copy/paste here is problematic, especially since the logic here is almost identical to that other function

pkg/repository/durable_events.go
```diff
@@ -626,6 +642,438 @@ func (r *durableEventsRepository) IngestDurableTaskEvent(ctx context.Context, op
...
+	dataToHash = append(dataToHash, []byte(triggerOpts.WorkflowName)...)
+
+	h := sha256.New()
+	h.Write(dataToHash)
+	hashBytes := h.Sum(nil)
+	idempotencyKey := make([]byte, hex.EncodedLen(len(hashBytes)))
+	hex.Encode(idempotencyKey, hashBytes)
+
+	return idempotencyKey, nil
+}
```
mrkaye97: I think we should be able to not duplicate this, and use the existing `createIdempotencyKey` instead

pkg/repository/durable_events.go
```diff
@@ -626,6 +642,438 @@ func (r *durableEventsRepository) IngestDurableTaskEvent(ctx context.Context, op
...
+	getParams := sqlcv1.BulkGetDurableEventLogEntriesParams{
+		Durabletaskids:         make([]int64, n),
+		Durabletaskinsertedats: make([]pgtype.Timestamptz, n),
+		Branchids:              make([]int64, n),
+		Nodeids:                make([]int64, n),
+	}
+
+	for i, m := range metas {
+		getParams.Durabletaskids[i] = task.ID
+		getParams.Durabletaskinsertedats[i] = task.InsertedAt
```
mrkaye97: I think these two params can be single variables, since we'll only ever do this for a single durable task id + inserted at at a time (might make the SQL slightly simpler)

pkg/repository/sqlcv1/durable_event_log.sql
```diff
@@ -161,6 +161,93 @@ RETURNING *
...
+        UNNEST(@durableTaskInsertedAts::TIMESTAMPTZ[]) AS durable_task_inserted_at,
+        UNNEST(@branchIds::BIGINT[]) AS branch_id,
+        UNNEST(@nodeIds::BIGINT[]) AS node_id,
+        UNNEST(@invocationCounts::INTEGER[]) AS invocation_count,
+        UNNEST(@idempotencyKeys::BYTEA[]) AS idempotency_key
+)
+UPDATE v1_durable_event_log_entry
+SET
+    invocation_count = inputs.invocation_count,
+    idempotency_key = inputs.idempotency_key
```
mrkaye97, line 246: I think maybe this was a relic of some old code I had, I'm not sure if we'll ever need to actually update the idempotency key so can maybe remove this from both queries (need to confirm)

mrkaye97: edit: I think maybe I missed this in a previous PR review (I think it was introduced in the eviction change) - I'm not 100% sure that we should be updating invocation counts for previous event log entries? I haven't really thought it all the way through though, but it seems like we should be treating previous entries in the event log as immutable. I'm also not sure we should be updating idempotency keys ever - would be curious to hear a situation in which we'd need to update an idempotency key that wouldn't throw a NonDeterminismError

mrkaye97 (commented):
Left a few comments on some of the things I think are the biggest here - I just missed some important implementation details in the eviction PR that affect the broader logic and also this specific one I think 🤦

pkg/repository/durable_events.go
grutt: whoops, sorry about that

pkg/repository/sqlcv1/durable_event_log.sql
grutt: i cant remember if we talked through this, but i'd vote to add a TODO here and in the tracker and then get this one merged so we can continue grinding on sdks

pkg/repository/durable_events.go
grutt: good call out, i intended to come back and refactor this out

pkg/repository/sqlcv1/durable_event_log.sql
mrkaye97: I'm on the fence because I feel like we won't end up circling back and it's a critical implementation detail, but I'm okay with this if we make sure to not drop it 😅

grutt: We wont drop it!

internal/services/dispatcher/v1/server.go
```diff
@@ -615,6 +590,109 @@ func (d *DispatcherServiceImpl) handleDurableTaskEvent(
...
+			if sigErr := d.triggerWriter.SignalCreated(ctx, invocation.tenantId, entry.CreatedTasks, entry.CreatedDAGs); sigErr != nil {
+				d.l.Error().Err(sigErr).Msgf("failed to signal created tasks/DAGs for durable run trigger at node %d", entry.NodeId)
+			}
+		}
+
+		ackResp.RunEntries = append(ackResp.RunEntries, &contracts.DurableTaskRunAckEntry{
+			NodeId:   entry.NodeId,
+			BranchId: entry.BranchId,
+		})
+	}
```
mrkaye97: I wonder if we can batch this / do it all in one shot? since if you spawn 10k tasks, we'll send 10k messages afaict

internal/services/dispatcher/v1/server.go
```diff
@@ -615,6 +590,109 @@ func (d *DispatcherServiceImpl) handleDurableTaskEvent(
...
+			if err := d.DeliverDurableEventLogEntryCompletion(
+				taskExternalId,
+				result.InvocationCount,
+				entry.BranchId,
+				entry.NodeId,
+				entry.ResultPayload,
+			); err != nil {
+				d.l.Error().Err(err).Msgf("failed to deliver callback completion for task %s node %d", taskExternalId, entry.NodeId)
+				return status.Errorf(codes.Internal, "failed to deliver callback completion: %v", err)
+			}
```
mrkaye97, line 637: a bit nervous about this too, I wonder if we could batch, although this one might be trickier since we're waiting for the individual responses?

pkg/repository/match.go
mrkaye97: honestly I'm kinda wary to touch much stuff in `match.go` 😅

sdks/python/hatchet_sdk/context/context.py
```diff
@@ -727,3 +734,54 @@ async def aio_memo(
...
+
+    def __init__(
+        self,
+        listener: DurableEventListener,
+        step_run_id: str,
+        invocation_count: int,
+        node_id: int,
+        branch_id: int,
+        workflow_name: str,
+        action_key: str,
```
mrkaye97: I'm not sold on this class - does this help us much vs. the previous thing of doing the `aio_result` method inline? Especially since we override a couple of the methods here and raise

sdks/python/hatchet_sdk/runnables/workflow.py
```diff
@@ -686,13 +686,7 @@ def run(
...
         :returns: The result of the workflow execution as a dictionary.
         """
-
-        ref = self.client._client.admin.run_workflow(
-            workflow_name=self.config.name,
-            input=self._serialize_input(input, target="string"),
-            options=self._create_options_with_combined_additional_meta(options),
-        )
-
+        ref = self.run_no_wait(input, options)
```
mrkaye97, line 689: thanks!

sdks/python/hatchet_sdk/runnables/workflow.py
```diff
@@ -709,6 +703,15 @@ async def aio_run_no_wait(
...
         """
+        durable_ctx = ctx_durable_context.get()
+        if durable_ctx is not None and durable_ctx._supports_durable_eviction:
+            config = WorkflowRunTriggerConfig(
+                workflow_name=self.config.name,
+                input=self._serialize_input(input, target="string"),
+                options=self._create_options_with_combined_additional_meta(options),
+            )
+            refs = await durable_ctx._spawn_children_no_wait(self, [config])
+            return refs[0]
```
mrkaye97: we probably need to handle the case where `refs` is empty here

pkg/repository/durable_events.go
```diff
@@ -46,28 +37,39 @@ type IngestDurableTaskEventOpts struct {
...
 	MemoKey []byte
 }
 
+type IngestDurableTaskEventEntry struct {
+	ResultPayload  []byte
+	CreatedTasks   []*V1TaskWithPayload
+	CreatedDAGs    []*DAGWithData
+	NodeId         int64
+	BranchId       int64
+	IsSatisfied    bool
```
mrkaye97, line 58: do we need these if they also exist on the `IngestDurableTaskEventResult`?

pkg/repository/durable_events.go
```diff
@@ -46,28 +37,39 @@ type IngestDurableTaskEventOpts struct {
...
-	IsSatisfied    bool
-	ResultPayload  []byte
-	AlreadyExisted bool
-
-	// Populated for RUNTRIGGERED: the tasks/DAGs created by the child spawn.
-	CreatedTasks []*V1TaskWithPayload
-	CreatedDAGs  []*DAGWithData
+	IsSatisfied     bool
+	ResultPayload   []byte
+	AlreadyExisted  bool
```
mrkaye97: how can this exist on the `Result` and also the `Entry`?

sdks/python/hatchet_sdk/context/context.py
grutt: yeah ill remove, i have no strong opinions here

internal/services/dispatcher/v1/server.go
grutt: yeah im not sure this is worth it right now...

pkg/repository/match.go
grutt: ehh yeah but we are using this EXACT thing in two places now and i can see this being a problem if we change something

grutt: i think id like to keep this helper

internal/services/dispatcher/v1/server.go
```diff
@@ -572,19 +553,19 @@ func (d *DispatcherServiceImpl) handleDurableTaskEvent(
...
 
-	if len(ingestionResult.CreatedTasks) > 0 || len(ingestionResult.CreatedDAGs) > 0 {
-		if sigErr := d.triggerWriter.SignalCreated(ctx, invocation.tenantId, ingestionResult.CreatedTasks, ingestionResult.CreatedDAGs); sigErr != nil {
-			d.l.Error().Err(sigErr).Msg("failed to signal created tasks/DAGs for durable run trigger")
-		}
+	if len(ingestionResult.Entries) == 0 {
+		return status.Errorf(codes.Internal, "failed to ingest durable task event: no entries returned")
 	}
 
+	entry := ingestionResult.Entries[0]
```
mrkaye97: is this because we can't have an overload on the Ingest method? feels like we need to make sure this doesn't have multiple values in it

internal/services/dispatcher/v1/server.go
```diff
@@ -597,24 +578,133 @@ func (d *DispatcherServiceImpl) handleDurableTaskEvent(
...
-			d.l.Error().Err(err).Msgf("failed to deliver callback completion for task %s node %d", taskExternalId, ingestionResult.NodeId)
+			d.l.Error().Err(err).Msgf("failed to deliver callback completion for task %s node %d", taskExternalId, entry.NodeId)
 			return status.Errorf(codes.Internal, "failed to deliver callback completion: %v", err)
 		}
 	}
 
 	return nil
 }
 
+func (d *DispatcherServiceImpl) handleDurableTaskRunEvent(
```
mrkaye97: I feel like at this point we can probably revert factoring out this method? it seems safe to return a list on the entires and then always iterate over them, even in the case where we only expect one, and then we can get rid of all this duplication (generating external ids, checking nondeterminism, calling ingest, sending the ack) - wdyt?

mrkaye97 (approved):

promptless-for-oss:

New suggestion: **[Document durable-aware bulk child spawning in Python SDK](https://app.gopromptless.ai/suggestions/9f02c2eb-dae6-42b6-bd9d-be905791fabf)**

---

PR #3175, opened 2026-03-05
[Python] Fix: Lifespan not shutting down properly on worker drain

### LINKED ISSUE #3166: [BUG] Python: lifespan pools race condition during graceful shutdown
**Describe the issue**
During graceful shutdown (`SIGTERM`), `_aio_start` and `exit_gracefully` race to clean up the `lifespan`, causing connection pools to be closed while tasks are still running. This results in connection pool errors as existing, running tasks try to acquire the connection.

I believe the race was introduced in PR #2908 which moved `_cleanup_lifespan` into `_aio_start` after the health check completes. The problem is that `_check_listener_health` exits its loop as soon as `self.killing` is set to `True`, which happens at the start of `exit_gracefully` — before `wait_for_tasks` has finished.

**Environment**
- SDK: 1.27.0
- Engine: Cloud

**Expected behavior**
Connection pools should remain open until all in-flight tasks have completed. `_cleanup_lifespan` should only run after `wait_for_tasks` finishes.

Any worker using a lifespan with connection pools will likely hit this on `SIGTERM` if tasks are in-flight, but I made a script that samples this. You need to BYOKey:

```
import asyncio
import os
from collections.abc import AsyncGenerator
from datetime import timedelta

hatchet = Hatchet()


class FakePool:
    """Simulates an async resource (e.g. a DB connection pool) that can be closed."""

    def __init__(self):
        self._closed = False

    def acquire(self):
        if self._closed:
            raise RuntimeError("the pool is already closed")
        print(f"  [pool] connection acquired")

    def close(self):
        self._closed = True
        print(f"  [pool] closed")

    @property
    def closed(self):
        return self._closed


async def lifespan() -> AsyncGenerator[FakePool, None]:
    pool = FakePool()
    print(f"[lifespan] Pool opened (pid={os.getpid()})")
    try:
        yield pool
    finally:
        pool.close()
        print(f"[lifespan] Pool torn down (pid={os.getpid()})")


@hatchet.task(
    name="repro.slow_task",
    execution_timeout=timedelta(minutes=5),
)
async def slow_task(_: EmptyModel, ctx: Context) -> dict:
    pool: FakePool = ctx.lifespan
    for i in range(30):
        print(f"[task] iteration {i}, acquiring connection...")
        pool.acquire()  # Will raise "pool is already closed" if lifespan cleaned up early
        print(f"[task] iteration {i} done, sleeping 1s...")
        await asyncio.sleep(1)
    return {"done": True}


if __name__ == "__main__":
    worker = hatchet.worker(
        "repro-shutdown-race",
        workflows=[slow_task],
        lifespan=lifespan,
        slots=5,
    )
    print(f"Worker PID: {os.getpid()}")
    worker.start()
```

Once starting this worker you then:
1. Create a `repro.slow_task` task
2. After a moment to let the iterations begin, you can send a `kill -SIGTERM <pid>`

Eventually it yields something like this depending on when the `SIGTERM` actually happens

```
...
[task] iteration 3 done, sleeping 1s...
[INFO]  🪓 -- 2026-03-04 19:10:05,847 - received signal SIGTERM...
[INFO]  🪓 -- 2026-03-04 19:10:06,360 - waiting for 1 tasks to finish...
[task] iteration 4, acquiring connection...
  [pool] connection acquired
[task] iteration 4 done, sleeping 1s...
  [pool] closed
[lifespan] Pool torn down (pid=84055)
[INFO]  🪓 -- 2026-03-04 19:10:07,361 - waiting for 1 tasks to finish...
[task] iteration 5, acquiring connection...
[ERROR] 🪓 -- 2026-03-04 19:10:07,364 - failed step run: repro.slow_task:repro.slow_task/ec9f9b85-6d04-4d05-ab3c-60a1d295a611
RuntimeError: the pool is already closed
...
```

mrkaye97:
# Description

Fixes #3166 

### Changed

- Fixes a bug where lifespans are shut down eagerly before the worker is drained, causing unexpected behavior in still-running tasks.

## Type of change

### DIFF
--- a/sdks/python/examples/lifespans/test_lifespans.py
+++ b/sdks/python/examples/lifespans/test_lifespans.py
@@ -1,6 +1,15 @@
+import asyncio
+import signal
+from subprocess import Popen
+from typing import Any
+
+import psutil
 import pytest
 
+from examples.lifespans.drain import lifespan_drain_task, DrainInput
 from examples.lifespans.simple import Lifespan, lifespan_task
+from hatchet_sdk import Hatchet
+from hatchet_sdk import RunStatus
 
 
 @pytest.mark.asyncio(loop_scope="session")
@@ -10,3 +19,47 @@ async def test_lifespans() -> None:
     assert isinstance(result, Lifespan)
     assert result.pi == 3.14
     assert result.foo == "bar"
+
+
+@pytest.mark.parametrize(
+    "on_demand_worker",
+    [
+        (
+            ["poetry", "run", "python", "examples/lifespans/drain.py"],
+            8009,
+        )
+    ],
+    indirect=True,
+)
+@pytest.mark.asyncio(loop_scope="session")
+async def test_lifespan_drain_on_sigterm(
+    hatchet: Hatchet,
+    on_demand_worker: Popen[Any],
+) -> None:
+    n = 6
+    ref = await lifespan_drain_task.aio_run_no_wait(DrainInput(n=n))
+
+    for _ in range(30):
+        run = await hatchet.runs.aio_get_details(ref.workflow_run_id)
+        if run.status == RunStatus.RUNNING:
+            break
+
+        await asyncio.sleep(1)
+    else:
+        pytest.fail("Task never entered RUNNING state")
+
+    await asyncio.sleep(1)
+
+    parent = psutil.Process(on_demand_worker.pid)
+    children = parent.children(recursive=True)
+
+    for proc in [parent] + children:
+        try:
+            proc.send_signal(signal.SIGTERM)
+        except psutil.NoSuchProcess:
+            pass
+
+    result = await ref.aio_result()
+
+    assert result.status == "ok"
+    assert result.iterations == n
--- a/sdks/python/CHANGELOG.md
+++ b/sdks/python/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to Hatchet's Python SDK will be documented in this changelog
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.28.1] - 2026-03-05
+
+### Changed
+
+- Fixes a bug where lifespans are shut down eagerly before the worker is drained, causing unexpected behavior in still-running tasks.
+
 ## [1.28.0] - 2026-03-02
 
 ### Added
--- a/sdks/python/examples/lifespans/drain.py
+++ b/sdks/python/examples/lifespans/drain.py
@@ -1,0 +1,68 @@
+import asyncio
+from collections.abc import AsyncGenerator
+from dataclasses import dataclass
+
+from hatchet_sdk import Context, Hatchet
+
+hatchet = Hatchet()
+
+
+class FakePool:
+    def __init__(self) -> None:
+        self._closed = False
+
+    def acquire(self) -> None:
+        if self._closed:
+            raise RuntimeError("the pool is already closed")
+
+    def close(self) -> None:
+        self._closed = True
+
+    @property
+    def closed(self) -> bool:
+        return self._closed
+
+
+async def drain_lifespan() -> AsyncGenerator[FakePool, None]:
+    pool = FakePool()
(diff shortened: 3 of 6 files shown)


promptless-for-oss:

New suggestion: **[Clarify lifespan cleanup waits for running tasks during graceful shutdown](https://app.gopromptless.ai/suggestions/5742e391-5b7b-49cd-b56d-c7253207f94b)**

---
_Tip: Create additional [Docs Collections](https://app.gopromptless.ai/doc-collections) to keep multiple repositories in sync 📚_

sdks/python/examples/lifespans/test_lifespans.py
```diff
@@ -10,3 +19,47 @@ async def test_lifespans() -> None:
...
+    [
+        (
+            ["poetry", "run", "python", "examples/lifespans/drain.py"],
+            8009,
+        )
+    ],
+    indirect=True,
+)
+@pytest.mark.asyncio(loop_scope="session")
+async def test_lifespan_drain_on_sigterm(
```
gregfurman, line 35: FYI this still fails when running against legacy mode. 

```python
   File "/path/to/hatchet/sdks/python/examples/lifespans/drain.py", line 16, in acquire
     raise RuntimeError("the pool is already closed")
RuntimeError: the pool is already closed
```

sdks/python/examples/lifespans/test_lifespans.py
```diff
@@ -10,3 +19,47 @@ async def test_lifespans() -> None:
     assert isinstance(result, Lifespan)
     assert result.pi == 3.14
     assert result.foo == "bar"
+
+
+@pytest.mark.parametrize(
+    "on_demand_worker",
+    [
+        (
+            ["poetry", "run", "python", "examples/lifespans/drain.py"],
```
gregfurman, line 28: nit: since this will be running in a python subprocess, I'm pretty sure that the current `venv` will be automatically inherited from the caller process. So this could simplify to:


```suggestion
            ["python", "examples/lifespans/drain.py"],
```

sdks/python/examples/lifespans/test_lifespans.py
```diff
@@ -10,3 +19,47 @@ async def test_lifespans() -> None:
     assert isinstance(result, Lifespan)
     assert result.pi == 3.14
     assert result.foo == "bar"
+
+
+@pytest.mark.parametrize(
+    "on_demand_worker",
```
gregfurman, line 25: non-blocking: Think we should consider using a `Path` based approach for these tests instead of hardcoding the relative path. Like, currently this'll force us to run tests from the `./sdks/python` directory. 

For example, in `sdks/python/conftest.py` we could have something like:

```python
from pathlib import Path

// ...

SDK_PATH = Path(__file__).resolve().parent

@pytest.fixture(scope="session", autouse=True)
def worker() -> Generator[Popen[bytes], None, None]:
    command = ["poetry", "run", "python", SDK_PATH / "examples/worker.py"]

    with hatchet_worker(command) as proc:
        yield proc
```

Then a similar approach could be done for the definition of `on_demand_worker` to correctly resolve.

Wdyt? Happy to open an issue for a refactor as well 👍

gregfurman (approved):
LGTM. Verified locally that this fixes the linked issue. Just a question about whether this fix should be included in the `legacy` SDK or not.

sdks/python/examples/lifespans/test_lifespans.py
mrkaye97: good call, thanks! I'll add it there too

sdks/python/examples/lifespans/test_lifespans.py
mrkaye97: sounds like a good plan to me! I'm not going to do this here, but feel free to open a PR, I think it's a good idea (same with the above thing about not relying on poetry!)

promptless-for-oss:

New suggestion: **[Document lifespan graceful shutdown behavior](https://app.gopromptless.ai/suggestions/df7bb1d6-3bc2-43f5-9600-37573c57b74b)**

---
_Tip: Request one-off documentation tasks in the Dashboard under [New Task](https://app.gopromptless.ai/new-task) 🚀_

PR #3177, opened 2026-03-05
Revised mechanics around cleaning up orphaned `v1_workflow_concurrency_slot`

mnafees:
# Description

- Do not call `cleanup_workflow_concurrency_slots()` SQL function in the hot-path of creating tasks
- Instead call it right before dropping old partitions
- New cron call to query `CleanupV1WorkflowConcurrencySlot` to make sure we get rid of orphaned `v1_workflow_concurrency_slot` records

## Type of change


mnafees:
Closing in favor of https://github.com/hatchet-dev/hatchet/pull/3318

PR #3185, opened 2026-03-06
RBAC v0

juliusgeo:
# Description

Fixes # (issue)

## Type of change

## What's Changed

- [x] Changes AuthZ middleware so that access checks are based on `rbac.yaml` instead of bespoke one off checks


api/v1/server/rbac/rbac.go
```diff
@@ -0,0 +1,166 @@
...
+				Message: fmt.Sprintf("%s exists in rbac.yaml but not in specs", operationId),
+			}
+		}
+	}
+	return nil
+}
+
+func LoadYaml() (*PermissionMap, error) {
+	_, yamlPath, _, _ := runtime.Caller(0)
+	yamlFile, err := os.ReadFile(filepath.Join(filepath.Dir(yamlPath), "rbac.yaml"))
```
mnafees: We need to make sure our Docker container has access to this YAML file. Or was the idea to mount this at runtime in the k8s pod spec itself?

api/v1/server/rbac/rbac.yaml
```diff
@@ -0,0 +1,142 @@
...
+      - ApiTokenCreate
+      - ApiTokenUpdateRevoke
+      - TenantInviteList
+      - TenantInviteCreate
+      - TenantMemberList
+      - TenantInviteUpdate
+      - TenantInviteDelete
+  MEMBER:
+    permissions:
+      - TenantAlertingSettingsGet
```
mnafees, line 18: This makes me feel like an additional step to do whenever we are adding a new handler / endpoint? I guess the test should check for missing entries that are present in the OpenAPI spec but not here.

mnafees (changes requested):

api/v1/server/rbac/rbac.yaml
juliusgeo: There's the test that checks, and then also there is a runtime check in `ValidateSpec` that checks it.

api/v1/server/rbac/rbac.go
juliusgeo: So tbh I'm not sure which of those options would be better--right now it's just grabbing it from the local directory which these files are in. Not sure how best to integrate it with Docker etc

api/v1/server/rbac/rbac.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	if curRole.Inherits != nil {
+		for _, inheritedRoleName := range *curRole.Inherits {
+			if p.HasPermission(inheritedRoleName, operation) {
+				return true
+			}
+		}
+	}
+	return false
+}
```
mnafees, line 86: I'm thinking, instead of this recursive + iterative approach, won't this be significantly faster to simply create a lookup table at init time and then using it for all perm lookups?

mnafees (changes requested):

api/v1/server/rbac/rbac.go
juliusgeo: I was debating between that approach and the current one--I think lookup table would be slightly faster but would also consume more memory. Not sure if the memory usage is really significant in real terms, though. Or are you thinking something that maps permission -> role, rather than role -> permission?

promptless-for-oss:

New suggestion: **[Document team member roles and permissions](https://app.gopromptless.ai/suggestions/84a4b019-9867-4e0c-827a-5194ee8533d6)**

---

go.mod
```diff
@@ -59,6 +59,7 @@ require (
 	go.opentelemetry.io/otel/trace v1.41.0
 	go.opentelemetry.io/proto/otlp v1.9.0
 	go.uber.org/goleak v1.3.0
+	go.yaml.in/yaml/v4 v4.0.0-rc.4
```
mnafees: Is this release candidate stable? Or should we be using `v3` for now?

juliusgeo: Ah, good catch, it is not stable. Switched to v3

mnafees (approved):

api/v1/server/authz/middleware.go
```diff
@@ -3,26 +3,31 @@ package authz
...
-
-	l *zerolog.Logger
+	rbac   *rbac.Authorizer
+	l      *zerolog.Logger
 }
 
 func NewAuthZ(config *server.ServerConfig) *AuthZ {
+	rbacAuthorizer, err := rbac.NewAuthorizer()
+	if err != nil {
+		panic(err)
```
abelanger5, line 25: while in this case a panic might be ok we generally don't have any panics in api or engine-side packages, we only panic from the top level (this guards against future usage of this library/package in a different context)

PR #3186, opened 2026-03-06
rip: remove unneeded durable event log update

grutt:
# Description

Removes incorrect, unneeded update on the event log

## Type of change


pkg/repository/match.go
```diff
@@ -155,45 +155,6 @@ type GroupMatchCondition struct {
 	Data []byte
 }
 
-func ChildTerminalMatchConditions(childExternalIdHints []string, readableDataKey string) []GroupMatchCondition {
```
mrkaye97, line 158: reverting this bc it scares me, can circle back later

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -449,6 +449,10 @@ def cleanup_run_id(self, key: ActionKey) -> None:
         if self.durable_eviction_manager is not None:
             self.durable_eviction_manager.unregister_run(key)
 
+        # Durable restores replay child spawns for the same action key, so we must
+        # reset the per-run spawn counter whenever an invocation exits early.
+        workflow_spawn_indices.pop(key, None)
```
grutt: this seems to fix locally, but not ci... also maybe we need similar on ts

pkg/repository/output.go
```diff
@@ -148,18 +148,18 @@ func ExtractOutputFromMatchData(data []byte) ([]byte, error) {
 
 	for _, keyMap := range outer {
 		for _, entries := range keyMap {
-			if len(entries) == 0 {
-				continue
+			for _, entry := range entries {
```
grutt: i think this is a side effect of the root cause, not the actual fix

mrkaye97 (approved):
TS is broken, but I'd say we ship it and fix that in a separate change to unblock

grutt:
Looks good! I'll TAL at ts

PR #3188, opened 2026-03-06
fix: flaky pull on old-engine

grutt:
# Description

Fixes flaky pull

## Type of change

### DIFF
--- a/.github/workflows/sdk-python.yml
+++ b/.github/workflows/sdk-python.yml
@@ -191,7 +191,18 @@ jobs:
 
       - name: Pull release images
         working-directory: .
-        run: docker compose -f docker-compose.yml -f docker-compose.release.yml pull hatchet-migrate hatchet-admin hatchet-engine hatchet-api
+        run: |
+          for i in 1 2 3; do
+            if docker compose -f docker-compose.yml -f docker-compose.release.yml pull hatchet-migrate hatchet-admin hatchet-engine hatchet-api; then
+              exit 0
+            fi
+            if [ "$i" -lt 3 ]; then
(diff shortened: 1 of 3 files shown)


.github/workflows/sdk-python.yml
```diff
@@ -191,7 +191,18 @@ jobs:
...
+            if docker compose -f docker-compose.yml -f docker-compose.release.yml pull hatchet-migrate hatchet-admin hatchet-engine hatchet-api; then
+              exit 0
+            fi
+            if [ "$i" -lt 3 ]; then
+              echo "Pull failed (attempt $i/3), retrying in 45s..."
+              sleep 45
+            fi
+          done
+          echo "Pull failed after 3 attempts"
+          exit 1
```
mrkaye97, line 205: this seems a little silly... do we really need this? feels like this should be pretty reliable

grutt: agreed, but it hasn't been 😓 ive seen this a few times now
https://github.com/hatchet-dev/hatchet/actions/runs/22771944323/job/66055080309?pr=3186

mrkaye97: idk, I don't think we should overcomplicate our CI to handle this - I think it's probably a ghcr blip, which is a bummer but 🤷 

we can merge if you want, but I can't imagine this will happen _super_ often, although who knows, Github seems to always be down

grutt: idk ive seen it enough that i think its worth the merge... also unclear if it will fix anything
could also be user error but idk how to restart the individual failed job until all are complete.. we should just migrate tests to hatchet already

grutt:
Closing this since it seems to be related to GH reliability

PR #3200, opened 2026-03-07
feat: handle server initiated eviction on workers

grutt:
# Description

Wires up eviction when engine detects stale workers


api-contracts/v1/dispatcher.proto
```diff
@@ -72,6 +72,14 @@ message DurableTaskAwaitedCompletedEntry {
     string durable_task_external_id = 1;
     int64 branch_id = 2;
     int64 node_id = 3;
+    int32 invocation_count = 4;
+}
+
+// Sent by the server to notify a worker that its invocation is stale and should be cancelled.
+message DurableTaskServerEvictNotification {
```
grutt: I'm concerned this is confusing that we have a similar named proto here, any ideas for naming?

internal/services/dispatcher/v1/server.go
```diff
@@ -882,6 +915,54 @@ func (d *DispatcherServiceImpl) handleWorkerStatus(
...
+			idInsertedAts := make([]v1.IdInsertedAt, 0, len(tasks))
+			taskIdToExternalId := make(map[v1.IdInsertedAt]uuid.UUID, len(tasks))
+
+			for _, t := range tasks {
+				key := v1.IdInsertedAt{ID: t.ID, InsertedAt: t.InsertedAt}
+				idInsertedAts = append(idInsertedAts, key)
+				taskIdToExternalId[key] = t.ExternalID
+			}
+
+			invocationCounts, err := d.repo.DurableEvents().GetDurableTaskInvocationCounts(ctx, invocation.tenantId, idInsertedAts)
```
grutt: i dont think there's any way to avoid this lookup unless i'm missing something 😅

internal/services/dispatcher/v1/server.go
```diff
@@ -864,13 +885,25 @@ func (d *DispatcherServiceImpl) handleWorkerStatus(
...
+
 	for _, cb := range req.WaitingEntries {
 		taskExternalId, err := uuid.Parse(cb.DurableTaskExternalId)
 		if err != nil {
 			d.l.Warn().Err(err).Msgf("invalid durable_task_external_id in worker_status: %s", cb.DurableTaskExternalId)
 			continue
 		}
+
+		if cb.InvocationCount > 0 {
+			uniqueExternalIds[taskExternalId] = cb.InvocationCount
```
grutt: need to add this back in for this

promptless-for-oss:

New suggestion: **[Document server-initiated eviction for stale worker invocations](https://app.gopromptless.ai/suggestions/705f9374-5650-4d24-984c-37f2d4f8b40e)**

---

sdks/typescript/src/v1/client/worker/eviction/eviction-cache.ts
```diff
@@ -3,7 +3,7 @@
 import { durationToMs } from '../../duration';
 import { EvictionPolicy } from './eviction-policy';
 
-export type ActionKey = string;
+export type ActionKey = `${string}/${number}`;
```
grutt: this scopes to taskExternalRunId/retry count but since invocation count is guaranteed to monotonically increase, i don't think we need to be doing this here or on py

grutt: furthermore, i think this case is impossible anyway since we only increment retry on failures which would cleanup on workers... going to remove these

grutt: ok.. i think i know why we added this (not durable), cleaning this up to better match py 😅

sdks/typescript/src/clients/listeners/durable-listener/durable-listener-client.ts
```diff
@@ -419,6 +430,16 @@ export class DurableListenerClient {
         pending.resolve();
         this._pendingEvictionAcks.delete(key);
       }
+    } else if (response.serverEvict) {
+      const evict = response.serverEvict;
+      this.logger.info(
```
grutt, line 424: maybe this should be warn?

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
```diff
@@ -556,8 +563,8 @@ async def wait_for_callback(
     ) -> DurableTaskEventLogEntryResult:
         key = (durable_task_external_id, invocation_count, branch_id, node_id)
 
-        if key in self._early_completions:
-            return self._early_completions.pop(key)
+        if key in self._buffered_completions:
+            return self._buffered_completions.pop(key)
```
mrkaye97, line 570: so I was thinking about this more - I wonder if we can remove it, because the worker will poll for a new completed event when it's waiting for one and it hasn't gotten it yet

mrkaye97: this also made me realize: we need to make sure that when you post a worker status and need many bulk spawn callbacks, they all get sent instead of going one at a time (1s elapsed per)

internal/services/dispatcher/v1/server.go
```diff
@@ -864,13 +885,20 @@ func (d *DispatcherServiceImpl) handleWorkerStatus(
...
 	waiting := make([]v1.TaskExternalIdNodeIdBranchId, 0, len(req.WaitingEntries))
+
 	for _, cb := range req.WaitingEntries {
 		taskExternalId, err := uuid.Parse(cb.DurableTaskExternalId)
 		if err != nil {
 			d.l.Warn().Err(err).Msgf("invalid durable_task_external_id in worker_status: %s", cb.DurableTaskExternalId)
 			continue
 		}
+
+		if cb.InvocationCount > 0 {
```
mrkaye97: this will always be true, we can remove the if statement

internal/services/dispatcher/v1/server.go
```diff
@@ -882,6 +910,57 @@ func (d *DispatcherServiceImpl) handleWorkerStatus(
...
 
+	if len(uniqueExternalIds) > 0 {
+		externalIds := make([]uuid.UUID, 0, len(uniqueExternalIds))
+		for extId := range uniqueExternalIds {
+			externalIds = append(externalIds, extId)
+		}
+
+		tasks, err := d.repo.Tasks().FlattenExternalIds(ctx, invocation.tenantId, externalIds)
+		if err != nil {
+			d.l.Warn().Err(err).Msg("failed to look up tasks for invocation count check in worker_status")
```
mrkaye97: I think this should return an error probably

internal/services/dispatcher/v1/server.go
```diff
@@ -882,6 +910,57 @@ func (d *DispatcherServiceImpl) handleWorkerStatus(
...
+
+			for _, t := range tasks {
+				key := v1.IdInsertedAt{ID: t.ID, InsertedAt: t.InsertedAt}
+				idInsertedAts = append(idInsertedAts, key)
+				taskIdToExternalId[key] = t.ExternalID
+			}
+
+			invocationCounts, err := d.repo.DurableEvents().GetDurableTaskInvocationCounts(ctx, invocation.tenantId, idInsertedAts)
+			if err != nil {
+				d.l.Warn().Err(err).Msg("failed to get invocation counts in worker_status")
```
mrkaye97: return error here too?

internal/services/dispatcher/v1/server.go
mrkaye97: yeah I think we need to do this

internal/services/dispatcher/v1/server.go
```diff
@@ -882,6 +910,57 @@ func (d *DispatcherServiceImpl) handleWorkerStatus(
...
+			idInsertedAts := make([]v1.IdInsertedAt, 0, len(tasks))
+			taskIdToExternalId := make(map[v1.IdInsertedAt]uuid.UUID, len(tasks))
+
+			for _, t := range tasks {
+				key := v1.IdInsertedAt{ID: t.ID, InsertedAt: t.InsertedAt}
+				idInsertedAts = append(idInsertedAts, key)
+				taskIdToExternalId[key] = t.ExternalID
+			}
+
+			invocationCounts, err := d.repo.DurableEvents().GetDurableTaskInvocationCounts(ctx, invocation.tenantId, idInsertedAts)
```
mrkaye97: naming thing: I've been liking naming maps like this something like `idInsertedAtToInvocationCount` so it's very clear the types of the keys and values

internal/services/dispatcher/v1/server.go
```diff
@@ -882,6 +910,57 @@ func (d *DispatcherServiceImpl) handleWorkerStatus(
...
+			if err != nil {
+				d.l.Warn().Err(err).Msg("failed to get invocation counts in worker_status")
+			} else {
+				for key, currentCount := range invocationCounts {
+					extId, ok := taskIdToExternalId[key]
+					if !ok || currentCount == nil {
+						continue
+					}
+					workerCount, has := uniqueExternalIds[extId]
```
mrkaye97: nit: confusing naming - this is the invocation count of that specific task running on some worker?

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
```diff
@@ -169,10 +171,15 @@ def __init__(self, config: ClientConfig, admin_client: AdminClient):
         self._running = False
         self._start_lock = asyncio.Lock()
 
+        self._on_server_evict: Callable[[str, int], None] | None = None
```
mrkaye97: do we need to pass this as a variable? or can we define it as a method inside of this class instead? (same question here as on line 180)

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
```diff
@@ -169,10 +171,15 @@ def __init__(self, config: ClientConfig, admin_client: AdminClient):
...
+
     @property
     def worker_id(self) -> str | None:
         return self._worker_id
 
+    def set_on_server_evict(self, callback: Callable[[str, int], None]) -> None:
+        self._on_server_evict = callback
```
mrkaye97: this is a very go-ish way of doing this, but I suspect we don't need this setter unless I'm missing something

api-contracts/v1/dispatcher.proto
grutt: @mrkaye97

sdks/typescript/src/clients/listeners/durable-listener/durable-listener-client.ts
grutt: @mrkaye97?

internal/services/dispatcher/v1/server.go
grutt: so i guess this is a decision.. basically do we best effort stale check or gurantee we only complete for the most recent runs.

without returning an error, we optimistically continue to callback... i think we keep it as warn since we'll eventually evict (i.e. on next callback insertion)

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
grutt: the poll will catch it, but what's the harm in keeping it? i've traced that we're cleaning it up and in my testing it seems like its a fairly common case which i'd prefer to not hit a poll for 

let me double check the second concern

grutt: looks okay to me, but probably worth bumping the poll immediately in this case and on reconnect

api-contracts/v1/dispatcher.proto
mrkaye97: I think this is fine 🤷 I sort of prefer "EvictionNotice" because I think that implies that it's server -> worker

sdks/typescript/src/clients/listeners/durable-listener/durable-listener-client.ts
mrkaye97: I think info is fine

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
mrkaye97: I sort of feel the opposite 😅 

IMO, this is a pretty big memory leak risk since we're storing payloads here, and I think polling in this case is probably fine. Also, it'd simplify the code a fair amount to not have it, and I also think that keeping it is really sort of supporting a weird situation we don't want to be in as opposed to having the code follow the happy path. Anyways yeah, I'd remove it for all those reasons if it doesn't affect the functionality beyond slight polling delays in certain cases

internal/services/dispatcher/v1/server.go
mrkaye97: how could `FlattenExternalIds` ever error if it wasn't something actually problematic? This will look up the task in the db, which should always return unless something is really broken, I think?

api-contracts/v1/dispatcher.proto
grutt: ya that's better

internal/services/dispatcher/v1/server.go
grutt: that's fair... i'll concede here

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
grutt: i think for now at least i'll compromise and log a warning with a todo to readdress before the final merge. if we see this warning enough then i think this is worth doing

grutt: https://github.com/hatchet-dev/hatchet/actions/runs/22831964959/job/66221569907?pr=3200

yeah... i think we need this OR this is flagging a bigger problem

grutt: i think i understand... it looks like we are basically guaranteed to hit this case on replays/restores since we will be sending the result with the ack. since we dont know node ids before we cant register earlier.

reverting the last commit...

mrkaye97: Ahh, I see. This does feel like either a bug or poor implementation to me. I think the ack should probably be sent back before any callbacks, or maybe the data should be sent as _part of the ack_ with a `WasAlreadySatisfied = true` flag or something. I'm just a bit nervous that this is too broad of a catch-all

mrkaye97 (approved):

PR #3204, opened 2026-03-08
chore: improve sdk setup and linting

grutt:
# Description

Adds tasks to setup and lint all sdks with consistent commands, lints and typechecks files that we've missed before

## Type of change


sdks/python/lint.sh
```diff
@@ -2,14 +2,16 @@
 
 set -eo pipefail
 
-echo "\nLinting with ruff"
+unset VIRTUAL_ENV
```
mrkaye97, line 5: this feels like an odd change to me, I don't think we should need this?

sdks/guides/python/pyproject.toml
```diff
@@ -26,11 +27,25 @@ browserbase = "^1.4.0"
...
+files = ["."]
+explicit_package_bases = true
+ignore_missing_imports = true
+check_untyped_defs = true
+disallow_incomplete_defs = true
+disallow_untyped_defs = true
+warn_return_any = true
+no_implicit_optional = true
+strict_equality = true
+warn_unused_ignores = true
```
mrkaye97, line 46: can we copy the mypy config from the main `pyproject.toml` instead? I think it just uses `strict=true`

sdks/guides/python/web_scraping/worker.py
```diff
@@ -2,29 +2,37 @@
 
 from hatchet_sdk import Context, EmptyModel, Hatchet
 from hatchet_sdk.rate_limit import RateLimit, RateLimitDuration
+from pydantic import BaseModel
 
-try:
-    from .mock_scraper import mock_scrape
-except ImportError:
-    from mock_scraper import mock_scrape
+from .mock_scraper import mock_scrape
```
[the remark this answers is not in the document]
mrkaye97: we're using absolute paths everywhere else, so I think we should use them here too

sdks/guides/python/web_scraping/worker.py
```diff
@@ -2,29 +2,37 @@
...
+
+scrape_wf = hatchet.workflow(name="ScrapeUrl", input_validator=UrlInput)
+process_wf = hatchet.workflow(name="ProcessContent", input_validator=ContentInput)
 
 
 # > Step 01 Define Scrape Task
 @scrape_wf.task(execution_timeout="2m", retries=2)
-async def scrape_url(input: dict, ctx: Context) -> dict:
-    return mock_scrape(input["url"])
+async def scrape_url(input: UrlInput, ctx: Context) -> dict:
```
mrkaye97, line 27: once we enable `strict=true`, I think `dict` with no type args will fail

.cursor/skills/guide-code-quality/SKILL.md
```diff
@@ -0,0 +1,146 @@
...
+Titles become snake_case keys: `# > Step 04 Rate Limited Scrape` → `step_04_rate_limited_scrape`. **All helper functions, models, and types used by a snippet must be inside the markers** so the generator captures them.
+
+After editing any guide file, regenerate: `cd frontend/snippets && python3 generate.py`
+
+## Zero Suppressions Policy
+
+No `type: ignore`, `@ts-ignore`, `nolint`, `disable_error_code`, or equivalent. Fix the code instead.
+
+- If types mismatch at a third-party SDK boundary, convert explicitly (e.g. role-narrowing helpers, `.model_dump()`)
+- If an import pattern triggers errors, restructure the code (e.g. `__init__.py` + relative imports instead of `try/except ImportError`)
```
mrkaye97, line 28: nit: we use absolute imports in Python

.cursor/skills/guide-code-quality/SKILL.md
```diff
@@ -0,0 +1,146 @@
...
+    done: bool
+
+class LLMService(ABC):
+    @abstractmethod
+    def complete(self, messages: list[ChatMessage]) -> CompletionResult: ...
+```
+
+### Package Structure
+
+Every guide subdirectory must have `__init__.py`. Use relative imports only — no `try/except ImportError` fallback pattern:
```
mrkaye97, line 90: shouldn't need this

examples/python/guides/streaming/client.py
```diff
@@ -5,9 +5,9 @@
 
 # > Step 03 Subscribe Client
 # Client triggers the task and subscribes to the stream.
-async def run_and_subscribe():
-    run = await hatchet.runs.create(workflow_name="stream_task", input={})
-    async for chunk in hatchet.runs.subscribe_to_stream(run.run_id):
+async def run_and_subscribe() -> None:
+    run = await hatchet.runs.aio_create(workflow_name="stream_task", input={})
```
mrkaye97, line 9: we shouldn't be using `hatchet.runs.create` - should call one of the `run` methods on the workflow instead

examples/python/guides/web_scraping/worker.py
```diff
@@ -44,25 +52,27 @@ async def scheduled_scrape(input: EmptyModel, ctx: Context) -> dict:
 
     results = []
     for url in urls:
-        scraped = await scrape_wf.aio_run(input={"url": url})
-        processed = await process_wf.aio_run(input={"url": url, "content": scraped["content"]})
+        scraped = await scrape_wf.aio_run(input=UrlInput(url=url))
+        processed = await process_wf.aio_run(
+            input=ContentInput(url=url, content=scraped["content"])
+        )
         results.append({"url": url, **processed})
```
mrkaye97, line 59: bulk spawn here instead of running these in a loop?

PR #3213, opened 2026-03-09
Observability overhaul + traces support

mnafees:
# Description

Introduces Hatchet o11y and changes for the same to Python, TypeScript, and Go SDKs.

## Type of change

---

> [!NOTE]
> **Medium Risk**
> Adds new persisted trace data (new DB table/partitioning) and new API endpoints surfaced in the UI; correctness and performance depend on batching/truncation and pagination behavior. Risk is mitigated by gating collection behind `HatchetO11y.Enabled`, but changes touch engine ingest, RBAC, and frontend navigation.
> 
> **Overview**
> Adds end-to-end OpenTelemetry trace support: the engine can now ingest/store spans (new `v1_otel_trace` table + enums/partitions) via the OTLP collector with configurable max batch size and retry-count correlation, gated by `HatchetO11y.Enabled`.
> 
> Exposes new stable APIs to fetch traces for a `task` or `workflow-run` (`GET .../trace` with pagination), wires them through RBAC and OpenAPI clients/models, and updates server handlers/transformers (including a CEL enum naming fix).
> 
> Replaces the workflow/task-run *Waterfall* UI with an *Observability* tab that fetches all spans, builds a span tree, and renders an interactive timeline/tree view; adds new example apps for Go/Python/TypeScript OTel instrumentation and small CI/lint tweaks (enable o11y in python SDK workflow, ignore flaky OTel tests, expand linter excludes).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 27aefc0b3f590bb4c9e225077b569df48c3d461d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


promptless-for-oss:

Updated existing suggestion: **[Document OpenTelemetry support for Python, TypeScript, Go, and Ruby SDKs](https://app.gopromptless.ai/suggestions/79f91512-3bd0-45b0-a570-f6c7f0d8a4e9)**

---
_Tip: See how your feedback shapes Promptless in [Agent Knowledge Base](https://app.gopromptless.ai/configure/settings) 🧠_

cmd/hatchet-migrate/migrate/migrations/20260310150948_v1_0_84.sql
```diff
@@ -0,0 +1,40 @@
+-- +goose Up
+
+CREATE TYPE v1_otel_span_kind AS ENUM ('UNSPECIFIED', 'INTERNAL', 'SERVER', 'CLIENT', 'PRODUCER', 'CONSUMER');
+CREATE TYPE v1_otel_status_code AS ENUM ('UNSET', 'OK', 'ERROR');
+
+CREATE TABLE v1_otel_traces (
```
mrkaye97: naming thing: I think this should probably be singular - `v1_otel_trace`

cmd/hatchet-migrate/migrate/migrations/20260310150948_v1_0_84.sql
```diff
@@ -0,0 +1,40 @@
...
+    id              BIGINT GENERATED ALWAYS AS IDENTITY,
+    tenant_id       UUID NOT NULL,
+    trace_id        TEXT NOT NULL,
+    span_id         TEXT NOT NULL,
+    parent_span_id  TEXT NOT NULL DEFAULT '',
```
mrkaye97: these should probably be nullable in the OSS. clickhouse doesn't support nullable fields well which is why we have it this way in Cloud, but in PG, we can make parent_span_id, status_message, etc. nullable instead of having the empty string default

cmd/hatchet-migrate/migrate/migrations/20260310150948_v1_0_84.sql
```diff
@@ -0,0 +1,40 @@
...
+    scope_name      TEXT NOT NULL DEFAULT '',
+    scope_version   TEXT NOT NULL DEFAULT '',
+    task_run_external_id    UUID,
+    workflow_run_external_id UUID,
+    start_time      TIMESTAMPTZ NOT NULL,
+    PRIMARY KEY (id, start_time)
+) PARTITION BY RANGE (start_time);
+
+CREATE INDEX idx_v1_otel_traces_task_lookup
+    ON v1_otel_traces (tenant_id, task_run_external_id)
```
mrkaye97: might need an index for the workflow run id too?

cmd/hatchet-migrate/migrate/migrations/20260310150948_v1_0_84.sql
```diff
@@ -0,0 +1,40 @@
...
+    WHERE task_run_external_id IS NOT NULL;
+
+CREATE INDEX idx_v1_otel_traces_trace
+    ON v1_otel_traces (tenant_id, trace_id, start_time);
+
+SELECT create_v1_range_partition('v1_otel_traces'::text, CURRENT_DATE::date);
```
mrkaye97: nit: might want to add one for tomorrow too, we've been in the habit of doing that

cmd/hatchet-migrate/migrate/migrations/20260316185928_v1_0_90.sql
mrkaye97, line 1: just flagging in general that if we allow any sort of filtering of traces, this is going to blow up db CPU - it'll need to do a sequential scan over basically everything. probably nothing we can do about that in the OSS, but 🤷

examples/python/opentelemetry_instrumentation/hatchet/trigger.py
```diff
@@ -0,0 +1,49 @@
...
+
+# Use the same console exporter so you can see trigger-side spans too
+resource = Resource(attributes={SERVICE_NAME: "hatchet-otel-pipeline-trigger"})
+provider = TracerProvider(resource=resource)
+provider.add_span_processor(SimpleSpanProcessor(ConsoleSpanExporter()))
+
+HatchetInstrumentor(
+    tracer_provider=provider,
+    enable_hatchet_otel_collector=True,
+).instrument()
```
mrkaye97: so I think ideally, we'd make the `tracer_provider` optional in the instrumentor and allow you to use it without doing any setup. that'd simplify this example pretty significantly, since you wouldn't need to create the `resource` or `provider`, wouldn't need the otel imports, etc.

examples/python/opentelemetry_instrumentation/hatchet/worker.py
```diff
@@ -0,0 +1,144 @@
...
+_tracer: Tracer | None = None
+
+
+def _get_tracer() -> Tracer:
+    global _tracer
+    if _tracer is None:
+        from opentelemetry.trace import get_tracer
+
+        _tracer = get_tracer(__name__)
+    return _tracer
```
mrkaye97: this is pretty gnarly - I think maybe we should have a third file we import both here and in the `trigger.py` that just sets up the instrumentor, and then in both places we can import `get_tracer` and call that

examples/python/opentelemetry_instrumentation/hatchet/worker.py
```diff
@@ -0,0 +1,144 @@
...
+        span.set_attribute("validation.records_checked", 42)
+        span.set_attribute("validation.errors", 2)
+        span.set_status(StatusCode.OK, "2 records failed validation")
+
+    with tracer.start_as_current_span("data.clean") as span:
+        time.sleep(0.01)
+        span.set_attribute("clean.records_dropped", 2)
+        span.set_attribute("clean.records_remaining", 40)
+
+    return {"valid_records": "40", "dropped": "2"}
```
mrkaye97: nit: awkward return type (should these be integers? or should we use a dataclass / pydantic model?)

frontend/app/src/components/v1/agent-prism/DetailsView/DetailsView.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { SquareTerminal, Tags, ArrowRightLeft } from 'lucide-react';
+import type { ReactElement, ReactNode } from 'react';
+import { useState } from 'react';
+
+type DetailsViewTab = 'input-output' | 'attributes' | 'raw';
+
+export interface DetailsViewProps {
+  /**
+   * The span data to display in the details view
+   */
```
mrkaye97: I think we should remove these sorts of comments, I don't find them that helpful (the naming of the props should make it clear enough what they're for, and we can add a comment to clarify if not)

frontend/app/src/components/v1/agent-prism/DetailsView/DetailsView.tsx
```diff
@@ -0,0 +1,140 @@
...
+  /**
+   * Custom header actions to render
+   * Can be a ReactNode or a render function that receives the data
+   */
+  headerActions?: ReactNode | ((data: TraceSpan) => ReactNode);
+
+  /**
+   * Optional custom header component to replace the default
+   */
+  customHeader?: ReactNode | ((props: { data: TraceSpan }) => ReactNode);
```
mrkaye97: this feels sort of clunky to me - is there a reason that this can take either a node or a function?

frontend/app/src/components/v1/agent-prism/DetailsView/DetailsViewHeader.tsx
```diff
@@ -0,0 +1,97 @@
...
+        </IconButton>
+      )}
+
+      <SpanBadge category={data.type} />
+
+      {typeof data.tokensCount === 'number' && (
+        <TokensBadge tokensCount={data.tokensCount} />
+      )}
+
+      {typeof data.cost === 'number' && <PriceBadge cost={data.cost} />}
```
mrkaye97: will we ever be sending cost or token data? / where is this supposed to be coming from?

frontend/app/src/components/v1/agent-prism/SpanCard/SpanCardTimeline.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+
+const timelineBgColors: Record<TraceSpanCategory, string> = {
+  llm_call: 'bg-agentprism-timeline-llm',
+  agent_invocation: 'bg-agentprism-timeline-agent',
+  tool_execution: 'bg-agentprism-timeline-tool',
+  chain_operation: 'bg-agentprism-timeline-chain',
+  retrieval: 'bg-agentprism-timeline-retrieval',
+  embedding: 'bg-agentprism-timeline-embedding',
+  guardrail: 'bg-agentprism-timeline-guardrail',
```
mrkaye97: are we supporting these sorts of things right now? seems pretty llm-specific? maybe we can remove it, idk

frontend/app/src/components/v1/agent-prism/Badge.tsx
mrkaye97, line 1: I feel like we should almost certainly not be making all of these new components here (badge, brand logo, button, collapsible, etc. - this whole section). we need to be importing things we already have if we can. this is going to lead to a split component library, which IMO is a very bad idea

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/otel-span-adapter.ts
```diff
@@ -0,0 +1,130 @@
...
+  const result: OtelSpan[] = [taskRootSpan];
+  const queue = [taskRootSpan.span_id];
+  while (queue.length > 0) {
+    const parentId = queue.shift()!;
+    const children = childrenByParent.get(parentId) || [];
+    for (const child of children) {
+      result.push(child);
+      queue.push(child.span_id);
+    }
+  }
```
mrkaye97: are we sure we want to do this in js on the frontend vs. changing the return type of the API?

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/task-run-trace.tsx
```diff
@@ -0,0 +1,119 @@
...
+  isRunning,
+}: {
+  taskExternalId: string;
+  isRunning: boolean;
+}) {
+  const tracesQuery = useQuery({
+    queryKey: ['task:trace', taskExternalId],
+    queryFn: () => fetchAllSpans(taskExternalId),
+    refetchInterval: isRunning ? 100 : false,
+  });
```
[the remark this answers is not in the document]
TehShrike: I changed it to 1s in my branch, 5s would probably be good to

frontend/app/src/components/v1/agent-prism/DetailsView/DetailsViewHeader.tsx
TehShrike: there's a lot of stuff like this from the original agent-prism implementation that I don't think we'll need, I'll try to delete as much of it as I notice while making it look closer to what we want

TehShrike: oh, specifically, I deleted the entire DetailsView directory in my branch 😅

frontend/app/src/components/v1/agent-prism/Badge.tsx
TehShrike: I haven't deleted this yet in my branch, but I have commented out the code that used the badge, since it's not in our designs.  If nobody asks for something similar in the near future I can do more deleting

api/v1/server/oas/transformers/v1/trace.go
```diff
@@ -0,0 +1,90 @@
...
+		result[i] = gen.OtelSpan{
+			TraceId:            s.TraceID,
+			SpanId:             s.SpanID,
+			ParentSpanId:       &s.ParentSpanID,
+			SpanName:           s.SpanName,
+			SpanKind:           gen.OtelSpanKind(s.SpanKind),
+			ServiceName:        s.ServiceName,
+			StatusCode:         gen.OtelStatusCode(s.StatusCode),
+			StatusMessage:      &s.StatusMessage,
+			Duration:           s.DurationNs,
```
TehShrike: can return a field named `durationNs` instead of `duration`?  `duration` is confusing in the front-end right now.

api/v1/server/oas/transformers/v1/trace.go
```diff
@@ -0,0 +1,99 @@
...
+
+	return result
+}
+
+func pgTextPtr(t pgtype.Text) *string {
+	if !t.Valid {
+		return nil
+	}
+	return &t.String
+}
```
mrkaye97: can we put this in `sqlchelpers`?

api/v1/server/oas/transformers/v1/trace.go
```diff
@@ -0,0 +1,99 @@
...
+		default:
+			b, err := json.Marshal(val)
+			if err == nil {
+				result[k] = string(b)
+			}
+		}
+	}
+
+	return result
+}
```
mrkaye97: we also have at least one (maybe multiple) `jsonToMap` implementations - maybe would be good to consolidate these?

examples/python/opentelemetry_instrumentation/hatchet/trigger.py
```diff
@@ -0,0 +1,39 @@
...
+    poetry run python -m examples.opentelemetry_instrumentation.hatchet.trigger
+"""
+
+from opentelemetry.trace import get_tracer
+
+from examples.opentelemetry_instrumentation.hatchet.worker import otel_workflow
+from hatchet_sdk.clients.admin import TriggerWorkflowOptions
+from hatchet_sdk.opentelemetry.instrumentor import HatchetInstrumentor
+
+HatchetInstrumentor(enable_hatchet_otel_collector=True).instrument()
```
mrkaye97: 😍 really nice!

frontend/app/src/components/v1/agent-prism/Button.tsx
mrkaye97, line 1: I'd still strongly prefer not duplicating `Button`, `CollapsibleSection`, `Tabs`, etc. if we can avoid it - we should have shadcn versions of these in our component library which we should either use as-is or extend

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/observability/observability.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+    const numPages = res.data.pagination?.num_pages ?? 1;
+    const currentPage = res.data.pagination?.current_page ?? 1;
+
+    if (currentPage >= numPages || rows.length === 0) {
+      break;
+    }
+
+    offset += PAGE_SIZE;
+  }
```
mrkaye97: this feels a little bit odd to me, there must be another way to fetch all the spans for a trace that doesn't require a `while(true)` loop on the client side. maybe we should enable infinite scroll and fetch as we go? or we should just fetch them all in one shot?

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/observability/task-run-trace.tsx
```diff
@@ -0,0 +1,60 @@
+import { TreeView } from '@/components/v1/agent-prism/TreeView';
+import { convertOtelSpansToOtelSpanTree } from '@/components/v1/agent-prism/convert-otel-spans-to-agent-prism-span-tree';
+import type { OtelSpanTree } from '@/components/v1/agent-prism/span-tree-type';
+import { OtelSpan } from '@/lib/api/generated/data-contracts';
+import { useEffect, useMemo, useState } from 'react';
+
+const getSpanIdsOfAllHatchetSpans = (spanTree: OtelSpanTree): string[] => {
+  if (!spanTree.spanName.startsWith('hatchet.')) {
```
mrkaye97: this makes me nervous because it's implicitly relying on span naming for figuring things out. Maybe we should set an attribute on internal spans like `instrumentor: hatchet` or something? some instrumentors do this I think

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/observability/task-run-trace.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const traceSpanTree = useMemo(
+    () => convertOtelSpansToOtelSpanTree(spans),
+    [spans],
+  );
+
+  const [expandedSpansIds, setExpandedSpansIds] = useState<string[]>([]);
+
+  useEffect(() => {
+    setExpandedSpansIds(getSpanIdsOfAllHatchetSpans(traceSpanTree));
+  }, [traceSpanTree]);
```
mrkaye97: why do we need this useEffect? this generally makes me wary, so making sure we really need it

sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
```diff
@@ -185,26 +213,87 @@ class HatchetInstrumentor(BaseInstrumentor):  # type: ignore[misc]
...
+            enable_hatchet_otel_collector=True,  # Also sends to Hatchet
+        )
     """
 
     def __init__(
         self,
         tracer_provider: TracerProvider | None = None,
         meter_provider: MeterProvider | None = None,
         config: ClientConfig | None = None,
+        enable_hatchet_otel_collector: bool = False,
```
mrkaye97: maybe we want to default this to true?

sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
```diff
@@ -185,26 +213,87 @@ class HatchetInstrumentor(BaseInstrumentor):  # type: ignore[misc]
...
     ):
         self.config = config or ClientConfig()
 
-        self.tracer_provider = tracer_provider or get_tracer_provider()
+        if tracer_provider is not None:
+            self.tracer_provider = tracer_provider
+        elif enable_hatchet_otel_collector:
+            from opentelemetry.trace import set_tracer_provider
```
mrkaye97: we can do this import at the top level

sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
```diff
@@ -310,18 +399,24 @@ async def _wrap_handle_start_step_run(
...
+                context=traceparent,
+                kind=SpanKind.CONSUMER,
+            ) as span:
+                result = await wrapped(*args, **kwargs)
+
+                if isinstance(result, Exception):
+                    span.set_status(StatusCode.ERROR, str(result))
+
+                return result
+        finally:
```
mrkaye97, line 503: surprising that we only need to do this on one of the wrapped methods? not too sure if it makes sense elsewhere too, but thought I'd check

sdks/python/hatchet_sdk/worker/runner/runner.py
```diff
@@ -312,8 +313,17 @@ async def async_wrapped_action_func(
...
+                # correct trace_id and parent_span_id from the active
+                # hatchet task run span.
+                ctx_snapshot = contextvars.copy_context()
                 loop = asyncio.get_event_loop()
-                return await loop.run_in_executor(self.thread_pool, pfunc)
+                return await loop.run_in_executor(
+                    self.thread_pool,
+                    ctx_snapshot.run,
+                    pfunc,
+                )
```
mrkaye97: might be worth taking a look at the lines above here - we're explicitly copying the context vars we need over, maybe we can extend that existing pattern, if it does what we want?

sdks/python/tests/otel_traces/test_otel_traces.py
```diff
@@ -0,0 +1,173 @@
...
+    resp = requests.get(f"http://localhost:{SPANS_PORT}/spans", timeout=5)
+    resp.raise_for_status()
+    result: list[dict[str, Any]] = resp.json()
+    return result
+
+
+def _clear_spans() -> None:
+    """Clear captured spans on the worker."""
+    resp = requests.delete(f"http://localhost:{SPANS_PORT}/spans", timeout=5)
+    resp.raise_for_status()
```
mrkaye97: we should be able to simplify this a lot, I think - e.g. to list spans, can't we use the API client on the SDK?

sdks/python/tests/otel_traces/test_otel_traces.py
```diff
@@ -0,0 +1,173 @@
...
+    await otel_simple_task.aio_run(
+        options=TriggerWorkflowOptions(
+            additional_metadata={"test_run_id": test_run_id},
+        ),
+    )
+
+    # Give the span processor a moment to flush
+    await asyncio.sleep(1)
+
+    spans = _get_spans()
```
mrkaye97: if we could, hypothetically, filter spans by attribute names over the API, that would make these tests dramatically simpler I think (no need to clear, etc.)

sdks/python/tests/otel_traces/test_otel_traces.py
```diff
@@ -0,0 +1,173 @@
...
+from uuid import uuid4
+
+import pytest
+import requests
+
+from hatchet_sdk import Hatchet, TriggerWorkflowOptions
+from hatchet_sdk.clients.rest.models.v1_task_status import V1TaskStatus
+from tests.otel_traces.worker import otel_retry_task, otel_simple_task
+
+SPANS_PORT = 8020
```
mrkaye97: are we making requests to an API on the _worker_ for spans?

sdks/python/tests/otel_traces/worker.py
```diff
@@ -0,0 +1,138 @@
...
+            self.send_response(404)
+            self.end_headers()
+
+    def log_message(self, format: str, *args: Any) -> None:
+        pass  # suppress request logs
+
+
+def _start_span_server(port: int) -> None:
+    server = HTTPServer(("0.0.0.0", port), SpanHandler)
+    server.serve_forever()
```
mrkaye97: I'm suspicious of all of this stuff - can't we sleep for a bit and then query the REST API for the spans?

also just in general, I've been putting the E2E tests in `examples/<< feature >>/(worker|test_<< feature >>).py` and feel like we should do that here too

mrkaye97 (commented):
Looking really good! left some comments, mostly on the SDK + the FE

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/observability/observability.tsx
TehShrike: The data structure is a tree inside of the trace view, it wouldn't be impossible to load in more after the fact, but loading in more as the user scrolled might be tricky.

Do we think there will ever be so many spans that it would be undesirable to hold them in memory?  If not, I'd be in favor of the endpoint just returning everything.

frontend/app/src/components/v1/agent-prism/Button.tsx
TehShrike: Sounds good, I'll swap them out in the morning.

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/observability/task-run-trace.tsx
TehShrike: Yeah, adding a span_attribute for that would be excellent.

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/observability/task-run-trace.tsx
TehShrike: Good catch, I think that function saw a bit of churn, so maybe it was justified somehow at the beginning.  Now, that should definitely just be the initial state for the expanded spans useState.

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/observability/observability.tsx
mrkaye97: honestly it's possible there could be like 10k or more per trace, depending on if e.g. you spawn a lot of children. but even in that case, we'll still fetch them all in this loop right?

TehShrike: Correct.  And I don't necessarily consider 10k to be unreasonable.  I'd probably rather get them all in one request, and then just discard the attributes that we don't need to keep in memory.

I can make the change to not bother holding on to the attributes that we aren't using in the view.

mrkaye97: sounds good! another thing we could do is just limit to some reasonable number of spans, like 1,000 or something, and assume the user won't scroll through 1000 spans! (only really relevant if there's a risk of crashing the page from the amount of data we fetch)

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/otel-span-adapter.ts
TehShrike: I've deleted the majority of the translation code from our OpenTelemetry shape to the data shape that the UI used to use, so if we change the server to return a tree structure, it will not be difficult to update the front-end.

mrkaye97: great stuff, thanks!

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/v2components/step-run-detail/observability/observability.tsx
TehShrike: If we changed the endpoint to return a tree, it would be easier to make more specific requests involving depth like "only return spans up to depth of 4".

If we want to optimize further then we could load in more spans when they start drilling down to a particular part of the tree, rather than when they scroll.

Probably not necessary for launch (?) but an option.

mrkaye97: this sounds like a pretty reasonable option to me

api/v1/server/oas/transformers/v1/cel.go
```diff
@@ -9,9 +9,9 @@ func ToV1CELDebugResponse(success bool, output *bool, err *string) gen.V1CELDebu
 	}
 
 	if success {
-		response.Status = gen.V1CELDebugResponseStatusSUCCESS
+		response.Status = gen.SUCCESS
```
mnafees: @mrkaye97 would this break anything?

mrkaye97: I think it's probably fine

cmd/hatchet-engine/engine/run.go
```diff
@@ -418,30 +418,34 @@ func runV0Config(ctx context.Context, sc *server.ServerConfig, cleanup *cleanup.
 			return fmt.Errorf("could not create admin service (v1): %w", err)
 		}
 
-		oc, err := otelcol.NewOTelCollector(
```
mnafees, line 419: This is fine since this was previously a no-op and now will only be enabled when the env var is set as in below.

mrkaye97:
going to leave some review comments for myself then will go through and fix them

.github/workflows/sdk-python.yml
```diff
@@ -102,6 +102,7 @@ jobs:
           export SERVER_DEFAULT_ENGINE_VERSION=V1
           export SERVER_MSGQUEUE_RABBITMQ_URL="amqp://user:password@localhost:5672/"
           export SERVER_OPTIMISTIC_SCHEDULING_ENABLED=${{ matrix.optimistic-scheduling }}
+          export SERVER_HATCHET_O11Y_ENABLED=true
```
mrkaye97: todo: change this to `OBSERVABILITY` instead of `O11Y`

.github/workflows/sdk-python.yml
```diff
@@ -244,7 +245,7 @@ jobs:
         run: |
           echo "Testing current SDK against engine ${{ env.LATEST_TAG }}"
           echo "Using HATCHET_CLIENT_NAMESPACE: $HATCHET_CLIENT_NAMESPACE"
-          poetry run pytest -s -vvv --maxfail=5 --capture=no --retries 3 --retry-delay 2 -n auto
+          poetry run pytest -s -vvv --maxfail=5 --capture=no --retries 3 --retry-delay 2 -n auto --ignore=tests/otel_traces
```
mrkaye97: todo: don't ignore this test

sdks/guides/python/ai_agents/__pycache__/worker.cpython-310.pyc
mrkaye97, line 1: not sure how these got checked in in the first place, but will check to make sure there aren't more

sdks/python/examples/run_details/test_run_detail_getter.py
```diff
@@ -23,7 +23,8 @@ async def test_run(hatchet: Hatchet) -> None:
 
     assert details.status == RunStatus.RUNNING
     assert details.input == mock_input.model_dump()
-    assert details.additional_metadata == meta
+    assert details.additional_metadata is not None
+    assert meta.items() <= details.additional_metadata.items()
```
mrkaye97, line 27: todo: use set contain for this assert

sdks/python/examples/opentelemetry_instrumentation/hatchet/trigger.py
```diff
@@ -0,0 +1,39 @@
+"""
```
mrkaye97: todo: do we need these new examples, or are the existing ones fine?

sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
```diff
@@ -53,12 +63,75 @@
...
 from hatchet_sdk.runnables.action import Action
+from hatchet_sdk.runnables.contextvars import ctx_hatchet_span_attributes
 from hatchet_sdk.utils.opentelemetry import OTelAttribute
 from hatchet_sdk.worker.runner.runner import Runner
 from hatchet_sdk.workflow_run import WorkflowRunRef
 
 hatchet_sdk_version = version("hatchet-sdk")
 
+
+_RETRY_AFTER_SECONDS = 5 * 60
```
mrkaye97: todo: make this a timedelta

sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
```diff
@@ -185,27 +258,118 @@ class HatchetInstrumentor(BaseInstrumentor):  # type: ignore[misc]
...
+            if isinstance(existing, SDKTracerProvider):
+                self.tracer_provider = existing
+            elif enable_hatchet_otel_collector:
+                self.tracer_provider = SDKTracerProvider()
+                set_tracer_provider(self.tracer_provider)
+            else:
+                self.tracer_provider = existing
+
+        if enable_hatchet_otel_collector:
+            self._add_hatchet_exporter()
```
mrkaye97, line 330: todo: don't use untyped kwargs here

sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
```diff
@@ -185,27 +258,118 @@ class HatchetInstrumentor(BaseInstrumentor):  # type: ignore[misc]
...
+            return
+
+        endpoint = self.config.host_port
+        insecure = self.config.tls_config.strategy == "none"
+        headers = (("authorization", f"Bearer {self.config.token}"),)
+
+        exporter_kwargs: dict[str, Any] = {
+            "endpoint": endpoint,
+            "headers": headers,
+        }
```
mrkaye97: todo: don't use untyped kwargs here

sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
```diff
@@ -329,18 +487,25 @@ async def _wrap_handle_start_step_run(
...
-        ) as span:
-            result = await wrapped(*args, **kwargs)
-
-            if isinstance(result, Exception):
-                span.set_status(StatusCode.ERROR, str(result))
+        hatchet_attrs = action.get_otel_attributes(self.config)
+        token = ctx_hatchet_span_attributes.set(hatchet_attrs)
 
-            return result
+        try:
```
mrkaye97, line 489: todo: confirm we need this try / finally

sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
```diff
@@ -329,18 +487,25 @@ async def _wrap_handle_start_step_run(
...
-            attributes=action.get_otel_attributes(self.config),
-            context=traceparent,
-            kind=SpanKind.CONSUMER,
-        ) as span:
-            result = await wrapped(*args, **kwargs)
-
-            if isinstance(result, Exception):
-                span.set_status(StatusCode.ERROR, str(result))
+        hatchet_attrs = action.get_otel_attributes(self.config)
+        token = ctx_hatchet_span_attributes.set(hatchet_attrs)
```
mrkaye97, line 487: todo: confirm this respects excluded attrs

sdks/python/tests/otel_traces/test_otel_traces.py
mrkaye97, line 1: todo: move this to the right place

mrkaye97 (commented):
and last one: need a minor version for python

grutt (approved):

promptless-for-oss:

Updated existing suggestion: **[Document OpenTelemetry support for Python, TypeScript, Go, and Ruby SDKs](https://app.gopromptless.ai/suggestions/79f91512-3bd0-45b0-a570-f6c7f0d8a4e9)**

---

PR #3221, opened 2026-03-09
fix: durable evict at capacity

grutt:
# Description

Prop drilling wrong value

## Type of change

### DIFF
--- a/sdks/python/examples/durable_eviction/test_durable_eviction.py
+++ b/sdks/python/examples/durable_eviction/test_durable_eviction.py
@@ -11,6 +11,8 @@ from __future__ import annotations
 import asyncio
 import signal
 import time
+from subprocess import Popen
+from typing import Any
 
 import psutil
 import pytest
@@ -29,7 +31,6 @@ from hatchet_sdk import Hatchet
 from hatchet_sdk.clients.admin import WorkflowRunDetail
 from hatchet_sdk.clients.rest.api.task_api import TaskApi
 from hatchet_sdk.clients.rest.models.v1_task_status import V1TaskStatus
-from tests.worker_fixture import hatchet_worker
 
 POLL_INTERVAL = 0.2
 MAX_POLLS = 150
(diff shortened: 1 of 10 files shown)


sdks/python/examples/durable_eviction/test_durable_eviction.py
```diff
@@ -381,6 +382,59 @@ async def test_evictable_cancel_after_eviction(hatchet: Hatchet) -> None:
...
+async def test_capacity_eviction_fires(hatchet: Hatchet) -> None:
+    """A task with ttl=None but allow_capacity_eviction=True should be evicted
+    under durable-slot pressure (durable_slots=1)."""
+    command = [
+        "poetry",
+        "run",
+        "python",
+        "-m",
+        "examples.durable_eviction.capacity_worker",
+    ]
```
mrkaye97: I think we should be able to use the `on_demand_worker` fixture for this instead, we shouldn't need to use this `hatchet_worker` thing directly I don't think (seems like I missed that in an earlier PR

sdks/python/examples/durable_eviction/test_durable_eviction.py
```diff
@@ -381,6 +382,59 @@ async def test_evictable_cancel_after_eviction(hatchet: Hatchet) -> None:
...
+        await _poll_until_status(hatchet, ref.workflow_run_id, V1TaskStatus.RUNNING)
+        details = await _poll_until_status(
+            hatchet, ref.workflow_run_id, V1TaskStatus.EVICTED
+        )
+        statuses = {t.status for t in details.task_runs.values()}
+
+        assert (
+            V1TaskStatus.EVICTED in statuses
+        ), f"Expected capacity eviction (ttl=None), got: {statuses}"
+
```
mrkaye97: I'm confused by this, is this testing TTL eviction?

mrkaye97: ah are we auto-evicting when we hit the max slot count, not when we go over? that makes sense

sdks/python/examples/durable_eviction/test_durable_eviction.py
grutt: nice! i didnt know we had this

mrkaye97: yeah I use it in a few places we need isolated workers

sdks/python/examples/durable_eviction/test_durable_eviction.py
```diff
@@ -381,6 +393,79 @@ async def test_evictable_cancel_after_eviction(hatchet: Hatchet) -> None:
...
+    await _poll_until_status(hatchet, ref.workflow_run_id, V1TaskStatus.RUNNING)
+    details = await _poll_until_status(
+        hatchet, ref.workflow_run_id, V1TaskStatus.EVICTED
+    )
+    statuses = {t.status for t in details.task_runs.values()}
+
+    assert (
+        V1TaskStatus.EVICTED in statuses
+    ), f"Expected capacity eviction (ttl=None), got: {statuses}"
```
mrkaye97, line 429: this is an API thing, but I feel like `details.status` should be `EVICTED` at this point. might make sense to update the detail getter on the gRPC API to do that correctly

mrkaye97 (approved):

PR #3239, opened 2026-03-11
Feat--consistent-analytics-events

grutt:
# Description

Improves analytics eventing consistency throughout applicaiton surface.

## Type of change


pkg/analytics/analytics.go
```diff
@@ -1,25 +1,138 @@
...
+	Tenant   Resource = "tenant"
+	Token    Resource = "api-token"
+	Workflow Resource = "workflow"
+
+	Event         Resource = "event"
+	WorkflowRun   Resource = "workflow-run"
+	TaskRun       Resource = "task-run"
+	Worker        Resource = "worker"
+	RateLimit     Resource = "rate-limit"
+	WebhookWorker Resource = "webhook-worker"
```
mrkaye97: webhook worker?

pkg/analytics/analytics.go
```diff
@@ -1,25 +1,138 @@
...
+	if tenantID != nil {
+		return "$tenant_" + tenantID.String()
+	}
+	return ""
+}
+
+// Props builds a property map from variadic key-value pairs, keeping all
+// non-nil values regardless of type. Keys must be strings; non-string keys
+// are skipped. Nil values are omitted.
+func Props(kvs ...interface{}) map[string]interface{} {
```
mrkaye97: can we use a more informative type than `interface{}` here?

mrkaye97 (changes requested):
blocking because this makes me _very_ nervous, both that we'll blow up our Posthog budget, and that we'll have a bunch of performance degradation, memory leak risk, etc. on the API, and, more importantly, on the engine from this. I think we need to discuss this more, and I also still feel pretty strongly that e.g. tracking counts of things like task runs, stream events, etc. in Posthog is not the right place to be putting them, but it's fine IMO if we want to make dashboards and stuff from them 🤷

mrkaye97 (approved):
discussed sync, let's give it a shot

(still one or two comments to address but wanted to unblock)

mrkaye97 (approved):
🚀

PR #3240, opened 2026-03-11
Fix--durable-feedback

grutt:
# Description

    * TS Confusing Sleep For doesn’t take MS
    * Clearer Non-determinism Error
    * Always evict on woker gracefulterm


promptless-for-oss:

New suggestion: **[Document TypeScript Duration milliseconds support and non-determinism error format](https://app.gopromptless.ai/suggestions/5827b33a-2178-4599-a9e1-c3203cd1fb24)**

---

pkg/repository/durable_events.go
```diff
@@ -165,6 +184,95 @@ func (e *StaleInvocationError) Error() string {
 	return fmt.Sprintf("invocation count mismatch for task %s: server has %d, worker sent %d", e.TaskExternalId.String(), e.ExpectedInvocationCount, e.ActualInvocationCount)
 }
 
+func formatCall(opts IngestDurableTaskEventOpts) string {
```
mrkaye97: should we make this a method on the `IngestDurableTaskEventOpts` instead of a standalone function?

pkg/repository/durable_events.go
```diff
@@ -165,6 +184,95 @@ func (e *StaleInvocationError) Error() string {
...
+	switch opts.Kind {
+	case sqlcv1.V1DurableEventLogKindRUN:
+		if opts.TriggerRuns != nil {
+			names := make([]string, 0, len(opts.TriggerRuns.TriggerOpts))
+			for _, t := range opts.TriggerRuns.TriggerOpts {
+				names = append(names, t.WorkflowName)
+			}
+			return "run(" + strings.Join(names, ", ") + ")"
```
mrkaye97: if someone bulk spawns 10k runs, this will cause problems I think or at least be very ugly

pkg/repository/durable_events.go
```diff
@@ -165,6 +184,95 @@ func (e *StaleInvocationError) Error() string {
...
+		if opts.WaitFor != nil {
+			parts := make([]string, 0, len(opts.WaitFor.WaitForConditions))
+			for _, c := range opts.WaitFor.WaitForConditions {
+				switch c.Kind {
+				case CreateExternalSignalConditionKindSLEEP:
+					if c.SleepFor != nil {
+						parts = append(parts, "sleep("+*c.SleepFor+")")
+					} else {
+						parts = append(parts, "sleep")
+					}
```
mrkaye97: maybe I'm missing something - if it's a `SLEEP` kind, how could `c.SleepFor` be `nil`?

pkg/repository/durable_events.go
```diff
@@ -165,6 +184,95 @@ func (e *StaleInvocationError) Error() string {
...
+				case CreateExternalSignalConditionKindUSEREVENT:
+					if c.UserEventKey != nil {
+						parts = append(parts, "waitForEvent("+*c.UserEventKey+")")
+					} else {
+						parts = append(parts, "waitForEvent")
+					}
```
mrkaye97: same here - how can this be nil?

pkg/repository/durable_events.go
```diff
@@ -165,6 +184,95 @@ func (e *StaleInvocationError) Error() string {
...
+
+func formatStoredPayload(kind sqlcv1.V1DurableEventLogKind, payload []byte) string {
+	if len(payload) == 0 {
+		return string(kind)
+	}
+
+	switch kind {
+	case sqlcv1.V1DurableEventLogKindRUN:
+		var triggerOpts WorkflowNameTriggerOpts
+		if json.Unmarshal(payload, &triggerOpts) == nil && triggerOpts.WorkflowName != "" {
```
mrkaye97: error handling here?

pkg/repository/durable_events.go
```diff
@@ -165,6 +184,95 @@ func (e *StaleInvocationError) Error() string {
...
+			return "run(" + triggerOpts.WorkflowName + ")"
```
mrkaye97, line 282: how will this work with bulk spawn?

pkg/repository/durable_events.go
```diff
@@ -165,6 +184,95 @@ func (e *StaleInvocationError) Error() string {
...
+		}
+	case sqlcv1.V1DurableEventLogKindWAITFOR:
+		var conditions []CreateExternalSignalConditionOpt
+		if json.Unmarshal(payload, &conditions) == nil && len(conditions) > 0 {
```
mrkaye97: error handling

pkg/repository/durable_events.go
```diff
@@ -165,6 +184,95 @@ func (e *StaleInvocationError) Error() string {
...
+					}
+				case CreateExternalSignalConditionKindUSEREVENT:
+					if c.UserEventKey != nil {
+						parts = append(parts, "waitForEvent("+*c.UserEventKey+")")
+					} else {
+						parts = append(parts, "waitForEvent")
+					}
+				default:
+					parts = append(parts, string(c.Kind))
+				}
```
mrkaye97: can we factor this out into a helper so we can share it?

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
```diff
@@ -225,6 +225,8 @@ async def ensure_started(self, worker_id: str) -> None:
     async def stop(self) -> None:
         self._running = False
 
+        self._fail_all_pending(Exception("DurableListener stopped"))
```
mrkaye97, line 231: will this prevent tasks that are running from draining normally on a sigterm?

pkg/repository/durable_events.go
grutt: gc

pkg/repository/durable_events.go
grutt: so i think this one is ok... or at least a pre existing problem? each run is a distinct row, right?

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
grutt: i'm not sure i'm understanding the concern. it will force the eviction path for durable tasks on sigterm

mrkaye97: yeah this is fair, I was thinking that some people have really long grace periods so everything can drain, but I guess in the durable case a reassign doesn't really matter anyways

pkg/repository/durable_events.go
mrkaye97: yep you're right, I misread 👍

mrkaye97 (approved):

PR #3251, opened 2026-03-12
fix(cli): honor active docker context when DOCKER_HOST is unset

### LINKED ISSUE #3113: [BUG] Hatchet CLI not using selected docker context (rancher)
**Describe the issue**
When I run `hatchet server start` I get the following error:

```
2026/02/26 14:23:39 FATA could not start hatchet-lite container: could not initialize network: could not list docker networks: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

Also worth noting the typo above ("FATA") if you want to fix that^ 

**Expected behavior**
The server should start 

This is because my docker context, which specifies the actual daemon socket I am using, is set to use rancher. Hatchet does not seem to honor this, but uses the default docker location.

```
$ docker context ls
NAME                DESCRIPTION                               DOCKER ENDPOINT                    ERROR
default             Current DOCKER_HOST based configuration   unix:///var/run/docker.sock        
rancher-desktop *   Rancher Desktop moby context              unix:///Users/xx/.rd/docker.sock   
```

**Additional context**

~~Even though I know it's not the cause, for posterity, setting `DOCKER_HOST` to the rancher socket does not fix it either.~~
My mistake! Setting DOCKER_HOST _does_ in fact set the correct socket and start the server - but the configured context is still ignored

BloggerBust:
# Description

Fixes #3113.

`hatchet server start` currently initializes the Docker client using `client.FromEnv`, which respects `DOCKER_HOST` but does not honor the active Docker context selected via `docker context use`. As a result, users running Docker through alternative contexts, for example Rancher Desktop, may see the CLI attempt to connect to the default socket (`/var/run/docker.sock`) instead of the configured context endpoint.

This change resolves the active Docker context when `DOCKER_HOST` is not set by reading the Docker CLI configuration (`~/.docker/config.json`) and associated context metadata. The resolved host is then passed to the Docker client via `client.WithHost`.

The precedence remains consistent with Docker CLI expectations:

1. `DOCKER_HOST` environment variable  
2. Active Docker context  
3. Default Docker behavior

Unit tests were added to validate context resolution behavior across several cases: default context, missing metadata, invalid configuration, and TCP endpoints.

## Type of change

## What's Changed

- Resolve active Docker context from Docker CLI configuration when `DOCKER_HOST` is unset
- Use the context-derived Docker host when creating the CLI Docker client
- Add unit tests for Docker context resolution scenarios
- Remove premature Docker client close during initialization

### DIFF
--- a/cmd/hatchet-cli/cli/internal/drivers/docker/context_test.go
+++ b/cmd/hatchet-cli/cli/internal/drivers/docker/context_test.go
@@ -1,172 +1,144 @@
 package docker
 
 import (
-	"crypto/sha256"
-	"encoding/hex"
+	"context"
 	"os"
 	"path/filepath"
 	"testing"
+
+	dockercontext "github.com/docker/go-sdk/context"
 )
 
-func TestResolveDockerHostFromContext(t *testing.T) {
-	t.Setenv("DOCKER_HOST", "")
-	t.Run("no docker config", func(t *testing.T) {
-		// Use a temp dir with no .docker folder
+func TestDockerContextResolution(t *testing.T) {
+	t.Run("no config returns error", func(t *testing.T) {
 		tmpDir := t.TempDir()
 		t.Setenv("HOME", tmpDir)
+		t.Setenv("USERPROFILE", tmpDir)
+		t.Setenv("DOCKER_HOST", "")
+		t.Setenv("DOCKER_CONTEXT", "")
 
-		result := resolveDockerHostFromContext()
-		if result != "" {
-			t.Errorf("expected empty string, got %q", result)
+		_, err := dockercontext.CurrentDockerHost()
+		if err == nil {
+			t.Error("expected error when no docker config exists")
 		}
 	})
 
-	t.Run("empty currentContext", func(t *testing.T) {
+	t.Run("default context returns default host", func(t *testing.T) {
 		tmpDir := t.TempDir()
 		t.Setenv("HOME", tmpDir)
-
-		// create .docker/config.json with no currentContext
-		dockerDir := filepath.Join(tmpDir, ".docker")
-		if err := os.MkdirAll(dockerDir, 0755); err != nil {
-			t.Fatal(err)
-		}
-		configContent := `{"auths": {}}`
-		if err := os.WriteFile(filepath.Join(dockerDir, "config.json"), []byte(configContent), 0644); err != nil {
-			t.Fatal(err)
-		}
-
-		result := resolveDockerHostFromContext()
-		if result != "" {
-			t.Errorf("expected empty string, got %q", result)
-		}
-	})
-
-	t.Run("default context", func(t *testing.T) {
-		tmpDir := t.TempDir()
-		t.Setenv("HOME", tmpDir)
-
-		// create .docker/config.json with currentContext = "default"
-		dockerDir := filepath.Join(tmpDir, ".docker")
-		if err := os.MkdirAll(dockerDir, 0755); err != nil {
-			t.Fatal(err)
-		}
-		configContent := `{"auths": {}, "currentContext": "default"}`
-		if err := os.WriteFile(filepath.Join(dockerDir, "config.json"), []byte(configContent), 0644); err != nil {
-			t.Fatal(err)
-		}
-
-		result := resolveDockerHostFromContext()
-		if result != "" {
-			t.Errorf("expected empty string for default context, got %q", result)
-		}
-	})
-
-	t.Run("non-default context with valid metadata", func(t *testing.T) {
-		tmpDir := t.TempDir()
-		t.Setenv("HOME", tmpDir)
-
-		contextName := "my-rancher"
-		expectedHost := "unix:///tmp/rancher.sock"
-
-		// create .docker/config.json
-		dockerDir := filepath.Join(tmpDir, ".docker")
-		if err := os.MkdirAll(dockerDir, 0755); err != nil {
-			t.Fatal(err)
-		}
-		configContent := `{"auths": {}, "currentContext": "` + contextName + `"}`
-		if err := os.WriteFile(filepath.Join(dockerDir, "config.json"), []byte(configContent), 0644); err != nil {
-			t.Fatal(err)
-		}
-
-		// create context metadata
-		hash := sha256.Sum256([]byte(contextName))
-		hashStr := hex.EncodeToString(hash[:])
-		metaDir := filepath.Join(dockerDir, "contexts", "meta", hashStr)
-		if err := os.MkdirAll(metaDir, 0755); err != nil {
-			t.Fatal(err)
-		}
-		metaContent := `{"Name":"` + contextName + `","Metadata":{},"Endpoints":{"docker":{"Host":"` + expectedHost + `","SkipTLSVerify":false}}}`
-		if err := os.WriteFile(filepath.Join(metaDir, "meta.json"), []byte(metaContent), 0644); err != nil {
-			t.Fatal(err)
-		}
-
-		result := resolveDockerHostFromContext()
-		if result != expectedHost {
-			t.Errorf("expected %q, got %q", expectedHost, result)
-		}
-	})
-
-	t.Run("non-default context with missing metadata", func(t *testing.T) {
-		tmpDir := t.TempDir()
-		t.Setenv("HOME", tmpDir)
-
-		// create .docker/config.json with non-default context but no metadata
-		dockerDir := filepath.Join(tmpDir, ".docker")
-		if err := os.MkdirAll(dockerDir, 0755); err != nil {
-			t.Fatal(err)
-		}
-		configContent := `{"auths": {}, "currentContext": "missing-context"}`
-		if err := os.WriteFile(filepath.Join(dockerDir, "config.json"), []byte(configContent), 0644); err != nil {
-			t.Fatal(err)
-		}
-
-		result := resolveDockerHostFromContext()
-		if result != "" {
-			t.Errorf("expected empty string for missing metadata, got %q", result)
-		}
-	})
-
-	t.Run("invalid config json", func(t *testing.T) {
-		tmpDir := t.TempDir()
-		t.Setenv("HOME", tmpDir)
+		t.Setenv("USERPROFILE", tmpDir)
+		t.Setenv("DOCKER_HOST", "")
+		t.Setenv("DOCKER_CONTEXT", "")
 
 		dockerDir := filepath.Join(tmpDir, ".docker")
 		if err := os.MkdirAll(dockerDir, 0755); err != nil {
 			t.Fatal(err)
 		}
-		// Invalid JSON
+		if err := os.WriteFile(filepath.Join(dockerDir, "config.json"), []byte(`{}`), 0644); err != nil {
+			t.Fatal(err)
+		}
+
+		driver, err := NewDockerDriver(context.Background())
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+		defer driver.apiClient.Close()
+
+		host := driver.apiClient.DaemonHost()
+		if host == "" {
+			t.Error("expected default Docker host, got empty string")
+		}
+	})
+
+	t.Run("non-default context resolves host", func(t *testing.T) {
+		dockercontext.SetupTestDockerContexts(t, 1, 1)
+		t.Setenv("DOCKER_HOST", "")
+		t.Setenv("DOCKER_CONTEXT", "")
+
+		driver, err := NewDockerDriver(context.Background())
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+		defer driver.apiClient.Close()
+
+		host := driver.apiClient.DaemonHost()
+		if host != "tcp://127.0.0.1:1" {
+			t.Errorf("expected tcp://127.0.0.1:1, got %q", host)
+		}
+	})
+
+	t.Run("DOCKER_HOST takes precedence", func(t *testing.T) {
+		dockercontext.SetupTestDockerContexts(t, 1, 1)
+		t.Setenv("DOCKER_HOST", "tcp://192.168.1.100:2375")
+
+		driver, err := NewDockerDriver(context.Background())
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+		defer driver.apiClient.Close()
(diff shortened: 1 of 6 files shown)


promptless-for-oss:

New suggestion: **[Document Docker context support for local server command](https://app.gopromptless.ai/suggestions/c7954e52-c7a6-4791-9a29-9dcdf165e0b8)**

---
_Tip: Use Vale? Add your vale.ini when setting up your [Docs Collection](https://app.gopromptless.ai/doc-collections)._

cmd/hatchet-cli/cli/internal/drivers/docker/context_test.go
```diff
@@ -0,0 +1,186 @@
...
+	origHome := os.Getenv("HOME")
+	origDockerHost := os.Getenv("DOCKER_HOST")
+	defer func() {
+		os.Setenv("HOME", origHome)
+		if origDockerHost != "" {
+			os.Setenv("DOCKER_HOST", origDockerHost)
+		} else {
+			os.Unsetenv("DOCKER_HOST")
+		}
+	}()
```
gregfurman: FYI you can use the [`t.SetEnv()` fixture](https://pkg.go.dev/testing#T.Setenv) for this instead.

cmd/hatchet-cli/cli/internal/drivers/docker/docker_context_test.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+
+		result := resolveDockerHostFromContext()
+		if result != "" {
+			t.Errorf("expected empty string for invalid JSON, got %q", result)
+		}
+	})
+
+	t.Run("tcp host endpoint", func(t *testing.T) {
+		tmpDir := t.TempDir()
```
gregfurman, line 138: Hmm... wonder if we can make use of the `https://pkg.go.dev/testing/fstest` instead of directly writing to the FS

cmd/hatchet-cli/cli/internal/drivers/docker/docker.go
```diff
@@ -36,13 +41,86 @@ func NewDockerDriver(ctx context.Context, optFns ...DockerDriverOpt) (*DockerDri
 }
 
 func (d *DockerDriver) init() error {
-	apiClient, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	clientOpts := []client.Opt{
```
gregfurman: TBH I think we should look into migrating over to using the new `github.com/docker/go-sdk` instead of `github.com/docker/docker`.

This newer package has much better support for lower level operations like this and improves interactions with the docker engine.

Like, setting the ctx is easy when creating a new client: https://pkg.go.dev/github.com/docker/go-sdk/client#WithDockerContext

cmd/hatchet-cli/cli/internal/drivers/docker/docker.go
```diff
@@ -36,13 +41,86 @@ func NewDockerDriver(ctx context.Context, optFns ...DockerDriverOpt) (*DockerDri
...
+		client.FromEnv,
+		client.WithAPIVersionNegotiation(),
+	}
+
+	// If DOCKER_HOST is not set, try to resolve from active Docker context
+	if os.Getenv("DOCKER_HOST") == "" {
+		if host := resolveDockerHostFromContext(); host != "" {
+			clientOpts = append(clientOpts, client.WithHost(host))
+		}
+	}
```
gregfurman: Think the [`context.CurrentDockerHost`](https://pkg.go.dev/github.com/docker/go-sdk/context#CurrentDockerHost) from the improved docker go SDK `github.com/docker/go-sdk` is relevant here i.e

```golang
	if os.Getenv("DOCKER_HOST") == "" {
		if host, _	:= context.CurrentDockerHost(); host != "" {
			clientOpts = append(clientOpts, client.WithHost(host))
		}
	}
```

This should replace the need to manually do this afaik (at the cost of an extra dependency!)

cmd/hatchet-cli/cli/internal/drivers/docker/docker.go
gregfurman, line 40: nit: While definitely out-of-scope for this PR, I think it's worth mentioning that we should probably migrate over to using the newer [`github.com/docker/go-sdk`](https://pkg.go.dev/github.com/docker/go-sdk/client) instead of this `github.com/docker/docker/client` at some point.

cmd/hatchet-cli/cli/internal/drivers/docker/context_test.go
```diff
@@ -0,0 +1,186 @@
...
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestResolveDockerHostFromContext(t *testing.T) {
```
gregfurman: IMO we should look into using https://pkg.go.dev/github.com/docker/go-sdk/context#SetupTestDockerContexts for testing instead of manually setting all of this up ourselves. WDYT?

cmd/hatchet-cli/cli/internal/drivers/docker/context_test.go
BloggerBust: Thanks, I switched the tests to `t.SetEnv()` to simplify the setup.

cmd/hatchet-cli/cli/internal/drivers/docker/docker_context_test.go
BloggerBust: Good idea. Since `resolveDockerHostFromContext()` currently reads from paths under `$HOME`, I kept the test using `t.TempDir()` and a real directory structure for now. Using `fstest` would probably mean introducing an `FS` abstraction, which felt a bit larger than I wanted for this fix.

cmd/hatchet-cli/cli/internal/drivers/docker/docker.go
BloggerBust: Agreed. I kept this fix self-contained to avoid broadening scope, but migrating toward the newer Docker SDK helpers seems worth considering separately.

cmd/hatchet-cli/cli/internal/drivers/docker/docker.go
BloggerBust: `context.CurrentDockerHost()` does look relevant here. I kept this fix self-contained to avoid adding a new dependency in this PR, but that helper seems like a good fit if you decide to migrate to `github.com/docker/go-sdk`.

cmd/hatchet-cli/cli/internal/drivers/docker/context_test.go
BloggerBust: The helper looks useful. For this PR I kept the tests self-contained so they don't depend on the Docker SDK test utilities, but `context.SetupTestDockerContexts` would simplify this setup. I am happy to make that change now or in a follow-up PR if you would prefer.

cmd/hatchet-cli/cli/internal/drivers/docker/docker.go
gregfurman: Appreciate the efforts to keep this self contained! 

Given this context to hostname translation function is _kinda_ complex I think it's worth using the context package instead

BloggerBust: I updated the implementation to use github.com/docker/go-sdk/context as suggested and also updated the tests to assert the configured Docker client host where possible.

One thing I wanted to call out explicitly is that the current package version is still tagged alpha. I assumed that was acceptable since this was the preferred direction, but let me know if you would rather avoid that dependency for now.

frontend/docs/pages/reference/cli/running-hatchet-locally.mdx
```diff
@@ -2,6 +2,8 @@
 
 The Hatchet CLI provides the `hatchet server` commands to run a local instance of Hatchet for development and testing purposes. This local instance relies on Docker to run the necessary services.
 
+When `DOCKER_HOST` is not set, Hatchet uses the active Docker context from your local Docker configuration. If `DOCKER_HOST` is set explicitly, it takes precedence.
```
gregfurman: Should we not also mention that the driver now respects the `DOCKER_CONTEXT` if `DOCKER_HOST` is not set? 

Granted, you added it as an inline comment but could be worth mentioning that we now support this

https://github.com/BloggerBust/hatchet/blob/4178fb71dd8b714ae89ff2e9a09eac8a136a905a/cmd/hatchet-cli/cli/internal/drivers/docker/docker.go#L45-L46

Here's the relevent snippet from the SDK code at [github.com/docker/go-sdk/context/context.go](https://github.com/docker/go-sdk/blob/f7329ac240a8b6b48192fad18f2ce38cb6446daa/context/context.go#L18-L22):

```golang
	// EnvOverrideContext is the name of the environment variable that can be
	// used to override the context to use. If set, it overrides the context
	// that's set in the CLI's configuration file, but takes no effect if the
	// "DOCKER_HOST" env-var is set (which takes precedence.
	EnvOverrideContext = "DOCKER_CONTEXT"
```

Also, see https://docs.docker.com/engine/manage-resources/contexts/#use-a-different-context

cmd/hatchet-cli/cli/internal/drivers/docker/docker.go
```diff
@@ -36,11 +37,21 @@ func NewDockerDriver(ctx context.Context, optFns ...DockerDriverOpt) (*DockerDri
...
+	clientOpts := []client.Opt{
+		client.FromEnv,
+		client.WithAPIVersionNegotiation(),
+	}
+
+	// Resolve Docker host from the current Docker context.
+	// This respects DOCKER_HOST, DOCKER_CONTEXT, and the active context in ~/.docker/config.json.
+	if host, err := dockercontext.CurrentDockerHost(); err == nil && host != "" {
+		clientOpts = append(clientOpts, client.WithHost(host))
+	}
```
gregfurman, line 49: praise: Nice! This is looking super clean 🙂

cmd/hatchet-cli/cli/internal/drivers/docker/context_test.go
```diff
@@ -0,0 +1,144 @@
...
+		tmpDir := t.TempDir()
+		t.Setenv("HOME", tmpDir)
+		t.Setenv("USERPROFILE", tmpDir)
+		t.Setenv("DOCKER_HOST", "")
+		t.Setenv("DOCKER_CONTEXT", "")
+
+		_, err := dockercontext.CurrentDockerHost()
+		if err == nil {
+			t.Error("expected error when no docker config exists")
+		}
```
gregfurman: Shouldn't we be testing the `NewDockerDriver` here instead of the SDK functionality?

cmd/hatchet-cli/cli/internal/drivers/docker/context_test.go
```diff
@@ -0,0 +1,144 @@
...
+			t.Fatal(err)
+		}
+		if err := os.WriteFile(filepath.Join(dockerDir, "config.json"), []byte("{invalid}"), 0644); err != nil {
+			t.Fatal(err)
+		}
+
+		_, err := dockercontext.CurrentDockerHost()
+		if err == nil {
+			t.Error("expected error for invalid config")
+		}
```
gregfurman: Again, shouldn't we be testing the behaviour of the actual `NewDockerDriver`? Like, what would it's behaviour be under the cirumstances outlined in the test?

cmd/hatchet-cli/cli/internal/drivers/docker/docker_context_test.go
gregfurman: Fair point 👍

cmd/hatchet-cli/cli/internal/drivers/docker/docker_context_test.go
```diff
@@ -0,0 +1,144 @@
...
+			t.Fatal(err)
+		}
+
+		driver, err := NewDockerDriver(context.Background())
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+		defer driver.apiClient.Close()
+
+		host := driver.apiClient.DaemonHost()
```
gregfurman, line 54: praise: Neat way of testing the context/host setting behaviour!

cmd/hatchet-cli/cli/internal/drivers/docker/context_test.go
```diff
@@ -0,0 +1,162 @@
...
+		t.Setenv("DOCKER_CONTEXT", "")
+
+		driver, err := NewDockerDriver(context.Background())
+		if err != nil {
+			t.Fatalf("expected driver to initialize, got error: %v", err)
+		}
+		defer driver.apiClient.Close()
+
+		host := driver.apiClient.DaemonHost()
+		if host == "" {
```
gregfurman: Shouldn't we be checking this actually falls back correctly, like with the  `"non-default context resolves host"` test?

cmd/hatchet-cli/cli/internal/drivers/docker/docker_context_test.go
gregfurman, line 1: Can this be renamed to `docker_test.go` or `docker_context_test.go` since we don't have a `context.go` file that corresponds to the filename

gregfurman (approved):
LGTM! Thanks for addressing all my feedback. If you have capacity, I've left 2 more comments. Otherwise, nothing blocking. Lmk!

gregfurman:
@BloggerBust Looks like you're getting merge conficts. Can you rebase against HEAD of `main` and see if that unblocks this?

promptless-for-oss:

Updated existing suggestion: **[Document Docker context support for local server command](https://app.gopromptless.ai/suggestions/c7954e52-c7a6-4791-9a29-9dcdf165e0b8)**

---

BloggerBust:
Hey @gregfurman, during the rebase I accepted the upstream changelog/version, then prepared a separate commit to bump the patch to `1.29.3` and re-add the changelog entry. I pushed that after the PR had already been merged. Would you like me to open a small follow-up PR to address that?

gregfurman:
Hey @BloggerBust. Which changelog are you referring to?

BloggerBust:
@gregfurman, I was thinking in terms of the SDK changelogs, but this change is only in the Go CLI path, so there’s no corresponding changelog/version update needed here.

PR #3260, opened 2026-03-12
Add cleanup module to handle graceful shutdown, improve logging experience

juliusgeo:
# Description

Fixes # (issue)

## Type of change

## What's Changed


pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (c *Cleanup) Add(fn func() error, name string) {
+	c.Fns = append(c.Fns, CleanupFn{
+		Name: name,
+		Fn:   fn,
+	})
+}
+
+func (c *Cleanup) Run() error {
+	log.Printf("waiting for all other services to gracefully exit...")
```
gregfurman: Should we rather be passing that `zerologger` and using it for logging ~since (afaik) this is logging to stdout~?

https://github.com/hatchet-dev/hatchet/blob/bf05890f0ba6f88de0c1e861a7ba0323ae7d2406/cmd/hatchet-engine/engine/run.go#L95

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,43 @@
...
+	for i, fn := range c.Fns {
+		log.Printf("shutting down %s (%d/%d)\n", fn.Name, i+1, len(c.Fns))
+		before := time.Now()
+		if err := fn.Fn(); err != nil {
+			return fmt.Errorf("could not teardown %s: %w", fn.Name, err)
+		}
```
gregfurman, line 63: q: What happens if we early exit without shutting down all processes? Is there a chance shutdown could hang?

cmd/hatchet-engine/engine/run.go
```diff
@@ -83,65 +85,49 @@ func init() {
...
 	}
 
 	var l = server.Logger
 
-	teardown, err := RunWithConfig(ctx, server)
+	err = RunWithConfig(ctx, server, &cleanup)
+	cleanup.Add(serverCleanup, "server")
+	cleanup.Add(func() error {
+		return server.Disconnect()
+	}, "database")
```
gregfurman: nit: why not pass the function as a param instead?

```suggestion
	cleanup.Add(server.Disconnect, "database")
```

juliusgeo: Good catch, fixed.

pkg/cleanup/cleanup.go
juliusgeo: So for this change, I'm trying to preserve the same behavior we had before: https://github.com/hatchet-dev/hatchet/pull/3260/changes/BASE..bf05890f0ba6f88de0c1e861a7ba0323ae7d2406#diff-a474308cfad07b710f145423c053e3cca49384ea95cb1144e3e18b4509808ba5L117
which also errors out when cleanup could not be completed. I don't think it would hang, though, it would just shutdown without gracefully cleaning up. Which if there was an error, is already the case.

pkg/cleanup/cleanup.go
juliusgeo: Fixed it so that it uses warn level log statements. Made it so that we only actually log the warnings if the total time limit is exceeded for cleanup.

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,54 @@
...
+func (c *Cleanup) Add(fn func() error, name string) {
+	c.Fns = append(c.Fns, CleanupFn{
+		Name: name,
+		Fn:   fn,
+	})
+}
+
+func (c *Cleanup) Run(l *zerolog.Logger) error {
+	lines := []string{}
+	start := time.Now()
```
gregfurman: IMO if we're adding a time-limit (or deadline) we should probably consider using a `context.WithTimeout`. 

Also, passing a `ctx` here gives us the opportunity to cancel early (if a force shutdown or whatever).

Could do something like:


```golang

func (c *Cleanup) Run(ctx context.Context) error {

// Could either assume the caller has set a deadline or attach one ourselves i.e
// ctx, cancel = context.WithTimeout(ctx, time.Second * 30)
// defer cancel()

// loop over all fns...

select {
   case <-ctx.Done:
      // optionally, if you want the cleanup exceeded error, check whether the error is context.DeadlineExceeded and return that formatted error.
      // (the deadline could be extracted via ctx.Deadline())
      // fmt.Errorf("cleanup exceeded time limit of %d seconds", ctx.)
      return ctx.Err()
   default:
      if err := fn.Fn(); err != nil {
			return fmt.Errorf("could not teardown %s: %w", fn.Name, err)
		}
}

}
```

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,54 @@
...
+	lines = append(lines, "waiting for all other services to gracefully exit...")
+	for i, fn := range c.Fns {
+		lines = append(lines, fmt.Sprintf("shutting down %s (%d/%d)", fn.Name, i+1, len(c.Fns)))
+		before := time.Now()
+		if err := fn.Fn(); err != nil {
+			return fmt.Errorf("could not teardown %s: %w", fn.Name, err)
+		}
+		lines = append(lines, fmt.Sprintf("successfully shutdown %s in %s (%d/%d)\n", fn.Name, time.Since(before), i+1, len(c.Fns)))
```
gregfurman: This makes sense! However... I'm concerned that we actually could want to see these logs happening live on shutdown. 

Like, what if a service is actually taking a while to shutdown but doesn't cause an error? Could be useful to optionally see that info.

Perhaps one of the other devs can weigh in but IMO it's nicer for devex to optionally see what is happening on startup/shutdown.

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func (c *Cleanup) Add(fn func() error, name string) {
+	c.Fns = append(c.Fns, CleanupFn{
+		Name: name,
+		Fn:   fn,
+	})
+}
+
+func (c *Cleanup) Run(l *zerolog.Logger) error {
```
gregfurman: nit: Thanks! Wonder if it could it be more idiomatic to make this an attribute of the `Cleanup` struct that we set on `New(l *zerolog.Logger)`? 👀

pkg/cleanup/cleanup.go
juliusgeo: So I'm not super sure about this because we don't want to cancel early, we just want to know *when* the cleanup took a long time to complete, so then we can look at the logs to figure out what service was taking the extra time. So using context.WithTimeout and then returning an error prior to all the cleanups finishing would lose info about which cleanup was taking that extra time.

pkg/cleanup/cleanup.go
juliusgeo: So in terms of seeing the logs live on shutdown--the main reason I'm not just dumping them to INFO is that this is intended for cloud deployments where the log level is set to WARN. I could make it so that it both logs to info and warn, but that would be confusing in the case someone is self hosting and has the log level set to INFO--they would get a bunch of repeated logs.

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,56 @@
...
+			return fmt.Errorf("could not teardown %s: %w", fn.Name, err)
+		}
+		lines = append(lines, fmt.Sprintf("successfully shutdown %s in %s (%d/%d)\n", fn.Name, time.Since(before), i+1, len(c.Fns)))
+	}
+	lines = append(lines, "all services have successfully gracefully exited")
+	if time.Since(start) > c.TimeLimit {
+		for _, line := range lines {
+			c.logger.Warn().Msg(line)
+		}
+		return fmt.Errorf("cleanup exceeded time limit of %d seconds", c.TimeLimit)
```
abelanger5: nit: since `TimeLimit` is a duration with a `String()` arg I think this can just be:

```
return fmt.Errorf("cleanup exceeded time limit of %s", c.TimeLimit)
```

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,56 @@
...
+	for i, fn := range c.Fns {
+		lines = append(lines, fmt.Sprintf("shutting down %s (%d/%d)", fn.Name, i+1, len(c.Fns)))
+		before := time.Now()
+		if err := fn.Fn(); err != nil {
+			return fmt.Errorf("could not teardown %s: %w", fn.Name, err)
+		}
+		lines = append(lines, fmt.Sprintf("successfully shutdown %s in %s (%d/%d)\n", fn.Name, time.Since(before), i+1, len(c.Fns)))
+	}
+	lines = append(lines, "all services have successfully gracefully exited")
+	if time.Since(start) > c.TimeLimit {
```
abelanger5: As-written, I don't think this gets us the behavior we need? The problem is that if we get a SIGTERM followed by a SIGKILL after 30 seconds, we're never going to reach this line, we'll have been killed already. I think we need to be checking the time limit exceeding async with `time.After` in a goroutine, and then safely printing the log lines that we do have

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,56 @@
...
+		for _, line := range lines {
+			c.logger.Warn().Msg(line)
```
abelanger5: I think this should be `Error()` as we won't see warnings unless we tail directly in our setup

pkg/cleanup/cleanup.go
juliusgeo: Ah that makes sense. I used @gregfurman's suggestion above to make it so that it uses a context with timeout that logs with Error when the timeout is exceeded. I also changed the timeout to 10 seconds because we wait 20 seconds in run.go for the server shutdown wait period. I also removed the error return because if the deadline is exceeded the process is already dead.

pkg/cleanup/cleanup.go
juliusgeo: Fixed.

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,58 @@
...
+		Name: name,
+		Fn:   fn,
+	})
+}
+
+func (c *Cleanup) Run() error {
+	lines := []string{}
+	lines = append(lines, "waiting for all other services to gracefully exit...")
+	go func() {
+		timeoutCtx, cancel := context.WithTimeout(context.Background(), c.TimeLimit)
```
gregfurman: Nice! I like this context based approach 👍 The ctx should probably be defined (and cancelled) outside of the goroutine though.

Also, think it could be re-worked a bit though since `lines` being accessed like this could actually cause a race condition where the only things that'll be logged are what was _currently_ captured in `lines` at the time of failure versus all shutdown logs.

Also, go's race detector could likely flag this since we're accessing `lines` in 2 seperate goroutines without mutex protection.

What do you think about converting `lines` into a channel? This approach is a bit more verbose but would also allow all shutdowns to be logged at the DEBUG level if no error and the ERROR level if the timeout is triggered  i.e

```golang
	lines := make(chan string)
	lines <- "waiting for all other services to gracefully exit..."

	ctx, cancel := context.WithTimeout(context.Background(), c.TimeLimit)
	defer cancel()

	go func() {
		// log at the debug level by default
		logger := c.logger.Debug()

		<-ctx.Done()

		switch ctx.Err() {
		case context.DeadlineExceeded:
			// if the ctx is cancelled due to a timeout, then promote the logs
			// to an error
			logger = c.logger.Error()
		}

		// ensure we drain the channel and log everything
		for line := range lines {
			logger.Msg(line)
		}
	}()

// ...
```

juliusgeo: That makes sense in terms of a race condition, but I'm pretty sure that's what we would want in this case. The goal is to be able to get logs in this scenario:
1. SIGINT gets sent to process
2. we wait for ShutdownWait seconds (20), and then clean up, but cleanup takes longer than 10 seconds
3. SIGKILL gets sent to the process at exactly 30 seconds


In this scenario, we would want to print all the logs we have at that moment in time (ie. what is *currently* in lines) and there is no extra time to wait for all of the shutdown logs to come in. So time of failure == time process gets killed, so the race condition would be irrelevant. However, I think your approach should work because the SIGKILL would just kill while waiting for log lines in:
```
for line := range lines {
	logger.Msg(line)
}
```

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,68 @@
...
+func (c *Cleanup) Add(fn func() error, name string) {
+	c.Fns = append(c.Fns, CleanupFn{
+		Name: name,
+		Fn:   fn,
+	})
+}
+
+func (c *Cleanup) Run() error {
+	// 1st and last line + 2 lines for each fn. Makes sure we don't block on the chan send
+	lines := make(chan string, (len(c.Fns)*2)+2)
```
gregfurman: IMO we can just make this unbuffered, which means we don't have to worry about blocking

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,68 @@
...
+	for i, fn := range c.Fns {
+		lines <- fmt.Sprintf("shutting down %s (%d/%d)", fn.Name, i+1, len(c.Fns))
+		before := time.Now()
+		if err := fn.Fn(); err != nil {
+			return fmt.Errorf("could not teardown %s: %w", fn.Name, err)
+		}
+		lines <- fmt.Sprintf("successfully shutdown %s in %s (%d/%d)\n", fn.Name, time.Since(before), i+1, len(c.Fns))
+	}
+	lines <- "all services have successfully gracefully exited"
+	close(lines)
```
gregfurman: This should be a deferred call since the function could shut down early due to an error, and we're left with a goroutine that never finishes and a channel that never closes.

i.e 

```golang
if err := fn.Fn(); err != nil {
            // log channel remains open, which means goroutine never exits.
			return fmt.Errorf("could not teardown %s: %w", fn.Name, err)
		}
```

So we could defer the `close`:

```golang
lines := make(chan string, (len(c.Fns)*2)+2)
defer close(lines)

ctx, cancel := context.WithTimeout(context.Background(), c.TimeLimit)
defer cancel()
```

Note, `defer` calls are LIFO so we probably want the channel to close before the ctx is cancelled but it's probably fine either way 🤷

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func (c *Cleanup) Add(fn func() error, name string) {
+	c.Fns = append(c.Fns, CleanupFn{
+		Name: name,
+		Fn:   fn,
+	})
+}
+
+func (c *Cleanup) Run() error {
```
gregfurman, line 37: Can we perhaps pass the ctx from the `run.go` call here? That way, if the caller context gets cancelled early we can stop the shutdown mechanism.

gregfurman:
@juliusgeo Left some more comments. This lifecycle stuff is deceivingly confusing so apologies if I'm being a pain here 😅 

If you need to get this in ASAP then I'm happy to approve and we can iterate on the cleanup mechanism

pkg/cleanup/cleanup.go
juliusgeo: Ah, for some reason when testing locally it was blocking without making it buffered. Seems to have resolved itself, so must have been something else. Fixed!

pkg/cleanup/cleanup.go
juliusgeo: Good catch, fixed!

pkg/cleanup/cleanup.go
juliusgeo: I was considering that, but it gets tricky because the context being passed in is the InterruptContext (if no graceful shutdown flag is enabled), which would mean that the cleanup context would get cancelled before it has a chance to run IIUC.

abelanger5 (approved):

gregfurman (approved):
Thanks for addressing all my feedback! LGTM

PR #3272, opened 2026-03-13
Fix  durable migration order

grutt:
# Description

Fixes migration order after merge into main

## Type of change


mrkaye97 (approved):
🥳

Issue #2573, opened 2025-11-26
[BUG] Workloads stuck in "RUNNING" state, no way to cancel or delete them


fourslashw:
**Describe the issue**
While under relatively high load, 4 workflows got stuck in glitched "RUNNING" state. Pressing "cancel" does nothing other then updating "Finished at" field and duration, while still leaving workflow as "running". 

Those stuck workflows are creating endless background activity of some kind on worker side while leaking memory which inevitably leads to workers crashing. We had to temporary create a new tenant and move everything there.

**Environment**
- SDK: @hatchet-dev/typescript-sdk: "1.6.3"
- Engine: Self hosted v0.73.34, using rabbitmq as message queue and amazon RDS m8g.xlarge postgresql as database.

**Expected behavior**
To cancel workflow or being able to delete it somehow.

**Code to Reproduce, Logs, or Screenshots**

Logs from incident time grepped by "error":
[hatchet_logs_errors.csv](https://github.com/user-attachments/files/23767184/hatchet_logs_errors.csv)
A lot of errors with SQLSTATE 57014 

Log snippet from engine and dashboard while trying to cancel task:
[hatchet_cancel_task_logs.csv](https://github.com/user-attachments/files/23766690/hatchet_cancel_task_logs.csv)
The only anomaly I can see there is this
```
"2025-11-26T10:25:12.000+01:00","jobs.prod.our.website","[90m2025-11-26T09:25:12.82Z[0m [31mERR[0m [1mAPI[0m [36merror=[0m[31m[1m""code=404, message=Task not found""[0m[0m [36mhost=[0mhatchet.prod.internal.ournetwork [36mlatency=[0m63.446482 [36mmethod=[0mGET [36mremote_ip=[0m79.125.5.174 [36mservice=[0mserver [36mstatus=[0m500 [36muri=[0m/api/v1/stable/tasks/e332fd2a-156d-44e9-94c4-5c46e3605a6d [36muser_agent=[0m""Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/142.0.0.0 Safari/537.36""","/hatchet-hatchet-dashboard-1","1","docker.hatchet.dashboard"
```
Which saying task `e332fd2a-156d-44e9-94c4-5c46e3605a6d` is not found while being on page
[...]

PR #3276, opened 2026-03-14
fix: exclude on_failure steps from DAG total_tasks to prevent stuck RUNNING status

### LINKED ISSUE #1996: [BUG] Run stuck at running even after cancelled or done
**Describe the issue**
Run is stuck at running, even after all tasks are cancelled or done.

**Environment**
- Engine: Self-hosted `v0.67.8` (tenant v1)

**Expected behavior**

I expected the run to be cancelled after all the tasks are cancelled.

Image

```
GET /api/v1/stable/workflow-runs/99d8ee83-5a3c-4a24-bf9f-34cff3725f30

{ "CronInput": null, "CronTriggers": null, "DefaultPriority": 1, "EventTriggers": null, "Name": "PDFChunkedTextExtractorWorkflow", "Sticky": null, "Tasks": [ { "Action": "pdfchunkedtextextractorworkflow:execute", "DesiredWorkerLabels": null, "Parents": [], "RateLimits": [], "ReadableId": "execute", "Retries": 2, "RetryBackoffFactor": 10, "RetryBackoffMaxSeconds": 86400, "ScheduleTimeout": "6h", "Timeout": "3h", "TriggerConditions": [] }, { "Action": "pdfchunkedtextextractorworkflow:procreate", "DesiredWorkerLabels": null, "Parents": [ "execute" ], "RateLimits": [], "ReadableId": "procreate", "Retries": 2, "RetryBackoffFactor": 10, "RetryBackoffMaxSeconds": 86400, "ScheduleTimeout": "6h", "Timeout": "15m", "TriggerConditions": [] }, { "Action": "pdfchunkedtextextractorworkflow:on-success", "DesiredWorkerLabels": null, "Parents": [ "execute", "procreate" ], "RateLimits": [], "ReadableId": "on-success", "Retries": 2, "RetryBackoffFactor": 10, "RetryBackoffMaxSeconds": 86400, "ScheduleTimeout": "6h", "Timeout": "5m", "TriggerConditions": [] } ], "concurrency": [ { "Expression": "additional_metadata.realm_id", "LimitStrategy": "GROUP_ROUND_ROBIN", "MaxRuns": 4096 } ], "description": "", "onFailureJob": { "Action": "pdfchunkedtextextractorworkflow:on-failure", "DesiredWorkerLabels": null, "Parents": [], "RateLimits": [], "ReadableId": "on-failure", "Retries": 2, "RetryBackoffFactor": 10, "RetryBackoffMaxSeconds": 86400, "ScheduleTimeout": "6h", "Timeout": "5m", "TriggerConditions": [] } }
```

Workflow config:
```
{"CronInput":null,"CronTriggers":null,"DefaultPriority":1,"EventTriggers":null,"Name":"cpu-PDFDataExtractorWorkflow","Sticky":null,"Tasks":[{"Action":"cpu-pdfdataextractorworkflow:execute","DesiredWorkerLabels":null,"Parents":[],"RateLimits":[],"ReadableId":"execute","Retries":2,"RetryBackoffFactor":10,"RetryBackoffMaxSeconds":86400,"ScheduleTimeout":"6h","Timeout":"3h","TriggerConditions":[]},{"Action":"cpu-pdfdataextractorworkflow:procreate","DesiredWorkerLabels":null,"Parents":["execute"],"RateLimits":[],"ReadableId":"procreate","Retries":2,"RetryBackoffFactor":10,"RetryBackoffMaxSeconds":86400,"ScheduleTimeout":"6h","Timeout":"15m","TriggerConditions":[]},{"Action":"cpu-pdfdataextractorworkflow:on-success","DesiredWorkerLabels":null,"Parents":["execute","procreate"],"RateLimits":[],"ReadableId":"on-success","Retries":2,"RetryBackoffFactor":10,"RetryBackoffMaxSeconds":86400,"ScheduleTimeout":"6h","Timeout":"5m","TriggerConditions":[]}],"concurrency":[{"Expression":"additional_metadata.realm_id","LimitStrategy":"GROUP_ROUND_ROBIN","MaxRuns":4096}],"description":"","onFailureJob":{"Action":"cpu-pdfdataextractorworkflow:on-failure","DesiredWorkerLabels":null,"Parents":[],"RateLimits":[],"ReadableId":"on-failure","Retries":2,"RetryBackoffFactor":10,"RetryBackoffMaxSeconds":86400,"ScheduleTimeout":"6h","Timeout":"5m","TriggerConditions":[]}}
```

**Additional context**

I've seen this problem twice, both times after the database was under extreme load.
To go around the problem, I scaled down the workers and started cancelling workflow runs.
But some wouldn't get cancelled, even after multiple tries.

From the status updates you can see that it was stuck at running even before it was cancelled.
`on-success` was completed and `on-failure` skipped, but the workflow was still running.

shaunpatterson:
## Summary

Disclaimer: Claude Code wrote this... throwing up this PR while I attempt to reproduce and test...

Workflow runs with `on_failure_task` handlers permanently stay in `RUNNING` status even after all tasks complete. This affects **every workflow type** that defines an on_failure handler.

We discovered this in production where 5 workflow runs across 4 different workflow types (summarization, fetch, ingest, discovery) were stuck in RUNNING for 9-12+ hours despite all tasks having completed successfully or failed.

Fixes #1996
Relates to #2573

## Root Cause

Two bugs in the DAG completion state machine work together to cause this:

### Bug 1: `on_failure` steps incorrectly counted in `total_tasks` (`trigger.go:629-637`)

When a DAG is created, the first loop iterates ALL steps (including `ON_FAILURE`) and adds them to `dagToTaskIds`. The length of this slice becomes `total_tasks` for the DAG:

```go
// BEFORE (buggy) - ALL steps added to dagToTaskIds, including ON_FAILURE
for _, step := range steps {
    if !isDag {
        stepsToExternalIds[i][step.ID] = tuple.externalId
    } else {
        externalId := uuid.New()
        stepsToExternalIds[i][step.ID] = externalId
        dagToTaskIds[tuple.externalId] = append(dagToTaskIds[tuple.externalId], externalId)
    }
}
```

However, `ON_FAILURE` steps are **conditionally triggered** — they only create a task when another step fails (via event match conditions at line 697). On the success path, the on_failure task is never created as an actual task row.

**Result**: For a workflow with 1 main task + 1 on_failure task: `total_tasks = 2`, but only 1 task is ever created.

### Bug 2: SQL uses `!=` instead of `<` (`olap.sql:1090`)

The `UpdateDAGStatuses` query determines DAG status with:

```sql
WHEN dtc.task_count != dtc.total_tasks THEN 'RUNNING'
```

This blocks **both** paths:
- **Success path**: `task_count=1, total_tasks=2` → `1 != 2` → RUNNING forever
- **Failure path** (after Bug 1 fix): on_failure creates an extra task, so `task_count=2, total_tasks=1` → `2 != 1` → RUNNING forever

### Data Flow

```
Workflow Definition (2 steps: main + on_failure)
    ↓
trigger.go: dagToTaskIds includes BOTH steps
    ↓
createDAGs: total_tasks = len(dagToTaskIds) = 2
    ↓
Only main_task created as actual task (on_failure waits for failure event match)
    ↓
main_task completes successfully
    ↓
UpdateDAGStatuses SQL:
    task_count = 1 (only main_task in v1_dag_to_task_olap)
    total_tasks = 2 (set at DAG creation, includes on_failure)
    → 1 != 2 → RUNNING forever ❌
```

### Fix 1: Exclude `ON_FAILURE` from `dagToTaskIds` (`trigger.go`)

```go
// AFTER - ON_FAILURE steps get external IDs but are NOT counted in total_tasks
for _, step := range steps {
    if !isDag {
        stepsToExternalIds[i][step.ID] = tuple.externalId
    } else {
        externalId := uuid.New()
        stepsToExternalIds[i][step.ID] = externalId

        if step.JobKind != sqlcv1.JobKindONFAILURE {
            dagToTaskIds[tuple.externalId] = append(dagToTaskIds[tuple.externalId], externalId)
        }
    }
}
```

`ON_FAILURE` steps still get external IDs (required for match condition creation at line 697), but are **not** added to `dagToTaskIds`, so `total_tasks` only counts unconditionally-triggered steps.

### Fix 2: Change SQL operator from `!=` to `<` (`olap.sql`)

```sql
-- BEFORE:
WHEN dtc.task_count != dtc.total_tasks THEN 'RUNNING'

-- AFTER:
WHEN dtc.task_count < dtc.total_tasks THEN 'RUNNING'
```

This ensures that when `on_failure` IS triggered (failure path), the extra task beyond `total_tasks` doesn't block the state machine. `task_count(2) < total_tasks(1)` is `FALSE`, allowing evaluation to proceed to `failed_count > 0 → FAILED`.

### Post-Fix Behavior Matrix

| Scenario | total_tasks | task_count | completed | failed | Result |
|----------|-------------|------------|-----------|--------|--------|
| Success (on_failure not triggered) | 1 | 1 | 1 | 0 | **COMPLETED** |
| Failure (on_failure triggered) | 1 | 2 | 1 | 1 | **FAILED** |
| Normal DAG (no on_failure) | 3 | 3 | 3 | 0 | **COMPLETED** |
| Normal DAG (tasks still creating) | 3 | 1 | 1 | 0 | **RUNNING** |
| Normal DAG (one failed) | 3 | 3 | 2 | 1 | **FAILED** |

## Production Evidence

5 stuck runs observed on our self-hosted Hatchet instance across 4 workflow types:

| Workflow | Task Status | Run Status | Stuck Duration |
|----------|-------------|------------|----------------|
| summarization | `summarize_task` COMPLETED | RUNNING | 9+ hours |
| discovery | `discover_task` COMPLETED | RUNNING | 12+ hours |
| ingest | `ingest_task` FAILED | RUNNING | 9+ hours |
| fetch | `on_fetch_failure` COMPLETED | RUNNING | 12+ hours |
| fetch | `on_fetch_failure` COMPLETED | RUNNING | 12+ hours |

All had 2 steps in the workflow shape but only 1 task ever created. The second step showed `taskExternalId: 00000000-0000-0000-0000-000000000000` (never instantiated).

### Unit Tests (`trigger_on_failure_test.go`) — 3 tests

- `Test_dagToTaskIds_excludesOnFailureSteps` — verifies ON_FAILURE excluded from task count (3 sub-cases)
- `Test_dagToTaskIds_onFailureStillGetsExternalId` — verifies ON_FAILURE still gets external ID for match conditions
- `Test_totalTasks_matchesCreateDAGOpts` — verifies createDAGOpts.TaskIds excludes ON_FAILURE

### Integration Tests (`trigger_on_failure_integration_test.go`) — 5 tests with real PostgreSQL via TestContainers

- `TestDAGStatusSQL_SuccessPath_OnFailureExcluded` — main completes → COMPLETED
- `TestDAGStatusSQL_FailurePath_OnFailureTriggered` — main fails + on_failure completes → FAILED
- `TestDAGStatusSQL_OldBug_OnFailureInTotalTasks` — documents that trigger.go fix alone is insufficient
- `TestDAGStatusSQL_OldOperator_WouldBlockOnFailureTriggered` — documents that `<` vs `!=` fix is also needed
- `TestDAGStatusSQL_NormalDAG_NoOnFailure` — regression tests for normal DAGs (5 sub-cases: all_completed, one_still_running, one_failed, not_all_tasks_created_yet, one_cancelled)

All 10 tests pass.

### Existing stuck runs will NOT self-heal
Already-stuck runs have `total_tasks=2` persisted in the OLAP database. The trigger.go fix only affects newly created DAGs. Existing stuck runs need manual cancellation or a data backfill to correct `total_tasks`.

### sqlc regeneration
The generated Go files were regenerated via `sqlc generate` (not hand-edited). The sqlc version bump from v1.29.0 → v1.30.0 is incidental.

### Future-proofing
Currently only `DEFAULT` and `ON_FAILURE` exist as `JobKind` values. If future conditional job kinds are added (e.g., `ON_SUCCESS`, `COMPENSATION`), they would need similar exclusion from `total_tasks`. Consider introducing a helper like `step.CountsTowardTotal()` to prevent this class of bug from recurring.

## Files Changed

| File | Change |
|------|--------|
| `pkg/repository/trigger.go` | Exclude `JobKindONFAILURE` from `dagToTaskIds` |
| `pkg/repository/sqlcv1/olap.sql` | Change `!=` to `<` in DAG status CASE expression |
| `pkg/repository/sqlcv1/*.go` | Regenerated via `sqlc generate` |
| `pkg/repository/trigger_on_failure_test.go` | 3 unit tests (NEW) |
| `pkg/repository/trigger_on_failure_integration_test.go` | 5 integration tests (NEW) |


shaunpatterson:
## Additional Findings from Production Investigation

After submitting this PR, we investigated further on our dev environment and found the on_failure bug has **cascading effects** beyond just stuck status display:

### Concurrency Slot Deadlock

The stuck RUNNING workflow runs **consume concurrency slots** that are never released. For our enrichment workflow with `GROUP_ROUND_ROBIN, max_runs=5`:

1. On_failure bug causes enrichment runs to stay RUNNING in OLAP
2. RUNNING enrichment runs hold concurrency slots for `key=abc`
3. Once `max_runs` slots are filled with phantom RUNNING runs, **all new enrichment tasks are permanently blocked**
4. We observed **21,000+ queued enrichment tasks** unable to process because 5 concurrency slots were held by completed-but-stuck runs

### OLAP Replication Lag Makes Cancellation Ineffective

Cancelling stuck runs via the SDK updates the core database, but the OLAP tables (which the concurrency scheduler reads from) don't update:

- `hatchet.runs.aio_cancel()` updates core DB ✓
- OLAP tables still show runs as RUNNING ✗
- Concurrency slot `is_filled` flag stays true ✗
- Engine re-assigns the same slots on each scheduling cycle ✗

### Database Fix Required for Existing Stuck Runs

After deploying this PR's fix (which prevents **new** runs from getting stuck), existing stuck runs require direct database intervention:

```sql
-- 1. Fix total_tasks on stuck OLAP DAGs
UPDATE v1_dags_olap
SET total_tasks = 1
WHERE readable_status = 'RUNNING'
  AND total_tasks = 2
  AND external_id IN ('<stuck_run_ids>');

-- 2. Release concurrency slots held by stuck runs  
DELETE FROM v1_workflow_concurrency_slot
WHERE workflow_id = '<workflow_id>'
  AND is_filled = true
  AND workflow_run_id IN ('<stuck_run_ids>');
```

### Scope of Impact

This is not just a cosmetic UI bug. The on_failure `total_tasks` miscounting causes:
1. **Stuck RUNNING status** (this PR's fix)
2. **Concurrency slot leaks** → queue starvation for all workflows using concurrency limits
3. **OLAP desync** → cancellation doesn't propagate, manual DB fixes required

### Recommendation

In addition to this PR's fix, consider:
1. A migration or startup check that detects and fixes existing DAGs with `total_tasks` that include ON_FAILURE steps
2. A periodic job that releases concurrency slots for runs that are in a terminal state in the core DB but still RUNNING in OLAP
3. Ensuring `UpdateDAGStatuses` OLAP reconciliation catches cancelled runs from the core DB

promptless-for-oss:

New suggestion: **[Add troubleshooting section for stuck RUNNING workflows with on-failure handlers](https://app.gopromptless.ai/suggestions/a863385d-ede9-481f-84b6-d5c4c8a46224)**

---

abelanger5:
Hi @shaunpatterson, thanks for the attempted fix! I'll take a look, but I don't think the premise of this PR is correct. On failure tasks are always created, even during a success, they're just created in a SKIPPED state. As a result, they should count towards the total_tasks. 

It's extremely likely that there are error logs on the Hatchet engine which indicate that an `internal-event` failed to process, which is the reason the workflow appears to be in a stuck state. The most common cause of failing to process an internal event is that it's not JSON serializable (from a Postgres perspective) - either containing null byte sequences or isn't a JSON object (instead an array or `null`). Another common cause is database queries failing, which is dependent on the Postgres setup. 

Could you share workflow code for a minimal reproducible example which shows this happening? I'm going to close this for now but happy to investigate further and dig into this PR if we can repro. Thanks!

shaunpatterson:
Sounds good.  I'll see if I can reproduce this week

shaunpatterson:
Yeah I think the root cause is something misconfigured on my end.  

I reproduced this again -- without the on_failure_task 
```
⏺ The gRPC stream drops because of the HATCHET_CLIENT_LISTENER_V2_TIMEOUT=180 we set. The worker intentionally disconnects every 3
  minutes (deadline exceeded, retrying subscription) and reconnects (listener established). That's working as designed — it prevents
   Azure's 4-minute proxy kill from causing a silent drop.

  The problem is what happens to tasks dispatched during the reconnect window. There's a ~1-3 second gap between deadline exceeded
  and listener established. If the engine dispatches a task during that gap:

  1. Engine marks the task as RUNNING and assigns it
  2. The gRPC stream is dead — the worker never receives the action
  3. The worker reconnects with a fresh listener, but the engine already "sent" the task
  4. The task sits in RUNNING forever — no worker is executing it, and the engine thinks it was delivered

  This is a fundamental race condition in Hatchet's dispatch model — there's no acknowledgment/nack protocol. The engine
  fires-and-forgets over the stream.
  ```
  
  So it looks like I have some investigation to do.  Thanks!

PR #3279, opened 2026-03-14
Gracefully handle empty bulk scheduled deletes

### LINKED ISSUE #3269: [FEAT] Gracefully handle missing ids in bulk_delete_scheduled
**Problem Summary**

In our hatchet usage, we rely on scheduled runs + bulk deleting as a means of implementing "dedupe+debounce" workflows. For example, we have a chat product that schedules a Hatchet workflow 15 minutes in the future keyed on a user id. If the user engages with chat again, we'll use `bulk_delete_scheduled` to delete (cancel) scheduled workflows keyed on that user id via `additional_metadata` and reschedule for 15 minutes in the future. 

When we call `bulk_delete_scheduled` with this user id in `additional_metadata` and no matching workflow IDs are found, the engine returns a 400 error with message `Provide scheduledWorkflowRunIds or filter.`
[link](https://github.com/hatchet-dev/hatchet/blob/344d7fa00fcbda729f7ae561c840de8e05c4e542/api/v1/server/handlers/workflows/bulk_delete_scheduled.go#L113)
```
if len(ids) == 0 {
	return gen.WorkflowScheduledBulkDelete400JSONResponse(apierrors.NewAPIErrors("Provide scheduledWorkflowRunIds or filter.")), nil
}
```

**Potential Solutions**
There are a few solutions that come to mind, and I'm open to hearing the maintainers' feedback.

1. At minimum, the error message is a bit misleading. It seems to indicate a malformed request payload (no `scheduledWorkflowRunIds` or filter) but is actually a runtime condition.
2. Rather than a 400, this condition could actually result in a 200 response, that is, "You wanted to delete something that doesn't exist, and it doesn't exist, so voila". This could be coupled with a `deleted_ids` response payload for clients to validate their own behavior. 
3.  Perhaps there's a request level flag `tolerate_missing_ids` which allows a client to explicitly opt-in to this behavior.

It's worth noting that the API technically does offer a pathway here to 
1. List scheduled workflows based on filter
2. Cancel based on ID if filter match
This is also reasonable, but could be subject to race conditions (net of other client application-level changes) in a concurrent execution environment where a separate thread creates workflows in between steps 1 and 2.

My personal lean is towards option 2 (don't treat this as a 400 error), although not too strongly held.

Happy to hear the maintainers' thoughts, and should be able to contribute a PR based on feedback.

avirajkhare00:
## Summary
- return 200 for filter-based bulk scheduled deletes when the filter matches no scheduled runs
- preserve per-item errors when a filter only finds non-API scheduled runs
- add targeted handler tests for empty filter matches and the missing ids/filter validation path

Fixes #3269

## Testing
- go test ./api/v1/server/handlers/workflows/...


api/v1/server/handlers/workflows/service.go
```diff
@@ -2,14 +2,22 @@ package workflows
...
+	if config != nil && config.V1 != nil {
+		workflowSchedules = config.V1.WorkflowSchedules()
+	}
+
 	return &WorkflowService{
-		config: config,
+		config:            config,
+		workflowSchedules: workflowSchedules,
 	}
 }
```
mrkaye97: let's not do this - I agree that this is a clean approach, but we don't have this pattern anywhere else so I'd prefer to be consistent

mrkaye97 (commented):
thanks for this change! I left one small comment on existing patterns, but broadly looks great 👍

api/v1/server/handlers/workflows/service.go
avirajkhare00: Updating PR

api/v1/server/handlers/workflows/bulk_delete_scheduled_test.go
```diff
@@ -0,0 +1,188 @@
+package workflows
```
mrkaye97: sorry for the double review 🤦 I should've flagged before, I don't think we really need to add all these tests, since it's such a small API change

avirajkhare00: shall i drop this test completely?

mrkaye97: go for it

avirajkhare00: done

mrkaye97 (approved):
great stuff! will let CI pass and then merge - thanks for the fix!

PR #3281, opened 2026-03-15
feat: add opt-in ListenV2 stream keepalive to prevent L7 proxy idle timeout

### LINKED ISSUE #3280: ListenV2 streams killed by L7 proxies with stream_idle_timeout (Azure Container Apps, Envoy)
## Problem

When running Hatchet behind an L7 proxy that enforces per-stream idle timeouts (e.g., Azure Container Apps' managed Envoy with a 4-minute `stream_idle_timeout`), the `ListenV2` gRPC stream is killed after a period of inactivity when no tasks are being dispatched.

### Root Cause

`ListenV2` is a server-streaming RPC that only sends `AssignedAction` messages when tasks are dispatched. When the queue is idle, **zero data frames flow on the stream**. L7 proxies like Envoy track idle time per HTTP/2 stream and terminate streams that have no DATA/HEADERS activity.

gRPC keepalive PING frames don't help because:
- They are **connection-level**, not stream-level
- Envoy responds to PINGs locally at the proxy hop and does not forward them ([envoyproxy/envoy#5142](https://github.com/envoyproxy/envoy/issues/5142))
- Azure Container Apps confirmed this as wontfix: [microsoft/azure-container-apps#113](https://github.com/microsoft/azure-container-apps/issues/113)

The worker's 4-second `Heartbeat` unary RPC also doesn't help because it creates separate HTTP/2 streams — it doesn't put data on the `ListenV2` stream.

### Impact

When the stream is killed:
1. Engine dispatches a task to the dead stream
2. Task is marked RUNNING but the worker never receives it
3. The worker reconnects, but the already-dispatched task is orphaned
4. Task stays in RUNNING forever — the reassignment logic (`ProcessTaskReassignments`) only catches tasks assigned to workers with expired heartbeats, not tasks with no worker ID

### Workaround Attempted

Setting `HATCHET_CLIENT_LISTENER_V2_TIMEOUT=180` causes the worker to proactively reconnect every 3 minutes (before the 4-minute proxy kill). This reduces silent stream death but creates a **1-3 second gap** during reconnection where dispatched tasks are lost.

## Architecture

```
Workers (Azure Container Apps) 
    → Azure Envoy Proxy (4-min stream_idle_timeout, not configurable)
    → Hatchet Engine (Azure Container Apps, same environment)
```

- Engine and workers are in the same Azure Container Apps environment
- Internal ingress uses HTTP/2 transport
- Azure Container Apps' managed Envoy proxy sits between all container apps (even internal traffic)
- The Envoy `stream_idle_timeout` is **not user-configurable** on Azure Container Apps (without premium ingress at ~$750/mo)
- TCP transport on ACA internal ingress does not work (connection timeouts)

## Context: Why ListenV2 Removed Server-Side Heartbeats

PR #308 moved heartbeats from server-side (on the Listen stream) to client-side (separate Heartbeat RPC) because proxies were **keeping dead worker connections alive** — the server heartbeats flowed through the proxy, masking worker death.

This was the right call. But it means the `ListenV2` stream has zero traffic when idle, making it vulnerable to proxy stream idle timeouts. The two problems are opposite sides of the same coin:

| | Listen v1 (server heartbeats) | ListenV2 (client heartbeats) |
|---|---|---|
| Dead worker detection | Proxy masks death (bad) | Client heartbeat stops immediately (good) |
| Stream idle timeout | Stream stays alive (good) | Proxy kills idle stream (bad) |

### Option A: Periodic no-op message on ListenV2 stream (engine-side)

Send a lightweight "keepalive" `AssignedAction` (or a new message type) every 30-60 seconds on the `ListenV2` stream. The SDK would ignore it.

This resets the proxy's stream idle timer without reintroducing the PR #308 problem, because:
- Worker liveness is still detected via the separate `Heartbeat` RPC (unaffected)
- The no-op stream message only prevents the proxy from killing the stream — it doesn't signal worker health to the engine
- If the worker dies, the `Heartbeat` RPC stops and the engine detects it normally

### Option B: Make-before-break dual streams (SDK-side)

The SDK could maintain two overlapping `ListenV2` streams with staggered lifetimes. When one approaches the proxy timeout, a new stream is opened before the old one is closed. This eliminates the reconnection gap.

This is more complex (fan-in, deduplication) and doesn't solve the root cause.

### Option C: Configurable server-side stream keepalive

Add a `SERVER_GRPC_STREAM_KEEPALIVE_INTERVAL` config option (default: disabled) that sends periodic no-op messages on active `ListenV2` streams. Users behind L7 proxies with stream idle timeouts can enable it. Users connecting directly (or with properly configured proxies) leave it disabled.

## Environment

- Hatchet Engine: v0.79.25 (self-hosted)
- Python SDK: latest
- Infrastructure: Azure Container Apps (managed Envoy proxy)
- gRPC keepalive: client 10s, server 30s (confirmed ineffective against stream_idle_timeout)

## Related Issues

- #308 — PR that moved heartbeats to client-side (created this vulnerability)
- #726 — Fix for reverting failed gRPC stream assignments (helps but doesn't catch proxy-killed streams)
- #1996 — Runs stuck in RUNNING after proxy issues
- #2432 — Task assigned after heartbeat timeout (related stream interruption)
- [microsoft/azure-container-apps#113](https://github.com/microsoft/azure-container-apps/issues/113) — Azure wontfix for HTTP/2 stream timeout
- [envoyproxy/envoy#5142](https://github.com/envoyproxy/envoy/issues/5142) — Envoy stream_idle_timeout breaks gRPC streaming

shaunpatterson:
## Summary

Adds an opt-in server-side stream keepalive for `ListenV2` to prevent L7 proxies from killing idle gRPC streams. When enabled, the engine sends periodic `STREAM_KEEPALIVE` messages (no-op `AssignedAction`) on idle streams, generating DATA frames that reset the proxy's stream idle timer.

Fixes #3280

## Problem

L7 proxies like Envoy (Azure Container Apps, AWS ALB, etc.) enforce per-stream idle timeouts (~4 minutes on ACA). When no tasks are being dispatched, the `ListenV2` stream has zero DATA frames and gets killed. gRPC keepalive PINGs don't help because they're connection-level and proxies respond to them locally ([envoyproxy/envoy#5142](https://github.com/envoyproxy/envoy/issues/5142), [microsoft/azure-container-apps#113](https://github.com/microsoft/azure-container-apps/issues/113)).

## Why this doesn't reintroduce #308

PR #308 removed server-side heartbeats from the stream because proxies were keeping dead worker connections alive — the server heartbeats flowed through the proxy, masking worker death.

This change is different:
- **Worker liveness is still determined solely by the separate Heartbeat RPC** (unchanged)
- The STREAM_KEEPALIVE message only prevents the proxy from killing the stream
- If a worker dies, the Heartbeat RPC stops and the engine detects it within 30 seconds regardless of stream state
- The engine does NOT use stream write-success as a liveness signal

## Configuration

```bash
SERVER_LISTENV2_STREAM_KEEPALIVE_INTERVAL=120s  # disabled by default (0s)
```

## Client Impact

Old SDKs receive ActionType=3 but don't match any handler — they skip it. No client changes required.

## Note

The dispatcher.pb.go was manually updated. Proper protobuf regeneration should be done before merging.

### DIFF
--- a/api-contracts/dispatcher/dispatcher.proto
+++ b/api-contracts/dispatcher/dispatcher.proto
@@ -124,6 +124,11 @@ enum ActionType {
     START_STEP_RUN = 0;
     CANCEL_STEP_RUN = 1;
     START_GET_GROUP_KEY = 2;
+    // STREAM_KEEPALIVE is a no-op message sent periodically on the ListenV2
+    // stream to prevent L7 proxies (e.g., Envoy) from killing the stream due
+    // to idle timeout. Clients should ignore this action type. Worker liveness
+    // is still determined by the separate Heartbeat RPC.
+    STREAM_KEEPALIVE = 3;
 }
 
 message AssignedAction {
--- a/cmd/hatchet-engine/engine/run.go
+++ b/cmd/hatchet-engine/engine/run.go
@@ -345,6 +345,7 @@ func runV0Config(ctx context.Context, sc *server.ServerConfig, cleanup *cleanup.
 			dispatcher.WithDefaultMaxWorkerLockAcquisitionTime(sc.Runtime.GRPCMaxWorkerLockAcquisitionTime),
 			dispatcher.WithWorkflowRunBufferSize(sc.Runtime.WorkflowRunBufferSize),
 			dispatcher.WithStreamEventBufferTimeout(sc.Runtime.StreamEventBufferTimeout),
+			dispatcher.WithListenV2StreamKeepaliveInterval(sc.Runtime.ListenV2StreamKeepaliveInterval),
 			dispatcher.WithVersion(sc.Version),
 			dispatcher.WithAnalytics(sc.Analytics),
 		)
@@ -772,6 +773,7 @@ func runV1Config(ctx context.Context, sc *server.ServerConfig, cleanup *cleanup.
 			dispatcher.WithDefaultMaxWorkerLockAcquisitionTime(sc.Runtime.GRPCMaxWorkerLockAcquisitionTime),
 			dispatcher.WithWorkflowRunBufferSize(sc.Runtime.WorkflowRunBufferSize),
 			dispatcher.WithStreamEventBufferTimeout(sc.Runtime.StreamEventBufferTimeout),
+			dispatcher.WithListenV2StreamKeepaliveInterval(sc.Runtime.ListenV2StreamKeepaliveInterval),
 			dispatcher.WithVersion(sc.Version),
 			dispatcher.WithAnalytics(sc.Analytics),
 		)
--- a/internal/services/dispatcher/contracts/dispatcher.pb.go
+++ b/internal/services/dispatcher/contracts/dispatcher.pb.go
@@ -82,6 +82,7 @@ const (
 	ActionType_START_STEP_RUN      ActionType = 0
 	ActionType_CANCEL_STEP_RUN     ActionType = 1
 	ActionType_START_GET_GROUP_KEY ActionType = 2
+	ActionType_STREAM_KEEPALIVE    ActionType = 3
 )
 
 // Enum value maps for ActionType.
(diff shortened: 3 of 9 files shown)


promptless-for-oss:

New suggestion: **[Document SERVER_LISTENV2_STREAM_KEEPALIVE_INTERVAL for L7 proxy support](https://app.gopromptless.ai/suggestions/9426ef82-8a6e-4768-97b5-bc9d6f9f75da)**

---
_Tip: Whenever you leave a comment tagged `@Promptless` on a Promptless PR, Promptless will remember it for future suggestions 🧠_

PR #3290, opened 2026-03-16
Remove dispatch backlog, replace with timeout lock acquisition

juliusgeo:
# Description

Prior to this change, there was a fixed backlog of roughly 20 simultaneous worker dispatches before they would start erroring out to be rescheduled. This was suboptimal as it was adding an additional buffer on top of the existing gRPC and TCP flow control buffers that exist internally. The `SendMsg` call will only block once 1) TCP buffer is exhausted 2) gRPC buffer is exhausted, thus making the additional worker backlog only activate once 19 additional worker sends were queued *after* both of those buffers were exhausted. This change makes it so that a timeout controls whether we send tasks back to the scheduler by checking how long it takes to acquire the lock surrounding SendMsg. 

Fixes # (issue)

## Type of change

## What's Changed

- [x] Removed maxWorkerBacklogSize
- [x] Adds WorkerLockAcquisitionTimeout


promptless-for-oss:

New suggestion: **[Update flow control env var from backlog size to lock acquisition timeout](https://app.gopromptless.ai/suggestions/0c6b61d6-0bf2-43a8-8890-93202494b897)**

---
_Tip: See how your feedback shapes Promptless in [Agent Knowledge Base](https://app.gopromptless.ai/configure/settings) 🧠_

internal/services/dispatcher/subscribed_worker_v1.go
```diff
@@ -19,6 +19,17 @@ import (
...
 )
 
+func (worker *subscribedWorker) tryAcquireSendLockWithTimeout(timeout time.Duration) bool {
+	stopTime := time.Now().Add(timeout)
+	for time.Now().Before(stopTime) {
+		if worker.sendMu.TryLock() {
+			return true
+		}
+		time.Sleep(5 * time.Millisecond) // small backoff to avoid busy spinning
+	}
```
abelanger5: I don't love this pattern for a few reasons:
1. It doesn't respect ordering, so workers can be crowded out if we happen to call `TryLock` from a different task send
2. It feels unpredictable what sort of CPU load we'll see from TryLock running this often

Is there a way we can implement the worker mutex as a semaphore on a channel instead, and have a channel call using `time.After`, and then have something like:

```go
select {
  <- worker.sendSemaphore:
  <- time.After:
}
```

(Not real code but hopefully clear what I mean)

The semaphore can be guarded on either send or receive. 

This should be efficient and also respect ordering?

pkg/config/server/server.go
```diff
@@ -185,9 +185,9 @@ type ConfigFileRuntime struct {
 	// GRPCMaxMsgSize is the maximum message size that the grpc server will accept
 	GRPCMaxMsgSize int `mapstructure:"grpcMaxMsgSize" json:"grpcMaxMsgSize,omitempty" default:"4194304"`
 
-	// GRPCWorkerStreamMaxBacklogSize is the maximum number of messages that can be queued for a worker before we start rejecting new
-	// messages. Default is 20.
-	GRPCWorkerStreamMaxBacklogSize int `mapstructure:"grpcWorkerStreamMaxBacklogSize" json:"grpcWorkerStreamMaxBacklogSize,omitempty" default:"20"`
+	// GRPCWorkerMaxLockAcquisitionTimeMS is the maximum number of milliseconds that the dispatcher will wait while attempting
+	// to send messages to workers. If it waits longer, the request will be rejected. Default is 250
+	GRPCWorkerMaxWorkerLockAcquisitionTimeMS int `mapstructure:"grpcWorkerMaxLockAcquisitionTimeMS" json:"grpcWorkerMaxLockAcquisitionTimeMS,omitempty" default:"250"`
```
abelanger5: Would it make more sense for this to be a `time.Duration`? I can't actually remember if viper supports unmarshalling into a `time.Duration` directly or if we need to parse a string. I think it might have the advantage of making the config slightly more readable

juliusgeo: Looks like Viper does support time.Duration unmarshalling. Fixed!

internal/services/dispatcher/subscribed_worker_v1.go
juliusgeo: That makes sense. I changed it so it uses a chan as a semaphore.

internal/services/dispatcher/subscribed_worker_v1.go
```diff
@@ -19,6 +19,21 @@ import (
...
+		return true
+	// timing out dequeues the semaphore send
+	case <-time.After(timeout):
+		return false
+	}
+}
+
+func (worker *subscribedWorker) releaseSendLock() {
+	<-worker.sendSemaphore
+}
```
abelanger5: this looks much cleaner, nice!

abelanger5 (approved):
nice!

internal/services/dispatcher/subscribed_worker_v1.go
```diff
@@ -19,6 +19,21 @@ import (
 	"github.com/hatchet-dev/hatchet/pkg/telemetry"
 )
 
+func (worker *subscribedWorker) tryAcquireSendLockWithTimeout(timeout time.Duration) bool {
+	select {
+	// attempt to send to the semaphore, blocks on contention because it has a buffer of 1
+	case worker.sendSemaphore <- struct{}{}:
+		return true
+	// timing out dequeues the semaphore send
+	case <-time.After(timeout):
```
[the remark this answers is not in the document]
abelanger5: This advice might be outdated:

> Before Go 1.23, this documentation warned that the underlying [Timer](https://pkg.go.dev/time#Timer) would not be recovered by the garbage collector until the timer fired, and that if efficiency was a concern, code should use NewTimer instead and call [Timer.Stop](https://pkg.go.dev/time#Timer.Stop) if the timer is no longer needed. As of Go 1.23, the garbage collector can recover unreferenced, unstopped timers. There is no reason to prefer NewTimer when After will do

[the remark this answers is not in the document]
abelanger5: But perhaps it's not a bad idea for the worker to manage a single timer and to cleanup the timer when we return from this method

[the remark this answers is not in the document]
juliusgeo: Would that work with multiple goroutines at the same time? It's slightly unclear from the documentation.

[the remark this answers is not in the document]
abelanger5: hmm, yeah I'm not entirely sure. can we try writing a Go benchmark to see how it performs? if we can do at least ~50k/second without much issue I think we'll be fine

abelanger5 (approved):

promptless-for-oss:

Updated existing suggestion: **[Update flow control env var from backlog size to lock acquisition timeout](https://app.gopromptless.ai/suggestions/0c6b61d6-0bf2-43a8-8890-93202494b897)**

---
_Tip: Worried about broken links? Ask Promptless to find and fix them automatically 🔗_

PR #3298, opened 2026-03-17
fix: silence tenant alert error

grutt:
# Description

Treats no row as inactive alert group.

```
2026-03-17T12:36:43.549Z ERR could not process tenant alerts error="could not check if tenant is active: no rows in result set" service=olap-controller
```

Fixes # (issue)

## Type of change

- [x] Chore (changes which are not directly related to


mrkaye97 (approved):
any risk here? not too familiar

PR #3318, opened 2026-03-18
fix: cleanup cleanup_workflow_concurrency_slots bug

grutt:
# Description

cleanup_workflow_concurrency_slots has a CARDINALITY(child_strategy_ids) <= COUNT(tasks) guard that prevents premature cleanup during DAG execution. But when cancelled tasks are purged from v1_task, the count drops to 0 and CARDINALITY <= 0 is always false -- so the workflow concurrency slot is never deleted.

## Type of change


ashsidhu:
Hi @grutt — we're hitting this exact bug in production.

Dispatch N DAG workflows with step-level concurrency — first batch completes the full pipeline, remaining workflows never get scheduled. Slots from completed workflows aren't freed. Re-dispatching as new workflows works immediately.

Would love to see this land — happy to test a pre-release build if that helps move it along.

PR #3321, opened 2026-03-18
Separate out partition cleanup helpers for visibility

juliusgeo:
# Description

Partition cleanup also cleans up a few cron jobs, which can block if there are in-flight tasks. Separating these out into separate cleanup functions provides more visibility in logs as to what is actually taking significant time.

Fixes # (issue)

## Type of change

## What's Changed

### DIFF
--- a/cmd/hatchet-engine/engine/run.go
+++ b/cmd/hatchet-engine/engine/run.go
@@ -152,8 +152,6 @@ func runV0Config(ctx context.Context, sc *server.ServerConfig, cleanup *cleanup.
 		return fmt.Errorf("could not create partitioner: %w", err)
 	}
 
-	p.Shutdown(cleanup)
-
 	var h *health.Health
 	healthProbes := sc.HasService("health")
 	if healthProbes {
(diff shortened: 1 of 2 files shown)


cmd/hatchet-engine/engine/run.go
```diff
@@ -152,10 +152,7 @@ func runV0Config(ctx context.Context, sc *server.ServerConfig, cleanup *cleanup.
 		return fmt.Errorf("could not create partitioner: %w", err)
 	}
 
-	cleanup.Add(
-		p.Shutdown,
-		"partitioner",
-	)
+	p.Shutdown(cleanup)
```
abelanger5: The naming here is super confusing and it seems like the callers are inverted -- perhaps we should get the cleanup methods back from the partitioner using something like `p.GetCleanupMethods` and then pass those into `cleanup.Add`?

It doesn't make sense to me for the partitioner to have a dependency on the cleanup object

abelanger5: If we want to leave it like this calling it `AddCleanupMethods` instead of `Shutdown` would also work

juliusgeo: I just changed the name--we only call the shutdown function from these two call sites

abelanger5 (approved):

PR #3333, opened 2026-03-19
Feat  o11y UI changes

grutt:
# Description

Mostly Ui and Bug Fixes
## Type of change


internal/services/controllers/olap/engine_spans.go
```diff
@@ -0,0 +1,292 @@
...
+	if len(events) == 0 {
+		return nil
+	}
+
+	taskIds := make([]int64, len(events))
+	for i, e := range events {
+		taskIds[i] = e.taskID
+	}
+
+	startedRows, err := tc.repo.OLAP().GetTaskStartedTimestamps(ctx, tenantId, taskIds)
```
abelanger5: This is likely a race condition since the started events can arrive out-of-order from the terminal events

internal/services/controllers/olap/engine_spans.go
```diff
@@ -0,0 +1,292 @@
...
+		resourceAttrs, _ := json.Marshal(map[string]string{
+			"service.name": "hatchet-engine",
+		})
+
+		spans = append(spans, &v1.SpanData{
+			TenantID:             tenantId,
+			TraceID:              traceIDBytes,
+			SpanID:               spanIDBytes,
+			ParentSpanID:         parentBytes,
+			Name:                 "hatchet.start_step_run",
```
abelanger5, line 256: We're sure we want to be using `step_run` here and everywhere else? Is this to preserve some kind of compatibility somewhere?

pkg/repository/sqlcv1/olap.sql
```diff
@@ -1894,6 +1894,20 @@ LEFT JOIN
...
+SELECT
+    task_id,
+    task_inserted_at,
+    retry_count,
+    MIN(event_timestamp)::timestamptz AS started_at
+FROM
+    v1_task_events_olap
+WHERE
+    tenant_id = @tenantId::uuid
+    AND task_id = ANY(@taskIds::bigint[])
```
abelanger5: This feels like it needs to be scoped to the retry count?

internal/services/controllers/olap/engine_spans.go
```diff
@@ -0,0 +1,292 @@
...
+	otelRepo := tc.repo.OTelCollector()
+	if otelRepo == nil || len(events) == 0 {
+		return
+	}
+
+	var queuedSpans []*v1.SpanData
+	var terminalEvents []engineSpanEvent
+
+	for i := range events {
+		e := &events[i]
```
abelanger5: We're doing this type of assignment in a number of places - we're also very inconsistent about the use of pointers in slices here. Why don't we stick with one (pointers or no pointers) and just use var assignment in the loop, `for _, e := range events`?

internal/services/controllers/olap/engine_spans.go
```diff
@@ -0,0 +1,292 @@
...
+
+		switch e.eventType {
+		case sqlcv1.V1EventTypeOlapSENTTOWORKER:
```
abelanger5, line 58: We have plenty of other event types, do we not want to add those? Could imagine it'd be super useful to know we're rate limited or in a concurrency queue...

grutt: i think we're going to need to write these as otel events since they will have a queued span already

internal/services/controllers/olap/engine_spans.go
grutt: compat on the pysdk

pkg/repository/sqlcv1/olap.sql
grutt: gc

abelanger5: this is not matching tuples, it's matching any retry count in the array. we need to be matching on `(task_id, retry_count) =`

internal/services/controllers/olap/engine_spans.go
abelanger5: If this is a race condition and we're merging with in-memory spans to prevent it, one of the following must be true:
1. The DB call is unnecessary, or
2. The in-memory is potentially incorrect or inaccurate

?

pkg/repository/sqlcv1/olap.sql
mrkaye97: just flagging too - probably better to match `(task_id, task_inserted_at, retry_count)` to prune partitions

PR #3347, opened 2026-03-20
feat: add server-side column sorting to task runs table

shaunpatterson:
## Summary
- Adds server-side sorting for **Task Name**, **Workflow**, **Created At**, **Started At**, **Finished At**, and **Duration** columns in the workflow runs list view
- New `order_by_field` and `order_by_direction` query parameters on the `v1-workflow-run:list` API endpoint
- Dynamic ORDER BY with whitelisted column mappings and LATERAL joins for event-derived fields (started_at, finished_at, duration)
- Frontend wires TanStack React Table sorting state through to the API, with type-safe field mapping and single-column sort only

## Changes across layers
- **OpenAPI spec**: Added `order_by_field` (enum) and `order_by_direction` (asc/desc) params
- **Go handler**: Extracts sort params in both `WithDags` and `OnlyTasks` paths
- **SQL**: Dynamic query builder with LATERAL joins for started_at/finished_at/duration, deterministic tie-breakers for stable pagination
- **Frontend**: Enabled sorting on 6 columns, default sort `createdAt DESC`, disabled multi-sort

## Test plan
- [ ] Verify sorting by each column (asc/desc) produces correct order
- [ ] Verify pagination remains stable across pages when sorting by non-unique columns
- [ ] Verify sorting by Started At / Finished At / Duration correctly handles NULL values (queued/running tasks)
- [ ] Verify default sort (Created At DESC) matches previous behavior
- [ ] Verify WithDags view falls back to createdAt sorting gracefully

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/api-contracts/openapi/paths/v1/workflow-runs/workflow_run.yaml
+++ b/api-contracts/openapi/paths/v1/workflow-runs/workflow_run.yaml
@@ -113,6 +113,28 @@ listWorkflowRuns:
         required: false
         schema:
           $ref: "../../../components/schemas/_index.yaml#/V1RunningFilter"
+      - description: The field to sort results by
+        in: query
+        name: order_by_field
+        required: false
+        schema:
+          type: string
+          enum:
+            - createdAt
+            - taskName
+            - workflow
+            - startedAt
+            - finishedAt
+            - duration
+      - description: The direction to sort results
+        in: query
+        name: order_by_direction
+        required: false
+        schema:
+          type: string
+          enum:
+            - asc
+            - desc
     responses:
       "200":
         content:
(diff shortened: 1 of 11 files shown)


mrkaye97:
hey @shaunpatterson, thanks for the suggestion here! unfortunately, I think this is a little more complex of a problem than this, because at scale, we need to be relying on primary keys or other indices for ordering and filtering and need the planner to be able to make intelligent decisions on tables with potentially hundreds of millions of rows (both for the joins and for where + order by clauses).

For instance, I ran the simplest and likely most common case here (no filters applied, order by finished at descending) against our production database, and got a terrible query plan back that took ~45 seconds to run, due to some sequential scans we have no choice but to do. I suspect, unfortunately, this might get worse with more complicated filters or ordering conditions.

You can pretty likely experiment with this if you dump a few million (maybe some more, not too sure) runs into the database and then explore how query performance changes as you apply different ordering and where conditions

mrkaye97:
going to close this as I don't think it's feasible

PR #3322, opened 2026-03-18
Fix: Add tenant id to context on v0 sdk

mrkaye97:
# Description

Adding this so we can use it on the internal worker

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

---

> [!NOTE]
> **Low Risk**
> Low risk: adds a simple accessor to expose `TenantId` from the existing `client.Action` payload and updates a test stub to satisfy the expanded interface.
> 
> **Overview**
> Adds `TenantId()` to the `HatchetContext` interface and implements it on `hatchetContext` by returning `action.TenantId` (marked deprecated like other v0 SDK internals).
> 
> Updates the middleware tests’ `testHatchetContext` stub to implement the new method so the test suite continues to compile.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5ec1bea3bf0e620b1550b1b3aa70e6b2e2636d86. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #3349, opened 2026-03-20
fix(sdk/go): Fallback when tenantID is zero UUID

gregfurman:
# Description

This PR checks if the supplied Tenant ID is either empty `""` or the zero `uuid.UUID` value `00000000-0000-0000-0000-000000000000` -- where the only the former was being checked which causing tenant IDs not to be extracted from supplied tokens with `v1.NewHatchetClient(...)` calls.

## Type of change

## What's Changed

- When a tenant ID is the zero value for a `uuid.UUID`, this triggers a fallback to extract from the passed in token.

## Testing

Using the below `client.yaml`, run the go script both with and without the change.

```yaml
token: "<token>"
log:
  level: debug
tls:
  base:
    tlsStrategy: none

```

```go
package main

import (
	"fmt"
	"os"

	clientloader "github.com/hatchet-dev/hatchet/pkg/client/loader"
	v1 "github.com/hatchet-dev/hatchet/pkg/v1"
)

func main() {
	configBytes, err := os.ReadFile("client.yaml")
	if err != nil {
		panic(err)
	}

	cf, err := clientloader.LoadClientConfigFile(configBytes)
	if err != nil {
		panic(err)
	}

	client, err := v1.NewHatchetClient(v1.Config{
		Token: cf.Token,
		TLS: &v1.TLSConfig{
			Base:          &cf.TLS.Base,
			TLSServerName: cf.TLS.TLSServerName,
		},
	})
	if err != nil {
		panic(err)
	}

	fmt.Printf("TenantID: %s\n", client.V0().TenantId())
}
```

You'll see HEAD of `main` return:

```
TenantID: 00000000-0000-0000-0000-000000000000
```

with these changes added:

```
TenantID: 707d0855-80ab-4e1f-a156-f1c4546cbf52 # or whatever tenant ID was used in your token generation
```

### DIFF
--- a/pkg/client/loader/loader.go
+++ b/pkg/client/loader/loader.go
@@ -6,6 +6,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/google/uuid"
+
 	"github.com/hatchet-dev/hatchet/pkg/config/client"
 	"github.com/hatchet-dev/hatchet/pkg/config/loader/loaderutils"
 )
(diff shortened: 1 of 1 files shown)


mrkaye97 (approved):
🙌

gregfurman:
This PR supersedes https://github.com/hatchet-dev/hatchet/pull/3322

PR #3357, opened 2026-03-21
Feat: Lookup table for trace ids

mrkaye97:
# Description

Adds a lookup table for trace ids which lets us grab the trace from an external id by querying the latest retry count and then filtering. Basically just lets us avoid bad CTEs in Clickhouse, and also simplifies reads on the API a bit

## Type of change


pkg/repository/otelcol.go
```diff
@@ -65,7 +66,22 @@ type ListSpansResult struct {
...
+}
+
+type otelLookupRepositoryImpl struct {
+	*sharedRepository
+}
+
+func newOTelLookupRepository(s *sharedRepository) OTelLookupRepository {
+	return &otelLookupRepositoryImpl{
+		sharedRepository: s,
+	}
```
mrkaye97, line 84: added this separate repo because we override the main OtelCollectorRepository, so this seemed like the next best option 🤷

pkg/repository/otelcol.go
```diff
@@ -178,25 +202,123 @@ func (o *otelCollectorRepositoryImpl) CreateSpans(ctx context.Context, tenantId
...
+
+	if err := commit(ctx); err != nil {
+		return fmt.Errorf("error committing transaction: %w", err)
+	}
+
+	return nil
+}
+
+func (o *otelLookupRepositoryImpl) CreateSpanLookupTableEntries(ctx context.Context, tenantId uuid.UUID, opts *CreateSpansOpts) error {
```
mrkaye97, line 217: don't love this not being tx-safe but I think it's fine

api/v1/server/handlers/v1/observability/traces.go
```diff
@@ -52,7 +37,17 @@ func (t *V1ObservabilityService) V1ObservabilityGetTrace(ctx echo.Context, reque
...
+
+	if errors.Is(err, pgx.ErrNoRows) {
+		return gen.V1ObservabilityGetTrace404JSONResponse(gen.APIErrors{
+			Errors: []gen.APIError{{Description: "Trace not found"}},
+		}), nil
+	} else if err != nil {
+		return nil, err
+	}
+
+	result, err := t.config.V1.OTelCollector().ListSpansByTraceId(ctx.Request().Context(), tenant.ID, traceId, offset, limit)
```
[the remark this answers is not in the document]
mrkaye97: @grutt is this the thing?

abelanger5 (approved):

PR #3365, opened 2026-03-23
Remove "account" dropdown', add notifications dropdown

TehShrike:
# Description

CleanShot 2026-03-23 at 10 07 46@2x

CleanShot 2026-03-26 at 09 39 24@2x

CleanShot 2026-03-26 at 09 40 13@2x

CleanShot 2026-03-26 at 09 37 08@2x

- "Logout"  and the theme changer are in the side nav now.
	- The theme changer still only toggles between light and dark, though passing the theme types through the nav component did push me to change use-theme slightly in a way that will make it more straightforward to change the theme to "system" in the future.
	- The side-nav items can have an `onClick` instead of a `to` (href) now
- there are three severity colors: green (invitations), yellow (tenant resource usage >= the alarm value), red (tenant resource useage >= the limit)
- New invitations no longer cause you to be immediately redirected to the invitation screen (unless you don't belong to an organization/tenant and have been invited to one)

Note: the resource limits for all tenants are being fetched right now, which requires a request per tenant.  We should probably add an endpoint that lets you see your limits for all tenants.  We could make it so that you only see notifications for limits on the current active tenant, but that seems weird to me, since the top nav/notification UI feels "global", and presumably most users are responsible for all the tenants they belong to to some degree.

## Type of change

## What's Changed


promptless-for-oss:

New suggestion: **[Document dashboard notifications dropdown for invites and resource limits](https://app.gopromptless.ai/suggestions/8ab01f78-08b7-4e6e-8ce7-bf6dba8cbdc3)**

---

frontend/app/src/components/hooks/use-theme.tsx
```diff
@@ -1,56 +1,63 @@
...
+  currentlyVisibleTheme: 'light',
 };
 
 const ThemeProviderContext = createContext<ThemeProviderState>(initialState);
 
+const getThemeToDisplay = (theme: SelectableTheme): DisplayableTheme => {
+  if (theme === 'system') {
+    return window.matchMedia('(prefers-color-scheme: dark)').matches
+      ? 'dark'
+      : 'light';
```
mrkaye97, line 32: thanks for the fix here!

frontend/app/src/components/v1/nav/notifications.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const { currentUser } = useCurrentUser();
+  const navigate = useNavigate();
+  const count = notifications.length;
+  const mostSevere = count > 0 ? getMostSevereColor(notifications) : null;
+  const title = `${count} notification${count !== 1 ? 's' : ''} for ${currentUser?.email ?? ''}`;
+
+  return (
+    <DropdownMenu>
+      <DropdownMenuTrigger asChild>
+        <button
```
mrkaye97: we should use a shadcn button here if we can (no worries if it's too much of an issue to do that while also matching the designs)

frontend/app/src/hooks/notifications/index.ts
```diff
@@ -0,0 +1,23 @@
...
+  const invites = useInviteNotifications();
+
+  const notifications = useMemo(
+    () =>
+      [...resourceLimits.notifications, ...invites.notifications].sort(
+        (a, b) => (b.timestamp > a.timestamp ? 1 : -1),
+      ),
+    [resourceLimits.notifications, invites.notifications],
+  );
+
```
mrkaye97, line 26: this is great 👍 should make it really easy for us to plug additional stuff

frontend/app/src/hooks/notifications/resource-limits.ts
```diff
@@ -0,0 +1,92 @@
...
+const TWO_MINUTES_MS = 2 * 60_000;
+
+const resourceLabels: Record<TenantResource, string> = {
+  [TenantResource.WORKER]: 'Total Workers',
+  [TenantResource.WORKER_SLOT]: 'Concurrency Slots',
+  [TenantResource.EVENT]: 'Events',
+  [TenantResource.TASK_RUN]: 'Task Runs',
+  [TenantResource.CRON]: 'Cron Triggers',
+  [TenantResource.SCHEDULE]: 'Schedule Triggers',
+  [TenantResource.INCOMING_WEBHOOK]: 'Incoming Webhooks',
```
mrkaye97: small thing: as I refactor more stuff in the codebase, I've been trying to use switch statements with an exhaustiveness check for this sort of thing as opposed to a map, so we don't need to handle the not found case implicitly. e.g.:


```suggestion
const resourceTypeToLabel = (resource: TenantResource): string => {
  switch (resource) {
    case ...
    ...
    
    default:
      const exhaustive: never = resource; // should never be non-never
      throw ...
  }
}
```

(typed this in Github, not sure if it'd compile but you get the idea :P )

frontend/app/src/hooks/notifications/resource-limits.ts
```diff
@@ -0,0 +1,92 @@
...
+    queries: tenants.map((tenant) => ({
+      ...queries.tenantResourcePolicy.get(tenant.metadata.id),
+      refetchInterval: TWO_MINUTES_MS,
+    })),
+  });
+
+  const notifications = useMemo(
+    () =>
+      tenants.flatMap((tenant, i) =>
+        (resourcePolicyQueries[i]?.data?.limits ?? [])
```
mrkaye97, line 80: flagging b/c this always sets of alarm bells 😅 - are we sure the tenants are guaranteed to be ordered the same as the resourcePolicyQueries?

frontend/app/src/pages/main/v1/index.tsx
```diff
@@ -21,15 +28,45 @@ function Main() {
...
+      await api.userUpdateLogout();
+    },
+    onSuccess: () => {
+      invalidateUserUniverse();
+      navigate({ to: appRoutes.authLoginRoute.to });
+    },
+    onError: handleApiError,
+  });
+
+  const onLogout = useCallback(() => logoutMutation.mutate(), [logoutMutation]);
```
mrkaye97: maybe this mutation + callback should live in the user universe hook?

mrkaye97 (commented):
a couple small UI things!

1. I think we should put the logout button either next to or above/below the help button on the bottom of the side nav
2. Small and annoying thing, but I wonder if we should nest the change theme button under general settings just for UI consistency :( Not great, hoping to improve the settings sidebar view thingy in general soon
3. It'd be great if we had some way to check if you finished the initial onboarding flow (maybe we can see if you have any workers / workflows / runs already as a hack), and if you haven't completed it, then have a notification / CTA telling you to do that. then we can also remove the overview tab from the sidebar which I really dislike since it's irrelevant for 95% of cases. WDYT?

frontend/app/src/hooks/notifications/resource-limits.ts
TehShrike: So as it is now, the type system catches it if there is any unhandled TenantResource – if you comment out any of those lines , there's a type error down at `resourceLabels[limit.resource]`.

I think TypeScript got smarter about map key types like this at some point a little bit back.

Personally in the case where I want to assert that there should never be a default case I like to use the type system to see it earlier.

Tangentially, this doesn't change much, but I also appreciate this pattern:

```ts
const resourceLabels = {
  [TenantResource.WORKER]: 'Total Workers',
  [TenantResource.WORKER_SLOT]: 'Concurrency Slots',
  [TenantResource.EVENT]: 'Events',
  [TenantResource.TASK_RUN]: 'Task Runs',
  [TenantResource.CRON]: 'Cron Triggers',
  [TenantResource.SCHEDULE]: 'Schedule Triggers',
  [TenantResource.INCOMING_WEBHOOK]: 'Incoming Webhooks',
} satisfies Record<TenantResource, string>
```

because then if you comment out one of those lines, not only do you get the type error down where `resourceLabels` is accessed, but you also get an error right there where the mapping is defined.

I have some preference for the mapping over the switch statement as long as I can get the early warning, but it's not a big deal, I'm happy to change it if you prefer the switch statements in general.

mrkaye97: Ah great stuff! No, either of these is good for me, as long as we get type errors instead of runtime errors if there’s a missing case, thanks for clarifying!

frontend/app/src/hooks/notifications/resource-limits.ts
TehShrike: So because the queries passed to `useQueries` is based on `tenants.map`, the indexes should be identical.

I don't particularly like this either – it reads too low-level than something I'd like to be composing here.  I could override the `queries.tenantResourcePolicy.get` query above so that the query data includes the tenant id along with the policy results, and then use that to get the relevant tenant, but that feels a little awkward too.

I think the main thing that needs to happen is an endpoint that returns the resource policies for all tenants in one go, including their tenantIds.

frontend/app/src/pages/main/v1/index.tsx
TehShrike: ah yeah, good call, I'll move the mutation there.  I just realized that there's no reason for the useCallback there though given the useMemo right below, wheee

TehShrike:
> I think we should put the logout button either next to or above/below the help button on the bottom of the side nav

Yeah, that makes sense, though I often find the "help" area eating into the vertical space on the sidebar a bit already – I'll probably try to reduce the vertical padding, at least.

> Small and annoying thing, but I wonder if we should nest the change theme button under general settings just for UI consistency :( Not great, hoping to improve the settings sidebar view thingy in general soon

Yeah, fair enough.  I can put theme under there with everything else – but maybe until a better categorization system comes along it would be okay to put everything at the top level, even if we recategorize everything soon after?  Putting it behind a click doesn't do much.

> It'd be great if we had some way to check if you finished the initial onboarding flow (maybe we can see if you have any workers / workflows / runs already as a hack), and if you haven't completed it, then have a notification / CTA telling you to do that. then we can also remove the overview tab from the sidebar which I really dislike since it's irrelevant for 95% of cases. WDYT?

I like the idea of supplanting the overview screen in the side nav – we do still force users to create a new tenant/organization if they don't have one and they don't have any invites, so until we change that they wouldn't ever see an onboarding notification for it.

Maybe if they don't have an API token, we could give them an onboarding notification that takes them to a "install the CLI + add a hatchet CLI profile" page?

mrkaye97:
> but maybe until a better categorization system comes along it would be okay to put everything at the top level, even if we recategorize everything soon after? Putting it behind a click doesn't do much.

I'm in favor of this - let's at least give it a shot and see how it looks (we discussed a bit in Slack yesterday, because I'd proposed the same)

> Maybe if they don't have an API token, we could give them an onboarding notification that takes them to a "install the CLI + add a hatchet CLI profile" page?

I feel like the simplest thing here would be to check if they have a token and maybe if they have some combination of workers / workflows / task runs and then link them to the current overview just so we don't need to rework that at all, but I agree in the future that linking to something more specific would be better

frontend/app/src/hooks/notifications/resource-limits.ts
mrkaye97: sounds good, works for me as long as it works as we intend :)

TehShrike:
The 05 Cypress test is failing in CI, but it passes on my machine.

mrkaye97 (approved):
🙌 thanks for the work here @TehShrike! sorry for just getting this over the line, have been swamped. excited to see this go live :)

promptless-for-oss:

Triggered by [PR #3365](https://github.com/hatchet-dev/hatchet/pull/3365)

Updated documentation for the dashboard notifications dropdown - fixed an inaccuracy about the Logout button location (now correctly described as being at the bottom of the sidebar rather than under Admin).

Review: [Document dashboard notifications dropdown for invites and resource limits](https://app.gopromptless.ai/suggestions/8ab01f78-08b7-4e6e-8ce7-bf6dba8cbdc3)

PR #3366, opened 2026-03-23
Add timeout context to OTel shutdown

juliusgeo:
# Description

Fixes # (issue)

## Type of change

## What's Changed

### DIFF
--- a/pkg/telemetry/telemetry.go
+++ b/pkg/telemetry/telemetry.go
@@ -92,23 +92,31 @@ func InitTracer(opts *TracerOpts) (func() error, error) {
 			return nil, fmt.Errorf("failed to parse traceIdRatio: %w", err)
 		}
 	}
-
-	otel.SetTracerProvider(
-		sdktrace.NewTracerProvider(
-			sdktrace.WithSampler(sdktrace.TraceIDRatioBased(traceIdRatio)),
-			sdktrace.WithBatcher(
-				exporter,
-				sdktrace.WithMaxQueueSize(sdktrace.DefaultMaxQueueSize*10),
-				sdktrace.WithMaxExportBatchSize(sdktrace.DefaultMaxExportBatchSize*10),
-			),
-			sdktrace.WithResource(resources),
+	tracerProvider := sdktrace.NewTracerProvider(
+		sdktrace.WithSampler(sdktrace.TraceIDRatioBased(traceIdRatio)),
+		sdktrace.WithBatcher(
+			exporter,
+			sdktrace.WithMaxQueueSize(sdktrace.DefaultMaxQueueSize*10),
+			sdktrace.WithMaxExportBatchSize(sdktrace.DefaultMaxExportBatchSize*10),
 		),
+		sdktrace.WithResource(resources),
+	)
+	otel.SetTracerProvider(
+		tracerProvider,
 	)
 
 	return func() error {
 		timeoutCtx, cancel := context.WithTimeout(context.Background(), time.Second*5)
 		defer cancel()
-		return exporter.Shutdown(timeoutCtx)
+		err := tracerProvider.Shutdown(timeoutCtx)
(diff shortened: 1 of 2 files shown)


pkg/telemetry/telemetry.go
```diff
@@ -92,20 +92,32 @@ func InitTracer(opts *TracerOpts) (func(context.Context) error, error) {
...
 		),
+		sdktrace.WithResource(resources),
+	)
+	otel.SetTracerProvider(
+		tracerProvider,
 	)
 
-	return exporter.Shutdown, nil
+	return func() error {
+		timeoutCtx, cancel := context.WithTimeout(context.Background(), time.Second*5)
```
gregfurman: Just wanna make sure it's desriable to be sharing the same context (and therefore timeout deadline) across both shutdown calls?

This'll mean that if the provider shutdown takes 4.5s, then the exporter shutdown could be cancelled early if it exceeds 500ms (due to the 5s timeout).

pkg/telemetry/telemetry.go
```diff
@@ -92,20 +92,32 @@ func InitTracer(opts *TracerOpts) (func(context.Context) error, error) {
...
+		defer cancel()
+		err := tracerProvider.Shutdown(timeoutCtx)
+		if err != nil {
```
gregfurman: nit: could always tighten up the `err` checks 🙂 

```suggestion
		if err := tracerProvider.Shutdown(timeoutCtx); err != nil {
```

pkg/telemetry/telemetry.go
```diff
@@ -92,20 +92,32 @@ func InitTracer(opts *TracerOpts) (func(context.Context) error, error) {
...
+			sdktrace.WithMaxExportBatchSize(sdktrace.DefaultMaxExportBatchSize*10),
 		),
+		sdktrace.WithResource(resources),
+	)
+	otel.SetTracerProvider(
+		tracerProvider,
 	)
 
-	return exporter.Shutdown, nil
+	return func() error {
```
gregfurman, line 108: Q: does shutdown order matter here? i.e can `traceProvider` be closed after `exporter`?

pkg/telemetry/telemetry.go
```diff
@@ -92,20 +92,32 @@ func InitTracer(opts *TracerOpts) (func(context.Context) error, error) {
...
+		defer cancel()
+		err := tracerProvider.Shutdown(timeoutCtx)
+		if err != nil {
+			return err
+		}
+		err = exporter.Shutdown(timeoutCtx)
+		if err != nil {
+			return err
+		}
+		return nil
```
gregfurman: nit: Unsure how the linter didn't catch this but since we aren't wrapping this `err` with any extra info, we really could just return the output of `export.Shutdown`:


```suggestion
		return exporter.Shutdown(timeoutCtx)
```

pkg/telemetry/telemetry.go
juliusgeo: Currently neither of these are finishing shutdown according to logs in live environments. Exporter doesn't shut down traceProvider, so I thought adding it in would be a good idea. There isn't anything in the Otel spec that indicates what order to shut them down, so I assumed shutting down the tracerProvider first makes sense.

pkg/telemetry/telemetry.go
juliusgeo: Yep, that is the goal.

pkg/telemetry/telemetry.go
gregfurman: Gotcha. You could also use an `errgroup` to shut down both concurrently and cancel if either return an error -- but not blocking or anything.

gregfurman (approved):
LGTM. Thanks for addressing my comment!

abelanger5 (approved):

PR #3370, opened 2026-03-23
feat(frontend): baseURL support

### LINKED ISSUE #2559: [FEAT] Allow API server and frontend to be served from URL path.
The hatchet API server and frontend require a dedicated hostname because their URL structures are fixed.  I'd like to host them at a different path behind a reverse proxy.

1a. Either (harder): Update the browser router to support a dynamic basename
1b. Or (easier): Update the dashboard.docker and frontend.dockerfile to accept a path that can be used to build the javascript code with a base path (otherwise the root path is hardcoded into the assets manifest)
2. Update hatchet-api to accept a base path to use for serving all apis (presumably by creating an echo group and adding all routes to that group)

1. Reverse proxy with response rewriting -> near impossible for the javascript, because the paths are embedded into js code

**Additional context**
N/A

gregfurman:
# Description

Allows a user to override the base URL of the hatchet frontend, allowing for hosting behind a path.

Fixes https://github.com/hatchet-dev/hatchet/issues/2559

## Type of change

## What's Changed

- Support runtime `BASE_PATH` override by dynamically injecting in a base URL for each page from the file-server without rebuilding the frontend image
- Extends `staticfileserver` to take in a `-base-path` flag, which will dynamically inject in the correct base-path into each served page.
- Includes a `BASE_PATH` environment variable in the `frontend.dockerfile`, passed to the `staticfileserver` via the `-base-url` flag

### Lite

1. Build the frontend, which should output a `./frontend/app/dist`:

```sh
$(cd frontend/app && npm run build)
```

2. Modify the `Caddyfile` to include a `handle /hatchet/*` instead of `handle /*` for `localhost:5173`

4. Run everything via hatchet lite pointing to build in `./frontend/app/dist` and with a custom base path of `/hatchet` (assuming environment is already seeded etc.):

```sh
task start-db goose-migrate && LITE_STATIC_ASSET_DIR=./frontend/app/dist LITE_FRONTEND_PORT=5173 LITE_FRONTEND_BASE_PATH="/hatchet" task start-lite
```

5. Navigate to [https://app.dev.hatchet-tools.com/hatchet](https://app.dev.hatchet-tools.com/hatchet).
6. Confirm you are routed to `https://app.dev.hatchet-tools.com/hatchet/auth/login`

### Container

1. Build and run the new container:

   ```sh
   docker build -f ./build/package/frontend.dockerfile -t ghcr.io/hatchet-dev/hatchet/hatchet-frontend:local .
   docker run --rm -e BASE_PATH="/hatchet" -p 5173:80 ghcr.io/hatchet-dev/hatchet/hatchet-frontend:local
   ```
2. Modify the `Caddyfile` to include a `handle /hatchet/*` instead of `handle /*` for `localhost:5173`
3. Start up a Hatchet engine and API instance (assuming environment is already seeded etc.):
   ```sh
    task start-db goose-migrate && task --parallel start-api start-engine
   ```
4. Navigate to [https://app.dev.hatchet-tools.com/hatchet](https://app.dev.hatchet-tools.com/hatchet).
5. Confirm you are routed to `https://app.dev.hatchet-tools.com/hatchet/auth/login`

### DIFF
--- a/build/package/dashboard-entrypoint.sh
+++ b/build/package/dashboard-entrypoint.sh
@@ -22,5 +22,9 @@ shutdown() {
 ./hatchet-api "$@" &
 HATCHET_API_PID=$!
 
+# Override the template-style {{ .BasePath }} with the envar $BASE_PATH
+: "${BASE_PATH:=/}"
+sed -i "s|{{ .BasePath }}|${BASE_PATH}|g" /usr/share/nginx/html/index.html
+
 # Start NGINX in the foreground
 nginx -g "daemon off;"
--- a/cmd/hatchet-staticfileserver/staticfileserver/server.go
+++ b/cmd/hatchet-staticfileserver/staticfileserver/server.go
@@ -18,7 +18,11 @@ func NewStaticFileServer(staticFilePath, basePath string) *chi.Mux {
 
 	fs := http.FileServer(http.Dir(staticFilePath))
 
-	r.Use(middleware.Logger)
+	r.Use(
+		middleware.Logger,
+		middleware.CleanPath,    // handles invalid multi-slashes ('/runs//<id>' -> '/runs/<id>')
+		middleware.StripSlashes, // strips trailing slash ('/hatchet/' -> '/hatchet')
+	)
 
 	basePath = strings.TrimRight(basePath, "/")
 	if basePath != "" {
--- a/cmd/hatchet-staticfileserver/staticfileserver/server.go
+++ b/cmd/hatchet-staticfileserver/staticfileserver/server.go
@@ -29,6 +29,7 @@ func NewStaticFileServer(staticFilePath, basePath string) *chi.Mux {
 		// Dynamcally build and serve the index.html and config.js when we have a custom basePath
 		r.Get(basePath, handleIndex(staticFilePath, basePath))
 		r.Get(basePath+"/config.js", handleJsConfig(basePath))
+		r.Get("/", http.RedirectHandler(basePath, http.StatusSeeOther).ServeHTTP)
 	}
 
 	r.Get(basePath+"/*", func(w http.ResponseWriter, r *http.Request) {
--- a/cmd/hatchet-staticfileserver/staticfileserver/server.go
+++ b/cmd/hatchet-staticfileserver/staticfileserver/server.go
@@ -1,8 +1,7 @@
 package staticfileserver
 
 import (
-	"encoding/json"
-	"fmt"
+	"html/template"
 	"net/http"
 	"os"
 	"path"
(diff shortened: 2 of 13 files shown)


promptless-for-oss:

Triggered by this PR

Added a new section to the self-hosting Docker Compose docs explaining how to configure a custom BASE_PATH by mounting a `config.js` file. This allows users to deploy Hatchet behind a reverse proxy with a subpath (e.g., `https://example.com/hatchet`) without rebuilding the Docker image.

Review at https://app.gopromptless.ai/suggestions/c40865f0-3585-41fd-92f3-e51fc279c06a

tvildo:
Hi, Thanks for your commits. I tested and when page is loaded all assets are hardcoded and have wrong path, because you have to change vite config as in this [commit](https://github.com/tvildop/hatchet/commit/fd8fe78b2a21c4ac5cd5bf3816b46fe60f8be3a0).  

here is how index html looks now when loaded under http://domain.com/hatchet
```
<!doctype html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <title>Hatchet</title>
    <meta
      name="description"
      content="Write workflows as code. Scale with Hatchet."
    />
    <link
      href="https://fonts.googleapis.com/css?family=Ubuntu:400,500,700"
      rel="stylesheet"
    />
    <meta property="og:title" content="Hatchet" />
    <link rel="icon" type="image/svg+xml" href="/favicon.svg" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <title>Hatchet</title>
    <script type="text/javascript" src="/config.js"></script>
    <script type="module" crossorigin src="/assets/index-BwLVrsIY.js"></script>
    <link rel="stylesheet" crossorigin href="/assets/index-9SuaZgKi.css">
  </head>
  <body>
    <div id="root"></div>
  </body>
</html>

```
Instead of /config.js it should be /hatchet/config.js or relative url like ./config.js

gregfurman:
@tvildo Thansk for testing this! Feel free to open a PR against my feature branch with those suggested changes?

gregfurman:
@tvildo I took your suggestions and integrated them into our `staticfileserver` -- so instead of using a custom entrypoint with `sed`, the server should handle the injection of a custom `BASE_URL` into each served page.

Let me know if all works now? I've added a `Testing` section to the PR description for a quickstart.

tvildo:
Hello, I tested the changes, but static assets are not loading correctly (see attached image).

It looks like the issue is related to absolute paths — switching to relative paths should resolve it, similar to what I implemented in my PR: https://github.com/gregfurman/hatchet/pull/98/changes

Also, from my perspective, the previous sed-based approach was quite straightforward and only ran once during container initialization, so it didn’t add much overhead.

Let me know what you think.

image

tvildo:
After new test now it answers with 404 on everything. I am not sure why...

image
image

gregfurman:
@tvildo Seems the liveness and readiness probes will fail for the frontend since they are expecting to hit `/` but we're now hosting behind a path, which could explain why your k8s deployment is failing:

See snippet from [hatchet-charts for the frontend-stack](https://github.com/hatchet-dev/hatchet-charts/blob/9381a5e0b42bbe318cb05ba5dec6938a503b7bb0/charts/hatchet-frontend/templates/deployment.yaml#L72-L81):

```yaml
        livenessProbe:
          httpGet:
            path: /
            port: {{ .Values.service.internalPort }}
          initialDelaySeconds: 15
          periodSeconds: 20
        readinessProbe:
          httpGet:
            path: /
            port: {{ .Values.service.internalPort }}
```

I did a super hacky `kubectl patch` and it seems to fix it:

```sh
kubectl patch deployment hatchet-stack-frontend --type=json -p='[
  {"op": "replace", "path": "/spec/template/spec/containers/0/livenessProbe/httpGet/path", "value": "/hatchet"},
  {"op": "replace", "path": "/spec/template/spec/containers/0/readinessProbe/httpGet/path", "value": "/hatchet"}
]'
```

Port forwarding Caddy from `80 => 8080`, we can access this fine:

image


Here's my `values.yaml`:

```yaml
frontend:
  image:
    tag: local
    pullPolicy: Never
  env:
    BASE_PATH: "/hatchet"
  sharedConfig:
    image:
      tag: ""
```

I did notice that a trailing `/` (i.e `/hatchet/`) gave a 404 so that needs to be fixed too. Did you see your healthchecks failing?

gregfurman:
Think this is ready for review.

Simplified the approach a bit to configure the `index.html` with a go-template variable that we dynamically execute each time a page is served from the `staticfilesever` instead of a mounted `config.js`.

I've tested on my local docker-desktop k8s cluster and all seems to be working fine so let me know if we'd like a `helm`/`k8s` example as well for this (using ingresses).

Bodoral:
any update on this PR?

cmd/hatchet-staticfileserver/staticfileserver/server.go
```diff
@@ -1,26 +1,30 @@
...
 
-func NewStaticFileServer(staticFilePath string) *chi.Mux {
+func NewStaticFileServer(staticFilePath, basePath string) *chi.Mux {
 	r := chi.NewRouter()
 
 	fs := http.FileServer(http.Dir(staticFilePath))
 
 	r.Use(middleware.Logger)
 
+	r.Get("/", indexHandler(staticFilePath, basePath))
```
BloggerBust, line 22: Only `/` renders through `indexHandler`. The `/*` fallback serves `index.html` straight from `http.FileServer` (line 30), so it skips templating. A hard refresh or deep link on a nested route returns the raw template with a literal `{{ .BasePath }}`. I confirmed this by running the built server with `-base-path /hatchet/`: `/` renders `base href="/hatchet/"`, while `/auth/login`, `/hatchet`, and `/hatchet/auth/login` all return `base href="{{ .BasePath }}"`.

Could you route the not-found SPA fallback through `indexHandler` so it gets the same templating?

cmd/hatchet-staticfileserver/staticfileserver/server.go
```diff
@@ -37,3 +41,15 @@ func NewStaticFileServer(staticFilePath string) *chi.Mux {
 
 	return r
 }
+
+func indexHandler(staticFilePath, basePath string) http.HandlerFunc {
+	t := template.Must(template.ParseFiles(filepath.Join(staticFilePath, "index.html")))
+	data := struct{ BasePath string }{basePath}
```
BloggerBust, line 47: Could you normalize the value used for `<base href>` so it always has exactly one trailing slash? With `base: './'` in Vite, `<base href="/hatchet">` makes relative assets resolve to `/assets/...` at the root, whereas `<base href="/hatchet/">` resolves them under `/hatchet/...`.

The TanStack Router `basepath` expects no trailing slash, which `config.ts` already strips. The HTML `<base href>` expects the trailing slash. Injecting the slash form here and letting `config.ts` strip it for the router covers both.

build/package/dashboard-entrypoint.sh
```diff
@@ -22,5 +22,9 @@ shutdown() {
 ./hatchet-api "$@" &
 HATCHET_API_PID=$!
 
+# Override the template-style {{ .BasePath }} with the envar $BASE_PATH
+: "${BASE_PATH:=/}"
+sed -i "s|{{ .BasePath }}|${BASE_PATH}|g" /usr/share/nginx/html/index.html
```
BloggerBust, line 27: Same `{{ .BasePath }}` placeholder here, so this needs the same trailing-slash normalization as the `<base href>` comment on `server.go`.

frontend/app/src/pages/auth/components/auth-layout.tsx
```diff
@@ -1,9 +1,11 @@
 import { HeroPanel } from './hero-panel';
+import { config } from '@/config';
 import { CSSProperties, PropsWithChildren } from 'react';
 
 export function AuthLayout({ children }: PropsWithChildren) {
   const bgContentStyle = {
-    '--noise-url': 'url("/noise.png")',
+    // HACK(gregfurman): Strip the trailing '/' from BASE_PATH, accounting for '//' and '/base/path/' cases.
+    '--noise-url': `url("${config.BASE_PATH.replace(/\/+$/, '')}/noise.png")`,
```
BloggerBust, line 8: `config.BASE_PATH` already strips trailing slashes, so the `.replace(/\/+$/, '')` here is redundant. The `HACK` comment above can be removed too.

BloggerBust (changes requested):
Thanks @gregfurman, this is looking better. I left a couple of change requests inline, and CI is failing on lint, so please address both.

I have two non-blocking nits. There is no Go test covering `indexHandler` or the SPA fallback, and given the deep-link issue, a small `httptest` case asserting that `/`, `/hatchet`, `/hatchet/`, and a nested route all render the templated `index.html` would lock the behavior in. The other is documentation. A short note that `/api` stays at the root and that the base path needs a trailing slash would help, since a missing slash breaks asset loading with no error. The docs can be a follow-up if you prefer.

Ping me when you are ready for a final review.

mnafees:
Closing this in favor of https://github.com/hatchet-dev/hatchet/pull/4558

PR #3380, opened 2026-03-24
Fix: Add sleep to python test to let otel traces propagate

mrkaye97:
# Description

Adds a sleep before we start polling

## Type of change


mnafees (approved):

grutt:
@mrkaye97 going to move this over to #3389 and update assertions

PR #3381, opened 2026-03-24
feat: enable tenant-scoped logs view

abelanger5:
# Description

Enable tenant-scoped logs view and adds a simple index for queries. 

## Type of change


mnafees (approved):

promptless-for-oss:

Updated existing suggestion: **[Document tenant-scoped logs view and run-level logs tab](https://app.gopromptless.ai/suggestions/ba91ae86-4bc7-4200-9203-ec7ed4d7cd76)**

---

cmd/hatchet-migrate/migrate/migrations/20260324135804_v1_0_92.go
```diff
@@ -0,0 +1,86 @@
+package migrations
```
mrkaye97, line 1: FYI that we don't need a go migration for this: https://github.com/hatchet-dev/hatchet/blob/main/cmd/hatchet-migrate/migrate/migrations/20251007220652_v1_0_47.sql

abelanger5: we definitely do. that case is different; the leaf partitions are static. but we have the constraint that we need to compute them dynamically but we can't run them inside of a transaction block with `CONCURRENTLY`. which means we can't use a function since it runs inside of an implicit tx. I think a Go migration is the only way.

mrkaye97: ooooh, good call, you're right 👍

PR #3384, opened 2026-03-24
Add in memory, keyed, locking queue to protect RunConcurrencyStrategy

### LINKED ISSUE #2866: [BUG] Multiple concurrency rules cause significant task queue overhead
**Describe the issue**
When a workflow is configured with **multiple concurrency rules**, each task experiences significant queue time (1-10 seconds), even when running a single isolated workflow with no concurrency contention.

 The overhead scales with the number of tasks in the workflow. A 4-task DAG workflow that should complete in ~15s takes ~60s due to queue overhead.

**Environment**
- SDK: TypeScript v5.9.3
- Engine: v0.77.30 Cloud (but also observed in self-hosted)

**Expected behavior**
Multiple concurrency rules should not add significant overhead when there's no actual concurrency contention. A single workflow running in isolation should have minimal queue time regardless of how many concurrency rules are configured.

```typescript
  export const workflow = hatchet.workflow({
    name: "test-concurrency",
    concurrency: [
      {
        expression: "input.key1",
        maxRuns: 1,
        limitStrategy: ConcurrencyLimitStrategy.CANCEL_IN_PROGRESS,
      },
      {
        expression: "input.key2",
        maxRuns: 5,
        limitStrategy: ConcurrencyLimitStrategy.GROUP_ROUND_ROBIN,
      },
    ],
  })
```

  // 4 sequential tasks: A → B → C → D

In this simple isolated run it already adds significant overhead. On real tasks on production, we have seen even more seconds. 

Image

**Additional context**
Obviously this could be fixed by leaving just one concurrency strategy, but having it defined at workflow level instead of application level makes it much easier to work with. e.g. de-duplicating tasks with CANCEL_IN_PROGRESS while also preventing resource hogging by user with GROUP_ROUND_ROBIN.

juliusgeo:
# Description

When multiple concurrency strategies are used, they will race to acquire the lock in `TryAdvisoryLock`, resulting in the strategies that lose being kicked back out, only to run on the next tick of the `loopConcurrency` loop. Results in large scheduling delays (https://github.com/hatchet-dev/hatchet/issues/2866). This fixes by adding an in memory queueing lock that protects RunConcurrencyStrategy, removing contention on the database locks, letting them all run during the same tick. 

Fixes #2866

## Type of change

## What's Changed


internal/services/shared/timeout_lock/timeout_lock.go
```diff
@@ -0,0 +1,68 @@
...
+	return &TimeoutLock{
+		sem:     make(chan struct{}, 1),
+		Timeout: timeout,
+	}
+}
+
+type KeyedTimeoutLock[T comparable] struct {
+	locks    map[T]*TimeoutLock
+	Timeout  time.Duration
+	lockLock *TimeoutLock
```
abelanger5: It seems like `lockLock` is adding unnecessary overhead here - it doesn't make sense to me that the `timeout` for the `lockLock` should necessarily be the same as the timeout that we're using to acquire the core lock. Mutexes are more performant than channels and we should never hold the `lockLock` for very long (unlike the timeout lock). So perhaps we can replace `lockLock` with a `sync.Mutex`?

pkg/scheduling/v1/concurrency.go
```diff
@@ -101,6 +107,27 @@ func (c *ConcurrencyManager) notify(ctx context.Context) {
...
 
+func (c *ConcurrencyManager) acquireStrategyLocks() bool {
+	acquired := c.advisoryLock.Acquire(c.strategy.ID)
+	if !acquired {
+		return acquired
+	}
+	if c.strategy.ParentStrategyID.Valid {
+		parentId := v1.PARENT_STRATEGY_LOCK_OFFSET + c.strategy.ParentStrategyID.Int64
+		return c.advisoryLock.Acquire(parentId)
+	}
```
abelanger5, line 122: there's a risk of a very strange deadlock here; I'll let you find it. but it would be arbitrarily resolved with a separate `advisoryLock` for the parent strategies ids (something that we can now do with the in-memory locks which we couldn't do before!)

internal/services/shared/timeout_lock/timeout_lock.go
```diff
@@ -0,0 +1,68 @@
...
+	Timeout time.Duration
+}
+
+func (l *TimeoutLock) Acquire() bool {
+	select {
+	// attempt to send to the semaphore, blocks on contention because it has a buffer of 1
+	case l.sem <- struct{}{}:
+		return true
+	// timing out dequeues the semaphore send
+	case <-time.After(l.Timeout):
```
abelanger5, line 20: something to think about: the timeout might be ok, but the more common use-case would be to pass in a `context.Context` instead. the duration approach is more difficult for callers because they need to synchronize the duration with the context in most cases

internal/services/shared/timeout_lock/timeout_lock.go
juliusgeo: hmm, I could go either way. I initially was using a sync.Mutex, but that breaks the ordering guarantee when there is heavy contention IIUC. Could also set the lockLock to just a very small timeout?

internal/services/shared/timeout_lock/timeout_lock.go
juliusgeo: I think with the current use case the context cancellation wouldn't be an issue, as we aren't using it any places with a TimeoutContext I believe, but that makes sense for a more general case.

sdks/python/tests/concurrency_strategies/test_concurrency_strategies.py
```diff
@@ -25,19 +25,16 @@ async def test_concurrency_strategy_scheduling(
...
+        ),
     )
     start = time.time()
-    results = run.result()
-    print(results)
+    run.result()
     end = time.time()
     elapsed_time = end - start
     print(f"Time taken: {elapsed_time:.4f} seconds")
+    assert elapsed_time < 100
```
abelanger5: this is less than 100 seconds? shouldn't we be checking for a smaller value?

sdks/python/tests/concurrency_strategies/test_concurrency_strategies.py
```diff
@@ -25,19 +25,16 @@ async def test_concurrency_strategy_scheduling(
...
+        input=InputModel(key1=key1, key2=key2),
+        options=TriggerWorkflowOptions(
+            additional_metadata={
+                "test_run_id": key1,
+            }
+        ),
     )
     start = time.time()
-    results = run.result()
-    print(results)
```
abelanger5: intentional removal of results? seems like this could be useful for debugging

pkg/scheduling/v1/concurrency.go
```diff
@@ -101,6 +109,30 @@ func (c *ConcurrencyManager) notify(ctx context.Context) {
 	}
 }
 
+func (c *ConcurrencyManager) acquireStrategyLocks() bool {
+	acquired := c.advisoryLock.Acquire(c.strategy.ID)
+	if !acquired {
+		return acquired
+	}
+	if c.strategy.ParentStrategyID.Valid {
+		parentId := v1.PARENT_STRATEGY_LOCK_OFFSET + c.strategy.ParentStrategyID.Int64
```
abelanger5: we don't need the offset anymore! since we're using separate locks. the point was always to avoid a collision between child strategy ids and parents.

juliusgeo: gotcha, fixed!

sdks/python/tests/concurrency_strategies/test_concurrency_strategies.py
juliusgeo: Oops, yeah, was testing it locally with much longer concurrency intervals. Should be less than 10 seconds

pkg/scheduling/v1/concurrency.go
juliusgeo: added a second lock for parent strategy ids!

internal/services/shared/timeout_lock/timeout_lock.go
juliusgeo: updated lockLock so it just has a very small timeout.

sdks/python/tests/concurrency_strategies/test_concurrency_strategies.py
juliusgeo: made the assertion message better, added back in results

internal/services/shared/timeout_lock/timeout_lock.go
abelanger5: I don't think the mutex would break the ordering guarantee? The mutex is just guarding the map reads and writes, correct? 

The ordering guarantee issue was from the dispatcher PR but was pretty specific to how that was implemented IIRC.

abelanger5 (approved):

abelanger5 (approved):

PR #3389, opened 2026-03-24
Fix: otel-bug-bash

grutt:
# Description

Implements feedback and bashes bugs for the new otel view.

### Backend
- **Separated engine spans from SDK spans** — resolves a race condition in OLAP span processing (`engine_spans.go`)
- **New OLAP repository queries** — added SQL queries for fetching OTEL spans (`otel.sql`, `otel.sql.go`, `olap.go`)
- **Fixed scheduling timeout engine span** literal edge cases in the traces handler

### Frontend — Observability / Trace Viewer
- **Trace timeline overhaul** — refactored `task-run-trace.tsx` (+519 lines) with expand/collapse, zoom lines, clear zoom button, and time-tick labels
- **Span detail redesign** — refactored `span-detail.tsx` to use the standard component library; improved large attribute value display
- **Minimap improvements** — added queued state/color, better affordances and state management
- **Span selection UX** — improved hover hierarchy, "show in context" feature, consistent label/clipping behavior
- **Timeline labels** — replaced `timeline-tick-header.tsx` with new `time-tick-labels.tsx`
- **Empty state cards**  — makes it clear how to enable and use features

### Frontend — Search & Navigation
- **New `use-run-detail-search` hook** — manages run detail state in query params
- **Trace search filter** — added "filter by span name" with autocomplete support
- **Keyboard shortcut** — clear search on Escape
- **Tab rename** — renamed tab to "Traces"; consistent log tab behavior

## Type of change


promptless-for-oss:

Updated existing suggestion: **[Document OpenTelemetry support for Python, TypeScript, Go, and Ruby SDKs](https://app.gopromptless.ai/suggestions/79f91512-3bd0-45b0-a570-f6c7f0d8a4e9)**

---
_Tip: Request one-off documentation tasks in the Dashboard under [New Task](https://app.gopromptless.ai/new-task) 🚀_

grutt:
image
image
image

pkg/repository/sqlcv1/otel.sql
```diff
@@ -117,3 +117,34 @@ ORDER BY start_time ASC
...
+
+-- name: ListSpansByTraceIds :many
+SELECT *
+FROM v1_otel_trace_olap
+WHERE
+    tenant_id = @tenantId::UUID
+    AND trace_id = ANY(@traceIds::BYTEA[])
+ORDER BY start_time ASC
+OFFSET COALESCE(@spanOffset::BIGINT, 0)
+LIMIT COALESCE(@spanLimit::BIGINT, 1000)
```
mrkaye97: not sure I'm following the need for these queries 🤔 

for the first one, the union won't do anything in non-oss, right? and without it it's the same as the existing trace lookup? and the second one is a plural version of span lookup (many traces instead of one) basically?

what's the situation where we'd want to look up many traces (assuming no global trace view yet)?

if we want to make this plural and have a single repo method that looks up the trace id(s) and then reads spans out that's fine, but unless I'm missing something, I don't think these query changes are needed

sdks/go/examples/opentelemetry_instrumentation/main.go
```diff
@@ -145,7 +145,7 @@ func main() {
 			chargeSpan.End()
 
 			paySpan.End()
-			return ChargePaymentOutput{}, fmt.Errorf("test error")
+			// return ChargePaymentOutput{}, fmt.Errorf("test error")
```
mrkaye97: intentional?

internal/services/controllers/olap/engine_spans.go
```diff
@@ -308,10 +241,22 @@ func safeUint64(v int64) uint64 {
 	return uint64(v) // nolint:gosec
 }
 
-func generateSpanID() string {
+// deriveEngineTraceID produces a deterministic 16-byte trace ID from
+// (externalID, retryCount) so all engine spans for a single step-run
+// retry share one trace instead of each getting a random one.
+func deriveEngineTraceID(externalID uuid.UUID, retryCount int32) []byte {
```
mrkaye97: will this work for DAGs? we need the trace id to be the same across all of the spans in a DAG's trace, so we'd need to be careful to make sure that we always use the same trace id (and I guess external id?) for all spans in a dag, in a tree of children, etc.

mrkaye97 (commented):
changes look good to me in general, thanks for the docs buttons + renaming to `Traces` 👍 a little suspicious of the API changes, I'm not sure we really need them and I'm not convinced the new queries are correct

sdks/go/examples/opentelemetry_instrumentation/main.go
grutt: yeah we fail out of one of the children so that case is covered

internal/services/controllers/olap/engine_spans.go
grutt: nope, made lots of other changes to make this and dags work....

pkg/repository/sqlcv1/otel.sql
grutt: i can explain sync if helpful, but we will need a separate trace for engine generated spans due to a few different race conditions. that said, you're right here that this isnt needed for now

sdks/python/examples/opentelemetry_instrumentation/test_otel_traces.py
```diff
@@ -91,20 +114,20 @@ async def test_otel_spans_created_on_task_run(hatchet: Hatchet) -> None:
     assert child_attrs.get("test.marker") == "hello"
     assert child_attrs.get("input.message") == message
 
-    run_workflow_spans = [s for s in spans if s.span_name == "hatchet.run_workflow"]
+    run_workflow_spans = [
+        s for s in spans if s.span_name == "hatchet.engine.workflow_run"
+    ]
```
mrkaye97, line 120: whoops, for some reason my comment on the commit doesn't seem to be showing up at the top level, but I think this test change (and some below it) are incorrect - this is a producer-side span coming from the SDK when we trigger the run. we definitely need to see a span with this name and make sure it exists / assert on it

sdks/python/examples/opentelemetry_instrumentation/test_otel_traces.py
```diff
@@ -133,14 +156,36 @@ async def test_otel_spans_on_event_triggered_run(hatchet: Hatchet) -> None:
 
     spans = await asyncio.to_thread(poll_for_trace, hatchet, run_id)
 
-    push_event_spans = [s for s in spans if s.span_name == "hatchet.push_event"]
```
mrkaye97, line 139: same here - producer-side span we need to see

sdks/python/examples/opentelemetry_instrumentation/test_otel_traces.py
grutt: good catch, introduced a bug for trace id inheritance and didnt think here

sdks/python/pyproject.toml
```diff
@@ -2,12 +2,6 @@
 name = "hatchet-sdk"
 version = "1.29.4"
 description = "This is the official Python SDK for Hatchet, a distributed, fault-tolerant task queue. The SDK allows you to easily integrate Hatchet's task scheduling and workflow orchestration capabilities into your Python applications."
-authors = [
-  "Alexander Belanger <alexander@hatchet.run>",
-  "Matt Kaye <matt@hatchet.run>",
-  "Gabe Ruttner <gabe@hatchet.run>",
-  "Mohammed Nafees <nafees@hatchet.run>",
-]
```
mrkaye97: looks like a merge issue

pkg/repository/trigger.go
```diff
@@ -472,20 +476,33 @@ func ensureTraceparent(additionalMetadata []byte, workflowRunExternalID uuid.UUI
...
+
+	var traceID string
 
-	if parentIDStr, ok := meta["hatchet__parent_workflow_run_id"].(string); ok {
-		if parsed, err := uuid.Parse(parentIDStr); err == nil {
-			traceOwnerID = parsed
+	if existingTP, ok := meta["traceparent"].(string); ok {
+		if tID, sID, ok := parseW3CTraceparent(existingTP); ok {
+			traceID = tID
+			meta["hatchet__traceparent_parent_span_id"] = sID
```
mrkaye97: I'm not sure I quite understand this part - why do we need to store a parent span id separately from the `traceparent`?

sdks/python/pyproject.toml
grutt: yep you're looking at an old commit?

mrkaye97: oh whoops 🤔 I thought I'd selected the four new ones... confusing - sorry!

pkg/repository/trigger.go
grutt: part of our issue is that we dont have the full tree context at various points in writing from the engine so this just helps us ensure everything is properly nested and we dont regenerated downstream

mrkaye97: 🤔 but doesn't the traceparent include the parent span id? I guess I don't see why this piece of metadata isn't redundant with the traceparent (asking especially b/c it'll be exposed to the user via additional meta)

grutt: mm, that's fair... let me see if i still really need this

internal/services/controllers/olap/engine_spans.go
```diff
@@ -40,119 +40,389 @@ type taskRetryKey struct {
 	retryCount int32
 }
 
-func (tc *OLAPControllerImpl) synthesizeEngineSpans(ctx context.Context, tenantId uuid.UUID, events []engineSpanEvent) {
+type eventEmittedAccumulator struct {
+	eventSeenAt                 time.Time
+	eventKey                    string
+	triggeredRunExternalIDs     []string
```
mrkaye97, line 46: uuid for this?

sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
```diff
@@ -127,8 +130,10 @@ def __init__(self, span_exporter: SpanExporter, **kwargs: Any) -> None:
     def on_start(self, span: Span, parent_context: Context | None = None) -> None:
         attrs = ctx_hatchet_span_attributes.get()
         if attrs and span.is_recording():
+            existing = getattr(span, "attributes", None) or {}
```
mrkaye97: shouldn't use `getattr` here

mrkaye97 (approved):
left one comment but seems to work fine against hdx in the new example at least, so works for me assuming we tested out the existing example and it looked the same on main / this branch

promptless-for-oss:

Triggered by [PR #3389](https://github.com/hatchet-dev/hatchet/pull/3389)

Updated documentation to reflect the new trace viewer features: renamed "Traces" tab, search bar for filtering spans by name/status/attributes with autocomplete suggestions, **Escape** keyboard shortcut to clear search, and zoom controls (click-drag on timeline, "clear zoom" button).

Review at https://app.gopromptless.ai/suggestions/79f91512-3bd0-45b0-a570-f6c7f0d8a4e9

PR #3396, opened 2026-03-25
fix(ruby): make runs.poll() cancellable and non-blocking; add poll_async (Fixes #3329)

### LINKED ISSUE #3329: [BUG]Ruby SDK: runs.poll() blocks the calling thread indefinitely with no cancellation option
**Describe the issue**
`runs.poll()` in the Ruby SDK is implemented as a bare synchronous `loop do ... sleep ... end` with no way to cancel, interrupt, or run asynchronously. Calling it in any multi-threaded Ruby application (Rails, Sidekiq, Puma) blocks the calling thread for the entire duration of the poll. If no timeout is set, the thread is blocked indefinitely. This was flagged by the original author during the PR review for #2086 with the comment "this is blocking.. need to think about this" but shipped without a fix.

**Environment**
- SDK: Ruby (initial release, PR #2086)
- Engine: Cloud or Self-hosted

**Expected behavior**
`poll()` should not block the calling thread indefinitely. At minimum it should be cancellable. Ideally it should support a block/callback or an async variant so callers are not forced to tie up a thread while waiting for run completion.

```ruby
# In a Sidekiq worker or Puma thread — this blocks the thread forever
# if the run never reaches a terminal state, with no way to cancel
hatchet.runs.poll(workflow_run_id)
```

In a Rails app with Puma, every call to `poll()` without an explicit timeout permanently consumes one Puma thread. With enough concurrent requests, the thread pool is exhausted and the app becomes unresponsive.

**Additional context**
The issue was acknowledged by the original author in the PR review for #2086:

> "this is blocking.. need to think about this"

The method was merged without resolving this. Any production Ruby application using Sidekiq, Puma, or any other threaded server is affected. Ref: #2086

Akhilesh29:
# Description

`runs.poll()` was implemented as a bare `loop do ... sleep ... end` with no way to cancel, interrupt, or run asynchronously. In any multi-threaded Ruby application (Rails/Puma, Sidekiq) this permanently consumed the calling thread
for the entire duration of the poll. With no timeout set, the thread was blocked indefinitely, acknowledged by the original author in PR #2086 but shipped
without a fix.

Fixes #3329

## Type of change

 - Bug fix (non-breaking change which fixes an issue)
-  New feature (non-breaking change which adds functionality)

## What's Changed

- Add `Hatchet::PollTimeoutError` — raised when the `:timeout` option elapses
- Add `Hatchet::PollCancelledError` — raised when a cancellation token is signalled
- Add `Hatchet::CancellationToken` — thread-safe stop flag (Mutex-protected boolean)
- Rewrite `runs.poll()` to check cancellation and timeout on every iteration instead of looping forever
- Fix timeout logic in `poll()` — deadline is now computed once upfront; sleep duration is capped to remaining budget so the method never overshoots its own timeout
- Add `runs.poll_async()` — non-blocking wrapper that spawns a background Thread and yields `(result, error)` to a block or callback; calling thread returns immediately
- Replace `puts` logging in `poll()` with `@config.logger` calls (debug/info/warn)
- All changes are fully backward-compatible — existing `poll(run_id)` calls with no arguments behave identically to before

### DIFF
--- a/sdks/ruby/src/lib/hatchet/features/runs.rb
+++ b/sdks/ruby/src/lib/hatchet/features/runs.rb
@@ -513,7 +513,27 @@ end
           break if event.respond_to?(:hangup) && event.hangup
         end
       end
+      def poll_async(workflow_run_id, interval: 1.0, timeout: nil,
+               on_complete: nil, on_error: nil,
+               cancellation_token: nil, &block)
+  token = cancellation_token || Hatchet::CancellationToken.new
 
+  thread = Thread.new do
+    begin
+      result = poll(workflow_run_id, interval: interval,
+                    timeout: timeout, cancellation_token: token)
+      block ? block.call(result, nil) : on_complete&.call(result, nil)
+    rescue Hatchet::PollTimeoutError, Hatchet::PollCancelledError => e
+      @config.logger.warn("poll_async stopped for #{workflow_run_id}: #{e.message}")
+      block ? block.call(nil, e) : on_error&.call(nil, e)
+    rescue => e
+      @config.logger.error("poll_async error for #{workflow_run_id}: #{e.message}")
+      block ? block.call(nil, e) : on_error&.call(nil, e)
+    end
+  end
+
+  [thread, token]
+end
       private
 
       # Build BulkCancelReplayOpts from keyword arguments.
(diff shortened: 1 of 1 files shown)


promptless-for-oss:

Updated existing suggestion: **[Add Ruby SDK setup documentation and reference section](https://app.gopromptless.ai/suggestions/132d7bb5-0773-42ee-aea9-3edc0165b73c)**

---
_Tip: Use Vale? Add your vale.ini when setting up your [Docs Collection](https://app.gopromptless.ai/doc-collections)._

mrkaye97:
Hey @Akhilesh29 - thank you for the contribution! As a matter of courtesy before requesting a review: have you run this code locally? The end to end tests are failing with an error indicating this code hasn't been run yet

PR #3401, opened 2026-03-25
fix: select only stable releases for install

Spid3rrr:
# Description

Fix installer version resolution so it only auto-selects stable Hatchet CLI releases in the format `vX.Y.Z`, and ignores scoped tags (for example `vscode/v0.1.2`) and prereleases (for example `v0.83.1-alpha.0`).

This updates the release lookup logic to use the releases list endpoint and apply strict stable tag filtering before choosing the latest version.

Fixes # (issue)

## Type of change

## What's Changed

- [x] Changed installer latest-version lookup to read from the releases list instead of relying on a single latest-tag response.
- [x] Added strict stable semver filtering (`^v[0-9]+\.[0-9]+\.[0-9]+$`) for auto-selected versions.
- [x] Ensured scoped tags and prerelease tags are excluded from automatic selection.
- [x] Updated failure messaging to indicate stable-version resolution failure when no valid tag is found.

### DIFF
--- a/hack/install/install.sh
+++ b/hack/install/install.sh
@@ -206,14 +206,16 @@ get_version() {
         log_info "Using requested version: $VERSION"
     else
         log_info "Fetching latest version..."
+        # Use /releases and select the first stable core tag (vX.Y.Z).
+        # This avoids non-CLI/scoped tags (e.g. vscode/v0.1.2) and prereleases.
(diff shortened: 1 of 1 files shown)


gregfurman (approved):
LGTM! Thanks for addressing this

PR #3408, opened 2026-03-26
Add more detailed logging to in memory advisory lock, configurable timeout

juliusgeo:
# Description

Fixes # (issue)

## Type of change

## What's Changed

### DIFF
--- a/pkg/scheduling/v1/concurrency.go
+++ b/pkg/scheduling/v1/concurrency.go
@@ -126,7 +126,9 @@ func (c *ConcurrencyManager) acquireStrategyLocks() bool {
 
 func (c *ConcurrencyManager) releaseStrategyLocks() {
 	c.advisoryLock.Release(c.strategy.ID)
-	c.advisoryParentLock.Release(c.strategy.ParentStrategyID.Int64)
+	if c.strategy.ParentStrategyID.Valid {
+		c.advisoryParentLock.Release(c.strategy.ParentStrategyID.Int64)
+	}
 }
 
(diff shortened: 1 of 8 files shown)


pkg/scheduling/v1/concurrency.go
```diff
@@ -126,9 +126,7 @@ func (c *ConcurrencyManager) acquireStrategyLocks() bool {
 
 func (c *ConcurrencyManager) releaseStrategyLocks() {
 	c.advisoryLock.Release(c.strategy.ID)
-	if c.strategy.ParentStrategyID.Valid {
-		c.advisoryParentLock.Release(c.strategy.ParentStrategyID.Int64)
-	}
+	c.advisoryParentLock.Release(c.strategy.ParentStrategyID.Int64)
```
abelanger5: Is this not going to cause a problem when we don't have a parent strategy id?

juliusgeo: If the key doesn't exist when we release, we just return. I was worried there may be a deadlock if the parentstrategyid becomes invalid between when we lock and unlock. Not sure if that's a valid concern though :)

abelanger5: hmm, that should never happen, the parent strategy id is immutable - I'd be a little concerned that we accidentally use `0` as the key which could cause all sorts of weird behavior?

juliusgeo: gotcha, will fix

abelanger5 (approved):

PR #3422, opened 2026-03-27
Add Python support for Claude/OpenAI MCP server tools

juliusgeo:
# Description

Add support for creating MCP server tools that are compatible with both Claude and OpenAI agent SDKs. 

Fixes # (issue)

## Type of change


sdks/python/examples/agent/agent.py
```diff
@@ -0,0 +1,40 @@
...
+    # Wrap the tool in an in-process MCP server
+    weather_server = create_sdk_mcp_server(
+        name="weather",
+        version="1.0.0",
+        tools=[temp_tool],
+    )
+
+    options = ClaudeAgentOptions(
+        mcp_servers={"weather": weather_server},
+        allowed_tools=["mcp__weather__get_temperature"],
```
mrkaye97: where does this name come from?

sdks/python/examples/agent/agent_claude.py
```diff
@@ -0,0 +1,40 @@
...
+        allowed_tools=["mcp__weather__get_temperature"],
+    )
+
+    async for message in query(
+        prompt="What's the temperature in San Francisco?",
+        options=options,
+    ):
+        print(message)
+        # ResultMessage is the final message after all tool calls complete
+        if isinstance(message, ResultMessage) and message.subtype == "success":
```
mrkaye97, line 32: I'm not familiar with the claude agent sdk so maybe this is recommended, but needing the `isinstance` check here feels unfortunate to me

sdks/python/examples/agent/worker.py
```diff
@@ -0,0 +1,54 @@
...
+        response = await client.get(
+            "https://api.open-meteo.com/v1/forecast",
+            params={
+                "latitude": input.coords.latitude,
+                "longitude": input.coords.longitude,
+                "current": "temperature_2m",
+                "temperature_unit": "fahrenheit",
+            },
+        )
+        data = response.json()
```
mrkaye97: I'm kinda surprised we don't need to `await` this? I haven't used `httpx` much 🤷

sdks/python/examples/agent/worker.py
```diff
@@ -0,0 +1,54 @@
...
+class TemperatureContent(BaseModel):
+    text: str
+
+
+temp_workflow = hatchet.workflow(
+    name="get_temperature", input_validator=TemperatureInput
+)
+
+
+@temp_workflow.task(name="get_temperature")
```
mrkaye97: small nit: I usually don't name these tasks - if you don't provide a name, we'll just use the function name

sdks/python/examples/agent/worker.py
```diff
@@ -0,0 +1,54 @@
...
+        )
+        data = response.json()
+
+    return TemperatureContent(
+        text=f"Temperature in {input.location_name}: {data['current']['temperature_2m']}°F"
+    )
+
+
+def main() -> None:
+    worker = hatchet.worker("test-worker", slots=1, workflows=[temp_workflow])
```
mrkaye97: small nit: I've been trying to remove `slots=1` from examples in various places

sdks/python/hatchet_sdk/runnables/workflow.py
```diff
@@ -1250,6 +1252,45 @@ def my_task(input, ctx) -> None:
...
+
+        def handle_field(field: type) -> Any:
+            if hasattr(field, "model_fields"):
+                return {
+                    k: handle_field(v.annotation) for k, v in field.model_fields.items()
+                }
+            return field
+
+        def basemodel_to_schema(validator: TypeAdapter[TWorkflowInput]) -> Any:
+            return handle_field(validator.core_schema["cls"])
```
mrkaye97: I wonder if we can do `validator.validate_python` instead? `TypeAdapter` is pretty magic, not sure if you've tried that

sdks/python/examples/agent/agent_claude.py
juliusgeo: This is copied directly from the Claude agent sdk example lol, but yeah it does feel weird.

sdks/python/examples/agent/agent.py
juliusgeo: Good catch, this name is hardcoded in the Claude examples but is always "mcp__{server_name}__{tool_name}" essentially

mrkaye97: gotcha, yeah it's similar in the currently-abandoned pydantic ai integration 🤷

sdks/python/examples/agent/worker.py
juliusgeo: fixed!

sdks/python/examples/agent/worker.py
juliusgeo: fixed!

sdks/python/hatchet_sdk/runnables/workflow.py
juliusgeo: Fixed by removing in favor of json_schema

sdks/python/examples/agent/worker.py
juliusgeo: Don't need to await this, it gets awaited at the `client.get` call

mrkaye97 (approved):
🙌

promptless-for-oss:

Triggered by [PR #3422](https://github.com/hatchet-dev/hatchet/pull/3422)

Created documentation for the new `mcp_tool()` method that enables Hatchet workflows and standalone tasks to be exposed as MCP tools for Claude's Agent SDK. The new guide covers installation (`pip install "hatchet-sdk[claude]"`), API reference, and usage patterns.

Review at https://app.gopromptless.ai/suggestions/9c18c163-0617-40df-a4e8-f78254fc35c9

sdks/python/examples/agent/agent_claude.py
```diff
@@ -0,0 +1,46 @@
...
+    ResultMessage,
+)
+
+from examples.agent.worker import temp_workflow, get_temperature_standalone
+
+
+async def main() -> None:
+    # You can use a workflow
+    temp_tool = temp_workflow.mcp_tool(
+        "Get the current temperature at a location",
```
grutt: should this description live on the workflow/task decorator?

sdks/python/examples/agent/agent.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+async def main() -> None:
+    # You can use a workflow
+    temp_tool = temp_workflow.mcp_tool(
+        "Get the current temperature at a location",
+    )
+
+    # Or a standalone task
+    temp_tool = get_temperature_standalone.mcp_tool(
```
grutt: nit: temp reads as temporary in my brain

sdks/python/examples/agent/agent.py
```diff
@@ -0,0 +1,46 @@
...
+    # Wrap the tool in an in-process MCP server
+    weather_server = create_sdk_mcp_server(
+        name="weather",
+        version="1.0.0",
+        tools=[temp_tool],
+    )
+
+    options = ClaudeAgentOptions(
+        mcp_servers={"weather": weather_server},
+        allowed_tools=[f"mcp__{weather_server["name"]}__{temp_tool.name}"],
```
grutt: helper for this weird string pattern?

sdks/python/examples/agent/agent_claude.py
juliusgeo: I think it should be an argument because this description is not just a general description of a workflow/task, but specifically the description that Claude will use to decide whether to pick this tool. I could go either way though

grutt: i do think its important that this at least lives with the workflow since this will likely be reused across multiple MCPs. we also already have a description on the wf/task so this is confusing imo.

ill commit what im thinking

sdks/python/hatchet_sdk/runnables/workflow.py
```diff
@@ -613,6 +625,75 @@ async def aio_delete(self) -> None:
...
+            raise ValueError(
+                f"Runnable '{self.config.name}' has no description. "
+                "Set description= when defining the workflow or task."
+            )
+        description = self.config.description
+        input_schema = self.input_validator.json_schema()
+        if isinstance(self, Workflow):
+            match provider:
+                case MCPProvider.CLAUDE:
+                    from .mcp.claude import workflow_to_claude_mcp
```
mrkaye97: let's use absolute imports here to keep consistent with the other ones

sdks/python/hatchet_sdk/runnables/mcp/claude.py
```diff
@@ -0,0 +1,59 @@
...
+    raise ModuleNotFoundError(
+        "To use the mcp_tool method with Claude as a provider, you must install Hatchet's `claude` extra using (e.g.) `pip install hatchet-sdk[claude]`"
+    ) from e
+
+
+def task_to_claude_mcp(
+    runnable: "Standalone[TWorkflowInput, R]",
+    input_schema: dict[str, Any],
+    description: str,
+    **kwargs: Any,
```
mrkaye97, line 21: ideally these wouldn't be untyped kwargs. I wonder if we can use `ParamSpec` for this somehow or something similar? It'd be nice to be able to show + get typing for the kwargs being passed along, but not a huge deal if it's not doable

mrkaye97 (approved):
couple small comments but looks great to me 👍

sdks/python/hatchet_sdk/runnables/workflow.py
juliusgeo: Oops, fixed. :)

sdks/python/hatchet_sdk/runnables/mcp/claude.py
juliusgeo: I think this would be good for the future, but I looked into it and it seems our only option is defining a TypedDict for just the parameters we aren't passing in. ParamSpec only works if all of the arguments are forwarded, IIUC.

sdks/python/hatchet_sdk/runnables/workflow.py
```diff
@@ -613,6 +625,75 @@ async def aio_delete(self) -> None:
...
+
+                    return workflow_to_claude_mcp(
+                        self, input_schema, description, **kwargs
+                    )
+                case MCPProvider.OPENAI:
+                    from hatchet_sdk.runnables.mcp.openai import workflow_to_openai_mcp
+
+                    return workflow_to_openai_mcp(
+                        self, input_schema, description, **kwargs
+                    )
```
mrkaye97, line 685: small thing: we might be able to import `typing.Never` and have an assertion here like:

```python
default:
    exhaustive_check: Never = provider
```

which would make sure that we never forget to handle one of these cases (same below here too)

mrkaye97 (approved):

PR #3541, opened 2026-04-03
Lint all supported Python versions

juliusgeo:
# Description

To avoid issues surrounding features available in later Python versions that aren't in 3.10, lint all supported Python versions

Fixes # (issue)

## Type of change

### DIFF
--- a/sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
+++ b/sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
@@ -1,5 +1,4 @@
 import json
-import sys
 from collections.abc import Callable, Collection, Coroutine, Sequence
 from datetime import datetime, timedelta, timezone
 from importlib.metadata import version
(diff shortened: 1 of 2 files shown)


sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
```diff
@@ -451,7 +452,7 @@ def extract_bound_args(
         args: tuple[Any, ...],
         kwargs: dict[str, Any],
     ) -> list[Any]:
-        try:
+        if sys.version_info >= (3, 14):
```
mrkaye97: is this reliable? I feel like we did the `try / except` because it wasn't, but might be misremembering

juliusgeo: hmm, I guess from the comments 3.14+ can fail to evaluate annotations lazily, so perhaps need both of these. Mainly switched because the try: except doesn't work with the type checker

mrkaye97: maybe safe to just add `type: ignore` for this line 🤷

mrkaye97 (approved):

PR #3554, opened 2026-04-05
Feat: Direct task + dag status updates

mrkaye97:
# Description

Removes the temp table logic for status updates in favor of just directly updating statuses when we process incoming task event messages. The basic ideas here are:

1. We can significantly simplify the query logic to hopefully take some pressure off the database for performing the updates
2. We can also take pressure off the db by removing all the intermediate CRUD into / out of the temp tables
3. If we do this directly, we don't need to pub any extra messages over the internal queue, which also would add pressure on Rabbit

## Type of change


internal/services/controllers/olap/controller.go
```diff
@@ -304,38 +304,38 @@ func (o *OLAPControllerImpl) Start() (func() error, error) {
...
-	pollIntervalSec := 2
+	// pollIntervalSec := 2
 
 	// Override with config value if available
-	if o.olapConfig != nil && o.olapConfig.PollInterval > 0 {
-		pollIntervalSec = o.olapConfig.PollInterval
-	}
+	// if o.olapConfig != nil && o.olapConfig.PollInterval > 0 {
+	// 	pollIntervalSec = o.olapConfig.PollInterval
+	// }
```
mrkaye97, line 312: will add these back later when I figure out how to cut over, just commented out for now for testing locally

pkg/repository/sqlcv1/olap.sql
```diff
@@ -1019,6 +1020,161 @@ SELECT
...
+        UNNEST(@workerIds::UUID[]) AS worker_id,
+        UNNEST(@retryCounts::INTEGER[]) AS retry_count
+), locked_tasks AS (
+    SELECT *
+    FROM v1_tasks_olap
+    WHERE (inserted_at, id, tenant_id) IN (
+        SELECT task_inserted_at, task_id, tenant_id
+        FROM inputs
+    )
+    FOR UPDATE
```
abelanger5: need an order by here I think

pkg/repository/sqlcv1/olap.sql
```diff
@@ -1019,6 +1020,161 @@ SELECT
...
+    SET
+        readable_status = i.readable_status,
+        latest_retry_count = i.retry_count,
+        latest_worker_id = CASE
+            WHEN i.worker_id != '00000000-0000-0000-0000-000000000000'::uuid THEN i.worker_id
+            ELSE t.latest_worker_id
+        END
+    FROM locked_tasks lt
+    JOIN inputs i ON (i.tenant_id, i.task_id, i.task_inserted_at) = (lt.tenant_id, lt.id, lt.inserted_at)
+    WHERE
```
abelanger5: shouldn't we be doing this join / where clause in the `locked_tasks` cte?

pkg/repository/sqlcv1/olap-overwrite.sql.go
```diff
@@ -586,3 +586,248 @@ func (q *Queries) FindV1OLAPPayloadPartitionsBeforeDate(ctx context.Context, db
...
+    external_id,
+    display_name,
+    input,
+    additional_metadata,
+    dag_id,
+    dag_inserted_at,
+    parent_task_external_id,
+    is_durable
+FROM inputs
+ON CONFLICT (inserted_at, id) DO NOTHING
```
mrkaye97, line 664: unclear if this is correct, gonna circle back

sql/schema/v1-olap.sql
```diff
@@ -157,7 +157,7 @@ CREATE TABLE v1_tasks_olap (
     parent_task_external_id UUID,
     is_durable BOOLEAN NOT NULL DEFAULT FALSE,
 
-    PRIMARY KEY (inserted_at, id, readable_status)
```
mrkaye97, line 160: drive by: forgot to do this in the last PR

pkg/repository/sqlcv1/olap-overwrite.sql.go
mrkaye97: update: think it's good since we write these with the defaults, so we never would want to update here

pkg/repository/sqlcv1/concurrency.sql
```diff
@@ -144,6 +144,13 @@ WHERE
 -- name: AdvisoryLock :exec
 SELECT pg_advisory_xact_lock(@key::bigint);
 
+-- name: AdvisoryLockMany :exec
+WITH keys AS (
+    SELECT UNNEST(@keys::BIGINT[]) AS key
+)
+
+SELECT pg_advisory_xact_lock(key) FROM keys;
+
```
abelanger5, line 158: curious if we just want an `order by` here to prevent deadlocks, instead of needing to manage it application-side?

pkg/repository/olap.go
```diff
@@ -1562,10 +1568,153 @@ func getCacheKey(event sqlcv1.CreateTaskEventsOLAPParams) string {
...
+		keys = append(keys, workflowRunAdvisoryInt(id))
+	}
+
+	sort.Slice(keys, func(i, j int) bool { return keys[i] < keys[j] })
+
+	for _, key := range keys {
+		if err := r.queries.AdvisoryLock(ctx, tx, key); err != nil {
+			return fmt.Errorf("failed to acquire advisory lock for workflow run: %w", err)
+		}
+	}
```
abelanger5: nit: this doesn't use the many function?

mrkaye97: oops 🤦 thanks, guess I forgot to change that over

pkg/repository/sqlcv1/concurrency.sql
mrkaye97: yeah, that's a good idea, will do

abelanger5 (approved):

PR #3581, opened 2026-04-08
refactor: make rbac package generic and reusable

abelanger5:
# Description

Moves the RBAC package to its own exported `/pkg/auth/rbac` directory so we can reuse it elsewhere. Decouples the OSS implementation from the permissions implementation. 

Drive-by fix for CI issues: 
- Runs the security check in a goroutine instead of `defer` - no change in risk since the method panicking out of `defer` would prevent startup regardless. There was some evidence locally of the security check blocking startup for a significant amount of time
- Rebalances inactive partitions during startup of the scheduler so that tenants which have never been active on a scheduler get started immediately. It seems that this + the security check being slow was preventing the tenant from scheduling workload immediately for the load tests
- Fixes a race condition on shutdown where we called `wg.Add` from a retry loop when our cleanup method was in `wg.Wait` 

## Type of change


juliusgeo (approved):
LGTM!

mnafees (approved):

PR #3584, opened 2026-04-09
Fix: Python worker pause on shutdown

mrkaye97:
# Description

### Changed

- Fixes a bug in the shutdown handlers that wouldn't correctly trigger graceful shutdown if only the parent process received a shutdown signal like `SIGTERM`, which might often be the case on e.g. Kubernetes.
- Fixes an issue where we generated protobufs using a more recent grpcio version than the minimum allowed, causing breakages on older versions.

## Type of change

### DIFF
--- a/sdks/python/hatchet_sdk/worker/worker.py
+++ b/sdks/python/hatchet_sdk/worker/worker.py
@@ -732,18 +732,8 @@ class Worker:
 
         self._killing = True
 
-        for process in [
-            self._action_listener_process,
-            self._durable_action_listener_process,
-        ]:
-            if process is not None and process.pid is not None:
-                try:
-                    if process.is_alive():
-                        os.kill(process.pid, signal.SIGTERM)
-                except Exception:
-                    pass
-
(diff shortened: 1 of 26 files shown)


sdks/python/hatchet_sdk/worker/worker.py
```diff
@@ -732,8 +733,19 @@ async def exit_gracefully(self) -> None:
...
+        # but only the listeners know what worker id we're on, so instead we need
+        # to propagate the SIGTERM to all child listener processes. Each child's SIGTERM
+        # handler calls pause_task_assignment() to pause assignment to this worker
+        for process in [
+            self._action_listener_process,
+            self._durable_action_listener_process,
+        ]:
+            if process is not None and process.pid is not None and process.is_alive():
+                with contextlib.suppress(ProcessLookupError):
+                    os.kill(process.pid, signal.SIGTERM)
```
mrkaye97, line 746: this is the change

grutt (approved):

PR #3137, opened 2026-03-02
Feat: Dynamic worker label assign

mrkaye97:
# Description

Adds trigger-time worker label assignment, so we can route tasks to a specific worker or worker pool at trigger time, instead of based on a statically defined preferred label (or set of labels).

Basic API is to do this:

```python
simple.run_no_wait(
    options=TriggerWorkflowOptions(
        desired_worker_label={"foo": DesiredWorkerLabel(value="baz", required=True)}
    )
)
```

## Type of change

- [x] New feature (non-breaking change which adds functionality)

PR #3591, opened 2026-04-10
Fix: Worker labels not respected on retry

mrkaye97:
# Description

In #3137, I made an update to `v1_task_update_function` but forgot to copy paste it into the migration, which caused worker labels to not be respected on retries. This should fix that!

## Type of change


pkg/repository/sqlcv1/tasks-overwrite.go
```diff
@@ -844,13 +844,13 @@ WITH input AS (
     INSERT INTO v1_queue_item (
         tenant_id, queue, task_id, task_inserted_at, external_id, action_id, step_id,
         workflow_id, workflow_run_id, schedule_timeout_at, step_timeout, priority,
-        sticky, desired_worker_id, retry_count
+        sticky, desired_worker_id, retry_count, desired_worker_label
```
mrkaye97, line 847: drive-by fix for eviction too, looks like this slipped when we implemented the eviction stuff

grutt (approved):

PR #3603, opened 2026-04-10
Fix: Remove status partitioning on `v1_(runs|dags|tasks)_olap`

mrkaye97:
# Description

Removes status partitioning, which we think will improve performance in multiple parts of the system a bunch. It'll make updates easier since we won't ever need to move rows between partitions (which postgres seems to not recommend), and it should improve query performance a bunch too. Also removed the old tenant index on the runs table in favor of the new one that's superseded it.

Same basic process as before:

1. Creates new tables that are basically copies of the existing ones, except that they have only `(inserted_at, id)` as the PK and are partitioned differently
2. Creates triggers to insert data into those tables
3. Backfills those tables
4. Progresses to a second migration where we drop the old tables rename the new ones to cut over

## Type of change


abelanger5 (approved):

cmd/hatchet-migrate/migrate/migrations/20260410190713_v1_0_97.go
```diff
@@ -0,0 +1,521 @@
...
+	workflow_version_id,
+	readable_status,
+	input,
+	additional_metadata,
+	parent_task_external_id,
+	total_tasks
+FROM v1_dags_olap
+ON CONFLICT (inserted_at, id) DO NOTHING`
+
+func up20260410190713(ctx context.Context, db *sql.DB) error {
```
abelanger5, line 494: idea to speed things up: we can do runs, tasks and dags in parallel by starting 3 separate transactions in parallel and running the migration in non-tx mode

cmd/hatchet-migrate/migrate/migrations/20260410190713_v1_0_97.go
```diff
@@ -0,0 +1,521 @@
...
+		PRIMARY KEY (inserted_at, id)
+	) PARTITION BY RANGE (inserted_at);
+
+	FOR r IN (
+		SELECT c.relname
+		FROM pg_class c
+		JOIN pg_inherits i ON c.oid  = i.inhrelid
+		JOIN pg_class p ON p.oid  = i.inhparent
+		WHERE
+			p.relname = '` + oldParent + `'
```
abelanger5, line 39: rare edge case, but what happens if this runs while we add or drop a partition? I'm guessing it won't cause much of an impact on the add partition since we have 24 hours before data starts getting written to the partition. But more curious about the case when we delete a partition.

cmd/hatchet-migrate/migrate/migrations/20260410190713_v1_0_97.go
```diff
@@ -0,0 +1,521 @@
...
+	inserted_at,
+	external_id,
+	readable_status,
+	kind,
+	workflow_id,
+	workflow_version_id,
+	additional_metadata,
+	parent_task_external_id
+FROM v1_runs_olap
+ON CONFLICT (inserted_at, id) DO NOTHING`
```
abelanger5: since the old `v1_runs_olap` could technically have multiple statuses, do we want to pick the larger status here or just not worry about that edge case?

mrkaye97: yeah, initially I was thinking that since the runs table doesn't have a retry count we don't know how to do this, but maybe we can naively just pick the largest one in case of a conflict and it'll work...

cmd/hatchet-migrate/migrate/migrations/20260410190713_v1_0_97.go
mrkaye97: I guess we'd either hang trying to drop the partition or the migration would fail with a no such table error and we'd retry

PR #3606, opened 2026-04-11
Proposal: DI-style declaration of task parents

mrkaye97:
# Description

Been considering this idea of using a DI-style syntax for declaring parent tasks in a DAG instead of the declarative way we have. Idea below:

```python
@dag_workflow.task()
async def step1(input: EmptyModel, ctx: Context) -> StepOutput:
    return StepOutput(random_number=random.randint(1, 100))


@dag_workflow.task()
async def step2(
    input: EmptyModel,
    ctx: Context,
    step_1: Annotated[StepOutput, Parent(step1)],
) -> RandomSum:
    one = step_1.random_number

    return RandomSum(sum=one + 1)
```

Decided to push this to collect feedback, but ultimately I think we should not adopt this for two reasons:

1. `Annotated` breaks the type checker
2. It's conceptually confusing that a) all of the other declarative config for the task lives on the decorator and b) if you want to use DAG conditions, you need to refer to parents on the decorator

So I'll probably close this in a few days

## Type of change

### DIFF
--- a/sdks/python/examples/dag/worker.py
+++ b/sdks/python/examples/dag/worker.py
@@ -1,10 +1,10 @@
 import random
 import time
 from datetime import timedelta
-
+from typing import Annotated
 from pydantic import BaseModel
 
-from hatchet_sdk import Context, EmptyModel, Hatchet
+from hatchet_sdk import Context, EmptyModel, Hatchet, Parent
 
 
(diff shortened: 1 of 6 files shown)


mrkaye97:
closing for now!

PR #3614, opened 2026-04-13
feat: sync repository for syncing tenants

abelanger5:
# Description

Adds a `SyncRepository` which allows tenants, users, tenant memberships and invites to be synced into a single Hatchet instance. This is required for some baseline multi-regional/sharded instance work. 

## Type of change


pkg/repository/sqlcv1/sync.sql
```diff
@@ -0,0 +1,149 @@
+-- name: SyncUpsertTenant :one
+INSERT INTO "Tenant" (
```
grutt, line 2: is there anyway we can "exhaustiveness check" if we add cols in the future....

grutt (approved):

PR #3617, opened 2026-04-14
fix: update skills install paths to align with agent defaults

wsehl:
# Description

hatchet skills are really nice, but right now `hatchet skills install` doesn't make much sense as most agent tooling expect either `.agents/skills` or `.claude/skills` so it doesn't integrate with Claude Code, Codex, OpenCode, Pi, etc, out of the box.

## Type of change

## What's Changed

here i just changed paths to install skills into `.agents/skills/hatchet-cli` with a `.claude/skills/hatchet-cli` symlink, but it would be nice to go further and move skills into the repo root `skills/hatchet-cli` to make it distributable via https://skills.sh/, which is really nice and widely used

### DIFF
--- a/cmd/hatchet-cli/cli/skills.go
+++ b/cmd/hatchet-cli/cli/skills.go
@@ -48,9 +48,10 @@ var skillsInstallCmd = &cobra.Command{
 	Short: "Install the Hatchet CLI agent skill package into your project",
 	Long: `Install Hatchet CLI agent skills into your project.
 
-Creates the skill directory structure under {dir}/skills/hatchet-cli/ and
-appends a reference section to the project AGENTS.md file.`,
-	Example: `  # Install to current directory (creates ./skills/hatchet-cli/)
+Creates the skill directory structure under {dir}/.agents/skills/hatchet-cli/,
+creates {dir}/.claude/skills/hatchet-cli as a symlink, and appends a reference
+section to the project AGENTS.md file.`,
+	Example: `  # Install to current directory
   hatchet skills install
 
   # Install to a custom base directory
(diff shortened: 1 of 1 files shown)


promptless-for-oss:

Triggered by this PR

Updated documentation for the `hatchet skills install` command to reflect the new default installation path (`.agents/skills/hatchet-cli/`) and the Claude Code symlink (`.claude/skills/hatchet-cli`). Changes include updates to the CLI reference and coding agents guide.

Review: [Add Hatchet Skills CLI Reference](https://app.gopromptless.ai/suggestions/3e7acc76-4b7b-421b-87f8-bdd90b9bcaf1)

cmd/hatchet-cli/cli/skills.go
```diff
@@ -67,7 +68,7 @@ func init() {
 	rootCmd.AddCommand(skillsCmd)
 	skillsCmd.AddCommand(skillsInstallCmd)
 
-	skillsInstallCmd.Flags().StringVarP(&skillsInstallDir, "dir", "d", ".", "Target base directory (skill installs under {dir}/skills/hatchet-cli/)")
+	skillsInstallCmd.Flags().StringVarP(&skillsInstallDir, "dir", "d", ".", "Target base directory (skill installs under {dir}/.agents/skills/hatchet-cli/)")
```
gregfurman, line 71: Should we not also mention `.claude` here?

gregfurman (approved):
Hey! Thanks for the contribution.

While not related to your changes speficially, I'm wondering about the devex of automatically assuming we should be installing for `.claude` 🤔  I'll follow up with the team on this.

Otherwise, LGTM. Should be fine to include in next release 👍

cmd/hatchet-cli/cli/skills.go
wsehl: well, yes, since there is no option it would be better to say explicitly what is going to happen.

updated the comment to reflect that.

(sorry for late reply)

cmd/hatchet-cli/cli/skills.go
```diff
@@ -67,7 +68,7 @@ func init() {
 	rootCmd.AddCommand(skillsCmd)
 	skillsCmd.AddCommand(skillsInstallCmd)
 
-	skillsInstallCmd.Flags().StringVarP(&skillsInstallDir, "dir", "d", ".", "Target base directory (skill installs under {dir}/skills/hatchet-cli/)")
+	skillsInstallCmd.Flags().StringVarP(&skillsInstallDir, "dir", "d", ".", "Target base directory (creates .agents/skills/hatchet-cli, .claude/skills/hatchet-cli, and AGENTS.md)")
```
gregfurman, line 71: @grutt wdyt of including a `.claude` by default?

PR #3619, opened 2026-04-14
feat(analytics): support explicit count values in Count method

### LINKED ISSUE #3382: [FEAT] Extend Analytics Interface

The Analytcis interface is quite constrained in its functionality. For example, if I wanted to count multiple times, I'd have to either call `Count(...)` in a loop (without concern for spamming the span table) OR try and set my value within `properties.Properties` as a label/attribute, unnecessarily increasing cardinality.

Extend the offering of the `analytics` package and associated `Analytics` interface to instead accept more types and options for recording of metrics/data/analytics.

i.e

```golang
// could make a private Property interface to ensure whatever props we pass in are valid i.e

type Property interface {
	private()
}

type Properties map[string]interface{}
func (Properties) private() {}

type Increment int
func (Increment) private() {}

type Analytics interface {
   // ...
   Count(ctx context.Context, resource Resource, action Action, props ...Property)
   // ...
}
```

Then we can increment this like:

```golang
oc.a.Count(ctx, analytics.OtelSpan, analytics.Create, analytics.Increment(len(spans)))
```

None that don't either require extending the interface or introducing some breaking changes.

**Additional context**

Follow-up from discussion in https://github.com/hatchet-dev/hatchet/pull/3375#discussion_r2981760483

wolf-06:
# Description
Adds support for explicit count values in `analytics.Count()` by extracting int64 values from the last `Properties` argument when multiple are provided.
## What's Changed
- Extract count from last Properties argument if multiple provided, summing all int64 values
- Fall back to count=1 when no explicit count is provided
- Add tests covering default, explicit, sum, and fallback behavior
- Update comment to explain the feature
## Type of change
---
FIXES #3382

### DIFF
--- a/pkg/analytics/posthog/posthog.go
+++ b/pkg/analytics/posthog/posthog.go
@@ -147,11 +147,27 @@ func (p *PosthogAnalytics) Count(ctx context.Context, resource analytics.Resourc
 			merged[k] = v
 		}
 	}
+
+	// Extract count from last properties arg if multiple are provided.
+	// Allows passing aggregated counts instead of always using 1.
+	var n int64
+	if len(props) > 1 && props[len(props)-1] != nil {
+		for _, v := range props[len(props)-1] {
+			if value, ok := v.(int64); ok {
+				n += value
+			}
+		}
+	}
+
+	if n == 0 {
+		n = 1
+	}
+
 	if source := analytics.SourceFromContext(ctx); source != "" {
 		merged["source"] = string(source)
 	}
 
-	p.aggregator.Count(resource, action, tid, tokenID, 1, merged)
+	p.aggregator.Count(resource, action, tid, tokenID, n, merged)
 }
 
 func (p *PosthogAnalytics) flushCount(resource analytics.Resource, action analytics.Action, tenantID uuid.UUID, tokenID *uuid.UUID, count int64, properties analytics.Properties) {
--- a/pkg/analytics/posthog/posthog_test.go
+++ b/pkg/analytics/posthog/posthog_test.go
@@ -1,0 +1,223 @@
+package posthog
+
+import (
+	"context"
+	"sync"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/rs/zerolog"
+
+	"github.com/hatchet-dev/hatchet/pkg/analytics"
+)
+
+var testLogger = zerolog.Nop()
+
+type countCall struct {
+	Resource   analytics.Resource
+	Action     analytics.Action
+	TenantID   uuid.UUID
+	TokenID    *uuid.UUID
+	Count      int64
+	Properties analytics.Properties
+}
+
+type countRecorder struct {
+	mu    sync.Mutex
+	calls []countCall
+}
+
+func (r *countRecorder) record(resource analytics.Resource, action analytics.Action, tenantID uuid.UUID, tokenID *uuid.UUID, count int64, properties analytics.Properties) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.calls = append(r.calls, countCall{
+		Resource:   resource,
+		Action:     action,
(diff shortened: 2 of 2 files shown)


pkg/analytics/posthog/posthog.go
```diff
@@ -147,11 +147,27 @@ func (p *PosthogAnalytics) Count(ctx context.Context, resource analytics.Resourc
 			merged[k] = v
 		}
 	}
+
+	// Extract count from last properties arg if multiple are provided.
+	// Allows passing aggregated counts instead of always using 1.
```
gregfurman, line 152: Tbh I still think we should be passing in a structured type to determine/override this value, where necessary instead of relying on a positional arg in the varadic params.

Like, it's more explicit that way i.e

```golang
a.Count(ctx, analytics.Log, analytics.Get, analytics.Properties{}, analytics.Increment(5))
```

versus:

```golang
a.Count(ctx, analytics.Log, analytics.Get, analytics.Properties{}, analytics.Properties{"": 5})
```

where the latter could be error prone as well since it means we'd need to document this positional behaviour.


This typed-approach also means we can extend supported operations i.e

```golang
a.Count(ctx, analytics.Log, analytics.Get, analytics.Properties{}, analytics.Decrement(5))
a.Count(ctx, analytics.Log, analytics.Get, analytics.Properties{}, analytics.Set(0)) // or maybe .Reset() ?
```

Then if we have other metric types (i.e a gauge or whatever) we can extend this approach to suit.

pkg/analytics/posthog/posthog.go
```diff
@@ -147,11 +147,27 @@ func (p *PosthogAnalytics) Count(ctx context.Context, resource analytics.Resourc
...
+	var n int64
+	if len(props) > 1 && props[len(props)-1] != nil {
+		for _, v := range props[len(props)-1] {
```
gregfurman, line 155: Unsure this is desirable due to the typing of `analytics.Properties`.

Like, `analytics.Properties` is a `map[string]any` which means (in the case of your changes) the last arg would still have to be correctly typed as such. So if I wanted to increment the count by `5` I'd need to pass a contrived map type where the keys will be ignored i.e

```golang
a.Count(ctx, analytics.Log, analytics.Get, analytics.Properties{},  analytics.Properties{"foo": 5}))
```

Also, if we pass in two (or more) properties, the caller has no way of knowing the final arg will have all of its values used for incrementing. This ambigious behaviour could lead to unintentionally passing props in that we expect to be included as tags/labels but get used to increment metrics.

pkg/analytics/posthog/posthog.go
```diff
@@ -147,11 +147,27 @@ func (p *PosthogAnalytics) Count(ctx context.Context, resource analytics.Resourc
...
+			if value, ok := v.(int64); ok {
```
gregfurman, line 156: This conversion means we're only ever supporting an `int64` being explicitly passed in, meaning other numeric types being used would get skipped over.

i.e

```golang
a.Count(ctx, analytics.Log, analytics.Get, analytics.Properties{},  analytics.Properties{"foo": uint64(5)}))
```

Would see `1` being added since a `uint64` is not an `int64` -- hence it's skipped as type-checking fails.

gregfurman (changes requested):
Thanks for picking this up! I think the current approach needs some more consideration since (IMO) it's prone to incorrect usage and is not very self-documenting.

Wdyt of the approach outlined in https://github.com/hatchet-dev/hatchet/issues/3382 that extends the `analytics.Properties` and uses a more explicit (and structured) typing system to determine values?

pkg/analytics/posthog/posthog.go
wolf-06: Fair enough, 
will move with original approach of using structured property, which ig will also solve to other concerns automatically.

gregfurman:
Closing as stale.

PR #3623, opened 2026-04-15
feat(docs): python support agent cookbook example

BloggerBust:
# Description

Adds a new cookbook docs page for building a support agent with Hatchet, along with the supporting Python example and trigger/example files.

This PR introduces a practical end-to-end workflow example that:
- receives a support ticket
- triages it
- generates an initial reply
- waits for either a customer reply event or a timeout
- resolves when a reply arrives
- escalates when the timeout fires first

It also adds a new Workflow Patterns section to the cookbooks index so this example is separate from the existing webhook-focused cookbook entries.

The Python example uses the snippet markers and was validated locally against a live Hatchet instance. The current implementation supports live Claude interaction when `ANTHROPIC_API_KEY` is set, and a fixed fallback reply when it is not.

## Type of change

## What's Changed

- [x] Add a new cookbook page: `workflow-support-agent.mdx`
- [x] Add a new "Workflow Patterns" section to the cookbooks index
- [x] Add a Python support agent example under `examples/python/support_agent/`
- [x] Add a snippet region for the trigger script
- [x] Register the support agent example in the central Python example worker
- [x] Add E2E coverage for the resolved and timeout/escalation paths
- [x] Document a durable workflow pattern that uses `consider_events_since` to safely capture early reply events

### DIFF
--- a/frontend/docs/pages/cookbooks/workflow-support-agent.mdx
+++ b/frontend/docs/pages/cookbooks/workflow-support-agent.mdx
@@ -1,0 +1,132 @@
+import { Steps } from "nextra/components";
+import { snippets } from "@/lib/generated/snippets";
+import { Snippet } from "@/components/code";
+
+# How to Create a Support Agent Using Hatchet
+
+Support workflows are a natural fit for Hatchet because they combine human input, long waits, and time-based escalation. In this cookbook, we will build a simple support agent that triages a ticket, generates an initial reply, then waits for either a customer response or a timeout. If the customer replies, the workflow resolves. If no reply arrives in time, the workflow escalates the ticket to a human team.
+
+## What this example builds
+
+This example implements the following durable support workflow:
+
+```
+      +-------------------------+
+      | Support ticket received |
+      +-------------------------+
+                  |
+                  v
+        +--------------------+
+        | Triage the ticket  |
+        +--------------------+
+                  |
+                  v
+      +--------------------------+
+      | Generate initial reply   |
+      +--------------------------+
+                  |
+                  v
+    +------------------------------+
+    | Wait for reply or timeout    |
+    +------------------------------+
+             /            \
+            v              v
++----------------+   +--------------------+
+| Customer reply |   | Timeout fires      |
++----------------+   +--------------------+
+        |                     |
+        v                     v
++----------------+  +----------------------+
+| Resolve ticket |  | Escalate to human    |
++----------------+  +----------------------+
+```
+
+The workflow is built around Hatchet’s durable execution model, child task spawning, and event waiting primitives.
+
+## Setup
+
+<Steps>
+
+### Prepare your environment
+
+To run this example, you will need:
+
+- a working local Hatchet environment or access to Hatchet Cloud
+- the Python SDK example environment
+- optionally, an `ANTHROPIC_API_KEY`
+
+Without `ANTHROPIC_API_KEY`, the example still runs using a canned fallback reply.
+
+### Define the models
+
+Start with a few small Pydantic models for the workflow input and outputs.
+
+<Snippet src={snippets.python.support_agent.worker.models} />
+
+These models keep the task boundaries explicit and make the workflow easier to follow.
+
+### Add the child tasks
+
+The durable workflow delegates its work to a few small child tasks.
+
+First, add a task to classify the incoming ticket:
+
+<Snippet src={snippets.python.support_agent.worker.triage_task} />
+
+Next, add a task to generate the initial support reply. This task calls Claude when `ANTHROPIC_API_KEY` is present, and falls back to a fixed response when it is not.
+
+<Snippet src={snippets.python.support_agent.worker.generate_reply_task} />
+
+Finally, add a task to represent escalation to a human team:
+
+<Snippet src={snippets.python.support_agent.worker.escalate_task} />
+
+### Build the durable workflow
+
+Now tie everything together in a durable Hatchet workflow.
+
+<Snippet src={snippets.python.support_agent.worker.support_agent_workflow} />
+
+This workflow:
+
+1. triages the ticket
+2. generates an initial reply
+3. waits for either a scoped customer reply event or a timeout
+4. resolves or escalates depending on which condition fires first
+
+The key detail that makes this workflow reliable is `consider_events_since` on the reply event condition. A customer reply could arrive while the workflow is still finishing the triage or reply-generation steps. By using `consider_events_since`, the workflow can still capture that reply when the wait becomes active.
+
+### Trigger the workflow
+
+The example also includes a small trigger script that starts the workflow, pushes a scoped reply event, and waits for the result.
+
+<Snippet src={snippets.python.support_agent.trigger.trigger_the_workflow} />
+
+Because the workflow uses `consider_events_since`, the trigger can push the reply event immediately after starting the support agent.
+
+### Test it
+
+This example includes two end-to-end tests against a live Hatchet instance:
+
+- a resolved path, where the customer reply event arrives before the timeout
+- a timeout path, where no reply arrives and the workflow escalates
+
+Together, these tests validate both branches of the support workflow and confirm that early reply events are handled safely without coordination sleeps.
+
+</Steps>
+
+## Why Hatchet fits this workflow
+
+This example is small, but it illustrates why support workflows map naturally to Hatchet:
+
+- the workflow needs to preserve state across multiple steps
+- it needs to wait for human input
+- it needs to branch on timeout versus event
+- it should not lose early reply events
+- it benefits from being inspectable and testable as a single durable workflow
+
+A simpler queue can handle isolated tasks, but the full interaction becomes much easier to model when the workflow itself is durable.
+
+## Next steps
+
+From here, you could extend this example by integrating with a real ticketing system, escalating only when the customer explicitly asks for a human and supporting multi-turn conversations instead of resolving on the first reply. For a first support-agent example, this version is enough to show the core idea clearly. Hatchet makes it straightforward to build a durable support workflow that can wait, react, and escalate without fragile coordination logic.
--- a/frontend/docs/pages/cookbooks/workflow-support-agent.mdx
+++ b/frontend/docs/pages/cookbooks/workflow-support-agent.mdx
@@ -65,7 +65,7 @@ Keeping triage, reply generation, and escalation as separate tasks keeps the wor
 
 ### Build the durable workflow
 
-Now tie everything together in a [durable Hatchet workflow](/v1/durable-execution).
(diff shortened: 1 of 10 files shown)


promptless-for-oss:

Triggered by [PR #3623](https://github.com/hatchet-dev/hatchet/pull/3623)

The PR adds the cookbook page but is missing the sidebar navigation update in `_meta.js`. This suggestion adds the "Workflow Patterns" separator and the `workflow-support-agent` entry so the new cookbook page will appear in the docs site navigation.

Review: [Add support agent meta entry](https://app.gopromptless.ai/suggestions/89221e19-56f4-416b-b691-a25d0eb3fc68)

BloggerBust:
Added the Workflow Patterns separator and the support agent entry to `frontend/docs/pages/cookbooks/_meta.js`.

sdks/python/examples/support_agent/worker.py
```diff
@@ -0,0 +1,198 @@
...
+# > Triage task
+@hatchet.task(input_validator=SupportTicketInput)
+async def triage_ticket(input: SupportTicketInput, ctx: Context) -> TriageOutput:
+    """Classify the ticket into a category and priority."""
+    subject = input.subject.lower()
+    body = input.body.lower()
+    text = subject + " " + body
+
+    if any(word in text for word in ["bill", "charge", "payment", "invoice"]):
+        category = "billing"
```
mrkaye97, line 56: maybe we can make this and the `priority` into enums if we're always returning one of a few categories? doesn't super duper matter, but one of those things that if we're going to be making recommendations via these examples for how to structure things, I'd love to recommend

sdks/python/examples/support_agent/worker.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+# !!
+
+
+# > Generate reply task
+@hatchet.task(input_validator=SupportTicketInput)
+async def generate_reply(input: SupportTicketInput, ctx: Context) -> ReplyOutput:
+    """Generate an initial support reply using Claude."""
+    api_key = os.environ.get("ANTHROPIC_API_KEY")
```
mrkaye97, line 79: ideally we'd prefer `pydantic-settings` for managing config IMO, or at the very least, read this on worker start and crash immediately so we don't fail tasks at runtime

frontend/docs/pages/cookbooks/workflow-support-agent.mdx
```diff
@@ -0,0 +1,135 @@
...
+            v              v
++----------------+   +--------------------+
+| Customer reply |   | Timeout fires      |
++----------------+   +--------------------+
+        |                     |
+        v                     v
++----------------+  +----------------------+
+| Resolve ticket |  | Escalate to human    |
++----------------+  +----------------------+
+```
```
abelanger5, line 42: nit: can we convert this to a mermaid diagram? we should have support for those and they're styled properly (for example https://docs.hatchet.run/v1/durable-tasks#how-it-works)

sdks/python/examples/support_agent/worker.py
```diff
@@ -0,0 +1,198 @@
...
+    if not api_key:
+        return ReplyOutput(
+            message=f"Thank you for contacting support about: {input.subject}. "
+            "We are looking into this and will get back to you shortly."
+        )
+
+    import importlib
+
+    anthropic = importlib.import_module("anthropic")
+    client = anthropic.AsyncAnthropic(api_key=api_key)
```
mrkaye97, line 90: what's the reason for burying this import? so we don't need to add any deps? (if yes totally fine, although maybe we could also set this example up as an extra or an optional dep so we could rework here a bit)

frontend/docs/pages/cookbooks/workflow-support-agent.mdx
```diff
@@ -0,0 +1,135 @@
+import { Steps } from "nextra/components";
+import { snippets } from "@/lib/generated/snippets";
+import { Snippet } from "@/components/code";
+
+# How to Create a Support Agent Using Hatchet
+
+Support workflows can become difficult to manage and scale once they involve long waits, human replies, and escalation rules. In this cookbook, we will build a simple support agent that triages a ticket, generates an initial reply, and then waits for either a customer response or a timeout. If the customer replies, the workflow resolves. If no reply arrives in time, the workflow escalates the ticket to human support agent.
```
abelanger5, line 7: Can we generalize this beyond support workflows? I think this is a great example of a complex workflow with many steps, timers, classification and routing, and would be good to zoom out a bit in this intro.

frontend/docs/pages/cookbooks/workflow-support-agent.mdx
```diff
@@ -0,0 +1,135 @@
...
+## Setup
+
+<Steps>
+
+### Prepare your environment
+
+To run this example, you will need:
+
+- a working local Hatchet environment or access to Hatchet Cloud
+- the Python SDK example environment
```
abelanger5, line 55: include internal links for the prerequisites? most likely to the quickstart.

sdks/python/examples/support_agent/worker.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+# !!
+
+
+# > Support agent workflow
+@hatchet.durable_task(input_validator=SupportTicketInput)
+async def support_agent(
+    input: SupportTicketInput, ctx: DurableContext
+) -> dict[str, Any]:
```
mrkaye97, line 133: would be great to return a pydantic model here too if we can

frontend/docs/pages/cookbooks/workflow-support-agent.mdx
```diff
@@ -0,0 +1,135 @@
...
+
+Start with a few small Pydantic models for the workflow input and outputs.
+
+<Snippet src={snippets.python.support_agent.worker.models} />
+
+The models keep the inputs and outputs for each task explicit, which makes the workflow easier to inspect and test.
+
+### Add the child tasks
+
+The durable workflow delegates its work to a few small child tasks.
```
abelanger5, line 70: Internal link to explaining child workflows/tasks?

sdks/python/examples/support_agent/worker.py
```diff
@@ -0,0 +1,198 @@
...
+                scope=input.ticket_id,
+                consider_events_since=consider_events_since,
+            ),
+        ),
+    )
+
+    # The or-group result is {"CREATE": {"<condition_key>": ...}}.
+    # Check whether the reply event condition was the one that resolved.
+    resolved_key = list(wait_result["CREATE"].keys())[0]
+    customer_replied = resolved_key == REPLY_EVENT_KEY
```
mrkaye97, line 159: so this is an issue on the SDK for sure, and I need to think about how to fix it, but what's the objective here? is it basically to wait for the event with a timeout? the raw `aio_wait_for` with an or group kind of sucks because it returns this thing that really should be internal, as you've discovered, so I might need to cook up some better way of doing this sort of thing... I wonder if adding a `timeout: timedelta` parameter to the wait could work, or something like that... I'm not too sure. I think this is okay for now, but ideally we could do this in a way where we don't need to use the internal `CREATE` match object, although that might be challenging

frontend/docs/pages/cookbooks/workflow-support-agent.mdx
```diff
@@ -0,0 +1,135 @@
...
+
+### Build the durable workflow
+
+Now tie everything together in a durable Hatchet workflow.
+
+<Snippet src={snippets.python.support_agent.worker.support_agent_workflow} />
+
+As you can see, the workflow runs triage first, generates an initial reply, and then waits for one of two things to happen: either a customer reply event arrives for that ticket, or the timeout fires. From there, the workflow either resolves the ticket or escalates it.
+
+The detail that matters most here is `consider_events_since` on the reply event condition. A customer reply could arrive while the workflow is still finishing triage or generating the first response. By using `consider_events_since`, the workflow can still pick up that reply once the wait becomes active instead of missing it because the event arrived slightly early.
```
abelanger5, line 94: Internal link to docs about `consider_events_since`?

mrkaye97 (commented):
left a couple of small comments on the python side! looks great broadly though, very cool

frontend/docs/pages/cookbooks/workflow-support-agent.mdx
```diff
@@ -0,0 +1,135 @@
...
+
+Finally, add a task to represent escalation to the support team:
+
+<Snippet src={snippets.python.support_agent.worker.escalate_task} />
+
+Keeping triage, reply generation, and escalation as separate tasks keeps the workflow itself small and makes each piece easier to reason about.
+
+### Build the durable workflow
+
+Now tie everything together in a durable Hatchet workflow.
```
abelanger5, line 88: I think this part could use an overview of why a durable task is appropriate here and what sorts of benefits a user might get from the durable workflow (being able to sleep for a long time/be evicted/resuming on a new worker if your worker crashes)

abelanger5 (commented):
💯 this looks great! Left some small comments and going to let @mrkaye97 leave comments on the code. In general I think explaining concepts a bit more and adding internal links to docs would be really helpful, but I like the clear and direct tone and it's easy to follow along.

sdks/python/examples/support_agent/worker.py
BloggerBust: That makes sense in production, but for this cookbook example I intentionally kept the provider / API key optional and supplied a fixed fallback response when it is not provided. I wanted the example to remain runnable even for users who do not have an Anthropic API key, so they can still see the intended workflow behavior end to end.

sdks/python/examples/support_agent/worker.py
BloggerBust: That is correct. I kept the Anthropic import local so the dependency remains optional for this cookbook example, which lets users run the example even if they do not have an API key or do not want to install the provider dependency just to follow the workflow pattern.

sdks/python/examples/support_agent/worker.py
BloggerBust: Yes, exactly, the objective here is to wait for the customer reply event with a timeout. I agree the raw `aio_wait_for(... or_(...))` result is a bit awkward to expose in an example, but this seemed like the most direct way to express the branching with the current SDK. As the SDK improves I would be happy to revisit the example.

sdks/python/examples/support_agent/worker.py
BloggerBust: Thank you for the feedback Matt! I considered enums here, but I kept these as simple strings to keep the example compact and focused on the workflow pattern, and to avoid adding extra modeling overhead in a small cookbook example.

sdks/python/examples/support_agent/worker.py
BloggerBust: Thanks, I agree that a typed Pydantic return model could be a nice refinement here. I left the workflow return as a simple dict to keep the example smaller and focused on the control-flow pattern.

abelanger5 (approved):

PR #3626, opened 2026-04-15
[Python] Hotfix: Retry tasks if `send_step_action_event` flakily fails

mrkaye97:
# Description

### Changed

- Fixes a bug where failures sending a completed or failed event from the worker to the engine would fail the task and bypass any retries, even if some were configured on the task

## Type of change

### DIFF
--- a/sdks/python/hatchet_sdk/clients/dispatcher/dispatcher.py
+++ b/sdks/python/hatchet_sdk/clients/dispatcher/dispatcher.py
@@ -152,13 +152,13 @@ class DispatcherClient:
             message = f"failed to send step action event {event_type} for action {action.action_id}."
 
             if was_completed:
-                message += "\n**IMPORTANT**: the task completed successfully on the worker, but the engine failed to receive the completed event."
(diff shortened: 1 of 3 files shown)


grutt (approved):

sdks/python/hatchet_sdk/clients/dispatcher/dispatcher.py
```diff
@@ -147,12 +147,18 @@ async def send_step_action_event(
...
             # for step action events, send a failure event when we cannot send the completed event
             if event_type in (STEP_EVENT_TYPE_COMPLETED, STEP_EVENT_TYPE_FAILED):
-                await self._try_send_step_action_event(
-                    action,
-                    STEP_EVENT_TYPE_FAILED,
-                    "Failed to send finished event: " + str(e),
-                    should_not_retry=True,
-                )
+                try:
+                    await self._try_send_step_action_event(
```
juliusgeo: I think it would be better to add the exception type that is causing this to error out in `tenacity_should_retry`

juliusgeo (approved):

juliusgeo (approved):

PR #3629, opened 2026-04-16
fix: only keep v1_step_concurrency active when it has work

grutt:
# Description

CheckStrategyActive was keeping any strategy on the latest workflow version pinned is_active=true regardless of whether any tasks were queued or running. This caused active strategy counts to grow with every deployment, each spinning up a ConcurrencyManager that polled Postgres every 500ms-5s. Strategies are now deactivated when idle and re-activated by the after_v1_concurrency_slot_insert trigger when new
work arrives.

In simpler words, v1_step_concurrency is deactivated if no future work exist.

## Type of change

### DIFF
--- a/pkg/repository/sqlcv1/concurrency.sql
+++ b/pkg/repository/sqlcv1/concurrency.sql
@@ -63,8 +63,9 @@ WHERE
 ;
 
 -- name: CheckStrategyActive :one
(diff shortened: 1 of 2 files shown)


mrkaye97:
Closing, superseded by the cron we added a couple days ago I think

PR #3631, opened 2026-04-16
fix: exchange token authz, remove n+1 tenant lookup for organizations, slack oauth casing

abelanger5:
# Description

- Fixes casing for the exchange token to not use the `restrictedWithBearerAuth` check when we have an exchange token 
- Removes individual tenant lookups in favor of using the `tenants` array returned from the get organization endpoint. 
- Cases slack properly on control plane
- Makes security for the metadata endpoint to list integrations optional

## Type of change


juliusgeo (approved):

frontend/app/src/lib/api/generated/Api.ts
```diff
@@ -1349,13 +1349,11 @@ export class Api<
    * @name MetadataListIntegrations
    * @summary List integrations
    * @request GET:/api/v1/meta/integrations
-   * @secure
    */
   metadataListIntegrations = Object.assign((params: RequestParams = {}) =>
     this.request<ListAPIMetaIntegration, APIErrors>({
       path: `/api/v1/meta/integrations`,
       method: "GET",
-      secure: true,
```
juliusgeo, line 1358: Why isn't this secure anymore?

frontend/app/src/pages/main/v1/tenant-settings/alerting/index.tsx
```diff
@@ -356,7 +358,13 @@ function SlackWebhooksList() {
         <h3 className="text-xl font-semibold leading-tight text-foreground">
           Slack Webhooks
         </h3>
-        <a href={'/api/v1/tenants/' + tenantId + '/slack/start'}>
+        <a
+          href={
+            isControlPlaneEnabled
+              ? '/api/v1/control-plane/tenants/' + tenantId + '/slack/start'
```
juliusgeo, line 364: nit: shouldn't this be `userUpdateSlackOauthStart`?

frontend/app/src/lib/api/generated/Api.ts
abelanger5: I wasn't able to access it with an exchange token since it's non-tenant scoped, but we read it from a tenant page. When I looked more closely, I realized it should have the same permissions boundary as the `metadata` endpoint (arguably it should be part of the metadata endpoint).

frontend/app/src/pages/main/v1/tenant-settings/alerting/index.tsx
abelanger5: yes, but I imagine we had it as a direct `href` for a reason in the old code and didn't want to mess with it

frontend/app/src/lib/api/generated/Api.ts
abelanger5: Thought about this a bit more and updated the exchange token behavior instead so that it operates like a PAT if there's no tenant scoping. I think this is cleaner.

frontend/app/src/pages/main/v1/tenant-settings/alerting/index.tsx
juliusgeo: ah yeah I see, it's supposed to be an actual link because it will redirect. Makes sense to me!

juliusgeo (approved):

PR #3335, opened 2026-03-19
docs: comprehensive durable execution documentation

promptless-for-oss:
[Open this suggestion in Promptless to view citations and reasoning process](https://app.gopromptless.ai/suggestions/2488198a-7630-4f23-a625-6d85cb5eabfb)

## Summary

Adds comprehensive documentation for Hatchet's durable execution features, covering eviction policies, determinism requirements, DurableContext methods, and more.

### Conceptual docs
- **Eviction policies** — TTL, capacity-based eviction, and priority configuration in `task-eviction.mdx`
- **Server-initiated eviction** — Stale invocation detection and split-brain prevention
- **Determinism requirements** — What causes `NonDeterminismError`, how to handle it, and best practices
- **Replaying from checkpoints** — `reset_durable_task` / `aio_reset_durable_task` for programmatic replay
- **Slot types** — Regular vs durable slot pools and deadlock prevention in `workers.mdx`

### SDK reference docs
- **Python `DurableContext`** — `aio_wait_for`, `aio_sleep_for` (with correct `Duration` param type and `SleepResult` return), `aio_wait_for_event` with `payload_validator` support
- **TypeScript `DurableContext`** — `sleepFor`, `sleepUntil`, `waitForEvent`, `waitFor`, `now()`, `SleepResult` type
- **`EvictionPolicy`** class reference with correct import path (`hatchet_sdk.runnables.eviction`)
- **`reset_durable_task`** — Added `branch_id` parameter, corrected return type to `V1BranchDurableTaskResponse`
- **`NonDeterminismError`** — Properties and handling guidance for both SDKs

### Fixes
- Fixed broken `/v1/durable-workflows` links → `/v1/durable-workflows-overview`
- Fixed eviction TTL default from 5 min → 15 min
- Fixed `NonDeterminismError` import path in TypeScript (`@hatchet-dev/typescript-sdk/v1`)

---
*Generated by [Promptless](https://gopromptless.ai)*

PR #3639, opened 2026-04-17
Feat: Durable execution frontend work and API improvements

mrkaye97:
# Description

Adding some new features to the FE + SDKs for durable execution:

1. Users can add a `label` when a wait is established, which is just a message that'll populate in the frontend
2. We also try to create a human-readable message from wait conditions, which we can populate there too
3. Durable events will now show up in the event log
4. The event log now uses the logging component instead of the previous `SimpleTable`
5. Improved overflow behavior a bunch in the side panel
6. Adds `is_durable` to `v1_tasks_olap` so we can determine if a task is durable on the FE more easily
7. Adds `satisfied_at` to the event log, to let us determine e.g. how long we waited, etc.

# Screenshots

## DAG view

Screenshot 2026-04-21 at 1 07 42 PM

## Side panel

Screenshot 2026-04-21 at 1 08 12 PM

## View worker button

Screenshot 2026-04-21 at 1 08 28 PM

## Type of change


promptless-for-oss:

Triggered by [PR #3639](https://github.com/hatchet-dev/hatchet/pull/3639)

Updated documentation for the new `label` parameter on durable waits (`sleepFor`, `waitFor`, `waitForEvent`). Added guide sections explaining how labels appear in the dashboard, and updated TypeScript and Python API references.

Review: [Consolidated durable execution documentation](https://github.com/hatchet-dev/hatchet/pull/3335)

api/v1/server/handlers/v1/durable-tasks/event_log.go
```diff
@@ -0,0 +1,140 @@
...
+	if taskInterface == nil {
+		return gen.V1DurableTaskEventLogList404JSONResponse{
+			Errors: []gen.APIError{{Description: "durable task not found"}},
+		}, nil
+	}
+
+	task, ok := taskInterface.(*sqlcv1.V1TasksOlap)
+	if !ok {
+		return nil, echo.NewHTTPError(500, "durable task type assertion failed")
+	}
```
abelanger5: should be good to remove all of this and replace it with `ctx.Get("durable-task").(*sqlcv1.V1TasksOlap)`. panics are caught easily with a single test during development and wouldn't take down the API in production (the only risk would be spawning a goroutine that panics)

api/v1/server/run/run.go
```diff
@@ -566,6 +569,27 @@ func (t *APIServer) registerSpec(g *echo.Group, spec *openapi3.T) (*populator.Po
...
+		}
+
+		task, err := config.V1.OLAP().ReadTaskRun(ctx, taskID)
+		if err != nil {
+			return nil, "", err
+		}
+
+		if task == nil {
+			return nil, "", echo.NewHTTPError(http.StatusNotFound, "durable task not found")
+		}
```
abelanger5, line 588: just double checking that this doesn't throw `pgx.ErrNoRows` when it's not found rather than returning a nil task?

pkg/repository/durable_events.go
```diff
@@ -186,59 +191,191 @@ func (e *StaleInvocationError) Error() string {
...
+	d, err := time.ParseDuration(*s)
+	if err != nil {
+		return nil
+	}
+
+	ms := d.Milliseconds()
+	return &ms
+}
+
+func formatDurationMs(ms int64) string {
```
abelanger5: Why not just use something like `time.Duration(ms).String()` to get a readable duration?

abelanger5 (approved):
nice!

api/v1/server/handlers/v1/durable-tasks/event_log.go
mrkaye97: ah yeah, I keep forgetting this 🤦 will fix, thanks!

pkg/repository/durable_events.go
mrkaye97: didn't think about it 🤦 thanks!

api/v1/server/run/run.go
mrkaye97: I think it does, but I think we would've failed before this in this case, since the task populator uses the same logic - just going to leave it for now, but agree it'd be good to clean up

PR #3650, opened 2026-04-20
fix(typescript-sdk): add subpath export for v1 to support ESM imports

### LINKED ISSUE #3613: [BUG] Typescript SDK: ESM import error for native ESM
**Describe the issue**  
The TypeScript SDK docs example uses `@hatchet-dev/typescript-sdk/v1`, but in native Node ESM this throws `ERR_UNSUPPORTED_DIR_IMPORT` at runtime. Node suggests `@hatchet-dev/typescript-sdk/v1/index.js` instead. This makes the documented import path incorrect or incomplete for ESM users.

**Environment**  
- SDK: `@hatchet-dev/typescript-sdk` `1.21.0`  
- Runtime: Node.js `24.14.1` (native ESM, `type: "module"`)  
- Engine: Cloud - latest

**Expected behavior**  
The documented import path `@hatchet-dev/typescript-sdk/v1` should work in ESM projects, or docs should explicitly mention the ESM-safe import path if different.

Repro repo: [repo](https://github.com/Spid3rrr/hatchet-dir-import-error)

Error observed at runtime: `ERR_UNSUPPORTED_DIR_IMPORT` for `@hatchet-dev/typescript-sdk/v1` in native ESM.

**Additional context**  
Behavior differs between module systems: CommonJS can resolve it, while native ESM cannot.

dhimanAbhi:
# Description
Fixes ESM import error when using: `import { HatchetClient } from '@hatchet-dev/typescript-sdk/v1';`

In native Node.js ESM environments, this import currently throws: `ERR_UNSUPPORTED_DIR_IMPORT`

This happens because Node.js ESM does not support directory imports without explicit subpath exports. This PR adds a proper subpath export for ./v1, mapping it to dist/v1/index.js, allowing the documented import path to work correctly in ESM.

Fixes #3613

## Type of change

## What's Changed
- Added subpath export for ./v1 in package.json
- Ensured compatibility with both ESM (import) and CommonJS (require)
- Verified fix using a minimal ESM reproduction setup


Leestex:
Confirming the underlying bug still reproduces at `@hatchet-dev/typescript-sdk@1.22.2` — `import('@hatchet-dev/typescript-sdk/v1')` from native ESM throws `ERR_UNSUPPORTED_DIR_IMPORT`, same as #3613. The shape of this patch matches what the package currently lacks (no `exports` map), so it should make the documented `/v1` path resolve. Would be great to see this merged.

BloggerBust:
@dhimanAbhi Thanks for contributing to #3613. The diagnosis looks right, and I think an `exports` map is the right mechanism for making `@hatchet-dev/typescript-sdk/v1` work under native Node ESM. However, there are a few things we need to address before this can merge.

Since the package previously had no `exports` map, adding one restricts which package subpaths can be imported. As written, this PR exports only `"."` and `"./v1"`, which fixes the reported `/v1` import but breaks existing subpath imports that currently work. I validated this with a local package-resolution test using the proposed export map.

Here are examples of imports that failed to resolve with this change:

- `@hatchet-dev/typescript-sdk/v1/index.js`: the workaround mentioned in #3613
- `@hatchet-dev/typescript-sdk/util/sleep`: used by example code
- `@hatchet-dev/typescript-sdk/opentelemetry`: used by the docs
- other existing subpaths such as `@hatchet-dev/typescript-sdk/sdk`, `@hatchet-dev/typescript-sdk/workflow`, and generated/protobuf paths

To preserve existing imports, we need a more complete export-map strategy. Ideally, we would explicitly enumerate the intended public subpaths. Alternatively, we could use a carefully tested wildcard export that we can prove maintains backward compatibility.

This branch has also fallen far behind `main` and should be updated before work continues.

Arnavkar:
Hi @BloggerBust - Came across this in the hatchet sdk on my end but saw that the corresponding issue for this went stale and got closed. Is there any plan to reopen this? 

Would be happy to open a new PR and fix the subpath issue you mentioned above if you reopen /replace the issue and assign to me. (@dhimanAbhi feel free to pipe in if you were still planning to fix this, otherwise I will assume that this PR is stale)

BloggerBust:
CC: @gregfurman

Arnavkar:
@gregfurman @BloggerBust bumping this thread - any thoughts regarding this?

PR #3653, opened 2026-04-21
docs(guide): add guide for durable tasks vs DAG workflows

BloggerBust:
# Description

Adds a new guide: **When to Use Durable Tasks vs. DAG Workflows**.

This guide is intended to help readers choose the right Hatchet workflow pattern based on the structure and maintainability of the workflow they are building. It explains where DAGs are the natural fit, where durable tasks become the better choice, and where composing both models together makes sense.

The guide also includes:
- a practical workflow-evaluation exercise
- concrete examples and boundary cases
- a decision checklist
- links to the most relevant supporting docs

## Type of change

## What's Changed

- [x] Add a new guide at `frontend/docs/pages/v1/durable-tasks-vs-dags.mdx`
- [x] Explain the distinction between DAG workflows and durable tasks as workflow authoring patterns
- [x] Add guidance for when a DAG is the better fit
- [x] Add guidance for when durable task composition is the better fit
- [x] Add a practical step-by-step workflow evaluation exercise
- [x] Add concrete examples, boundary cases, and a decision checklist
- [x] Add internal links to related Hatchet docs
- [x] Add the new guide to the v1 docs navigation in `_meta.js`


frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
```diff
@@ -0,0 +1,103 @@
+import { Steps } from "nextra/components";
+
+# When to Use Durable Tasks vs. DAG Workflows
+
+Hatchet supports two ways to orchestrate durable multi-step workflows: [directed acyclic graphs (DAGs)](/v1/directed-acyclic-graphs) and [durable tasks](/v1/durable-tasks). Hatchet persists task state and results so workflows can recover without re-running completed work. The important difference is how you author the workflow, which means making an architectural decision early: should this workflow be modeled as a DAG, or composed at runtime from tasks? Both are forms of [durable execution](/v1/durable-execution) in Hatchet, but they fit different kinds of workflows. The choice comes down to whether the workflow’s control flow and dependencies are known at implementation time. If they are, a DAG is usually the natural choice. If they are not, task composition is usually the better fit. In that model, your code runs, makes decisions, waits for things, and spawns children while Hatchet persists progress along the way.
```
abelanger5: > If they are not, task composition is usually the better fit.

It's unclear what task composition refers to, perhaps we should say `durable tasks are usually the better fit`

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+## When the structure is known up front
+
+DAGs work best when the workflow already looks like a graph before execution starts. The steps are known in advance, the dependencies between them are explicit, and parallelizable stages are easy to identify. Document processing, ETL pipelines, and CI/CD pipelines are good examples. Their major stages are known ahead of time. The interesting part is how those stages depend on one another and which ones can run concurrently. DAGs are also a good choice when fan-out and fan-in are central to the workflow. If one stage produces a variable number of items and the next stage processes them in parallel before joining, that maps directly onto a declared graph. Hatchet schedules the parallel work, passes outputs from parents to children, and gives you a clear visual representation of the whole pipeline in the dashboard.
```
abelanger5: I think there are two phrases here which need some rewording:

> parallelizable stages are easy to identify

even if a stage is identified as parallelizable, it doesn't guarantee that DAGs are a good fit if the fanout is a dynamic number of items, since that would require child spawning

> If one stage produces a variable number of items and the next stage processes them in parallel before joining, that maps directly onto a declared graph

sort of the same as the above, but producing a variable number of items implies a dynamic fanout which you would need to solve with child spawning. so child spawning is always the better fit for a variable fanout pattern. 

Just as a heads up, we're planning to change some of the behavior of child spawning from a regular task so it can have optional idempotency guarantees without full durability guarantees, in the hope of making this even more clear! That way, you can achieve dynamic fanout without a durable context.

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+Use a DAG when you want the structure of the workflow to be the thing you express directly. If you find yourself drawing boxes and arrows on a whiteboard to explain the workflow, a DAG is probably the right fit.
```
abelanger5: I think this is a bit redundant and can be cut

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+## When declarative structure stops being practical
```
abelanger5, line 9: for this section, I think breaking up the text to include some mermaid diagrams/visuals rather than having three long blocks of text would be helpful!

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+Durable tasks handle these cases by letting you write the control flow directly in code. Hatchet checkpoints durable task progress around waits and [child spawning](/v1/child-spawning), evicts the task from the worker slot while it waits, and resumes from the checkpoint when the wait resolves. The workflow can pause for arbitrary durations, pick up exactly where it left off, and make its next decision based on what actually happened rather than what was predicted at design time. Use a durable task when the branching logic is complex enough that maintaining it declaratively would be fragile, when the logic changes often enough that a static graph would be a maintenance burden, or when the workflow requires loops or runtime-determined iteration.
+
+## Evaluating your own workflow
+
+When you are staring at a new workflow problem and are not sure which model fits, try this exercise. It works on paper, a whiteboard, or a text file, the point is to make the structure visible before you start coding.
```
abelanger5: Personally I think this section is unnecessary (particularly given the tables / examples further down), I'd be curious what others think -- the most useful part is the `Mixed patterns` section, which I think would be worth hoisting into its own section which might replace this section instead.

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+## Boundary cases
+
+The line between the two is not always sharp, and the presence of waits or branches alone does not determine which model to use. DAGs in Hatchet handle waits natively through or groups, a DAG node can wait for a sleep timer, a user event, or a combination of both before proceeding. DAGs also handle branching through parent conditions and conditional skipping. A workflow with fixed stages, a wait for human approval in the middle, and a two-way branch afterward is still a perfectly valid DAG. The point where a DAG starts to strain is not when you add your first wait or your first branch. It is when the branching logic becomes dense enough that the declarative graph is harder to understand than equivalent imperative code, when the rules change often enough that updating the graph is a recurring source of mistakes, or when the workflow needs to loop or spawn a number of children that is only known at runtime. Those are the structural signals that the workflow has moved past what a static graph can comfortably express.
```
abelanger5: This section feels a bit redundant and ambiguous, so I'd also be in favor of cutting it

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+## Mistakes to avoid
+
+The most common mistake is treating durable tasks as the default advanced choice and DAGs as the simple beginner option. A DAG is often the better design when the workflow can genuinely be represented as an acyclic graph with stable, understandable branching. Choosing imperative orchestration when declarative orchestration already fits makes the workflow harder to read, test, and visualize in the dashboard.
+
+The second most common mistake is treating DAGs as if they can only express rigid linear pipelines. DAGs in Hatchet support branching via parent conditions, event and sleep waits via or groups, and conditional skipping. A workflow that waits for human input or branches based on a previous step's output is not automatically a durable task problem as DAGs handle those patterns natively. Strong evidence that a DAG is the wrong choice includes loops, branching logic that is too complex or volatile to maintain declaratively, and cases where the set of steps cannot be known until runtime.
```
abelanger5: Would consider condensing this section and cutting it down significantly, I think this is also a bit repetitive based on what's above

abelanger5 (commented):
@BloggerBust great, this looks very solid! Style-wise for these pages I think we want to be more direct, practical, and to the point, so the examples and overview are good but it's a bit wordy in other parts. I flagged the sections I felt could be cut down or stripped.

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
BloggerBust: Thank you! I agree and changed the wording.

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
BloggerBust: I agree with all points and revised the wording, thank you for the great feedback!

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
BloggerBust: Done!

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
BloggerBust: Agreed. I captured the bit you found valuable into its own `## Mixing both models` section.

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
BloggerBust: I shortened this section and added a mermaid diagram to the new Mixing both models section where I thought a visual would be more helpful.

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
BloggerBust: Agreed. I cut it.

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
BloggerBust: I cut this section. Thanks for the helpful feedback!

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
BloggerBust: I ended up adding a Mermaid diagram here to make the point that branching grows beyond maintainability more clear. I kept it fairly simple, since Mermaid only gives me so much flexibility for showing branching that continues to fan out. The leaf labels end with ellipses to suggest that the diagram keeps expanding.

Here is a screenshot:

image

If you don’t think it quite works, I am happy to remove it. Another option would be to have your designer turn the idea into a custom diagram if you think the concept is worth illustrating but needs improvement.

abelanger5: I do like this, but I think it might be overwhelming given the complexity of the workflow we're showing. One potential solve is to place it into an accordion which is collapsed by default. But otherwise it might be helpful to show just a simple cycle in this diagram, since the immediate sentence before the diagram is `That requires a cycle, which a DAG cannot express.`

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+Support workflows, approval flows, and agentic loops are good examples. Consider a support agent: it triages a ticket, sends an initial response, waits for a customer reply or a timeout, then either resolves or escalates. You could express a simple version of that as a DAG with or groups and [parent conditions](/v1/directed-acyclic-graphs#branching-with-parent-conditions). But in practice the branching logic tends to grow beyond maintainability, with multiple escalation paths, multi-turn conversations, and context-dependent retries. At that point, writing it as imperative code with checkpointing is less work. Agentic loops are an even clearer case: the number of iterations is unknown, each action depends on the previous result, and the stopping condition is evaluated at runtime. That requires a cycle, which a DAG cannot express.
```
abelanger5: nit: internal link to the support workflow cookbook?

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
```diff
@@ -0,0 +1,89 @@
...
+| Agentic loop / iterative tool-calling                                                                                                                                   | Durable task | Contains a cycle. The number of iterations is unknown and the stopping condition is evaluated at runtime. A DAG cannot express this.                                            |
+| Dynamic runtime fan-out                                                                                                                                                 | Both         | A durable task decides at runtime what work needs to happen, then spawns a DAG for each structured sub-pipeline.                                                                |
+
+## Decision checklist
+
+| Question                                                                   | Lean DAG                           | Lean Durable Task                                                 |
```
abelanger5: nit: `Lean DAG` has a double meaning which confused me at first (in that lean is sort of synonymous with lightweight), but this might just be me!

abelanger5 (approved):
Nice! This is looking really good -- left two minor nits and a comment, but this should be good to merge.

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
BloggerBust: Thanks, I added the link.

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
BloggerBust: Thank you, I improved the labeling so that they are not ambiguous. Nice catch!

frontend/docs/pages/v1/durable-tasks-vs-dags.mdx
BloggerBust: @abelanger5 I updated the diagram to show a simpler agentic loop and laid it out horizontally so it takes up less vertical space and keeps the surrounding text in view. If you’re happy with this, I think the PR is ready to merge.

image

PR #3658, opened 2026-04-21
fix: use exchange token interceptor for cloud endpoints

abelanger5:
# Description

Updates the usage of `cloudApi` to include an interceptor for tenant-scoped endpoints, so we can use the exchange token mechanism if the control plane is enabled. 

## Type of change


mrkaye97 (approved):
neat, didn't know wee could do this

PR #3663, opened 2026-04-21
Cleanup old workers via daily gocron

mnafees:
# Description

Adds a new retention gocron job to get rid of workers based on tenant retention period. Also starts using `tenant_id` for query `ListTasksToReassign` to start using index of `v1_task_runtime`.

## Type of change

### DIFF
--- a/internal/services/controllers/retention/worker.go
+++ b/internal/services/controllers/retention/worker.go
@@ -2,53 +2,25 @@ package retention
 
 import (
 	"context"
-	"fmt"
 	"time"
-
-	"github.com/hatchet-dev/hatchet/pkg/repository/sqlcv1"
-	"github.com/hatchet-dev/hatchet/pkg/telemetry"
 )
-
-const cleanupOldWorkersSleepBetweenBatches = time.Second
 
 func (rc *RetentionControllerImpl) runCleanupOldWorkers(ctx context.Context) func() {
 	return func() {
 		rc.l.Debug().Ctx(ctx).Msg("retention controller: cleaning up old workers")
 
-		err := rc.ForTenants(ctx, rc.runCleanupOldWorkersForTenant)
-		if err != nil {
-			rc.l.Err(err).Ctx(ctx).Msg("could not cleanup old workers")
-		}
-	}
-}
+		ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+		defer cancel()
 
-func (rc *RetentionControllerImpl) runCleanupOldWorkersForTenant(ctx context.Context, tenant sqlcv1.Tenant) error {
-	ctx, span := telemetry.NewSpan(ctx, "cleanup-old-workers-tenant")
-	defer span.End()
+		shouldContinue := true
+		var err error
 
-	cutoff, err := GetDataRetentionExpiredTime(tenant.DataRetentionPeriod)
-	if err != nil {
-		return fmt.Errorf("could not get cutoff for tenant %s: %w", tenant.ID.String(), err)
-	}
-
-	for {
-		if ctx.Err() != nil {
-			return ctx.Err()
-		}
-
-		hasMore, err := rc.repo.Workers().DeleteOldWorkers(ctx, tenant.ID, cutoff)
-		if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/repository/worker.go
```diff
@@ -723,22 +723,39 @@ func (w *workerRepository) UpsertWorkerLabels(ctx context.Context, workerId uuid
 	return affinities, nil
 }
 
-func (w *workerRepository) DeleteOldWorkers(ctx context.Context, tenantId uuid.UUID, lastHeartbeatBefore time.Time) (bool, error) {
-	hasMore, err := w.queries.DeleteOldWorkers(ctx, w.pool, sqlcv1.DeleteOldWorkersParams{
+func (w *workerRepository) CleanupOldWorkers(ctx context.Context, tenantId uuid.UUID, lastHeartbeatBefore time.Time) (bool, error) {
+	const timeout = 1000 * 60 * 3 // 3 minutes
```
mrkaye97, line 727: I feel like if this is taking 3 mins, we have major problems... this should probably be a lot shorter imo

pkg/repository/worker.go
```diff
@@ -723,22 +723,39 @@ func (w *workerRepository) UpsertWorkerLabels(ctx context.Context, workerId uuid
...
+
+	tx, commit, rollback, err := sqlchelpers.PrepareTxWithStatementTimeout(ctx, w.pool, w.l, timeout)
+	if err != nil {
+		return false, fmt.Errorf("error beginning transaction for %s: %w", lockName, err)
+	}
+	defer rollback()
+
+	acquired, err := w.queries.TryAdvisoryLock(ctx, tx, hash(lockName))
+	if err != nil {
+		return false, fmt.Errorf("error acquiring advisory lock for %s: %w", lockName, err)
```
mrkaye97: do we need an advisory lock for this?

mrkaye97 (commented):
couple small things but otherwise looks good, thanks for fixing this!

pkg/repository/worker.go
mnafees: Yeah so I kept this at 3 minutes because on cloud with 1.8M zombie workers the first few iterations took around 80-90 seconds. So wanted to fool proof it against other big DBs.

pkg/repository/worker.go
mnafees: Good catch, since this is running for controller partitioned tenants, this won't be required.

mrkaye97 (approved):

internal/services/controllers/retention/worker.go
```diff
@@ -0,0 +1,48 @@
...
+func (rc *RetentionControllerImpl) runCleanupOldWorkers(ctx context.Context) func() {
+	return func() {
+		rc.l.Debug().Ctx(ctx).Msg("retention controller: cleaning up old workers")
+
+		ctx, cancel := context.WithTimeout(ctx, 30*time.Minute)
+		defer cancel()
+
+		if err := rc.ForTenants(ctx, rc.cleanupOldWorkersForTenant); err != nil {
+			rc.l.Err(err).Ctx(ctx).Msg("could not cleanup old workers")
+		}
```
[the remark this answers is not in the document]
mrkaye97: this could be legit btw

pkg/repository/sqlcv1/workers.sql
```diff
@@ -379,36 +379,16 @@ SET
...
-    DELETE FROM "WorkerAssignEvent" wae
-    WHERE wae."workerId" IN (SELECT "id" FROM expired_with_limit)
-    RETURNING wae."id"
+-- name: CleanupOldWorkers :execresult
+WITH old_workers AS (
+    SELECT "id"
+    FROM "Worker"
+    WHERE "tenantId" = @tenantId::uuid
+      AND "lastHeartbeatAt" < @lastHeartbeatBefore::timestamp
+    LIMIT @batchSize::int
```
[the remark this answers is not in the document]
mrkaye97: this too - if we have no FKs, we'll just have a bunch of orphaned rows in these tables. would be good to check on that!

PR #3666, opened 2026-04-22
Batch process worker cleanup and also cleanup `v1_worker_slot_config`

mnafees:
# Description

Follow up to https://github.com/hatchet-dev/hatchet/pull/3663

This PR
- Adds cleanup step for old `v1_worker_slot_config` which would otherwise be orphaned once the `"Worker"` records were deleted
- Makes sure to batch process 50 tenant worker cleanups concurrently at a time

## Type of change

### DIFF
--- a/internal/services/controllers/retention/shared.go
+++ b/internal/services/controllers/retention/shared.go
@@ -2,7 +2,9 @@ package retention
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"sync"
 	"time"
 
 	"golang.org/x/sync/errgroup"
@@ -20,9 +22,7 @@ func GetDataRetentionExpiredTime(duration string) (time.Time, error) {
 	return time.Now().UTC().Add(-d), nil
 }
 
-func (rc *RetentionControllerImpl) ForTenants(ctx context.Context, f func(ctx context.Context, tenant sqlcv1.Tenant) error) error {
-
-	// list all tenants
+func (rc *RetentionControllerImpl) ForTenants(ctx context.Context, perTenantTimeout time.Duration, f func(ctx context.Context, tenant sqlcv1.Tenant) error) error {
 	tenants, err := rc.p.ListTenantsForController(ctx, sqlcv1.TenantMajorEngineVersionV0)
 
 	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/services/controllers/retention/shared.go
```diff
@@ -20,29 +22,36 @@ func GetDataRetentionExpiredTime(duration string) (time.Time, error) {
...
-	// list all tenants
+func (rc *RetentionControllerImpl) ForTenants(ctx context.Context, perTenantTimeout time.Duration, f func(ctx context.Context, tenant sqlcv1.Tenant) error) error {
 	tenants, err := rc.p.ListTenantsForController(ctx, sqlcv1.TenantMajorEngineVersionV0)
 
 	if err != nil {
 		return fmt.Errorf("could not list tenants: %w", err)
 	}
 
 	g := new(errgroup.Group)
+	g.SetLimit(50)
```
mnafees, line 33: Explicitly set a limit of 50 concurrent goroutines.

internal/services/controllers/retention/shared.go
```diff
@@ -20,29 +22,36 @@ func GetDataRetentionExpiredTime(duration string) (time.Time, error) {
...
-		index := i
+	for _, tenant := range tenants {
 		g.Go(func() error {
-			return f(ctx, *tenants[index])
+			tenantCtx, cancel := context.WithTimeout(ctx, perTenantTimeout)
+			defer cancel()
+
+			if err := f(tenantCtx, *tenant); err != nil {
+				mu.Lock()
+				errs = append(errs, fmt.Errorf("tenant %s: %w", tenant.ID.String(), err))
```
mnafees, line 47: This mechanism is to make sure we aggregate any errors and not return when any error occurs thereby making sure we process all tenants that's possible.

internal/services/controllers/retention/worker.go
```diff
@@ -16,7 +16,7 @@ func (rc *RetentionControllerImpl) runCleanupOldWorkers(ctx context.Context) fun
 		ctx, cancel := context.WithTimeout(ctx, 30*time.Minute)
 		defer cancel()
 
-		if err := rc.ForTenants(ctx, rc.cleanupOldWorkersForTenant); err != nil {
+		if err := rc.ForTenants(ctx, 5*time.Minute, rc.cleanupOldWorkersForTenant); err != nil {
```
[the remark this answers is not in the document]
mnafees: That is per tenant timeout.

mrkaye97 (approved):

PR #3667, opened 2026-04-22
Cache CEL programs in order to avoid expensive heap allocations on each event match

mnafees:
# Description

Parsing and compiling CEL expressions on each event match leads to significant memory pressure. This PR attempts to relieve this pressure by introducing a non-expiring per-process LRU cache to reuse already parsed and compiled CEL programs (expressions).

## Type of change

### DIFF
--- a/pkg/repository/match.go
+++ b/pkg/repository/match.go
@@ -920,8 +920,8 @@ func (m *sharedRepository) processCELExpressions(ctx context.Context, events []C
 		if !ok {
 			ast, issues := m.env.Compile(expr)
 
-			if issues != nil {
-				m.l.Error().Ctx(ctx).Msgf("failed to compile CEL expression: %s", issues.String())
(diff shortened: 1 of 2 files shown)


pkg/repository/match.go
```diff
@@ -915,18 +915,25 @@ func (m *sharedRepository) processCELExpressions(ctx context.Context, events []C
...
-			m.l.Error().Ctx(ctx).Err(err).Msgf("failed to create CEL program: %s", expr)
-			continue
+			compiled, err := m.env.Program(ast)
+
+			if err != nil {
+				m.l.Error().Ctx(ctx).Err(err).Msgf("failed to create CEL program: %s", expr)
+				continue
+			}
+
+			m.celProgramCache.Add(expr, compiled)
```
grutt: should we hash the expression?

mrkaye97 (approved):

PR #3684, opened 2026-04-23
fix(python/sdk): use tool.poetry in pyproject.toml

gregfurman:
# Description

The current `pyproject.toml` is in PEP 621 format for both `poetry` and `uv` compatability. However, dependabot is unable to disambiguate the package manager type from the TOML, and therefore skips the regeneration of the `poetry.lock` -- breaking CI on updates as the `pyproject.toml` and lock files become out-of-sync.

This PR reintroduces a `tool.poetry` section which indicates to dependabot that it should regenerate the `poetry.lock` on SDK updates.

## Type of change

## What's Changed

- Reintroduces a `tool.poetry` section, indicating to dependabot that `poetry.lock`should be updated


sdks/python/pyproject.toml
```diff
@@ -96,6 +96,9 @@ asyncio_default_fixture_loop_scope = "session"
 [tool.black]
 extend_exclude = "(hatchet_sdk/contracts/|hatchet_sdk/v0/contracts/|site)"
 
+[tool.poetry]
+package-mode = false
```
juliusgeo: do we want package mode to be false?

gregfurman: Ah I was thinking we used `hatchling` when packaging. Reverted!

juliusgeo (approved):

PR #3698, opened 2026-04-24
Make sure we call for durable task invocation only for durable tasks in the scheduler

mnafees:
# Description

Example error log
```
could not schedule step runs error="could not get durable task invocation counts for assigned tasks: failed to get log files: context canceled"
```

We were calling `GetDurableTaskInvocationCounts` while scheduling step runs even for non-durable tasks.

We were also not emitting `task-failed` MQ message if this call failed.

## Type of change


internal/services/scheduler/v1/scheduler.go
```diff
@@ -431,16 +431,25 @@ func (s *Scheduler) scheduleStepRuns(ctx context.Context, tenantId uuid.UUID, re
...
 
-		invocationCounts, invCountErr := s.repov1.DurableEvents().GetDurableTaskInvocationCounts(ctx, tenantId, invCountOpts)
-		if invCountErr != nil {
-			return fmt.Errorf("could not get durable task invocation counts for assigned tasks: %w", invCountErr)
+		invocationCounts := make(map[repov1.IdInsertedAt]*int32, len(invCountOpts))
+
+		if len(invCountOpts) > 0 {
+			invocationCounts, err = s.repov1.DurableEvents().GetDurableTaskInvocationCounts(ctx, tenantId, invCountOpts)
+			if err != nil {
+				s.internalRetry(ctx, tenantId, res.Assigned...)
```
mnafees, line 449: Make sure, just like in other call sites, we emit a `task-failed` event if this DB call fails.

internal/services/scheduler/v1/scheduler.go
```diff
@@ -431,16 +431,25 @@ func (s *Scheduler) scheduleStepRuns(ctx context.Context, tenantId uuid.UUID, re
...
+				})
+			}
 		}
 
-		invocationCounts, invCountErr := s.repov1.DurableEvents().GetDurableTaskInvocationCounts(ctx, tenantId, invCountOpts)
-		if invCountErr != nil {
-			return fmt.Errorf("could not get durable task invocation counts for assigned tasks: %w", invCountErr)
+		invocationCounts := make(map[repov1.IdInsertedAt]*int32, len(invCountOpts))
+
+		if len(invCountOpts) > 0 {
```
mnafees, line 446: We only care about calling this if there are any durable tasks.

mrkaye97 (approved):
thanks for the fix!

PR #2772, opened 2026-01-10
fix: actually reconnect pg_notify listener to postgres if connection fails.

m-kostrzewa:
# Description

Postgres connection for pg_notify listener is not restored after it fails.

The bug is in pgxlisten.Listener setup. Connection is acquired once at startup. The pgxlisten library's reconnect logic ([pgxlisten.go](https://github.com/jackc/pgxlisten/blob/master/pgxlisten.go)) expects `Connect()` to return a **fresh connection** on each retry. But Hatchet's implementation returns the same broken connection, defeating the reconnect mechanism.

Instead, connection should be acquired in pgxlisten.Listener's Connect callback.

Fixes https://github.com/hatchet-dev/hatchet/issues/2771

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation change (pure documentation change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking changes to code which doesn't change any behaviour)
- [ ] CI (any automation pipeline changes)
- [ ] Chore (changes which are not directly related to any business logic)
- [ ] Test changes (add, refactor, improve or change a test)
- [ ] This change requires a documentation update

## What's Changed

- [X] Acquire new connection inside pgxlisten.Listener Connect callback instead of once.

Issue #2771, opened 2026-01-10
[BUG] pg_notify listener fails to reconnect to postgres


m-kostrzewa:
**Describe the issue**

When PostgreSQL crashes and recovers, the Hatchet engine's `pg_notify` listener fails to reconnect, even though the database is accepting new connections. The listener keeps retrying every 10 seconds but always fails with "conn closed" because the same stale connection is reused.

Impact:
- All workers become INACTIVE (can't receive task notifications)
- Jobs get stuck in QUEUED/RUNNING state indefinitely
- Hatchet engine stays running but is non-functional
- Only fix is manual restart of hatchet-engine container

**Environment**
- Engine: self-hosted v0.74.9
- PostgreSQL version: 15.6 (no rabbitmq)
- Using `SERVER_MSGQUEUE_KIND: "postgres"` (pg_notify for task dispatch)

**Expected behavior**

Connection is restored and hatchet engine keeps processing new requests.

**Code to Reproduce, Logs, or Screenshots**

Originally the issue was found when Postgres failed during load tests of our backend application:

```
# postgres logs
2026-01-09 17:50:18.689 UTC [1] LOG:  server process (PID 560720) exited with exit code 2
2026-01-09 17:50:18.718 UTC [1] LOG:  terminating any other active server processes
2026-01-09 17:50:21.476 UTC [1] LOG:  all server processes terminated; reinitializing
2026-01-09 17:50:22.238 UTC [560754] LOG:  database system was interrupted; last known up at 2026-01-09 17:46:03 UTC
2026-01-09 17:50:22.864 UTC [560754] LOG:  database system was not properly shut down; automatic recovery in progress
...
2026-01-09 17:50:25.851 UTC [560755] LOG:  checkpoint starting: end-of-recovery immediate wait
2026-01-09 17:50:26.985 UTC [1] LOG:  database system is ready to accept connections
```

but hatchet failed to restore the connection after DB was ready:

```
# hatchet-engine logs

2026-01-09T17:50:20.039Z ERR failed to rollback transaction error="failed to deallocate cached statement(s): conn closed" service=database
2026-01-09T17:50:20.04Z ERR could not list finalized workflow runs error="unexpected EOF" service=dispatcher
2026-01-09T17:50:20.04Z WRN error in listener error="waiting for notification: unexpected EOF" service=database
[...]

PR #3701, opened 2026-04-24
fix(repository/multiplexer): hijack listener conn to avoid pool slot leak on reconnect

### LINKED ISSUE #3694: [BUG] Multiplexed LISTEN connection leaks pool slots on reconnect under server-side idle connection termination
## Describe the issue

`pkg/repository/multiplexer.go`'s `Connect` callback hands the raw `*pgx.Conn` from an acquired pool slot to `pgxlisten.Listener` without retaining the `*pgxpool.Conn` wrapper:

https://github.com/hatchet-dev/hatchet/blob/v0.83.41/pkg/repository/multiplexer.go#L58-L66

```go
Connect: func(ctx context.Context) (*pgx.Conn, error) {
    poolConn, err := m.pool.Acquire(ctx)
    if err != nil {
        return nil, err
    }
    return poolConn.Conn(), nil          // *pgxpool.Conn wrapper falls out of scope
},
```

The wrapper goes out of scope with no `Release()` call, so pgxpool's internal bookkeeping permanently counts the slot as "acquired." When the server-side terminates the listener's conn (`idle_session_timeout`, pgbouncer `server_idle_timeout`, an L7 proxy idle kill), pgxlisten's `listen()` returns with an error, `defer conn.Close(ctx)` closes the raw conn — but the pool wrapper was already orphaned.

On reconnect, pgxlisten calls `Connect` again and acquires a *fresh* slot. The previous slot is still marked as acquired from pgx's POV, but its underlying TCP is dead. **Every reconnect cycle leaks one pool slot.**

This is a separate bug from #2771 (fixed by #2772). #2772 ensured reconnect happens at all. This issue concerns the slot-leak *consequence* of how reconnect is wired — #2772 didn't address it.

## Environment

- **Engine image:** `ghcr.io/hatchet-dev/hatchet/hatchet-lite:v0.79.16`. Reviewed `main` / `v0.83.41` — multiplexer code is unchanged, same bug present.
- **PostgreSQL:** 15.10 (AWS RDS), with `idle_session_timeout = 3600000` (1 h) set on the parameter group.
- **Database config:** all Hatchet defaults — `DATABASE_MAX_CONNS=50`, `DATABASE_MIN_CONNS=1`, `DATABASE_MAX_QUEUE_CONNS=50`, `DATABASE_MIN_QUEUE_CONNS=10`, `DATABASE_MAX_CONN_LIFETIME=15m`, `DATABASE_MAX_CONN_IDLE_TIME=1m`.
- **Msgqueue kind:** `postgres` (pg_notify via multiplexed listener).

## Expected behavior

A long-running Hatchet engine should be stable under server-side connection termination policies. Each listener-conn death should cleanly return its pool slot and reconnect without any persistent resource accumulation.

## Actual behavior

**Hourly on the dot**, the listener logs:

```
WRN error in listener
  error="waiting for notification: FATAL: terminating connection due to idle-session timeout (SQLSTATE 57P05)"
  service=database
```

pgxlisten reconnects cleanly (no error cascade visible — #2772's fix is working as intended). But pgxpool accumulates:

- **`DatabaseConnections` on the Postgres server** drains by ~1 per hour. In our case it went from ~30 at engine startup to ~12 over 60 hours — a clean, monotonic decline with no recovery.
- **`pgxpool.Stat().AcquiredConns()`** grows monotonically as wrapper-orphans accumulate.
- After ~60 cycles, the pool's "acquired" counter saturates at `MaxConns=50` while only ~12 actual TCP conns exist server-side. New `pool.Acquire()` calls block against the cap.

Once the pool is effectively wedged, downstream failure is severe:

1. Queue/scheduler/controller subsystems' periodic queries time out with `context deadline exceeded`.
2. Partition-heartbeat `UPDATE`s into `SchedulerPartition` / `ControllerPartition` / `TenantWorkerPartition` stop landing.
3. The periodic `rebalanceInactive*Partitions` job (`internal/services/partition/partition.go`) runs and — being a simpler single-row `DELETE` — sometimes still wins a live slot. It successfully deletes the engine's own stale partition rows.
4. `ON DELETE SET NULL` on `Tenant → Partition` FKs cascades: `Tenant.schedulerPartitionId`, `Tenant.controllerPartitionId`, `Tenant.workerPartitionId` all go NULL.
5. Worker `RegisterWorker` RPCs can no longer bind to a partition; worker heartbeat RPC handlers can't complete within the SDK's 5 s deadline.
6. Workers' `/health` endpoint returns 503 (Python SDK `action_listener_process._compute_health` → `last_heartbeat_succeeded == False`), container health check fails, ECS (or equivalent) restarts the worker, cycle repeats every ~3 min.
7. Restarting the worker doesn't help — the engine's control plane is dead. Only an engine restart (rebuilds pool from zero, re-runs partition bootstrap at startup) recovers.

## Reproducer

1. On a Postgres 14+ instance, set `idle_session_timeout = 300000` (5 min) via parameter group or `ALTER SYSTEM`.
2. Start `hatchet-lite` (any recent version, including `v0.83.41`) with `SERVER_MSGQUEUE_KIND=postgres`.
3. Every 5 min you'll see `WRN error in listener … FATAL: terminating connection due to idle-session timeout` in engine logs. Reconnect happens after `ReconnectDelay=10s`.
4. Query `SELECT count(*) FROM pg_stat_activity WHERE application_name LIKE '%hatchet%'` — count drifts downward monotonically, ~1 per listener cycle.
5. After enough cycles (depends on pool max + real demand), engine subsystems start logging `context deadline exceeded` on all operations.
6. Eventually partition tables get emptied by the engine's own rebalance cron and the engine becomes functionally dead until restart.

Bonus failure amplifier: `DATABASE_MIN_QUEUE_CONNS=10` (default) ensures 10 queue-pool conns are never recycled by the client-side idle timer, so they pin up against the 1 h Postgres kill and all die in lockstep once per hour, accelerating the leak rate.

## Root cause

pgx v5's `pgxpool` uses explicit `Acquire`/`Release` semantics with no wrapper finalizer. The contract is "whoever acquires must release" — pgx has no background reaper for conns held indefinitely. In the multiplexer's `Connect`:

```go
poolConn, err := m.pool.Acquire(ctx)
// ...
return poolConn.Conn(), nil
```

`poolConn` is a local variable. When the function returns, the `*pgxpool.Conn` wrapper is unreachable. pgxlisten holds only the raw `*pgx.Conn`. There is no path back to `poolConn.Release()`.

When `pgxlisten.Listener.listen()` returns on error, its `defer conn.Close(ctx)` closes the raw conn — but this sends Terminate on the underlying protocol, it does not call `pgxpool.Conn.Release()`. pgxpool's bookkeeping never decrements `acquired`, and the slot is never recycled into the idle queue or destroyed.

Neither `MaxConnLifetime`, `MaxConnIdleTime`, nor `HealthCheckPeriod` can recover this slot. All three operate on pgxpool's *idle* list via `puddle.AcquireAllIdle()` — the dead slot is in the *acquired* state and is invisible to these mechanisms.

## Suggested fix

Two independent options, either works:

### Option A — retain and release the wrapper

Keep a reference to the `*pgxpool.Conn` alive for the duration of the raw conn, and release it when the raw conn is closed. pgxlisten doesn't expose a release hook, so a minimal workaround is to track the previous pool connection in the `Connect` closure and release it on the next acquisition:

```go
var currentPoolConn *pgxpool.Conn
var poolConnMu sync.Mutex

listener := &pgxlisten.Listener{
    Connect: func(ctx context.Context) (*pgx.Conn, error) {
        poolConnMu.Lock()
        defer poolConnMu.Unlock()

        // Previous conn is dead (that's why we're here); release its slot.
        if currentPoolConn != nil {
            currentPoolConn.Release()
        }

        poolConn, err := m.pool.Acquire(ctx)
        if err != nil {
            currentPoolConn = nil
            return nil, err
        }
        currentPoolConn = poolConn
        return poolConn.Conn(), nil
    },
    // ...
}
```

pgxpool's `Release()` on a conn that's been externally closed is safe: `Release()` inspects conn state (`IsClosed() || IsBusy() || TxStatus() != 'I'`) and, if any is true, calls `res.Destroy()` which closes and removes the slot cleanly.

### Option B — use a dedicated conn outside the pool

Replace `pool.Acquire(ctx)` in the Connect callback with `pgx.Connect(ctx, connString)`:

```go
Connect: func(ctx context.Context) (*pgx.Conn, error) {
    return pgx.Connect(ctx, m.connString)
},
```

The listener's conn is then independent of pgxpool. Its lifecycle is controlled by pgxlisten alone — `defer conn.Close(ctx)` closes the raw conn, the next `Connect` opens a fresh one. No pool accounting involved.

Trade-off: the listener conn no longer counts against `DATABASE_MAX_CONNS`. Operators need to know this when sizing Postgres `max_connections` for Hatchet. The addition is minimal (one conn per engine instance).

Option A is less invasive. Option B is conceptually cleaner. Either closes the leak.

### Complementary recommendation — listener keepalive

Independent of the leak fix, the listener will still die every hour under `idle_session_timeout` because Postgres considers any session with no *client-issued queries* to be idle, regardless of incoming `NotificationResponse`s. A periodic `SELECT 1` from the listener goroutine (e.g. every 10 min) would reset the server-side idle timer and eliminate the death loop entirely. This is a common pattern for LISTEN/NOTIFY consumers.

pgxlisten doesn't offer a keepalive hook, so this lives in Hatchet. Option B composes well with this — once the listener owns its conn outright, adding a keepalive is straightforward.

## Evidence from a production-adjacent reproduction

Our dev environment matched the reproducer exactly. We added `idle_session_timeout = 3600000` via infrastructure-as-code, the engine ran fine for ~60 hours (workload was light enough that the leak took time to bite), then pool saturation → partition wipe → worker flap → engine wedge. Forensics:

- Listener kills once per hour, on the hour (UTC), for 60+ hours straight before the visible failure.
- RDS `DatabaseConnections` metric: 30 (at engine startup) → 12 (by failure).
- pgxpool `acquired_connections=50` when the pool logged saturation, while RDS simultaneously reported 7 real conns — the 43-count gap is the zombie population.
- Partition rows rebootstrap cleanly on engine restart. `ON DELETE SET NULL` cascade on `Tenant` FKs confirmed as the mechanism via DB inspection.

Happy to share sanitized logs, the full `pg_stat_user_tables` snapshot of the partition tables during the wedge, and the RDS connection-count decline graph if that helps triage.

## Related

- #2771 / #2772 — prior reconnect bug, already fixed. This issue is orthogonal.
- #3280 — L7 proxy killing `ListenV2` on the worker side. Same shape of problem (idle kill + client that doesn't detect cleanly) but different connection.

I'm happy to open a PR for either Option A or Option B if the project prefers one direction. Let me know.

voidborne-d:
Closes #3694.

## Root cause

`pkg/repository/multiplexer.go`'s `Connect` callback handed the raw `*pgx.Conn` from an acquired pool slot to `pgxlisten.Listener` without retaining the `*pgxpool.Conn` wrapper:

```go
// before
Connect: func(ctx context.Context) (*pgx.Conn, error) {
    poolConn, err := m.pool.Acquire(ctx)
    if err != nil {
        return nil, err
    }
    return poolConn.Conn(), nil        // wrapper falls out of scope
},
```

The `*pgxpool.Conn` wrapper went out of scope with no `Release()`, so pgxpool permanently counted the slot as acquired. When the server-side terminated the listener conn (`idle_session_timeout`, pgbouncer `server_idle_timeout`, an L7 proxy idle kill), pgxlisten's `listen()` returned with an error and its `defer conn.Close(ctx)` closed the raw conn — but the pool wrapper was already orphaned.

On reconnect pgxlisten called `Connect` again and acquired a fresh slot. **Every reconnect cycle leaked one pool slot.**

This is a separate bug from #2771 (fixed by #2772). #2772 ensured reconnect happens at all. This issue concerns the slot-leak *consequence* of how reconnect is wired — #2772 didn't address it.

Issue reporter's environment: AWS RDS PG 15.10 with `idle_session_timeout = 1h` triggered exactly one reconnect-per-hour, with one `AcquiredConns` increment per cycle.

## Fix

Extract the Connect body into a small `acquireListenerConn` helper that uses `poolConn.Hijack()` instead of `poolConn.Conn()`. `Hijack` transfers ownership of the raw conn out of the pool immediately, so the slot is released right after `Acquire`. pgxlisten's existing `defer conn.Close(ctx)` still closes the raw conn cleanly on listener exit — it just no longer leaks bookkeeping.

```go
// after
func acquireListenerConn(ctx context.Context, pool *pgxpool.Pool) (*pgx.Conn, error) {
    poolConn, err := pool.Acquire(ctx)
    if err != nil {
        return nil, err
    }
    return poolConn.Hijack(), nil
}
```

Behavioral differences vs. the old code:

- Pool's `AcquiredConns` count correctly drops back to 0 after `Connect` returns. No per-reconnect leak.
- Hijack reduces `TotalConns` by one (the listener conn is no longer pool-tracked); the next `Acquire` anywhere in the process opens a fresh backend if needed, up to `MaxConns`. This is the same net effect as the old code in the steady state, just without the orphaned bookkeeping.
- Legitimate pgxlisten reconnect path is unchanged — the raw `*pgx.Conn` it gets and the `defer conn.Close(ctx)` it runs on exit work identically.

## Tests

Added `pkg/repository/multiplexer_listen_test.go` with two testcontainer-backed regression guards:

1. **`TestAcquireListenerConn_ReleasesPoolSlotImmediately`** — spins up `postgres:15.6`, creates a pool (MaxConns=5), calls `acquireListenerConn`, and asserts `pool.Stat().AcquiredConns()` is `0` immediately after. Under the old code this would be `1` (and stay at 1 for the life of the process unless the raw conn was somehow returned to the pool — it isn't).

2. **`TestAcquireListenerConn_SurvivesReconnectCyclePastPoolLimit`** — simulates pgxlisten's reconnect loop by running `MaxConns*4` acquire → close cycles on a `MaxConns=3` pool. With the slot-leak bug the pool starves by iteration 4; with the fix all 12 iterations succeed and `AcquiredConns` stays at 0 throughout.

Both tests follow the existing testcontainers pattern in `task_partition_test.go` (no migrations needed, just a raw pgxpool) and run under the default `!e2e && !load && !rampup && !integration` build tag alongside the other multiplexer unit tests.

## Local gates

- `go build ./...` — clean
- `go vet ./pkg/repository/` — clean
- `go test ./pkg/repository/` (non-docker multiplexer tests) — all existing tests pass
- `golangci-lint run --config=.golangci.yml ./pkg/repository/...` — the two pre-existing warnings on the `multiplexedListener` struct (fieldalignment) and `newMultiplexedListener` (gosec G118 on the never-called `cancel`) are unchanged; my new code adds zero warnings

Docker was not available on my local machine to run the testcontainer tests end-to-end, but they compile clean and will run in CI under the `unit` job (same infra as `task_partition_test.go`).

### DIFF
--- a/pkg/repository/multiplexer.go
+++ b/pkg/repository/multiplexer.go
@@ -17,6 +17,26 @@ import (
 // multiplexChannel is a single channel used for all multiplexed messages.
 const multiplexChannel = "hatchet_listener"
 
+// acquireListenerConn acquires a connection from the pool and transfers
+// ownership to the caller via Hijack, so the pool slot is released immediately.
+//
+// pgxlisten takes a raw *pgx.Conn from Connect and closes it via
+// `defer conn.Close(ctx)` when Listen exits (e.g. the server-side kills the
+// conn via idle_session_timeout). If we returned poolConn.Conn() without
+// Hijack, the *pgxpool.Conn wrapper would fall out of scope with no Release()
+// call, leaving pgxpool's bookkeeping permanently counting the slot as
+// "acquired." Each reconnect cycle would then leak one slot until the pool
+// is exhausted. Hijack transfers ownership out of the pool immediately; the
+// raw conn is closed cleanly by pgxlisten, and the next Connect call acquires
+// a fresh slot without any orphaned bookkeeping.
+func acquireListenerConn(ctx context.Context, pool *pgxpool.Pool) (*pgx.Conn, error) {
+	poolConn, err := pool.Acquire(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return poolConn.Hijack(), nil
+}
+
 // multiplexedListener listens for messages on a single Postgres channel and
 // dispatches them to the appropriate handlers based on the queue name.
 type multiplexedListener struct {
@@ -57,12 +77,7 @@ func (m *multiplexedListener) startListening() {
 	// listen for multiplexed messages
 	listener := &pgxlisten.Listener{
 		Connect: func(ctx context.Context) (*pgx.Conn, error) {
-			// Acquire a new connection each time
-			poolConn, err := m.pool.Acquire(ctx)
-			if err != nil {
-				return nil, err
-			}
-			return poolConn.Conn(), nil
+			return acquireListenerConn(ctx, m.pool)
 		},
 		LogError: func(innerCtx context.Context, err error) {
 			m.l.Warn().Err(err).Msg("error in listener")
--- a/pkg/repository/multiplexer_listen_test.go
+++ b/pkg/repository/multiplexer_listen_test.go
@@ -1,0 +1,123 @@
+//go:build !e2e && !load && !rampup && !integration
+
+package repository
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/stretchr/testify/require"
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/modules/postgres"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+// setupPostgresPlain spins up a fresh Postgres 15.6 container and returns a
+// pgxpool configured with the given MaxConns. Unlike setupPostgresWithMigration,
+// it does NOT run hatchet migrations — these tests only need a raw connection
+// to exercise pgxpool bookkeeping.
+func setupPostgresPlain(t *testing.T, maxConns int32) (*pgxpool.Pool, func()) {
+	t.Helper()
+
+	ctx := context.Background()
+
+	postgresContainer, err := postgres.Run(ctx,
+		"postgres:15.6",
+		postgres.WithDatabase("hatchet"),
+		postgres.WithUsername("hatchet"),
+		postgres.WithPassword("hatchet"),
+		testcontainers.WithWaitStrategy(
+			wait.ForLog("database system is ready to accept connections").
+				WithOccurrence(2).
+				WithStartupTimeout(30*time.Second),
+		),
+	)
+	require.NoError(t, err)
+
+	connStr, err := postgresContainer.ConnectionString(ctx, "sslmode=disable")
+	require.NoError(t, err)
+
+	config, err := pgxpool.ParseConfig(connStr)
+	require.NoError(t, err)
+	config.MaxConns = maxConns
+	config.MinConns = 0
+
+	pool, err := pgxpool.NewWithConfig(ctx, config)
+	require.NoError(t, err)
+
+	err = pool.Ping(ctx)
+	require.NoError(t, err)
+
+	cleanup := func() {
+		pool.Close()
+		_ = postgresContainer.Terminate(ctx)
+	}
+
+	return pool, cleanup
+}
+
+// TestAcquireListenerConn_ReleasesPoolSlotImmediately verifies that
+// acquireListenerConn returns a raw *pgx.Conn detached from the pool, so that
+// pgxpool's "acquired" bookkeeping drops back to zero right after the call.
+//
+// Regression guard for #3694: the previous implementation returned
+// poolConn.Conn() without hijacking, letting the *pgxpool.Conn wrapper fall
+// out of scope without Release() — the slot stayed permanently counted as
+// acquired even though the raw conn would later be closed by pgxlisten.
+func TestAcquireListenerConn_ReleasesPoolSlotImmediately(t *testing.T) {
+	pool, cleanup := setupPostgresPlain(t, 5)
+	defer cleanup()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	require.Equal(t, int32(0), pool.Stat().AcquiredConns(),
+		"pool should start with zero acquired conns")
+
+	raw, err := acquireListenerConn(ctx, pool)
+	require.NoError(t, err)
+	require.NotNil(t, raw)
+	defer raw.Close(ctx)
+
+	require.Equal(t, int32(0), pool.Stat().AcquiredConns(),
+		"Hijack should transfer the conn out of the pool so AcquiredConns drops to zero")
+}
+
+// TestAcquireListenerConn_SurvivesReconnectCyclePastPoolLimit simulates the
+// exact scenario reported in #3694: pgxlisten repeatedly calls Connect after
+// each server-side reconnect, and each returned conn is eventually closed.
+// A slot-leak bug would exhaust the pool within MaxConns iterations because
+// the orphaned *pgxpool.Conn wrappers would never release their slots.
+//
+// With acquireListenerConn's Hijack, each call is independent of pool state:
+// we run many more iterations than MaxConns and assert the pool never becomes
+// exhausted and AcquiredConns returns to zero after each cycle.
+func TestAcquireListenerConn_SurvivesReconnectCyclePastPoolLimit(t *testing.T) {
+	const maxConns int32 = 3
+	const iterations = int(maxConns) * 4 // well past MaxConns
+
+	pool, cleanup := setupPostgresPlain(t, maxConns)
+	defer cleanup()
+
+	for i := 0; i < iterations; i++ {
+		// Each iteration uses its own short timeout so a would-be slot leak
+		// surfaces as a deadline-exceeded error rather than hanging the test.
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		raw, err := acquireListenerConn(ctx, pool)
+		require.NoErrorf(t, err, "iteration %d acquire should succeed; slot leak would starve the pool here", i)
+		require.NotNil(t, raw)
+
+		require.Equalf(t, int32(0), pool.Stat().AcquiredConns(),
+			"iteration %d: AcquiredConns must be zero after Hijack", i)
+
+		// Simulate pgxlisten's `defer conn.Close(ctx)` when Listen exits.
+		err = raw.Close(ctx)
+		require.NoError(t, err, "iteration %d close should not error", i)
+		cancel()
+	}
+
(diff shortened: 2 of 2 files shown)


pkg/repository/multiplexer.go
```diff
@@ -57,12 +77,7 @@ func (m *multiplexedListener) startListening() {
 	// listen for multiplexed messages
 	listener := &pgxlisten.Listener{
 		Connect: func(ctx context.Context) (*pgx.Conn, error) {
-			// Acquire a new connection each time
-			poolConn, err := m.pool.Acquire(ctx)
-			if err != nil {
-				return nil, err
-			}
-			return poolConn.Conn(), nil
```
gregfurman, line 65: Can we not just change this `Conn` to a `Hijack`?

pkg/repository/multiplexer_listen_test.go
gregfurman, line 1: I'm wondering if there's instead an e2e test we could write. Do you know how we could manually induce this error case?

pkg/repository/multiplexer.go
```diff
@@ -17,6 +17,26 @@ import (
...
+//
+// pgxlisten takes a raw *pgx.Conn from Connect and closes it via
+// `defer conn.Close(ctx)` when Listen exits (e.g. the server-side kills the
+// conn via idle_session_timeout). If we returned poolConn.Conn() without
+// Hijack, the *pgxpool.Conn wrapper would fall out of scope with no Release()
+// call, leaving pgxpool's bookkeeping permanently counting the slot as
+// "acquired." Each reconnect cycle would then leak one slot until the pool
+// is exhausted. Hijack transfers ownership out of the pool immediately; the
+// raw conn is closed cleanly by pgxlisten, and the next Connect call acquires
+// a fresh slot without any orphaned bookkeeping.
```
gregfurman: This seems a bit excessive for private function. We should rather be linking to a GH issue or comment instead of inlining this whole explanation i.e

```golang
// NOTE: Hijack allows us to transfer ownership of the conn to the caller, so the pool slot is released immediately.
// See github.com/hatchet-dev/hatchet/<issue>
```

PR #3710, opened 2026-04-24
fix(typescript-sdk): add workflow span fallback for dashboard-triggered runs

### LINKED ISSUE #3706: [BUG] [TS SDK OTEL] No workflow-level parent span when a run is triggered from the Hatchet dashboard
## Summary

When a workflow run is triggered from the Hatchet dashboard (manual event push or manual run), the resulting trace in the observability backend has no workflow-level parent span. The step runs appear as orphaned siblings under the trace root, because the only place the TS SDK emits a workflow-level span is at the trigger site — and the dashboard is not an instrumented trigger site.

## Current behavior

Given a workflow `find-subprocessors` triggered by pushing `subprocessor.created` from the dashboard:

Image

The step spans share a trace ID (the engine propagates a `traceparent` via `action.additionalMetadata`), but the parent span ID they reference was never exported — so no workflow envelope shows up in the trace viewer.

## Expected behavior

A workflow-level span should exist around all step runs of a dashboard-triggered run, e.g.:

Trace 76f869...
└── hatchet.workflow_run  (find-subprocessors, workflow_run_id=…)
├── hatchet.start_step_run.find-subprocessors:resolve-parent-company
├── hatchet.start_step_run.find-subprocessors:search-company-subprocessors
└── ...

## Why this matters

The dashboard is the canonical way to trigger runs outside the SDK — retries, one-off replays, operator debugging, first-run-during-setup, etc. all go through it. Today every one of those runs loses its workflow envelope in tracing, which is exactly when you most want to see it.

## Root cause

In [`sdks/typescript/src/opentelemetry/instrumentor.ts`](https://github.com/hatchet-dev/hatchet/blob/main/sdks/typescript/src/opentelemetry/instrumentor.ts), workflow-level spans (`hatchet.push_event`, `hatchet.run_workflow`, `hatchet.schedule_workflow`) are created only at the trigger site. Worker-side instrumentation only emits per-step `hatchet.start_step_run.*` spans. The dashboard is not a trigger site that participates in the user's trace export, so the parent span simply doesn't exist.

## Suggested fix

Have the dashboard's "push event" / "trigger run" action emit a span (or call through an instrumented path) so the `traceparent` it injects into `additionalMetadata` points at a span that actually gets exported. The simplest form: the dashboard backend opens a `hatchet.dashboard_trigger` (or equivalent) span when the user clicks trigger, and the existing propagation machinery in the TS SDK will chain the step spans under it automatically.

onthebed:
Dashboard-triggered runs arrive at the worker with a propagated `traceparent`, but no exported producer span from the TS SDK side. That left step spans attached to a missing parent and the workflow envelope never showed up in traces.

This adds a worker-side `hatchet.workflow_run` fallback span when a valid upstream trace context exists but the metadata was not marked as coming from an SDK-exported trigger span. SDK-triggered runs still use their existing producer spans because instrumented trigger paths now mark the injected metadata before the worker sees it.

Verification:
- `pnpm test:unit -- --runInBand src/opentelemetry/instrumentor.test.ts`
- `pnpm lint:check && npx tsc`
- `pnpm test:unit` is still noisy in this environment because unrelated admin-client tests hit local gRPC connection failures (`ECONNREFUSED 127.0.0.1:50051`).

Fixes #3706

### DIFF
--- a/sdks/typescript/src/opentelemetry/instrumentor.test.ts
+++ b/sdks/typescript/src/opentelemetry/instrumentor.test.ts
@@ -1,0 +1,71 @@
+import { propagation } from '@opentelemetry/api';
+import { W3CTraceContextPropagator } from '@opentelemetry/core';
+import {
+  BasicTracerProvider,
+  InMemorySpanExporter,
+  SimpleSpanProcessor,
+} from '@opentelemetry/sdk-trace-base';
+import { HatchetInstrumentor } from './instrumentor';
+
+describe('HatchetInstrumentor worker spans', () => {
+  it('creates a workflow-level parent span for dashboard-triggered step runs', async () => {
+    const exporter = new InMemorySpanExporter();
+    const provider = new BasicTracerProvider({
+      spanProcessors: [new SimpleSpanProcessor(exporter)],
+    });
+
+    propagation.setGlobalPropagator(new W3CTraceContextPropagator());
+
+    const dashboardTracer = provider.getTracer('dashboard-test');
+    const dashboardSpan = dashboardTracer.startSpan('dashboard-trigger');
+    const dashboardSpanContext = dashboardSpan.spanContext();
+    const carrier: Record<string, string> = {
+      traceparent: `00-${dashboardSpanContext.traceId}-${dashboardSpanContext.spanId}-01`,
+    };
+    dashboardSpan.end();
+
+    class FakeWorker {
+      workerId = 'worker-1';
+
+      async handleStartStepRun(_action: unknown) {
+        return undefined;
+      }
+
+      async handleCancelStepRun() {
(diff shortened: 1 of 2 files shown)


mrkaye97:
hi @onthebed, thanks for the contribution! I'd like a little more context here: If you're triggering from the dashboard, why would there be a parent span from the SDK? The SDK doesn't know anything about this run until it lands on a worker

gregfurman:
@onthebed 👋 Do you still have capacity to address @mrkaye97's comment? If not, perhaps we can place this into draft if there is still more to consider with this approach 👀 

Lmk!

gregfurman:
Closing this as stale. Feel free to re-open when you have capacity to address the review.

PR #3713, opened 2026-04-26
fix: increase max password length to 64 and improve validation error message

### LINKED ISSUE #3712: [BUG] Passwords have an undocumented 32 character limit
**Describe the issue**
When attempting to log in with a password longer than 32 characters, login fails with:

> Invalid password. Passwords must be at least 8 characters in length, contain an upper and lowercase letter, and contain at least one number.

**Environment**
- SDK: N/A
- Engine: self-hosted v0.83.48

**Expected behavior**

Either arbitrary length passwords should be supported, or the error message should be updated to list the maximum password length.

Set a password to a value that contains more than 32 characters

33 character long password:

Image

32 character long password:

Image

**Additional context**

Tangentially related: https://github.com/hatchet-dev/hatchet/issues/3711

NathanDrake007:
# Description
## Summary
Fixes a bug where login silently fails for passwords longer than 32 characters
with a misleading error message that gives no indication a maximum length exists.
Modern security guidelines (like NIST and OWASP) now recommend maximum atleast 64 character length.
## Problem
The `passwordValidation` function enforced a hard cap of 32 characters, but the
error message only described minimum requirements:
> "Passwords must be at least 8 characters in length, contain an upper and
> lowercase letter, and contain at least one number."
Users with valid passwords longer than 32 characters had no way to understand
why login was failing.
Fixes #3712
## Type of change
## What's Changed
- [x] Raised the maximum password length from 32 to 64 characters in `passwordValidation`.
- [x] Updated the validation error message to state the full range: _"Passwords must be between 8 and 64 characters in length...".
- [x] Added unit tests covering: passwords in the 32–64 range, exactly 64 characters, and passwords exceeding 64 characters

**Note:** 64 was chosen after auditing downstream constraints:
- `bcrypt.GenerateFromPassword` silently truncates at 72 bytes — 64 ASCII chars (64 bytes) is safely under this limit

### DIFF
--- a/frontend/app/src/pages/auth/login/components/user-login-form.tsx
+++ b/frontend/app/src/pages/auth/login/components/user-login-form.tsx
@@ -10,7 +10,13 @@ import { z } from 'zod';
 
 const schema = z.object({
   email: z.string().email('Invalid email address'),
-  password: z.string().min(8, 'Password must be at least 8 characters long'),
+  password: z
+    .string()
+    .min(8, 'Password must be at least 8 characters long')
+    .max(64, 'Password must be at most 64 characters long')
+    .regex(/[A-Z]/, 'Password must contain at least one uppercase letter')
+    .regex(/[a-z]/, 'Password must contain at least one lowercase letter')
+    .regex(/[0-9]/, 'Password must contain at least one number'),
 });
 
 interface UserLoginFormProps {
--- a/frontend/app/src/pages/auth/login/components/user-login-form.tsx
+++ b/frontend/app/src/pages/auth/login/components/user-login-form.tsx
@@ -10,13 +10,7 @@ import { z } from 'zod';
 
 const schema = z.object({
   email: z.string().email('Invalid email address'),
-  password: z
-    .string()
-    .min(8, 'Password must be at least 8 characters long')
-    .max(64, 'Password must be at most 64 characters long')
-    .regex(/[A-Z]/, 'Password must contain at least one uppercase letter')
-    .regex(/[a-z]/, 'Password must contain at least one lowercase letter')
-    .regex(/[0-9]/, 'Password must contain at least one number'),
+  password: z.string().min(8, 'Password must be at least 8 characters long'),
 });
 
 interface UserLoginFormProps {
--- a/pkg/validator/default.go
+++ b/pkg/validator/default.go
@@ -15,7 +15,7 @@ import (
 
 const (
 	EmailErr       = "Invalid email address"
-	PasswordErr    = "Invalid password. Passwords must be at least 8 characters in length, contain an upper and lowercase letter, and contain at least one number."
+	PasswordErr    = "Invalid password. Passwords must be between 8 and 64 characters in length, contain an upper and lowercase letter, and contain at least one number."
 	UUIDErr        = "Invalid UUID reference"
 	HatchetNameErr = "Hatchet names must match the regex ^[a-zA-Z0-9\\.\\-_]+$"
 	ActionIDErr    = "Invalid action ID. Action IDs must be in the format <integrationId>:<verb>"
--- a/pkg/validator/validator.go
+++ b/pkg/validator/validator.go
@@ -104,7 +104,7 @@ func passwordValidation(pw string) bool {
 		}
 	}
 
-	return hasNumber && hasUpper && hasLower && pwLen >= 8 && pwLen <= 32
+	return hasNumber && hasUpper && hasLower && pwLen >= 8 && pwLen <= 64
 }
 
 func IsValidUUID(u string) bool {
--- a/pkg/validator/validator_test.go
+++ b/pkg/validator/validator_test.go
@@ -68,6 +68,63 @@ func TestValidatorValidDuration(t *testing.T) {
 	assert.NoError(t, err, "no error")
 }
 
+type passwordResource struct {
+	Password string `validate:"password"`
+}
+
+func TestValidatorValidPassword(t *testing.T) {
+	v := newValidator()
+
+	err := v.Struct(&passwordResource{
+		Password: "ValidPass1",
+	})
+
+	assert.NoError(t, err, "should accept a valid password")
+}
+
+func TestValidatorPasswordTooShort(t *testing.T) {
+	v := newValidator()
+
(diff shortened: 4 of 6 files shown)


pkg/validator/validator_test.go
```diff
@@ -68,6 +68,63 @@ func TestValidatorValidDuration(t *testing.T) {
 	assert.NoError(t, err, "no error")
 }
 
+type passwordResource struct {
+	Password string `validate:"password"`
+}
+
+func TestValidatorValidPassword(t *testing.T) {
```
gregfurman: suggestion(non-blocking): Can we rather use a test-table here? Think it'll work better for this boundary testing approach.

pkg/validator/validator.go
```diff
@@ -104,7 +104,7 @@ func passwordValidation(pw string) bool {
 		}
 	}
 
-	return hasNumber && hasUpper && hasLower && pwLen >= 8 && pwLen <= 32
+	return hasNumber && hasUpper && hasLower && pwLen >= 8 && pwLen <= 64
```
gregfurman, line 107: suggestion(non-blocking): Could be a nice addition to include this as a constraint in the frontend login and create user forms 👀

gregfurman (approved):
LGTM! Have manually e2e tested this locally (negative and positive cases for account creation and login) and all works fine.

I've left 2 non-blocking suggestions. Let me know if you'd have the capacity to address this otherwise, I'm happy to merge 🙂

pkg/validator/validator_test.go
NathanDrake007: refactored the test cases.

pkg/validator/validator.go
NathanDrake007: Nice catch, I have updated the forms to use this validation.

pkg/validator/validator_test.go
gregfurman: Ahh I was referring specifically to the password validation case. See you've refactored all cases now. Will give a look

solidDoWant:
Hatchet's password hashing function, [bcrypt](https://github.com/hatchet-dev/hatchet/blob/3d357967fdd28ed08ca4011c0d4ae8fe6268facc/pkg/repository/user.go#L75), has [a max length of 72 bytes](https://pkg.go.dev/golang.org/x/crypto/bcrypt#GenerateFromPassword), and runes in Go are 4 bytes wide. Is the actual max password length 18 characters?

frontend/app/src/pages/auth/login/components/user-login-form.tsx
```diff
@@ -10,7 +10,13 @@ import { z } from 'zod';
...
 const schema = z.object({
   email: z.string().email('Invalid email address'),
-  password: z.string().min(8, 'Password must be at least 8 characters long'),
+  password: z
+    .string()
+    .min(8, 'Password must be at least 8 characters long')
+    .max(64, 'Password must be at most 64 characters long')
+    .regex(/[A-Z]/, 'Password must contain at least one uppercase letter')
+    .regex(/[a-z]/, 'Password must contain at least one lowercase letter')
+    .regex(/[0-9]/, 'Password must contain at least one number'),
```
gregfurman: Taking these changes for a spin made me realise that (from a devex perspective) I'm not sure it actually makes sense to include validations on `login` 😅 Bad suggestion on my part. Can we revert back to how it originally was?

NathanDrake007:
> Hatchet's password hashing function, [bcrypt](https://github.com/hatchet-dev/hatchet/blob/3d357967fdd28ed08ca4011c0d4ae8fe6268facc/pkg/repository/user.go#L75), has [a max length of 72 bytes](https://pkg.go.dev/golang.org/x/crypto/bcrypt#GenerateFromPassword), and runes in Go are 4 bytes wide. Is the actual max password length 18 characters?

Actually, the 72-byte limit in bcrypt doesn't strictly mean 18 characters. In Go, strings are UTF-8 encoded, so while a rune is 4 bytes, most standard characters (ASCII) only use 1 byte.

If a user uses emojis or complex symbols, those do scale up to 4 bytes. By capping the limit at 64 characters in the frontend and backend, this shouldn't be a problem.

pkg/validator/default.go
```diff
@@ -15,7 +15,7 @@ import (
 
 const (
 	EmailErr       = "Invalid email address"
-	PasswordErr    = "Invalid password. Passwords must be at least 8 characters in length, contain an upper and lowercase letter, and contain at least one number."
+	PasswordErr    = "Invalid password. Passwords must be between 8 and 64 characters in length, contain an upper and lowercase letter, and contain at least one number."
```
gregfurman, line 18: I just realised we're returning this `PasswordErr` at the API layer -- which isn't great IMO. Like, we shouldn't really be making our validations on a password string public.

See where this is returned:

https://github.com/hatchet-dev/hatchet/blob/3d357967fdd28ed08ca4011c0d4ae8fe6268facc/api/v1/server/handlers/users/update_login.go#L27-L31

Perhaps we change this to instead return a generic `ErrInvalidCredentials` instead of returning exactly what was wrong with the password on validation failure i.e

```golang
	// validate the request
	if apiErrors, err := u.config.Validator.ValidateAPI(request.Body); err != nil {
		return nil, err
	} else if apiErrors != nil {
		return gen.UserUpdateLogin400JSONResponse(
			apierrors.NewAPIErrors(ErrInvalidCredentials),
		), nil
	}
```

Wdyt?

gregfurman (changes requested):
Think the API should be returning a generic error when a `login` fails, but allow for specifics when a `create-user` fails. Have left a comment for this.

Also, am unsure the devex of password validation on `login` at the form level makes much sense 🤔

pkg/validator/default.go
NathanDrake007: I mean, yeah, technically anyone with a burner email and two minutes can reverse-engineer the rules. 😂 It’s basically security through obscurity, but why make it easy for them? 😈

Also, we don't even need to bother users with these complex error messages on the login form anyway. At that point, we're just checking if the password matches what's in the DB, not validating it against the registration rules again.

I'll make the changes.

gregfurman (approved):
LGTM!

PR #3760, opened 2026-04-28
Chore: move durable tasks vs. dags docs section to the cookbooks

mrkaye97:
# Description

Moving this page to the cookbooks under workflow patterns

## Type of change


abelanger5 (approved):
@mrkaye97 probably do want to add a redirect here, but otherwise looks good

PR #3774, opened 2026-04-28
Fix: Attempt at improving worker actions query

mrkaye97:
# Description

Lots of slow traces, which I think are happening because we're pulling a ton of rows down out of the db. I figured we could hash the action names and store them with the worker, and then try to only look up actions for each hash once, and then map them back to the relevant worker that way, which should cut down the number of rows we need to pull out of the db by a factor of the number of replicas of a worker any given tenant has

Unfortunately we can't make the action hash non-null and backfill so we need to also have a lookup by worker id for backwards compat for now

## Type of change


pkg/repository/sqlcv1/workers.sql
```diff
@@ -247,19 +247,28 @@ GROUP BY "tenantId"
...
-LEFT JOIN "Action" a ON aw."A" = a.id
+JOIN "_ActionToWorker" aw ON w.id = aw."B"
+JOIN "Action" a ON aw."A" = a.id
+WHERE
+    a."tenantId" = @tenantId::UUID
+    AND w.id = ANY(@workerIds::UUID[])
+;
+
+-- name: GetWorkerActionsByWorkerActionHash :many
+SELECT DISTINCT ON (w."actionHash")
```
mrkaye97: goal here is we'll just move to this completely in a little bit, once all the workers have hashes

grutt (approved):

PR #3781, opened 2026-04-29
Fix--control-plane-bash

grutt:
# Description

Small fixes while setting up tooling for running the control plane

## Type of change


internal/services/ticker/ticker.go
```diff
@@ -137,6 +137,15 @@ func (t *TickerImpl) Start() (func() error, error) {
 
 	t.l.Debug().Ctx(ctx).Msgf("starting ticker %s", t.tickerId)
 
+	_, err := t.repov1.Ticker().CreateNewTicker(ctx, &v1.CreateTickerOpts{
```
grutt, line 147: @juliusgeo just wanted to confirm there wasnt a reason for the old order? afict there's a FK for cron refs on tickers so we need to create one first

frontend/app/vite.config.ts
```diff
@@ -3,6 +3,11 @@ import path from 'path';
 import react from '@vitejs/plugin-react';
 import { defineConfig } from 'vite';
 
+const apiProxyTarget =
+  process.env.VITE_API_PROXY_TARGET ?? 'http://127.0.0.1:8080';
```
grutt, line 7: quality of life exposing config for local tooling

cmd/hatchet-admin/cli/token.go
```diff
@@ -41,10 +42,9 @@ var tokenCreateAPICmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(tokenCmd)
 	tokenCmd.AddCommand(tokenCreateAPICmd)
-	tenantId := tokenTenantId.String()
 
 	tokenCreateAPICmd.PersistentFlags().StringVar(
-		&tenantId,
+		&tokenTenantIdStr,
```
grutt, line 47: maybe im missing something but this just didnt work before...

frontend/app/src/lib/api/api.ts
```diff
@@ -69,6 +69,18 @@ function readStoredTenantId(): string | null {
   }
 }
 
+function readTenantIdFromUrl(url: string | undefined): string | null {
```
grutt: @abelanger5 i think these endpoints should use the id in the url -> localstore

OSS tenant routes: /api/v1/tenants/:tenant/...
stable tenant routes: /api/v1/stable/tenants/:tenant/...
cloud tenant routes: /api/v1/cloud/tenants/:tenant/...
billing tenant routes: /api/v1/billing/tenants/:tenant/...

cmd/hatchet-admin/cli/token.go
abelanger5: probably broke after the uuid migration, this looks correct now

frontend/app/src/lib/api/api.ts
abelanger5: why is this necessary?

grutt: seeing weird behavior locally after sign in

internal/services/ticker/ticker.go
juliusgeo: Yeah, if I remember correctly I just used the ordering that existed prior.

frontend/app/src/lib/api/api.ts
abelanger5: what's the behavior?

grutt: i dont remember if 403s or 404s, usually it was after i reset local volumes so i'm assuming we have an old tenant id in local store that no longer exists/no permissions

juliusgeo (approved):

frontend/app/src/lib/api/api.ts
abelanger5: shouldn't we handle that case then? we should assume that any request using the OSS or cloud API should be proxied in control plane mode, so this really should be unnecessary. if we no longer have permissions on a tenant that's in local storage, we should probably choose a different tenant to automatically navigate to?

grutt: maybe, im not really sure i understand why we're going to local store for this tenant id in these cases though... the tenant id is in the url context

grutt: let me take a closer look at this. it was a quick patch for a case i was hitting fairly consistently but maybe there's a broader issue here

frontend/app/src/pages/authenticated.tsx
```diff
@@ -262,12 +266,22 @@ function AuthenticatedInner() {
...
+        if (loaderData.isControlPlaneEnabled) {
+          localStorage.removeItem(CONTROL_PLANE_TENANT_STORAGE_KEY);
+        }
       }
 
       const targetTenant =
         lastTenantInMemberships ?? tenantMemberships[0].tenant;
 
       if (targetTenant) {
+        if (loaderData.isControlPlaneEnabled) {
```
grutt, line 278: @abelanger5 afict i think this should resolve the other conversation

abelanger5 (approved):

frontend/app/src/pages/authenticated.tsx
abelanger5: yep, this looks correct

PR #3782, opened 2026-04-29
fix: workflow reschedule update fail

jishnundth:
# Description

Schedule workflow update is failing due to wrong if condition logic

## Type of change


mrkaye97 (approved):
thanks!

PR #3783, opened 2026-04-29
Added validation for seed password

### LINKED ISSUE #3711: [BUG] ADMIN_PASSWORD env var allows for invalid passwords
**Describe the issue**
When seeing the database with admin credentials via `ADMIN_PASSWORD`, the seed job succeeds with any password value. If this doesn't meet the "at least 8 characters in length, contain an upper and lowercase letter, and contain at least one number" requirement, then users cannot log in.

Fixing this requires deleting the database and re-seeding, because the seed job will not update an existing admin user.

**Environment**
- SDK: N/A
- Engine: self-hosted v0.83.48

**Expected behavior**
The seed job should reject the password.

Ideally there would be a flag to force the seed job to update the admin user password as well, to aid in recovery.

Seed a new password with the env var `ADMIN_PASSWORD=mypass` and try to log in.

Image

**Additional context**
N/A

NathanDrake007:
# Description

Fixes the seed job accepting any value for ADMIN_PASSWORD without validation, which resulted in an admin account that could not be used to log in. The seed job now rejects passwords that don't meet complexity requirements before hashing and writing to the database.

Fixes #3711

Type of change

What's Changed
- Renamed passwordValidation to ValidatePassword in pkg/validator/validator.go to export it for reuse outside the package
- Added password validation to SeedDatabase in cmd/hatchet-admin/cli/seed/seed.go — the seed job now returns a clear error and exits early if ADMIN_PASSWORD does not meet complexity requirements (8–64 characters, at least one uppercase letter, one lowercase letter, and one number)

### DIFF
--- a/cmd/hatchet-admin/cli/seed/seed.go
+++ b/cmd/hatchet-admin/cli/seed/seed.go
@@ -19,7 +19,12 @@ func SeedDatabase(dc *database.Layer) error {
 
 	if shouldSeedUser {
 		// validate the password meets complexity requirements before hashing
-		if !validator.ValidatePassword(dc.Seed.AdminPassword) {
+		v := validator.NewDefaultValidator()
+		opts := struct {
+			Password string `validate:"password"`
+		}{dc.Seed.AdminPassword}
+
+		if err := v.Validate(opts); err != nil {
 			return fmt.Errorf("ADMIN_PASSWORD does not meet requirements: must be between 8 and 64 characters and contain at least one uppercase letter, one lowercase letter, and one number")
 		}
 
--- a/cmd/hatchet-admin/cli/seed/seed.go
+++ b/cmd/hatchet-admin/cli/seed/seed.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/hatchet-dev/hatchet/pkg/config/database"
 	v1 "github.com/hatchet-dev/hatchet/pkg/repository"
+	"github.com/hatchet-dev/hatchet/pkg/validator"
 )
 
 func SeedDatabase(dc *database.Layer) error {
@@ -17,6 +18,11 @@ func SeedDatabase(dc *database.Layer) error {
 	var userID uuid.UUID
 
 	if shouldSeedUser {
+		// validate the password meets complexity requirements before hashing
+		if !validator.ValidatePassword(dc.Seed.AdminPassword) {
+			return fmt.Errorf("ADMIN_PASSWORD does not meet requirements: must be between 8 and 64 characters and contain at least one uppercase letter, one lowercase letter, and one number")
+		}
+
 		// seed an example user
 		hashedPw, err := v1.HashPassword(dc.Seed.AdminPassword)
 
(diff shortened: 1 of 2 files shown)


cmd/hatchet-admin/cli/seed/seed.go
```diff
@@ -10,13 +10,19 @@ import (
...
 
 func SeedDatabase(dc *database.Layer) error {
 	shouldSeedUser := dc.Seed.AdminEmail != "" && dc.Seed.AdminPassword != ""
 	var userID uuid.UUID
 
 	if shouldSeedUser {
+		// validate the password meets complexity requirements before hashing
+		if !validator.ValidatePassword(dc.Seed.AdminPassword) {
+			return fmt.Errorf("ADMIN_PASSWORD does not meet requirements: must be between 8 and 64 characters and contain at least one uppercase letter, one lowercase letter, and one number")
+		}
```
gregfurman: While this is technically valid, I'm thinking that we should rather just return a `DefaultValidator` and pass in an anonymous struct with a password field i.e

```golang
validator := validator.NewDefaultValidator()
opts := struct{Password string `json:"password"`}{dc.Seed.AdminPassword}
// Or maybe even just use the already existing CreateUserOpts 👇 
// opts := repository.CreateUserOpts{Password: dc.Seed.AdminPassword} 
err := validator.Validate(opts)
// ...
```

That way we don't have to expose this function publicly and can just rely on our normal validation mechanism for this check.

gregfurman (changes requested):

cmd/hatchet-admin/cli/seed/seed.go
NathanDrake007: updated the code

pkg/validator/validator.go
```diff
@@ -89,7 +89,10 @@ func newValidator() *validator.Validate {
 	return validate
 }
 
-func passwordValidation(pw string) bool {
+// ValidatePassword returns true if the password meets complexity requirements:
+// between 8 and 64 characters, with at least one uppercase letter, one lowercase
+// letter, and one number.
+func ValidatePassword(pw string) bool {
```
gregfurman: Can we revert this now that we're using the validator approach?

```suggestion
func passwordValidation(pw string) bool {
```

pkg/validator/validator.go
```diff
@@ -28,7 +28,7 @@ func newValidator() *validator.Validate {
 	})
 
 	_ = validate.RegisterValidation("password", func(fl validator.FieldLevel) bool {
-		return passwordValidation(fl.Field().String())
+		return ValidatePassword(fl.Field().String())
```
gregfurman:
```suggestion
		return passwordValidation(fl.Field().String())
```

gregfurman (changes requested):

pkg/validator/validator.go
NathanDrake007: Yeah, sorry, forgot to revert this.

gregfurman (approved):

PR #3794, opened 2026-04-30
docs(ts): use ctx.logger.info in v1 TypeScript examples

ltsSmitty:
# Description

Updates v1 TypeScript examples under `sdks/typescript/src/v1/examples` so task and workflow handlers log through `ctx.logger.info` instead of `console.log`, matching the SDK’s supported logging path and keeping run output visible in Hatchet’s logging pipeline.

I'm not sure if this was passed over (in what I assume was an upgrade from the non-versioned examples to v1), or if it was intentional; but seeing last month's celebration posts of new logging capabilities, I thought it was worth updating the examples to show it. In my new experience using Hatchet, my LLM which had access to the docs was able to set up logging using the provided ctx.logger without any issue, but I'd like for the examples to be clearer on what I assume is best practice.

Where handlers did not already take context, the second parameter `ctx` was added (e.g. webhook tasks, rate-limit examples, the non-durable task in the durable workflow example). Middleware `before` hooks and abort-signal examples now use the same logger on the provided context. Driver scripts (`run.ts`, `event.ts`, e2e helpers, agents, etc.) were left unchanged because they do not receive task context.

## Type of change

### DIFF
--- a/sdks/typescript/src/v1/examples/bulk_operations/workflow.ts
+++ b/sdks/typescript/src/v1/examples/bulk_operations/workflow.ts
@@ -4,7 +4,7 @@ export const bulkReplayTest1 = hatchet.task({
   name: 'bulk-replay-test-1',
   retries: 1,
   fn: async (_input, ctx) => {
-    ctx.logger.info('retrying bulk replay test task', ctx.retryCount());
+    ctx.logger.info('retrying bulk replay test task', { retryCount: ctx.retryCount() });
     if (ctx.retryCount() === 0) {
       throw new Error('This is a test error to trigger a retry.');
     }
@@ -15,7 +15,7 @@ export const bulkReplayTest2 = hatchet.task({
   name: 'bulk-replay-test-2',
   retries: 1,
   fn: async (_input, ctx) => {
-    ctx.logger.info('retrying bulk replay test task', ctx.retryCount());
+    ctx.logger.info('retrying bulk replay test task', { retryCount: ctx.retryCount() });
     if (ctx.retryCount() === 0) {
       throw new Error('This is a test error to trigger a retry.');
     }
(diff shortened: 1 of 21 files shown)


promptless-for-oss:

Triggered by this PR

Updated documentation to reflect the `ctx.logger` pattern instead of the deprecated `ctx.log()` method. Changes include updates to the logging guide, middleware documentation, and added comprehensive `logger` property documentation to the TypeScript Context reference.

**3 files changed:**
- `frontend/docs/pages/v1/logging.mdx`
- `frontend/docs/pages/v1/middleware.mdx`
- `frontend/docs/pages/reference/typescript/Context.mdx`

Review: [Update TypeScript logger documentation](https://app.gopromptless.ai/suggestions/c07100ff-9415-4cdf-b027-0206fd3349ee)

sdks/typescript/src/v1/examples/bulk_operations/workflow.ts
```diff
@@ -4,7 +4,7 @@ export const bulkReplayTest1 = hatchet.task({
   name: 'bulk-replay-test-1',
   retries: 1,
   fn: async (_input, ctx) => {
-    console.log('retrying bulk replay test task', ctx.retryCount());
+    ctx.logger.info('retrying bulk replay test task', ctx.retryCount());
```
gregfurman: The `ctx.logger` interface expects these extra args to be in JSON form. See https://docs.hatchet.run/v1/logging#using-the-contextlog-method

Where the log events being sent to the engine would fail with

```
[WARN/Dispatcher] Could not put log: /EventsService/PutLog INVALID_ARGUMENT: Invalid metadata JSON: json: cannot unmarshal number into Go value of type map[string]interface {}
```
Instead, you'd need to interpolate this variable into the log message beforehand:

```ts
    ctx.logger.info(`retrying bulk replay test task ${ctx.retryCount()}`);
```

Or pass in the extra params in JSON form (which I'm incidentally unsure how to actually see these extra kwargs in the dashboard 🤷):

```ts
    ctx.logger.info('retrying bulk replay test task', {"retryCount": ctx.retryCount()});
```

gregfurman (changes requested):

ltsSmitty:
Great catch, sloppy of me to miss that. I'll address that.

On further digging, it looks like the metadata passed to the BYO structured logger isn't being passed to the LogLines, which explains your difficultly finding the kwargs. I'll submit that as a separate issue/PR.

image

gregfurman (approved):
LGTM :shipit: Thanks for addressing my comments and opening that extra issue!

PR #3804, opened 2026-05-02
feat(otel): add useLinksInsteadOfParent option to HatchetInstrumentor

### LINKED ISSUE #3641: [FEAT] OTel: option to use span links instead of parent context for fire-and-forget child runs
## Problem                                                                                                              
                                                                                                                          
  `HatchetInstrumentor` currently wires every spawned run as a parent-child OTel relationship: the caller's trace context 
  is injected into `additionalMetadata` (`instrumentor.js`, `injectContext`) and the worker extracts it and uses it as
   the **parent** of `hatchet.start_step_run` (`instrumentor.js`, passed as `parentContext` to `startActiveSpan`).    
                                                            
  This is the right default for synchronous, awaited orchestration. It is the **wrong** default for fire-and-forget       
  patterns, which are very common with Hatchet:
                                                                                                                          
  - Calling `runNoWait` in a loop to fan out N child runs.                                                                
  - Children whose start time depends on rate limits, queue depth, or worker availability.
  - Long-running children spawned from a short-lived parent.                                                              
                                                                                                                          
  In all of these, the parent trace's duration is artificially extended to whenever the **slowest** child finishes,       
  because OTel backends keep a trace "open" while any descendant span is still active. Concretely:                        
                                                                                                                          
  - Aggregate p50/p95 of the parent transaction become meaningless (they reflect downstream rate-limit waits, not the     
  parent's own work).
  - The waterfall view is dominated by N deeply nested `hatchet.run_workflow` / `hatchet.start_step_run` spans.           
  - Trace-search and trace-duration alerts on the parent become noisy / useless.                                          
                                                                                                                          
  OpenTelemetry has a first-class concept that is designed exactly for this: **span links**. A link references the        
  spawning span without making the child causally part of the same trace, so each child becomes its own trace (cheap to   
  query, navigable from the parent), and the parent trace closes when the spawn loop returns.                             
                                                            
  ## Proposal

  Add a config option to `HatchetInstrumentor` so users can opt specific task types into link semantics instead of parent 
  semantics. A predicate is the most flexible shape:
                                                                                                                          
  ```ts                                                     
  new HatchetInstrumentor({
    // Return true to use a span link instead of parent context for this task's                                           
    // hatchet.start_step_run span. The actionId is the worker's task identifier.
    useLinksInsteadOfParent: (actionId: string) => boolean,                                                               
  });                                                       
  ```                                                                                                                     
                                                            
  Behavior when the predicate returns `true` for a given action:                                                          
                                                            
  - In `instrumentor.js`, instead of calling `startActiveSpan(name, opts, parentContext, fn)`, call `startSpan(name, {
   ...opts, links: [{ context: spanContextFromParent }] })` (no parent context argument). The child span becomes a root in
   a new trace, with a link pointing back to the spawning span.                                                           
  - Caller side (`instrumentor.js:325`) is unchanged — `traceparent` is still injected into `additionalMetadata`. The
  worker just chooses to interpret it as a link rather than a parent.                                                     
  
  Default: `() => false`, preserving today's behavior exactly.                                                            
                                                            
  ## Why per-task granularity (not a global flag)                                                                         
                                                            
  A single Hatchet app legitimately wants both behaviors. Synchronous orchestrators (`await workflow.run(...)`) want      
  parent-child so latency rolls up. Fan-out spawners (`workflow.runNoWait(...)` in a loop) want links so the parent trace
  doesn't get held open. A global boolean would force one or the other; a per-task predicate lets each workflow declare   
  what it is.                                               

  Per-call granularity (a `runNoWait(input, { trace: "link" })` option) would be strictly more flexible, but it's a bigger
   API surface change and the per-task predicate covers the vast majority of cases — usually the "kind of relationship" is
   a property of the task, not the call site.                                                                             
                                                            
  ## Workaround today

  ```ts
  import { context, ROOT_CONTEXT } from "@opentelemetry/api";
                                                                                                                          
  context.with(ROOT_CONTEXT, () => {
    void someWorkflow.runNoWait(input);                                                                                   
  });                                                                                                                     
  ```
                                                                                                                          
  This detaches the child from the current trace, so it starts as its own root. It works but loses the cross-trace link   
  entirely (no clickable jump back from child to spawner), and it pushes OTel plumbing into application code at every
  spawn site.

purva-8:
Closes #3641

## Problem

`HatchetInstrumentor` currently forces every spawned run into a parent→child OTel relationship. For fire-and-forget patterns (`runNoWait` fan-outs, long-running children, rate-limited queues) this is wrong: the parent trace stays open until the slowest child finishes, which inflates p50/p95 metrics and pollutes the waterfall view.

OTel's [span links](https://opentelemetry.io/docs/concepts/signals/traces/#span-links) are the right primitive here — the child starts a new root trace with a navigable link back to the spawning span, so the parent trace closes as soon as the spawn loop returns.

## Solution

Add a `useLinksInsteadOfParent` predicate to `HatchetInstrumentationConfig`:

```ts
new HatchetInstrumentor({
  // Return true to use a span link instead of parent context for this task.
  useLinksInsteadOfParent: (actionId) => actionId.startsWith('fan-out-worker:'),
});
```

**When the predicate returns `true` for a given `actionId`:**
- `handleStartStepRun` extracts the `SpanContext` from the incoming `additionalMetadata` (same as before)
- Calls `startActiveSpan(name, { kind, attributes, links: [{ context: parentSpanCtx }] }, fn)` — **no parent context argument**, so the child span becomes a root in a new trace with a link pointing back to the spawning span
- If the extracted span context is invalid (e.g. no `traceparent` in metadata), the links array is empty and the span still starts cleanly as a root

**When the predicate returns `false` (or is not provided):**
- Existing `startActiveSpan(name, opts, parentContext, fn)` path is used - **zero behaviour change**

Default: `undefined` → all spans use parent-child semantics, fully backwards compatible.

## Changes

- `sdks/typescript/src/opentelemetry/instrumentor.ts` - new config option + updated `_patchHandleStartStepRun`
- `sdks/typescript/src/opentelemetry/instrumentor.test.ts` - 5 unit tests (default, predicate=false, predicate=true, actionId forwarding, invalid-context fallback)

## Test plan

- [x] `pnpm test:unit` - all 195 tests pass (2 pre-existing skips)
- [x] `tsc --noEmit` - no type errors
- [ ] Manual verification with a real Hatchet worker + an OTel backend (Jaeger / Honeycomb) to confirm child spans appear as separate root traces with a link

### DIFF
--- a/sdks/typescript/src/opentelemetry/instrumentor.links.test.ts
+++ b/sdks/typescript/src/opentelemetry/instrumentor.links.test.ts
@@ -1,0 +1,274 @@
+/**
+ * Tests for HatchetInstrumentor.useLinksInsteadOfParent option.
+ *
+ * Verifies that fire-and-forget child runs use OTel span links instead of
+ * parent-child relationships when the predicate returns true, while
+ * preserving the default parent-child behaviour when it returns false.
+ *
+ * These tests use jest module mocks so they live in a separate file from
+ * instrumentor.test.ts, which uses a real OTel SDK provider.
+ */
+
+// Minimal OTel span mock
+const makeMockSpan = () => ({
+  end: jest.fn(),
+  recordException: jest.fn(),
+  setStatus: jest.fn(),
+  spanContext: jest.fn(() => ({
+    traceId: 'a'.repeat(32),
+    spanId: 'b'.repeat(16),
+    traceFlags: 1,
+  })),
+});
+
+// Build a mock tracer whose startActiveSpan captures its arguments.
+const makeTracerMock = (span = makeMockSpan()) => {
+  const calls: IArguments[] = [];
+  const tracer = {
+    _calls: calls,
+    _span: span,
+    startActiveSpan: jest.fn(function (...args: unknown[]) {
+      // The last argument is always the callback (fn).
+      const fn = args[args.length - 1] as (span: unknown) => unknown;
+      return fn(span);
+    }),
+  };
+  return tracer;
+};
+
+// Dummy action used in all tests.
+const makeAction = (actionId = 'my-worker:my-task') => ({
+  actionId,
+  tenantId: 'tenant-1',
+  workflowRunId: 'run-1',
+  taskId: 'task-1',
+  taskRunExternalId: 'ext-1',
+  retryCount: 0,
+  parentWorkflowRunId: undefined,
+  childWorkflowIndex: undefined,
+  childWorkflowKey: undefined,
+  actionPayload: '{}',
+  jobName: 'my-task',
+  taskName: 'my-task',
+  workflowId: 'wf-1',
+  workflowVersionId: 'wfv-1',
+  // Encode a fake traceparent in the metadata so extractContext finds a valid context.
+  additionalMetadata: JSON.stringify({
+    traceparent: '00-' + 'a'.repeat(32) + '-' + 'b'.repeat(16) + '-01',
+  }),
+});
+
+// ---------------------------------------------------------------------------
+// Sentinel object used as ROOT_CONTEXT in the mock — allows tests to assert
+// that the implementation explicitly passes ROOT_CONTEXT, not the active ctx.
+// ---------------------------------------------------------------------------
+const MOCK_ROOT_CONTEXT = { _isRootContext: true } as const;
+
+jest.mock('@opentelemetry/api', () => {
+  const validSpanCtx = {
+    traceId: 'a'.repeat(32),
+    spanId: 'b'.repeat(16),
+    traceFlags: 1,
+  };
+
+  // SpanKind.CONSUMER = 4, SpanStatusCode.OK = 1, SpanStatusCode.ERROR = 2
+  return {
+    SpanKind: { INTERNAL: 0, SERVER: 1, CLIENT: 2, PRODUCER: 3, CONSUMER: 4 },
+    SpanStatusCode: { UNSET: 0, OK: 1, ERROR: 2 },
+    ROOT_CONTEXT: MOCK_ROOT_CONTEXT,
+    context: {
+      active: jest.fn(() => ({})),
+      with: jest.fn((_ctx: unknown, fn: () => unknown) => fn()),
+    },
+    propagation: {
+      extract: jest.fn(() => ({ _extracted: true })),
+      inject: jest.fn(),
+    },
+    trace: {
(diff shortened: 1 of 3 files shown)


promptless-for-oss:

Added the new `useLinksInsteadOfParent` option to the TypeScript OpenTelemetry configuration table, documenting how users can use span links instead of parent-child relationships for fire-and-forget patterns.

Review: [Document useLinksInsteadOfParent option](https://app.gopromptless.ai/suggestions/06513bc4-2803-4dbc-a1b1-d7ba309df6fd)

BloggerBust:
Hi @purva-8,  thank you for your contribution. The predicate based opt-in seems like a sensible API for the fire and forget use case from #3641. Using `ROOT_CONTEXT` with span links seems sensible as well.

I found one blocking issue in the tests: the test file currently fails locally for me.

`pnpm exec jest src/opentelemetry/instrumentor.test.ts`

Two tests still expect the old 3 argument `startActiveSpan(name, opts, fn)`, but the implementation now calls:

`startActiveSpan(name, opts, otelApi.ROOT_CONTEXT, fn)`

So these tests need to be updated to match the 4 argument root context change.

Suggested fix:

1. Add `ROOT_CONTEXT` to the `@opentelemetry/api` mock.
2. Update the assertions from `toHaveLength(3)` to `toHaveLength(4)`.
3. Assert that `args[2]` is the mocked `ROOT_CONTEXT`.
4. Update the comment that still describes link mode as the 3 argument overload.
5. Validate that the tests pass by locally running: `pnpm exec jest src/opentelemetry/instrumentor.test.ts`

gregfurman:
👋 @purva-8 Do you still have capacity to address @BloggerBust's review?

purva-8:
Hi @BloggerBust, thanks for the thorough review! All four points are addressed in the latest commit:

- Added `ROOT_CONTEXT` to the `@opentelemetry/api` mock
- Updated all link-mode assertions to `toHaveLength(4)`
- Added `expect(args[2]).toBe(MOCK_ROOT_CONTEXT)` to verify the explicit root context
- Updated the JSDoc comment in `getStartStepRunArgs` to reflect the 4-argument signature

`pnpm exec jest src/opentelemetry/instrumentor.test.ts` passes locally with all 5 tests green.

purva-8:
Hi @gregfurman, yes! Just pushed the fix addressing @BloggerBust's review. All tests pass now.

PR #3816, opened 2026-05-04
feat: swap links to control plane

grutt:
# Description

Swaps all public links to use the control plane url

## Type of change


mrkaye97 (approved):
🙌

PR #3826, opened 2026-05-05
Chore: Auto-hide more generated code in PR reviews

mrkaye97:
# Description

Sets up more linguist rules to auto-hide generated stuff in reviews

## Type of change


grutt (approved):
cool

PR #3830, opened 2026-05-05
fix: Slow cold start when workflows are inactive > 1 day

juliusgeo:
# Description

When a workflow is triggered that was last triggered > 1 day prior, the call to `l.lr.ListQueues` in `acquireQueueLeases` does not pick up that workflow's queue lease. `acquireQueueLeases` is called with a polling interval of 5 seconds, which means that if a workflow is triggered, it would have to wait up to 5 seconds before `acquireQueueLeases` will be called, pick up the queue lease, notify the tenant manager, and finally have `addQueuer` get called in `TenantManager.listenForQueueLeases` to actually create the queue and have it get picked up in `runOptimisticScheduling`. This PR changes the logic in the TenantManager such that when a `check-tenant-queue` message is received, instead of just waking up the existing queues that exist in memory, it also creates new queues for queues that do not exist in memory, thus avoiding the aforementioned wait for the acquireQueueLeases poll. 

Fixes # (issue)

## Type of change

## What's Changed

### DIFF
--- a/pkg/scheduling/v1/queuer.go
+++ b/pkg/scheduling/v1/queuer.go
@@ -103,20 +103,15 @@ func (q *Queuer) Cleanup() {
 }
 
 func (q *Queuer) queue(ctx context.Context) {
-	if ok := q.queueMu.TryLock(); !ok {
-		return
+	ctx, span := telemetry.NewSpan(ctx, "notify-queue")
+	defer span.End()
+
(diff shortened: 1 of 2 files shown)


juliusgeo:
The failing e2e test fails on main as well, so not related to these changes.

grutt (approved):

PR #3831, opened 2026-05-06
Chore--update-dependencies

grutt:
# Description

Updates outdated dependenceis

## Type of change


examples/python/quickstart/pyproject.toml
```diff
@@ -9,6 +9,7 @@ package-mode = false
 [tool.poetry.dependencies]
 python = "^3.10"
 hatchet-sdk = "1.33.3"
+python-dotenv = ">=1.2.2"
```
mrkaye97, line 12: do we need to add this?

sdks/python/examples/quickstart/pyproject.toml
```diff
@@ -9,6 +9,7 @@ package-mode = false
 [tool.poetry.dependencies]
 python = "^3.10"
 hatchet-sdk = "1.33.3"
+python-dotenv = ">=1.2.2"
```
mrkaye97, line 12: maybe this is the actual line 🤦 getting confused with the generated stuff

sdks/python/pyproject.toml
```diff
@@ -16,15 +16,15 @@ authors = [
 dependencies = [
     "grpcio>=1.76.0,<2",
     "grpcio-tools>=1.76.0,<2",
-    "protobuf>=6.30.2,<7",
+    "protobuf>=6.33.5,<7",
```
mrkaye97, line 19: I think if we do this we might need to run the `generate.sh` again - we've seen issues from this in the past

sdks/python/pyproject.toml
```diff
@@ -41,6 +41,7 @@ lint = [
     "mypy==1.20.0",
     "types-protobuf>=5.28.3.20241030",
     "black==26.3.1",
+    "click>=8.3.3",
```
mrkaye97, line 44: is this one a transitive dep that we need to pin?

mrkaye97 (approved):
left a couple small comments, the main thing is the protobuf one

PR #3844, opened 2026-05-07
Fix: Persist dashboard column visibility across reloads (#3822)

### LINKED ISSUE #3822: [FEAT] Persist dashboard column visibility across reloads
## Summary

Column visibility toggles in the dashboard reset on every page reload and on navigation away from the table. Hiding columns on Events, Workers, Workflows, Scheduled Runs, Recurring, Filters, Rate Limits, or Workflow Runs only sticks for the current mount of that page.

## Repro

1. Open `/events` (or any other table page).
2. Click the column-toggle menu, hide one or more columns.
3. Hard reload, or navigate away and back.
4. All columns are visible again.

Reproduced on a self-hosted instance running `ghcr.io/hatchet-dev/hatchet/hatchet-dashboard:latest`. Same code path applies to Hatchet Cloud since it ships the same React app.

## Root cause

Every table page wires `columnVisibility` to plain `useState`, never to a persistence layer. Confirmed by code search across `frontend/app/src/pages/main/v1/`:

| File | Line |
|------|------|
| `events/index.tsx` | 59 |
| `workers/index.tsx` | 45 |
| `scheduled-runs/index.tsx` | 63 |
| `workflows/index.tsx` | 19 |
| `filters/index.tsx` | 41 |
| `recurring/index.tsx` | 28 |
| `rate-limits/index.tsx` | 24 |
| `workflow-runs-v1/hooks/runs-provider.tsx` | 97 |

`DataTable` itself (`frontend/app/src/components/v1/molecules/data-table/data-table.tsx`) is correctly designed as a controlled component: it accepts `columnVisibility` and `setColumnVisibility` as props. The persistence responsibility sits with the parent, and no parent provides a persistent setter.

## Suggested fix

The repo already has `frontend/app/src/hooks/use-local-storage-state.tsx`, which:

- reads the initial value from `window.localStorage`,
- writes on every set,
- syncs across tabs via the native `storage` event plus a custom `hatchet:local-storage` event for same-tab listeners,
- falls back gracefully on JSON parse / quota errors.

It is already used for sidebar collapse (`use-sidebar.tsx`), the refetch interval (`refetch-interval-context.tsx`), and the side panel (`nav/side-panel.tsx`).

Swapping `useState<VisibilityState>(default)` for `useLocalStorageState<VisibilityState>(\"hatchet:columns:<page>\", default)` at each of the 8 sites above should be enough. Each page would need a stable key, e.g. `hatchet:columns:events`, `hatchet:columns:workers`, etc.

## Why it matters

Anyone who customizes their table view has to redo it every time they reload or change pages. The hook to fix it is in-tree, the change is mechanical, and no schema or backend work is involved. Happy to send a PR if the team is open to that approach.

sajdakabir:
# Description

Persists dashboard column-visibility selections to `localStorage` so they survive page reloads and navigation. Reuses the existing `useLocalStorageState` hook (extended to accept updater functions, matching `useState`'s API).

For `RunsProvider`, persistence is opt-in via a new `persistColumnVisibilityKey` prop — only `/runs` opts in, since other consumers use dynamic `tableKey`s like `task-runs-${id}` that would bloat `localStorage`.

Fixes #3822

## Type of change

## What's Changed

- [x] Extended `useLocalStorageState` to accept updater functions
- [x] Persisted column visibility on 7 table pages: events, workers, scheduled-runs, workflows, filters, recurring, rate-limits
- [x] Added opt-in `persistColumnVisibilityKey` prop to `RunsProvider`; enabled on `/runs`

## Test plan

Verified locally against a full dev stack:

- Hide column → reload → stays hidden (events, workers, runs)
- Navigate away and back → stays hidden
- `localStorage` shows `hatchet:columns:*` keys for stable pages only; dynamic tableKeys don't persist
- Cross-tab sync works
- Lint, prettier, typecheck, unit tests all clean

### DIFF
--- a/frontend/app/src/hooks/use-local-storage-state.tsx
+++ b/frontend/app/src/hooks/use-local-storage-state.tsx
@@ -2,10 +2,12 @@ import { useEffect, useState } from 'react';
 
 const LOCAL_STORAGE_EVENT = 'hatchet:local-storage';
 
+type SetStateValue<T> = T | ((prev: T) => T);
+
 export function useLocalStorageState<T>(
   key: string,
   defaultValue: T,
-): [T, (value: T) => void] {
+): [T, (value: SetStateValue<T>) => void] {
   const [state, setState] = useState<T>(() => {
     try {
       const item = window.localStorage.getItem(key);
@@ -15,15 +17,21 @@ export function useLocalStorageState<T>(
     }
   });
 
-  const setValue = (value: T) => {
+  const setValue = (value: SetStateValue<T>) => {
     try {
-      setState(value);
-      window.localStorage.setItem(key, JSON.stringify(value));
-      // Ensure other hook instances in the same tab update too (the native `storage`
-      // event does not fire within the same document).
-      window.dispatchEvent(
-        new CustomEvent(LOCAL_STORAGE_EVENT, { detail: { key, value } }),
-      );
+      setState((prev) => {
+        const next =
+          typeof value === 'function' ? (value as (prev: T) => T)(prev) : value;
+        window.localStorage.setItem(key, JSON.stringify(next));
+        // Ensure other hook instances in the same tab update too (the native `storage`
+        // event does not fire within the same document).
+        window.dispatchEvent(
+          new CustomEvent(LOCAL_STORAGE_EVENT, {
+            detail: { key, value: next },
+          }),
+        );
+        return next;
+      });
     } catch (error) {
       console.warn(`Error setting localStorage key "${key}":`, error);
     }
--- a/frontend/app/src/pages/main/v1/events/index.tsx
+++ b/frontend/app/src/pages/main/v1/events/index.tsx
@@ -20,6 +20,7 @@ import { SimpleTable } from '@/components/v1/molecules/simple-table/simple-table
 import { Button } from '@/components/v1/ui/button';
 import { CodeHighlighter } from '@/components/v1/ui/code-highlighter';
 import { Separator } from '@/components/v1/ui/separator';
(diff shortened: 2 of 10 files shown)


gregfurman:
Hey! Any chance you could provide some screenshots for this? Perhaps just a before and after case showing how your changes will change the UX.

sajdakabir:
Hey! Here's a quick demo showing the after state - hiding columns, hard-reloading, and the selections persisting:

https://github.com/user-attachments/assets/e5c5e489-3e91-47ec-8469-094984f01780

Before this change, hidden columns would reset to visible on every reload (the bug described in #3822). Let me know if you'd like a side-by-side before/after as well!

mrkaye97 (commented):
hi @sajdakabir - thanks for the contribution! how do you think about the tradeoffs between using local storage for this vs. using search params in the URL, like we do for the filters?

sajdakabir:
> 

umm..., column visibility feels more like a personal preference to me , if we do it like filters, the user's choice resets when they come back the next day. But if that's the behavior you want, happy to switch to URL params ,let me know!

mrkaye97:
> > 
> 
> umm..., column visibility feels more like a personal preference to me , if we do it like filters, the user's choice resets when they come back the next day. But if that's the behavior you want, happy to switch to URL params ,let me know!

Fair enough! I think that's reasonable - IMO, the tradeoff here is if you copy / paste a URL to share with someone, those settings won't be persisted when they open it, but I agree with you. Will take a look through now!

mrkaye97 (approved):
this looks great to me, thanks for the contribution!

PR #3847, opened 2026-05-07
Fix: load correct credentials for mtls in otel

vnghia:
# Description

When using mtls mode, otel trace does not work because the client credentials aren't loaded into it.

## Type of change

## What's Changed

- add a `load_channel_credentials` function
- use that function for otel and grpc connection

### DIFF
--- a/sdks/python/hatchet_sdk/connection.py
+++ b/sdks/python/hatchet_sdk/connection.py
@@ -50,7 +50,7 @@ def new_conn(config: ClientConfig, aio: Literal[True]) -> grpc.aio.Channel: ...
 
 
 def new_conn(config: ClientConfig, aio: bool) -> grpc.Channel | grpc.aio.Channel:
-    credentials: grpc.ChannelCredentials | None = load_channel_credentials(config)
+    credentials = load_channel_credentials(config)
 
     start = grpc if not aio else grpc.aio
 
--- a/sdks/python/hatchet_sdk/connection.py
+++ b/sdks/python/hatchet_sdk/connection.py
@@ -6,15 +6,7 @@ import grpc
 from hatchet_sdk.config import ClientConfig
 
 
-@overload
-def new_conn(config: ClientConfig, aio: Literal[False]) -> grpc.Channel: ...
-
-
-@overload
-def new_conn(config: ClientConfig, aio: Literal[True]) -> grpc.aio.Channel: ...
-
-
-def new_conn(config: ClientConfig, aio: bool) -> grpc.Channel | grpc.aio.Channel:
+def load_channel_credentials(config: ClientConfig) -> grpc.ChannelCredentials | None:
     credentials: grpc.ChannelCredentials | None = None
 
     # load channel credentials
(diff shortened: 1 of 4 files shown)


sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
```diff
@@ -347,12 +348,9 @@ def _add_hatchet_exporter(self) -> None:
         endpoint = self.config.host_port
         insecure = self.config.tls_config.strategy == "none"
         headers = (("authorization", f"Bearer {self.config.token}"),)
-        credentials: grpc.ChannelCredentials | None = None
-
-        if self.config.tls_config.root_ca_file:
-            with open(self.config.tls_config.root_ca_file, "rb") as f:
-                root_certs = f.read()
-            credentials = grpc.ssl_channel_credentials(root_certificates=root_certs)
+        credentials: grpc.ChannelCredentials | None = load_channel_credentials(
```
mrkaye97: nit: I don't think we need the type hint here

sdks/python/hatchet_sdk/connection.py
```diff
@@ -46,6 +38,20 @@ def new_conn(config: ClientConfig, aio: bool) -> grpc.Channel | grpc.aio.Channel
...
+@overload
+def new_conn(config: ClientConfig, aio: Literal[False]) -> grpc.Channel: ...
+
+
+@overload
+def new_conn(config: ClientConfig, aio: Literal[True]) -> grpc.aio.Channel: ...
+
+
+def new_conn(config: ClientConfig, aio: bool) -> grpc.Channel | grpc.aio.Channel:
+    credentials: grpc.ChannelCredentials | None = load_channel_credentials(config)
```
mrkaye97: nit: I don't think we need the type hint here

sdks/python/hatchet_sdk/connection.py
```diff
@@ -68,7 +74,7 @@ def new_conn(config: ClientConfig, aio: bool) -> grpc.Channel | grpc.aio.Channel
         # See discussion: https://github.com/hatchet-dev/hatchet/pull/2057#discussion_r2243233357
         os.environ["GRPC_POLL_STRATEGY"] = "poll"
 
-    if config.tls_config.strategy == "none":
+    if not credentials:
```
mrkaye97, line 77: nit: maybe we should have some sort of check where we assert on credentials not being present if the strategy is `none` or some such - maybe could do this at the validator level in the `TLSConfig`?

mrkaye97 (commented):
this looks great, thanks for the fix! would you mind bumping the patch version of the package and adding a changelog entry for the fix?

sdks/python/hatchet_sdk/connection.py
vnghia: Thank you for your review. I don't think we really need to do that because `load_channel_credentials` has already done a similar thing where `none` strategy won't load the credential at all.

mrkaye97 (approved):
Thanks for the fix!

PR #3848, opened 2026-05-07
Fix durable listener race

igor-kupczynski:
# Description

Fixes a Go SDK durable event listener reconnect race where `retrySend` held `clientMu.RLock()` across retry backoff sleeps. When the stream disconnected, `retryListen` could not acquire the write lock to replace the broken stream, so durable event replies could be delayed until retries were exhausted.

## Type of change

## What's Changed

This adds unit-tests for this case and changes `retrySend` to acquire the read lock only around each send attempt, allowing reconnects to proceed between retries. Adds unit coverage for reconnecting during retry backoff and delivering the durable event after reconnect.

The fix is cheery-picked from https://github.com/hatchet-dev/hatchet/pull/3845

Test commit only:
```
❯ go test ./pkg/client -run '^TestDurableEventsListener' -v
=== RUN   TestDurableEventsListenerReconnectsWhileRetrySendBacksOff
    durable_listener_test.go:156: expected durable event listener to reconnect while retrySend is backing off
--- FAIL: TestDurableEventsListenerReconnectsWhileRetrySendBacksOff (0.21s)
=== RUN   TestDurableEventsListenerDeliversEventAfterReconnectDuringRetryBackoff
    durable_listener_test.go:238: expected durable event to be delivered after reconnect during retrySend backoff
--- FAIL: TestDurableEventsListenerDeliversEventAfterReconnectDuringRetryBackoff (0.21s)
FAIL
FAIL    github.com/hatchet-dev/hatchet/pkg/client       0.713s
FAIL
```

With the cherry picked fix:
```
❯ go test ./pkg/client -run '^TestDurableEventsListener' -v
=== RUN   TestDurableEventsListenerReconnectsWhileRetrySendBacksOff
--- PASS: TestDurableEventsListenerReconnectsWhileRetrySendBacksOff (0.01s)
=== RUN   TestDurableEventsListenerDeliversEventAfterReconnectDuringRetryBackoff
--- PASS: TestDurableEventsListenerDeliversEventAfterReconnectDuringRetryBackoff (0.01s)
PASS
ok      github.com/hatchet-dev/hatchet/pkg/client       0.339s
```

**Note**: leaving as draft, as I want to think it through once more. Open for feedback for the time being.

### DIFF
--- a/sdks/go/e2e/durable_test.go
+++ b/sdks/go/e2e/durable_test.go
@@ -294,6 +294,7 @@ func TestDurableReplayReset(t *testing.T) {
 			require.NoError(t, err)
 
 			start := time.Now()
+			pollUntilRunStatus(t, ctx, sharedClient, ref.RunId, string(rest.V1TaskStatusRUNNING))
 			resetResult, err := ref.Result()
 			require.NoError(t, err)
 			resetElapsed := time.Since(start).Seconds()
@@ -340,7 +341,7 @@ func TestDurableBranchingOffBranch(t *testing.T) {
 	require.NoError(t, err)
 
 	start := time.Now()
-	time.Sleep(1 * time.Second)
+	pollUntilRunStatus(t, ctx, sharedClient, ref.RunId, string(rest.V1TaskStatusRUNNING))
 	resetResult, err := ref.Result()
 	require.NoError(t, err)
 	resetElapsed := time.Since(start).Seconds()
@@ -359,7 +360,7 @@ func TestDurableBranchingOffBranch(t *testing.T) {
 	require.NoError(t, err)
 
 	start = time.Now()
-	time.Sleep(1 * time.Second)
+	pollUntilRunStatus(t, ctx, sharedClient, ref.RunId, string(rest.V1TaskStatusRUNNING))
 	resetResult2, err := ref.Result()
 	require.NoError(t, err)
 	resetElapsed2 := time.Since(start).Seconds()
--- a/sdks/typescript/src/v1/examples/bulk_operations/bulk_replay.e2e.ts
+++ b/sdks/typescript/src/v1/examples/bulk_operations/bulk_replay.e2e.ts
@@ -6,7 +6,7 @@ import { bulkReplayTest1, bulkReplayTest2, bulkReplayTest3 } from './workflow';
 describe('bulk-replay-e2e', () => {
   const hatchet = makeE2EClient();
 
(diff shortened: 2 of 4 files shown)


mrkaye97 (approved):

sdks/go/e2e/durable_test.go
```diff
@@ -294,6 +294,7 @@ func TestDurableReplayReset(t *testing.T) {
 			require.NoError(t, err)
 
 			start := time.Now()
+			pollUntilRunStatus(t, ctx, sharedClient, ref.RunId, string(rest.V1TaskStatusRUNNING))
```
igor-kupczynski, line 297: My understanding is that `BranchDurableTask` only guarantees the replay was enqueued, not that the replayed execution has started. Waiting for `RUNNING` avoids `ref.Result()` immediately returning the previous completed result (this was likely flaky beforehand; not really related to this fix (although it might have changed the timing slightly).

sdks/typescript/src/v1/examples/bulk_operations/bulk_replay.e2e.ts
```diff
@@ -41,6 +42,9 @@ describe('bulk-replay-e2e', () => {
     );
 
     expect(initialRuns.rows).toHaveLength(expectedTotal);
+    const initialRetryCounts = new Map(
```
igor-kupczynski, line 45: This is a drive by, not see how it could be related to my change. I'm not up to speed with the TS SDK, so please take an extra care when reviewing. AI suggests that the old-engine job does not reliably expose the newer attempt field on list rows. The test now records each task’s initial `retryCount` and verifies bulk replay by waiting for that count to increase.

sdks/go/e2e/durable_test.go
mrkaye97: thanks for fixing this :)

sdks/typescript/src/v1/examples/bulk_operations/bulk_replay.e2e.ts
mrkaye97: This could definitely be the case, so if it works, I say we should keep it 👍 we have a handful of small API bugs where we return a zero value or something, this could be one of those

mrkaye97 (approved):
🚀

igor-kupczynski:
Thanks @mrkaye97 I'm merging this, as I think it fixes a real issue.

One thing we need can consider at some point is to compare the `durable_listener.go` with `listener.go` which handles this differently. We should probably align them.

PR #3849, opened 2026-05-07
NO MERGE: Align durable_listener reconnect with workflow runs listener

igor-kupczynski:
No merge: this is AI generated PR to see if we can align the patterns in two listeners. I do NOT intend to merge it as is, but rather to discuss the right way forward.

# Description

Aligns `DurableEventsListener` (`pkg/client/durable_listener.go`) with the reconnect pattern that `WorkflowRunsListener` (`pkg/client/listener.go`) already uses, so the two listeners share the same hardened lock + reconnect strategy.

This is a follow-up to #3848. That PR fixed the immediate read-lock-held-across-backoff race in `retrySend` with a minimal change. This PR generalizes it: lock-free `Send`/`Recv`, generation-counter snapshots, and `singleflight`-coalesced reconnects between `Listen` and `retrySend`.

Fixes # (n/a — refactor on top of #3848)

## Type of change

## What's Changed

- Switch `DurableEventsListener` to the snapshot + generation + `singleflight` reconnect pattern already used by `WorkflowRunsListener`.
- `Send`/`Recv` now run on a `getClientSnapshot()` copy, so the listener mutex is never held across gRPC I/O or backoff sleeps.
- Concurrent reconnects from `Listen` and `retrySend` are coalesced via `singleflight`; `retrySend` skips its own reconnect when it sees the generation bumped between attempts.
- `Listen` bails out after 10 consecutive reconnect failures and `Close` tolerates a nil client, matching `WorkflowRunsListener`.
- Tests cover the generation fast-path and singleflight coalescing in addition to the cases from #3848.

## Why

After #3848 there were two listeners with very similar shape but two different reconnect strategies. Aligning them:

- Removes the only remaining place where `clientMu` is held across `Send` (`doRetrySubscribe` still holds it briefly to swap the client and replay subscriptions, matching `listener.go`).
- Prevents redundant reconnects when both `Listen` and `retrySend` race to recover a broken stream.
- Makes future maintenance straightforward — both listeners now share the same mental model.

### DIFF
--- a/pkg/client/durable_listener.go
+++ b/pkg/client/durable_listener.go
@@ -13,6 +13,7 @@ import (
 	grpc_retry "github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/retry"
 	"github.com/rs/zerolog"
 	"golang.org/x/sync/errgroup"
+	"golang.org/x/sync/singleflight"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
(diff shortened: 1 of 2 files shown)


igor-kupczynski:
A better approach would be to have a same implementation for both listeners (abstract the common parts).

igor-kupczynski:
Closing in favor of https://github.com/hatchet-dev/hatchet/pull/4040

PR #3850, opened 2026-05-07
Test for listener recovery after Listen exits

igor-kupczynski:
# Description

After a Hatchet dispatcher outage long enough to trigger `Listen`'s `maxConsecutiveErrors` exit (or a graceful `io.EOF`), the gRPC connection reconnects and unary calls work again, but the cached `*WorkflowRunsListener` / `*DurableEventsListener` no longer has a goroutine reading its stream.

Adding new subscriptions on those listeners then either silently drops the server's events (workflow runs) or fails `AddSignal` after the full retry window (durable events). User-reported as "tasks run on the dashboard but `Result()` never returns".

## Type of change

## What's Changed

This PR adds two failing repros:
- `TestWorkflowRunsListenerEventsLostAfterListenExits`
- `TestDurableEventsListenerEventsLostAfterListenExits`

```
❯ go test ./pkg/client -run '^TestWorkflowRunsListenerEventsLostAfterListenExits$|^TestDurableEventsListenerEventsLostAfterListenExits$' -v

=== RUN   TestDurableEventsListenerEventsLostAfterListenExits
    durable_listener_test.go:305:
                Error Trace:    /Users/igor/src/hatchet-dev/hatchet/pkg/client/durable_listener_test.go:305
                Error:          Received unexpected error:
                                could not send to the worker after 5 retries
                Test:           TestDurableEventsListenerEventsLostAfterListenExits
                Messages:       AddSignal must not fail on a listener whose Listen has exited; the listener should recover or report a recoverable error to the caller
--- FAIL: TestDurableEventsListenerEventsLostAfterListenExits (20.00s)

=== RUN   TestWorkflowRunsListenerEventsLostAfterListenExits
    listener_test.go:616: workflow run event was not dispatched to the handler after Listen exited; the listener accepted the subscription but does not deliver events
--- FAIL: TestWorkflowRunsListenerEventsLostAfterListenExits (5.50s)

FAIL
FAIL    github.com/hatchet-dev/hatchet/pkg/client       25.798s
FAIL
```

Fix TBD

### DIFF
--- a/pkg/client/durable_listener.go
+++ b/pkg/client/durable_listener.go
@@ -13,6 +13,7 @@ import (
 	grpc_retry "github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/retry"
 	"github.com/rs/zerolog"
 	"golang.org/x/sync/errgroup"
+	"golang.org/x/sync/singleflight"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
@@ -26,8 +27,16 @@ type DurableEventHandler func(e DurableEvent) error
 type DurableEventsListener struct {
 	constructor func(context.Context) (contracts.V1Dispatcher_ListenForDurableEventClient, error)
 
-	client   contracts.V1Dispatcher_ListenForDurableEventClient
-	clientMu sync.RWMutex
+	client     contracts.V1Dispatcher_ListenForDurableEventClient
+	clientMu   sync.Mutex
+	generation uint64
+
+	reconnectGroup singleflight.Group
+
+	listenMu  sync.Mutex
+	listening bool
+	started   bool
+	closed    bool
 
 	l *zerolog.Logger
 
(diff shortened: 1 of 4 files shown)


igor-kupczynski:
<details>
  <summary>AI generated plan for the fix</summary>


# Listener Recovery Fix

## Diagnosis

PR 3850 adds repros for the same lifecycle bug in both listeners:

- [`pkg/client/listener.go`](pkg/client/listener.go): `WorkflowRunsListener.Listen` can return on `io.EOF` or after repeated reconnect failures. Later, `AddWorkflowRun` can rebuild the stream through `retrySubscribe` and successfully send a subscription, but no goroutine is reading `Recv` from that rebuilt stream.
- [`pkg/client/durable_listener.go`](pkg/client/durable_listener.go): `DurableEventsListener.Listen` has the same receive-loop exit problem, and `AddSignal` currently retries `Send` without rebuilding/restarting the receive loop after it has stopped.
- The factory goroutines in [`pkg/client/listener.go`](pkg/client/listener.go) also `Close` and clear cached listener pointers after `Listen` exits. That is risky once listeners can self-recover, because a stale goroutine can close a newly rebuilt stream.

## Implementation Plan

1. Add explicit receive-loop lifecycle state to both listener structs.
   - Track whether a receive loop is currently running with a mutex-protected boolean.
   - Keep the existing workflow `generation` model, and add the same client generation/snapshot pattern to the durable listener.
   - Add `singleflight.Group` to the durable reconnect path so concurrent recovery attempts coalesce like workflow reconnects already do.

2. Split each public `Listen(ctx)` into lifecycle management plus an internal receive loop.
   - `Listen(ctx)` marks the listener as running, executes the existing receive loop, and marks it stopped in a `defer` before returning.
   - The internal loop keeps the current behavior of returning `nil` for `io.EOF` and `codes.Canceled`, so existing `Listen` tests keep their contract.
   - The loop must keep using a client snapshot during blocking `Recv`, rather than holding a client lock across `Recv`; durable should be adjusted to match workflow here.

3. Add an `ensureListening(ctx)` helper to both listeners and call it from `AddWorkflowRun` / `AddSignal`.
   - After registering the handler, if no receive loop is running, call `retrySubscribe` / `retryListen` to create a fresh stream and resend all stored subscriptions, then start exactly one background receive loop.
   - If a receive loop is already running, keep the existing per-subscription send path.
   - This makes the PR repro pass: after the manual `Listen` call exits, the next add call rebuilds the stream, restarts the reader, and the event sent on the new stream is dispatched.

4. Make durable retry semantics match workflow retry semantics.
   - Rename or wrap `retryListen` as the durable reconnect operation, increment durable `generation` on successful replacement, and use generation checks in `retrySend`.
   - On durable `Send` failure, trigger reconnect/resubscribe rather than waiting out all retries on the stale stream.
   - Preserve the existing durable tests that verify reconnect during retry backoff and delivery on the new stream.

5. Adjust the subscribe-client factory goroutines for self-recovering listeners.
   - In `getWorkflowRunsListener` and `getDurableEventsListener`, avoid unconditional `Close` and cache clearing when `Listen` exits due to a transient stream EOF/error.
   - Only close and clear the cached pointer when the factory context is canceled or the listener is explicitly being shut down, and clear conditionally only if the cached pointer still points to that listener.
   - This prevents an old `Listen` goroutine from closing a stream rebuilt later by `AddWorkflowRun` or `AddSignal`.

## Validation

- Run the PR repros directly with `go test ./pkg/client -run '^TestWorkflowRunsListenerEventsLostAfterListenExits$|^TestDurableEventsListenerEventsLostAfterListenExits$' -count=1 -v`.
- Run neighboring listener tests with `go test ./pkg/client -run 'TestRetrySend|TestRetrySubscribe|TestListen_|TestDurableEventsListener' -count=1 -v`.
- Run `task lint-go` if the targeted tests pass. There is no package-scoped Taskfile test target, so the targeted `go test` commands are the practical first pass before broader validation.

</details>

igor-kupczynski:
Closing in favor of https://github.com/hatchet-dev/hatchet/pull/4040

PR #3856, opened 2026-05-08
chore: Bump to Go 1.26

gregfurman:
# Description

Updates Hatchet engine to Go 1.26. Also includes updates to relevant CI and Dockerfiles in `build/packages`.

## Type of change

## What's Changed

- Updates go.mod to 1.26.
- Updates associated CI tasks/workflows in `.github` to use 1.26 from 1.25.9.
- Bumps dockerfiles in `build/packages` to use go version 1.26 from 1.25.


.github/workflows/gen-examples.yml
```diff
@@ -17,9 +17,9 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Setup Go
-        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
```
mnafees: Seems like a regression

go.mod
```diff
@@ -42,7 +42,6 @@ require (
 	github.com/pingcap/errors v0.11.4
 	github.com/posthog/posthog-go v1.12.1
 	github.com/pressly/goose/v3 v3.27.0
-	github.com/sashabaranov/go-openai v1.41.2
```
mnafees, line 45: Are we sure of this removal?

mnafees (changes requested):

go.mod
gregfurman: Definitely unintentional. Have reverted

.github/workflows/test.yml
```diff
@@ -496,10 +496,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
-          go-version: "1.25.9"
-
-      - name: Go deps
-        run: go mod download
```
BloggerBust, line 502: Did we intend to drop the `go mod download` step here? I think this should still work without it because `go run` can download modules on demand. One thing to consider is that the explicit download was useful self-documentation and may also help slightly by cahcing the module early.

BloggerBust (dismissed):
Good upgrade. The previous review concerns also look resolved: the `setup-go` action version is no longer regressed, and `go-openai` is still present in `go.mod`.

There are a few stale Go 1.25 references that should be updated for consistency:

| File                                            | Current                       | Expected             |
| ----------------------------------------------- | ----------------------------- | -------------------- |
| `.github/workflows/go-benchmarks.yml`           | `go-version: "1.25.9"`        | `go-version: "1.26"` |
| `.github/workflows/sdk-python.yml`              | second `go-version: "1.25.9"` | `go-version: "1.26"` |
| `build/package/e2e-go.dockerfile`               | `golang:1.25-alpine`          | `golang:1.26-alpine` |
| `cmd/hatchet-cli/cli/templates/go/go.mod.embed` | `go 1.25.9`                   | `go 1.26`            |

Everything else looks correct. I am happy to re-rreview after the refs are updated.

mrkaye97:
Any risk here from switching versions everywhere? I assume they're good at avoiding anything breaking and regressions, but...

gregfurman:
@mrkaye97 I'd be pretty confident in this but if you want, we could upgrade in CI to 1.26 first?

mrkaye97:
> @mrkaye97 I'd be pretty confident in this but if you want, we could upgrade in CI to 1.26 first?

That sounds good to me

gregfurman:
@mrkaye97 On second thought... I think we should probably just upgrade. We can let this run in CI for a bit before doing a release a bit later in the week which IMO would give us confidence that all is working fine.

mrkaye97:
> @mrkaye97 On second thought... I think we should probably just upgrade. We can let this run in CI for a bit before doing a release a bit later in the week which IMO would give us confidence that all is working fine.

Whatever works!

gregfurman:
@mrkaye97 image

mrkaye97 (approved):

PR #3861, opened 2026-05-08
Fix: Allow partial successes on status updates

mrkaye97:
# Description

Allowing partial successes on status update messages, and re-publishing failed payloads after so they get requeued and can retry. The goal here is to load shed, in hopes that the vast majority of payloads will succeed and only a small proportion will need to be re-pubbed

## Type of change

### DIFF
--- a/internal/services/controllers/olap/controller.go
+++ b/internal/services/controllers/olap/controller.go
@@ -537,7 +537,7 @@ func (tc *OLAPControllerImpl) handleCreatedTask(ctx context.Context, tenantId uu
 		result, workflowRunIdsOfLocksNotAcquired, err := tc.repo.OLAP().CreateTasks(ctx, tenantId, createTaskOpts)
 
 		if err != nil {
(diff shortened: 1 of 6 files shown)


internal/services/controllers/olap/controller.go
```diff
@@ -529,7 +529,28 @@ func (tc *OLAPControllerImpl) handleCreatedTask(ctx context.Context, tenantId uu
...
+			succeededOpts = append(succeededOpts, opt)
+		}
+	}
+
+	tc.emitStandaloneTaskRootSpans(ctx, tenantId, succeededOpts)
+
+	for _, opt := range failedOpts {
+		msg, err := tasktypes.CreatedTaskMessage(tenantId, opt)
+		if err != nil {
+			return err
```
abelanger5: returning an `err` here means retrying the same batch of tasks, is that the correct behavior?

mrkaye97: I think this is correct, although maybe not ideal? if we don't error here, won't we lose some data potentially?

juliusgeo (approved):

PR #3865, opened 2026-05-10
feat: extend analytics interface to support incrementing counts

### LINKED ISSUE #3382: [FEAT] Extend Analytics Interface

AKASH10513:
This PR extends the Analytics interface to support incrementing counts without looping.

## Changes
- Added  interface with private method for type safety
- Implemented  and  types that satisfy 
- Updated  to accept  instead of 
- Modified  to parse  and  from props
- Maintained backward compatibility with existing  usage

## Usage
Now you can count multiple times without looping:

Resolves #3382


pkg/analytics/analytics.go
```diff
@@ -56,6 +56,16 @@ const (
 
 type Properties map[string]interface{}
 
+type Property interface {
+	private()
+}
```
gregfurman, line 59: nit: can we move the `Property` interface above these implementations? i.e

```golang
type Property interface {
	private()
}

type Properties map[string]interface{}

func (Properties) private() {}
// ...
```

pkg/analytics/analytics.go
gregfurman, line 1: Could we add a simple test for this? Wonder if in `aggregating_test.go` would be appropriate or whether we create a new test file `analytics_test.go`

gregfurman (changes requested):
Looks great! This would benefit a lot from a test though 🙂

pkg/analytics/analytics.go
AKASH10513: Yep, I can add a simple test for this. I think analytics_test.go would probably keep it more isolated/clear instead of mixing it into aggregating_test.go

pkg/analytics/analytics_test.go
gregfurman, line 1: Ahh I see this is actually quite contrived to test since `posthog` is really the only implementation here of the `Analytics` interface.

I think that testing this would actually require a bit more infrastructure -- perhaps a mock analytics implementation with an aggregator we can use to track calls to the interface.

Otherwise, perhaps just some manual steps that I can run through to ensure multi-counts are correctly propogated.

Both of these could be a bit out-of-scope for a good first issue though 🤔 Wdyt?

gregfurman (approved):

gregfurman:
Hey @AKASH10513. Have discussed this internally and unfortunately this change will break some upstream internal deps. Sorry for not keeping the original issue in sync with the upstream and for any time wasted here!

PR #3876, opened 2026-05-11
fix: scheduled timeout is not being propogated to task.result()

### LINKED ISSUE #3922: [BUG] `"Scheduling timed out"` does not show as an error
**Describe the issue**
When using `aio_run_many` and the children task can not run because of scheduling timed out, the status of the parent is still success. This is very misleading and could lead to bigger problem because the error is not addressed correspondingly.

**Environment**
- SDK: Python v1.33.5
- Engine: Selfhosted v0.86.7

**Expected behavior**
When the children task can not because of scheduling timed out, the status of the parent should show error

```
# set child_wf concurrency to 1
# spawn 100 child tasks
# sleep like 10s
# schedule timeout to 10s
child_wf.aio_run_many([])
```

**Additional context**
Thank you very much for Hatchet :D

jishnundth:
# Description

The reason for task failures caused by schedule timeouts is only visible in the dashboard as an OLAP event. It cannot be verified or checked from the SDK via task.result(), because no error message is set for such failures.

For task cancellation caused by concurrent limit and scheduled timeouts, an event message is propogated to the controller from the scheduler to be set as the error message. 

## Type of change


pkg/repository/task.go
```diff
@@ -1119,7 +1119,7 @@ func (r *TaskRepositoryImpl) ListFinalizedWorkflowRuns(ctx context.Context, tena
 	return res, nil
 }
 
-func (r *TaskRepositoryImpl) CancelTasks(ctx context.Context, tenantId uuid.UUID, tasks []TaskIdInsertedAtRetryCount) (*FinalizedTaskResponse, error) {
+func (r *TaskRepositoryImpl) CancelTasks(ctx context.Context, tenantId uuid.UUID, tasks []TaskIdInsertedAtRetryCount, reasons map[int64]string) (*FinalizedTaskResponse, error) {
```
gregfurman, line 1130: IMO we should probably rather be including an optional `EventMessage` or `Reason` attribute within the `TaskIdInsertedAtRetryCount` instead of changing the interface definition.

Otherwise, if we are going to be breaking the interface def, perhaps a `CancelTaskOpt` struct could be a better idea.

See how `FailTaskOpts` does it by embedding the aforementioned struct and including some more attributes for error handling:

https://github.com/hatchet-dev/hatchet/blob/779c7c5d0bc10d76fc085723d2fa1982253ced13/internal/services/controllers/task/controller.go#L506-L515

WDYT?

gregfurman (changes requested):
Apologies for the delay here. I'm wondering whether breaking the interface with an extra parameter is the best idea here. 

IMO if we're going to be changing the def, it probably makes more sense to do this at the struct level -- either with a new embedded/wrapped struct like `CancelTaskOpt` OR by introducing a backwards compatible change with an extra attribute on `TaskIdInsertedAtRetryCount`.

gregfurman:
Hey @jishnundth. Do you still have capacity to address my review? Otherwise, I'm happy to push this PR over the line. Lmk!

jishnundth:
> Hey @jishnundth. Do you still have capacity to address my review? Otherwise, I'm happy to push this PR over the line. Lmk!

Hey @gregfurman apologies for the late reply. Could you please take this over. I'm not available at the moment.

gregfurman:
@jishnundth No worries! We squash commits so you'll be the author in anycase 🙂

PR #3877, opened 2026-05-11
fix(templates/ts): pnpm and sdk version mismatch

gregfurman:
# Description

The `templates/typescript/pnpm` overrides the `qs` dependency to exclude an upper limit with `>=6.14.2`, conflicting with `sdks/typescript` bounding this via a caret `^6.14.2`. 

Additionally, we need to pass `--ignore-scripts` when installing since some transative deps have install scripts that aren't explicitly approved -- circumventing the `ERR_PNPM_IGNORED_BUILDS` error.

## Type of change

## What's Changed

- Removes conflicting `qs` override in `templates/typescript/pnpm` and regenerates the lockfile.
- Passes a `--ignore-scripts` flag when running the `pnpm` integration tests in the  `templates/typescript/shared/Dockerfile`

### DIFF
--- a/cmd/hatchet-cli/cli/templates/typescript/pnpm/package.json
+++ b/cmd/hatchet-cli/cli/templates/typescript/pnpm/package.json
@@ -19,8 +19,9 @@
     "@hatchet-dev/typescript-sdk": "1.21.2"
   },
   "pnpm": {
-    "overrides": {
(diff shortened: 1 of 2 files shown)


gregfurman:
Now that we've moved these templates to another repo, no longer relevant.

PR #3878, opened 2026-05-11
feat(migrate): redact DSNs and emit structured phase errors in hatchet-migrate

igor-kupczynski:
# Description

`hatchet-migrate` previously logged migration failures as `goose: failed to <thing>: <err>`, with no indication of which DSN it was talking to or which phase of the migration was running. That made the logs hard to diagnose.

This PR introduces a small `pkg/migratediag` package and rewires every error path in `cmd/hatchet-migrate/migrate/run.go` through it.

## Type of change

## What's Changed

Operator-visible behavior, a failed migration will show something like this:
```
2026/05/11 11:21:03 hatchet-migrate[DATABASE_URL oss]: failed during apply migrations on postgres://<redacted>@db.internal:5432/hatchet?keys=application_name,tls: context deadline exceeded
exit status 1
```

### DIFF
--- a/cmd/hatchet-migrate/migrate/run.go
+++ b/cmd/hatchet-migrate/migrate/run.go
@@ -19,6 +19,7 @@ import (
 	"github.com/sethvargo/go-retry"
 
 	_ "github.com/hatchet-dev/hatchet/cmd/hatchet-migrate/migrate/migrations" // register go migrations
+	"github.com/hatchet-dev/hatchet/pkg/migratediag"
 )
 
 //go:embed migrations/*.sql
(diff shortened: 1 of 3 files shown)


igor-kupczynski:
> ⚠️ GitGuardian has uncovered 1 secret following the scan of your pull request.

This is a test case to test the secrets are not leaked 😅  we should probably ignore this package from GitGuardian

igor-kupczynski:
CI failure looks unrelated to this PR.

The failing check is `test / integration`, specifically `pkg/auth/cookie`:

`TestSessionStoreLogoutClearsCookieWhenSessionRowMissing` is saving a session with a random `user_id`, which violates the `UserSession_userId_fkey` constraint because there is no matching `User` row.

This PR only touches migration diagnostics code, so I don’t think this should block this change. We’ll handle the broken test separately.

abelanger5 (approved):
💯 looks great

PR #3887, opened 2026-05-12
Refactor: use shared Dialog for plan upgrade modal

jishnundth:
# Description
A refactor to use the common Dialog utility on plan upgrade modal.

## Type of change

### DIFF
--- a/frontend/app/src/pages/main/v1/managed-workers/index.tsx
+++ b/frontend/app/src/pages/main/v1/managed-workers/index.tsx
@@ -2,6 +2,14 @@ import { BillingRequired } from './components/billing-required';
 import { ManagedWorkersTable } from './components/managed-workers-table';
 import { MonthlyUsageCard } from './components/monthly-usage-card';
 import { Button } from '@/components/v1/ui/button';
+import {
+  Dialog,
+  DialogContent,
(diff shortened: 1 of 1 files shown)


frontend/app/src/pages/main/v1/managed-workers/index.tsx
```diff
@@ -194,7 +164,34 @@ export default function ManagedWorkers() {
           />
         </div>
       </div>
-      <UpgradeModal />
+      <Dialog open={showUpgradeModal} onOpenChange={setShowUpgradeModal}>
```
gregfurman, line 167: Thanks for the refactor here! Any chance you can post a screenshot of a before and after for these changes? I'd like to see how the UI changes now that we're using these internal components.

gregfurman (changes requested):

frontend/app/src/pages/main/v1/managed-workers/index.tsx
jishnundth: image


It's the same only, the popup looks little more wider now. This is not visible in the OSS version, you've to override some variables to see managed compute on the sidebar, where this is located.

gregfurman (approved):

PR #3892, opened 2026-05-12
feat(ci): Better dependabot grouping

gregfurman:
# Description

Optimizes dependabot groupings to include all deps within the `cli/template` sections, which need to be bumped to the latest hatchet sdk version.

Also, we need to support `uv` ecosystems and tighten up unnecessary groupings.

## Type of change

## What's Changed

- Groups together all SDK example updates into a single PR using multi-ecosystem groupings.
- Ensures we also update `uv` lockfile for python version bumps.
- Removes refs to non-existent directories & ensures we only target `sdks/**/examples` instead of the autogenerated `examples/**`
- Correctly updates all `cli` template types for each ecosystem (including `bun`, `uv`, etc.)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,12 @@
 version: 2
+
+multi-ecosystem-groups:
+  hatchet-sdk-examples:
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    open-pull-requests-limit: 1
+
 updates:
   - package-ecosystem: "github-actions"
     directory: "/"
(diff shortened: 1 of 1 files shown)


BloggerBust:
Nice work on this, Greg. I reviewed this in the context of some quickstart and template work I have in progress.

From my review, once this PR merges, Go remains the only uncovered CLI template because it uses `go.mod.embed`, which Dependabot cannot find given the current logic in the [file fetcher](https://github.com/dependabot/dependabot-core/blob/main/go_modules/lib/dependabot/go_modules/file_fetcher.rb). I would be happy to address that in a follow-up PR.

This PR looks good to me @grutt.

mrkaye97 (approved):

BloggerBust:
@gregfurman #4087 is the follow-up PR for the remaining Go CLI template dependency update issue caused by `go.mod.embed`.

PR #3895, opened 2026-05-12
fix: parse multi-unit duration strings in convert_duration_to_interval

### LINKED ISSUE #3894: [BUG] Multi-unit duration strings are enforced with the wrong timeout
**Describe the issue**

The engine accepts `42m30s` during workflow registration, stores that value, then enforces it as roughly `42s`.

The mismatch is in `convert_duration_to_interval(duration text)`. It extracts the leading integer and chooses the unit from the final suffix, so `42m30s` becomes `42 seconds`.

**Environment**
- SDK: `@hatchet-dev/typescript-sdk@1.17.2`
- Latest SDK checked: `@hatchet-dev/typescript-sdk@1.22.1`, published 2026-05-05
- Engine: self-hosted `ghcr.io/hatchet-dev/hatchet/hatchet-lite:latest`
- Local engine image digest: `sha256:fd87ee914ad64845557b2f6130c749fbe99e132c568e1dac1ec2e49b0cbfaa96`

**Expected behavior**

If registration accepts `42m30s`, runtime enforcement should apply a timeout of 42 minutes and 30 seconds.

If multi-unit strings are not supported, registration should reject them instead of storing a value that runtime enforcement parses differently.

```ts
import { Hatchet } from "@hatchet-dev/typescript-sdk";

const hatchet = Hatchet.init();

const task = hatchet.task({
  name: "multi-unit-timeout-repro",
  executionTimeout: 2_550_000,
  fn: async () => {
    await new Promise((resolve) => setTimeout(resolve, 120_000));
    return { ok: true };
  },
});
```

`2_550_000` ms serializes to `42m30s`. The task is cancelled after roughly 43 seconds, even though the configured timeout is `42m30s`.

Event log from the failed task run:

```json
[
  {
    "eventType": "TIMED_OUT",
    "message": "Task exceeded timeout of 42m30s",
    "timestamp": "2026-04-30T09:31:59.157Z"
  },
  {
    "eventType": "STARTED",
    "timestamp": "2026-04-30T09:31:16.302Z"
  }
]
```

The event timestamps are 42.86 seconds apart, not `42m30s`.

The runtime SQL helper is where the stored string is converted to an interval for enforcement. A direct query shows the enforced interval:

```sql
SELECT
  convert_duration_to_interval('11m') AS eleven_m,
  convert_duration_to_interval('42m30s') AS forty_two_m_thirty_s,
  convert_duration_to_interval('1h30m5s') AS one_h_thirty_m_five_s;
```

```text
 eleven_m | forty_two_m_thirty_s | one_h_thirty_m_five_s
----------+----------------------+-----------------------
 00:11:00 | 00:00:42             | 00:00:01
```

Single-unit strings work as expected. For example, `660_000` ms serializes to `11m` and is enforced as 11 minutes.

**Re-verification (2026-05-12)**

Pulled `ghcr.io/hatchet-dev/hatchet/hatchet-lite:latest` today, digest `sha256:9511a4175d6d229ba5c14e8ed79f5ef4c458c86d2b2f7b3f8a6a0ed3f0350b7e`. The image was built on 2026-04-30T15:21:28Z, so `:latest` has not been rebuilt since the original observation.

Running the same query against a postgres migrated by the image's bundled `/hatchet-migrate` returns the same result:

```text
 eleven_m | forty_two_m_thirty_s | one_h_thirty_m_five_s
----------+----------------------+-----------------------
 00:11:00 | 00:00:42             | 00:00:01
(1 row)
```

**Additional context**

The TypeScript SDK accepts and emits multi-unit duration strings:

```ts
type TwoUnitDurations = `${number}h${number}m` | `${number}h${number}s` | `${number}m${number}s`;
type ThreeUnitDurations = `${number}h${number}m${number}s`;
```

Numeric durations are normalized to the same Go-style string format. For example, `2_550_000` ms becomes `42m30s`.

Registration validation accepts these strings because the `duration` validator uses Go's `time.ParseDuration`, which accepts values such as `42m30s` and `1h30m5s`.

The public timeout docs describe a single-unit format, such as `10s`, `4m`, and `1h`, but the TypeScript SDK type/tests and engine validation accept multi-unit values.

The current SQL helper does not enforce the same grammar:

```sql
num_value := substring(duration from '^\d+');

RETURN CASE
    WHEN duration LIKE '%ms' THEN make_interval(secs => num_value::float / 1000)
    WHEN duration LIKE '%s' THEN make_interval(secs => num_value)
    WHEN duration LIKE '%m' THEN make_interval(mins => num_value)
    WHEN duration LIKE '%h' THEN make_interval(hours => num_value)
    WHEN duration LIKE '%d' THEN make_interval(days => num_value)
    WHEN duration LIKE '%w' THEN make_interval(days => num_value * 7)
    WHEN duration LIKE '%y' THEN make_interval(months => num_value * 12)
    ELSE '5 minutes'::interval
END;
```

Because that helper is reused, affected surfaces include execution timeouts, schedule timeouts, `ctx.refreshTimeout`, durable sleeps, `onFailure` task timeouts, and `onSuccess` task timeouts.

An engine-side fix is needed because existing stored values and non-TypeScript clients can still send multi-unit strings. Runtime enforcement should match registration validation, or registration should reject values that runtime enforcement cannot parse correctly.

juanuicich:
## Summary

`convert_duration_to_interval(duration text)` only inspects the leading integer and the final unit suffix. Multi-unit Go duration strings such as `42m30s` therefore parse as roughly `42s`. Registration accepts the value through `time.ParseDuration` and the engine stores the original string, so timeouts silently shrink at enforcement time.

This PR closes the gap from both sides:

1. The SQL helper is replaced with one that walks the whole string and sums every component.
2. The `duration` validator is restricted to the grammar the SQL helper parses, so registration only accepts what enforcement can enforce.

Full parity with `time.ParseDuration` was deliberately not the goal. Chasing it would mean supporting nanoseconds, microseconds and signed values in plpgsql, none of which make sense for a timeout, and it would leave two hand-maintained parsers that can drift. Narrowing registration instead means the two sides agree by construction, and anything outside the grammar fails at registration with a clear error rather than being silently replaced at enforcement time.

Fixes #3894

## What changed

- New migration `20260611112431_v1_0_117.sql` does `CREATE OR REPLACE FUNCTION convert_duration_to_interval(duration text)`. The replacement parser accepts one or more `<number><unit>` components with units `ms`, `s`, `m`, `h`, including fractional values, and sums them into a single fixed interval. The function is now marked `IMMUTABLE PARALLEL SAFE`.
- Legacy single-unit suffixes `d`, `w`, `y` keep their prior calendar semantics for values stored before this migration. They are only accepted as the entire input (`'1d'`, `'1y'`); mixed forms such as `'30s1d'` fall through to the standard fallback.
- The `duration` validator (`pkg/validator/validator.go`) now matches the same grammar before calling `time.ParseDuration` (kept for overflow detection). It rejects signed values, bare numbers, `d`/`w`/`y`, and sub-millisecond units.
- The validation error message (`pkg/validator/default.go`) now states the real grammar; it previously claimed only single-unit `s`/`m`/`h` were valid.
- `frontend/docs/pages/v1/timeouts.mdx` and `frontend/docs/pages/v1/error-handling/timeouts.mdx` describe the multi-unit, decimal format and note that signed values and bare numbers are rejected.
- The `-- +goose Down` block restores the body of the prior migration (`20240321215205_v0_17_1.sql`) verbatim.

## Behavior changes

For the SQL helper (affects values already stored, and any clients that bypass validation):

| Input | Before | After |
|---|---|---|
| `42m30s` | 42 seconds | 42 minutes 30 seconds |
| `1h30m5s` | 1 second | 1 hour 30 minutes 5 seconds |
| `1.5h` | 1 hour (`num_value` was `INT`) | 1 hour 30 minutes |
| `-1h`, `abch` (unparseable number, known suffix) | `NULL` interval, meaning no timeout at all | 5 minute fallback |
| `1.5d` | 1 day | 5 minute fallback |
| oversized values (e.g. hundreds of digits) | raised an error inside the calling query | 5 minute fallback |
| `1d`, `2w`, `1y` | calendar interval | unchanged |
| invalid input | 5 minute fallback (when it did not return `NULL` or raise) | 5 minute fallback, now without the `NULL` and error cases |

For registration validation (affects all fields tagged `duration`: execution and schedule timeouts, `refreshTimeout`, durable sleeps, alert `maxFrequency`, tenant retention period, API token `expiresIn`):

| Input | Before | After |
|---|---|---|
| `42m30s`, `1.5h`, `1500ms`, `0s` | accepted | accepted |
| `-1.5h`, `+1h` | accepted, enforced as no timeout | rejected |
| `100ns`, `100us`, `100µs` | accepted, enforced as 5 minutes or 100 seconds | rejected |
| `0` (bare zero) | accepted, enforced as 5 minutes | rejected |

In every newly rejected case the previous behavior was silently enforcing something other than what the user asked for, so failing at registration is strictly better. SDK serialization formats are unaffected: the Go SDK emits `<int>s` (and `0s` for zero, `<int>ms` for sleep conditions), and the TypeScript SDK emits `h`/`m`/`s`/`ms` combinations; all fit the grammar.

The fix is engine-side because stored values and other clients can already send multi-unit strings. An SDK-only change would not cover them.

## Regression coverage

`pkg/repository/duration_test.go` runs the full migration chain against a containerized Postgres via the existing `setupPostgresWithMigration(t)` helper, then asserts the core invariant on a shared corpus: every string the Go validator accepts must convert to the intended number of seconds, and every string it rejects must hit the 5 minute fallback rather than parse to something else. Legacy `d`/`w`/`y` cases additionally assert that the validator rejects them while the SQL function keeps calendar semantics for stored values.

`pkg/validator/validator_test.go` covers the validator in isolation, including the newly rejected forms (signs, bare zero, `ns`/`us`, `d`, overflow).

## Test plan

- [x] `go test -count=1 ./pkg/repository -run TestConvertDurationToInterval -v`: 25 cases, all passing
- [x] `go test -count=1 ./pkg/validator`: passing
- [x] `task test` (full unit suite, post-rebase): all packages pass
- [x] `gofmt -s -l` on changed files: no diff
- [x] `go vet ./pkg/validator/ ./pkg/repository/`: clean
- [x] `task lint-go`: `golangci-lint run ./... --config .golangci.yml` (v2.8.0) reports 153 pre-existing issues, none in files this PR touches
- [x] `task test-integration`: full integration suite against docker compose Postgres + RabbitMQ, all packages pass


juanuicich:
The two failing CI checks (`old-engine-new-sdk`, `test-e2e`) both fail on the same unrelated test:

```
FAIL src/v1/examples/bulk_operations/bulk_replay.e2e.ts (121s)
> 90 |  throw new Error(`Timed out waiting for ${label} after ${timeoutMs}ms`);
```

This is a TypeScript SDK bulk-replay e2e timing out at 121s. This PR only touches Go code (migration, `convert_duration_to_interval`, validator) and a couple of docs pages, so there's no path from these changes to that test.

The Vercel check is an authorization prompt, not a build failure.

Could someone with re-run permissions retry the failed jobs? If `bulk_replay.e2e.ts` is flaky on `main` too, happy to open a separate issue.

juanuicich:
@abelanger5 Rebased onto current main and revised, addressing two problems with the previous revision:

1. **Migration renumbered.** Main has moved to `v1_0_116` and took the `v1_0_106` label in the meantime, and the old timestamp had become out of order, which `goose.Up` rejects on already-migrated databases. The migration is now `20260611112431_v1_0_117.sql`. The regression test runs the full migration chain, so the ordering is exercised.

2. **The fix now closes the gap from both sides.** The previous revision widened the SQL parser towards `time.ParseDuration` but still left mismatches (`ns`/`us`, bare `0`, signed values), and it added negative-duration support that nothing needs. This revision drops sign support from the SQL helper and instead restricts the `duration` validator to the grammar the helper parses, so registration and enforcement agree by construction. Inputs that previously slipped through and were enforced as something else now fail at registration with a clear error. The validator error message and the timeout docs were brought in line.

Also new in this revision: the SQL helper no longer raises on oversized values (it falls back like other invalid input), it is marked `IMMUTABLE PARALLEL SAFE`, and the two test suites now share a corpus that asserts validator and SQL agreement directly. The behavior-change table in the description is rewritten to cover all of this, including the old `NULL`-interval path I had previously missed (inputs like `-1h` used to disable the timeout entirely).

juanuicich:
@mrkaye97 following up from #4017. This is rebased onto current main and the migration is renumbered to `v1_0_117`, so it applies cleanly on databases already at `v1_0_116`. One change since then: rather than making the SQL function accept everything `time.ParseDuration` does, the `duration` validator now only accepts what the function can parse, so the two can't drift apart. The comment above has the details.

Unit and integration suites pass locally. Could you take a look, or let me know what else it needs?

cmd/hatchet-migrate/migrate/migrations/20260616120000_v1_0_119.sql
```diff
@@ -0,0 +1,92 @@
...
+    -- Legacy single-unit suffixes (d, w, y) keep calendar semantics for
+    -- values stored before this migration. Only accepted as the entire input.
+    m := regexp_match(duration, '^([0-9]{1,8})(d|w|y)$');
+    IF m IS NOT NULL THEN
+        val := m[1]::double precision;
+        unit := m[2];
+        CASE unit
+            WHEN 'd' THEN RETURN make_interval(days => val::int);
+            WHEN 'w' THEN RETURN make_interval(days => (val * 7)::int);
+            WHEN 'y' THEN RETURN make_interval(months => (val * 12)::int);
```
mrkaye97, line 25: just confirming we don't need to support other possibilities, e.g. `m`

cmd/hatchet-migrate/migrate/migrations/20260616120000_v1_0_119.sql
```diff
@@ -0,0 +1,92 @@
...
+
+        IF m IS NULL THEN
+            RETURN '5 minutes'::interval;
+        END IF;
+
+        -- Digit cap keeps the cast and the sum well inside double precision,
+        -- so oversized values fall back instead of raising.
+        IF length(m[1]) > 15 THEN
+            RETURN '5 minutes'::interval;
+        END IF;
```
mrkaye97, line 47: we might want to raise here or something instead, this could cause unexpected behavior (although obviously shouldn't really happen in practice)

cmd/hatchet-migrate/migrate/migrations/20260612112431_v1_0_118.sql
```diff
@@ -0,0 +1,92 @@
...
+
+        total_seconds := total_seconds + val * factor;
+
+        rest := substring(rest from length(m[1]) + length(m[2]) + 1);
+    END LOOP;
+
+    -- Anything beyond Go's maximum duration (~292 years) cannot have passed
+    -- registration validation; fall back rather than risk interval overflow.
+    IF total_seconds > 9223372036 THEN
+        RETURN '5 minutes'::interval;
```
mrkaye97: same here, maybe raise?

mrkaye97 (commented):
left a couple small comments, but this looks like a really nice fix to me! would you mind also copy-pasting the function definition into `v1-core.sql` and `v1-olap.sql` (it should've been there in the first place but wasn't)

juanuicich:
Thank you @mrkaye97 for the reviews. Caught me travelling this weekend but I'll pick them up on Monday.

mrkaye97:
> Thank you @mrkaye97 for the reviews. Caught me travelling this weekend but I'll pick them up on Monday.

Sounds good! No rush

gregfurman:
@juanuicich lmk if you still have capacity to address @mrkaye97's review 🙂

juanuicich:
@gregfurman thanks for the nudge, sorry about the delay. Pushed an update.

@mrkaye97: rebased onto current main and renumbered the migration to `v1_0_119` (both `v1_0_117` and `v1_0_118` got claimed while this was open).

Took both `RAISE` suggestions. The digit-cap and 292-year branches now `RAISE EXCEPTION` with the offending input in the message, matching the partition helpers and the rest of the project's plpgsql. Test corpus picks up a `wantRaise` flag for the two cases that now raise.

Copied the function into `sql/schema/v1-core.sql` and `sql/schema/v1-olap.sql` with a comment on each noting they mirror the migration.

On the legacy regex (`^([0-9]{1,8})(d|w|y)$`): kept as-is. The branch exists only for units with calendar semantics; `d`/`w`/`y` go through `make_interval(days => ...)` and `make_interval(months => ...)` and respect month and DST boundaries. Anything fixed-length, `m` included, goes through the multi-unit loop and ends up as the right number of seconds, so no separate path needed.

mrkaye97 (approved):
lgtm, thanks for the fix @juanuicich, and sorry for the delay here. going to get this merged and deployed!

juanuicich:
No worries! Thank you @mrkaye97 and sorry for the delays on my side. I am happy to be able to contribute and also fix the issue that I was hitting in production.

PR #3906, opened 2026-05-13
Fix: Deadlock on trigger workflows

mrkaye97:
# Description

Fixes a deadlock on trigger workflows where we still used `r.pool` instead of the open tx

## Type of change

### DIFF
--- a/pkg/repository/trigger.go
+++ b/pkg/repository/trigger.go
@@ -619,13 +619,13 @@ func (r *sharedRepository) triggerWorkflows(
 		workflowVersionIds = append(workflowVersionIds, id)
 	}
 
-	var listStepsTx sqlcv1.DBTX = r.pool
+	var preflightTx sqlcv1.DBTX = r.pool
 
 	if existingTx != nil {
-		listStepsTx = existingTx.tx
+		preflightTx = existingTx.tx
 	}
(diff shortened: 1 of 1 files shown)


pkg/repository/trigger.go
```diff
@@ -669,7 +669,7 @@ func (r *sharedRepository) triggerWorkflows(
 	stepsToAdditionalMatches := make(map[uuid.UUID][]*sqlcv1.V1StepMatchCondition)
 
 	if len(stepsWithAdditionalMatchConditions) > 0 {
-		additionalMatches, err := r.queries.ListStepMatchConditions(ctx, r.pool, sqlcv1.ListStepMatchConditionsParams{
+		additionalMatches, err := r.queries.ListStepMatchConditions(ctx, listStepsTx, sqlcv1.ListStepMatchConditionsParams{
```
abelanger5: nit: let's rename `listStepsTx` to something more like `preflightTx` since we're using it across multiple queries now?

mrkaye97: yeah, good idea

mrkaye97: fixed!

abelanger5 (approved):

PR #3914, opened 2026-05-14
Refactor: Use `externalId` for retrieving payloads

mrkaye97:
# Description

Adds the `ExternalId` to the `RetirevePayloadOpts` everywhere now that it's always non-null and we have the index. Once we do this, we can start passing the external id through to the external store implementation as the key, and then use it to do the magic index file retrieve

Also adds yet another expensive migration to make the external ids in the v1_task_event and v1_task_events_olap not null, since right now we're implicitly (unsafely) assuming those are set

## Type of change

### DIFF
--- a/cmd/hatchet-migrate/migrate/migrations/20260511174313_v1_0_108.sql
+++ b/cmd/hatchet-migrate/migrate/migrations/20260511174313_v1_0_108.sql
@@ -1,7 +1,6 @@
 -- +goose NO TRANSACTION
 -- +goose Up
 -- +goose StatementBegin
-ALTER TABLE v1_payload ALTER COLUMN external_id SET DEFAULT gen_random_uuid();
 ALTER TABLE v1_payload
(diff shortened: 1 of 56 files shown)


cmd/hatchet-migrate/migrate/migrations/20260511174313_v1_0_108.sql
```diff
@@ -27,6 +26,7 @@ END $$;
 -- +goose StatementBegin
 ALTER TABLE v1_payload VALIDATE CONSTRAINT v1_payload_external_id_not_null;
 ALTER TABLE v1_payload ALTER COLUMN external_id SET NOT NULL;
+ALTER TABLE v1_payload ALTER COLUMN external_id SET DEFAULT gen_random_uuid();
```
mrkaye97: drive by - I think this has to happen after backfilling and validating the constraint, otherwise we might lock the table for a long time (important for oss)

abelanger5 (approved):

Issue #3913, opened 2026-05-14
UserSession table in Hatchet Postgres grows unboundedly from Bearer API requests


luislee818:
## Problem

In our self-hosted Hatchet deployment, the `UserSession` table in the Hatchet Postgres instance has grown to 2.6M rows (~1 GB) over ~30 days.

```
SELECT count(*) FILTER (WHERE "userId" IS NULL), count(*) FILTER (WHERE "userId" IS NOT NULL)
FROM "UserSession";
-- 2,604,487 anonymous   |   9 with userId (4 real users)
```

- ~1 row/sec sustained insertion rate, matching our REST API call volume.
- Every row has `userId = NULL`, `expiresAt = createdAt + 30d`, `createdAt == updatedAt` (written once, never read again).
- The only `DELETE` in `pkg/repository/sqlcv1/users.sql` is `DeleteUserSession` by id (explicit logout). Nothing purges by `expiresAt`. Tenant retention (`SERVER_LIMITS_DEFAULT_TENANT_RETENTION_PERIOD`) doesn't cover this table.

## Likely cause

Bearer-authenticated REST requests (SDK calls, worker callbacks) appear to trip the cookie session middleware, which calls `Save()` → no cookie → fresh `UserSession` row inserted. Our setup: ALB OIDC for dashboard humans, Bearer tokens for SDK/workers — so essentially every row is from Bearer traffic.

## Suggested fix

Skip cookie session creation for Bearer-authenticated requests, and/or add a periodic `DELETE FROM "UserSession" WHERE "expiresAt" < NOW()` to the retention cron.

Version: `hatchet-lite:v0.82.3`.

gregfurman:
Thanks for the report here! I believe I've recreated this issue in my local environment 👇 

<details>

<summary><b>Expand for recreation steps</b></summary>

1. Build the frontend, start the DB, and then launch hatchet-lite locally:

```sh
(cd frontend/app && npm run build)
task start-db migrate seed
LITE_STATIC_ASSET_DIR=./frontend/app/dist LITE_FRONTEND_PORT=5173 task start-lite
```

2. Connect to the postgres instance and execute the count query

```sh
PGPASSWORD=hatchet psql -p 5431 -h localhost -U hatchet hatchet -c 'SELECT count(*) FILTER (WHERE "userId" IS NULL), count(*) FILTER (WHERE "userId" IS NOT NULL)
```

```
 count | count 
-------+-------
     0 |     0
```

[...]

PR #3923, opened 2026-05-15
fix(server/authn): Re-order auth flow to prevent side effects on handleCookieAuth

gregfurman:
# Description

Currently, any HTTP request to a route that accepts cookie-based auth creates a new entry in the `UserSession` table, even when the request is ultimately authenticated via a different strategy (e.g. bearer auth using `HATCHET_CLIENT_API_TOKEN`).

This happens because `handleCookieAuth` saves an unauthenticated session as a side effect before falling through to the next auth strategy in the chain. For programmatic clients making many bearer-auth requests, this produces an unauthenticated `UserSession` row per request that is never referenced again and never cleaned up.

Hence, we should only create new `UserSession` entries when the request is actually attempting cookie-based auth.

Partial fix for #3913. See original issue for reproduction steps. 

## Type of change

## What's Changed

- When multiple authentication types are possible, ensure `cookieAuth` occurs last.

### DIFF
--- a/api/v1/server/authn/middleware.go
+++ b/api/v1/server/authn/middleware.go
@@ -60,22 +60,6 @@ func (a *AuthN) authenticate(c echo.Context, r *middleware.RouteInfo) error {
 		return a.handleNoAuth(c)
 	}
 
-	var cookieErr error
-
-	if r.Security.CookieAuth() {
-		cookieErr = a.handleCookieAuth(c)
-
-		c.Set("auth_strategy", "cookie")
-
-		if cookieErr == nil {
-			return nil
-		}
-	}
-
-	if cookieErr != nil && !r.Security.BearerAuth() && !r.Security.CustomAuth() {
-		return cookieErr
-	}
-
 	var bearerErr error
 
 	if r.Security.BearerAuth() {
@@ -88,7 +72,7 @@ func (a *AuthN) authenticate(c echo.Context, r *middleware.RouteInfo) error {
 		}
 	}
 
-	if bearerErr != nil && !r.Security.CustomAuth() {
+	if bearerErr != nil && !r.Security.CustomAuth() && !r.Security.CookieAuth() {
 		return bearerErr
 	}
 
@@ -104,8 +88,26 @@ func (a *AuthN) authenticate(c echo.Context, r *middleware.RouteInfo) error {
 		}
 	}
 
-	if customErr != nil {
+	if customErr != nil && !r.Security.CookieAuth() {
 		return customErr
+	}
+
+	var cookieErr error
+
+	if r.Security.CookieAuth() {
+		// FIXME(gregfurman): handleCookieAuth (practically) always creates a new UserSession entry.
+		// Hence, the ordering of bearer -> custom -> cookie intentionally prevents this.
+		cookieErr = a.handleCookieAuth(c)
+
+		c.Set("auth_strategy", "cookie")
+
+		if cookieErr == nil {
+			return nil
+		}
+	}
+
+	if cookieErr != nil {
+		return cookieErr
 	}
 
 	return fmt.Errorf("no auth strategy found")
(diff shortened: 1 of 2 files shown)


gregfurman:
❓ Should we be including a migration to clean-up orphaned session entries (i.e those with  `userId = NULL`) in the `UserSession` table?

api/v1/server/authn/middleware.go
```diff
@@ -141,6 +141,10 @@ func (a *AuthN) handleCookieAuth(c echo.Context) error {
 
 	store := a.config.SessionStore
 
+	if _, err := c.Cookie(store.GetName()); err != nil {
+		return forbidden
+	}
```
abelanger5: Are we sure this is correct? There are some endpoints where we would like to call `Set-Cookie`, for example when beginning an oauth flow, we need to store the oauth `state` parameter. 

It's quite possible we don't hit this case on any such endpoints, in which case I wonder why we call `SaveUnauthenticated` later on (which we do a few lines later). Though it's possible that the error block is there so that we can invalidate a cookie which can no longer be decrypted.

abelanger5:
> ❓ Should we be including a migration to clean-up orphaned session entries (i.e those with `userId = NULL`) in the `UserSession` table?

I think a cron job to clean up those entries is not unreasonable, though note that not every session which has `userId = NULL` should necessarily be cleaned up (again for the purposes of the oauth flow). So it seems like we should clean up when:
1. The session has expired (assuming we have no mechanism to refresh a session, which I don't think we do), or
2. The session has `userId = NULL` and was created before a certain time, like older than 24 hours ago

api/v1/server/authn/middleware.go
gregfurman: > Are we sure this is correct?

Not entirely 🫠 Since the bug here is that we shouldn't be creating a new `UserSession` entry when there is no session to be persisted -- which will always occur regardless of the auth type since we assume cookie auth whose fallback has the aforementioned side-effect of a new row being added.

I could rejig the logic here to instead start off assuming bearer auth (which has no side-effects) i.e from `cookie -> bearer -> custom` to `bearer -> custom -> cookie`

> It's quite possible we don't hit this case on any such endpoints, in which case I wonder why we call SaveUnauthenticated later on (which we do a few lines later).

At a glance, I couldn't really figure out how to induce a fallback to `SaveUnauthenticated` which makes me think this handler logic deserves a revisit.


For this PR, let me not make any changes to the individual auth handler logic and instead try the aforementioned handler re-ordering. Wdyt?

gregfurman:
> I think a cron job to clean up those entries is not unreasonable, though note that not every session which has `userId = NULL` should necessarily be cleaned up (again for the purposes of the oauth flow).

Ah yes. I omitted the `expiredAt` constraint as well. Sessions that have expired are already cleaned up on logout due to our `Save` logic:

https://github.com/hatchet-dev/hatchet/blob/26a44c2e19c55914d6019013f07758225ebfd842/pkg/auth/cookie/sessionstore.go#L218-L226

Granted, there are probably some edge-cases here like a user never logging back in after their session expired which would probably just leave the expired session in the table. Since there isn't really a way of determining whether a session entry was incorrectly included on a token-authenticated request, perhaps a cron makes sense.

Do we have a preference on whether to manage cleanups in the Go code or would a `pg_cron` be better suited? Suppose this could be included in the `DataRetention` cleanup as well?


Also, I re-worked the logic here based on https://github.com/hatchet-dev/hatchet/pull/3923#discussion_r3252780996 -- lmk if we're happy with the approach!

abelanger5 (approved):

abelanger5:
> Do we have a preference on whether to manage cleanups in the Go code or would a pg_cron be better suited? Suppose this could be included in the DataRetention cleanup as well?

Definitely would prefer to manage this in Go code, we don't want to depend on postgres extensions if we can avoid them. I think a simple `go-cron` on the API layer would be fine. I don't think we have this on any of the API surface yet, only on the engine, but those should be a good reference.

PR #3930, opened 2026-05-16
feat: enhance partition management with advisory locks and connection handling

grutt:
# Description

When running multiple instances against a single OLAP DB, partition maintenance can race across controllers and issue overlapping DDL. This change runs OLAP partition updates on one dedicated DDL connection guarded by a session advisory lock, so only one instance performs the work at a time while still allowing DETACH PARTITION CONCURRENTLY outside an explicit transaction. If releasing the lock fails, the connection is closed instead of returned to the pool to avoid leaking a lock-holding session.

Fixes # (issue)

## Type of change

## What's Changed

- Implemented advisory locks to prevent concurrent OLAP partition operations across instances.
- Refactored connection handling to ensure DDL operations are executed with appropriate statement timeouts.
- Updated methods to use the new connection management for OLAP partition creation and detachment.


pkg/repository/sqlchelpers/connection.go
```diff
@@ -0,0 +1,90 @@
...
+	"time"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/rs/zerolog"
+)
+
+const defaultStatementTimeoutMs = 30000
+
+func AcquirePoolConnectionWithStatementTimeout(ctx context.Context, pool *pgxpool.Pool, l *zerolog.Logger, timeoutMs int) (*pgxpool.Conn, error) {
```
igor-kupczynski, line 15: Nit, but `Exported types and functions must have godoc comments.` https://github.com/hatchet-dev/hatchet/blob/main/.cursor/rules/code-comments.mdc#L21 (here and next)

pkg/repository/task_partition_test.go
```diff
@@ -307,6 +307,41 @@ func TestUpdateTablePartitions_RealPartitionCreation(t *testing.T) {
 	t.Logf("Partitions before: %d, after: %d (increase: %d)", countBefore, countAfter, countAfter-countBefore)
 }
 
+func TestUpdateTablePartitions_SingleConnectionDDLPoolCreatesPartition(t *testing.T) {
```
igor-kupczynski, line 310: Should we drop `TestUpdateTablePartitions_AdvisoryLockBehavior` give that it tests the old code path (which is no longer a production path?)

pkg/repository/task_partition_test.go
```diff
@@ -307,6 +307,41 @@ func TestUpdateTablePartitions_RealPartitionCreation(t *testing.T) {
...
+	pool, cleanup := setupPostgresWithMigration(t)
```
igor-kupczynski, line 311: Should the olap.go get a similar test?

pkg/repository/sqlchelpers/connection.go
```diff
@@ -0,0 +1,90 @@
+package sqlchelpers
```
igor-kupczynski, line 1: I suggest we add tests to these new helpers

igor-kupczynski (commented):
I understand that the core idea behind this PR is to replace the transactioned scoped lock with a session lock (and moreover, to make sure the DDL commands run on the same connection, previously they didn't). This makes sense to me and I'm +1 for the change.

Raise a few nits around testing and docs, but given it's such a critical change, it's worth addressing IMHO.

mrkaye97:
Closing this, superseded now

PR #3932, opened 2026-05-18
fix(ts-sdk): honor options.priority and options.scope in EventClient.bulkPush

### LINKED ISSUE #3931: [BUG][Typescript SDK] EventClient.bulkPush silently drops options.priority and options.scope
**Describe the issue**
`EventClient.bulkPush(type, inputs, options)` silently drops `options.priority` and `options.scope`. The implementation reads `priority` and `scope` only from each per-input `EventWithMetadata` entry and never falls back to the `options` argument. `options.additionalMetadata` is honored as a per-input fallback, but the other two fields in the same `PushEventOptions` shape are not, which is inconsistent with `EventClient.push`.

**Environment**
- SDK: TypeScript v1.22.1

**Expected behavior**
When `options.priority` or `options.scope` is provided to `bulkPush`, the value should be applied to every event in the batch that does not set its own override, mirroring how `push` applies the same options to a single event. At minimum the inconsistency should be documented on `PushEventOptions` and `EventClient.bulkPush`.

```ts
await client.events.bulkPush(
  "user:created",
  [
    { payload: { id: "1" } },
    { payload: { id: "2" } },
  ],
  { priority: 5, scope: "tenant-acme" },
);
```

The resulting gRPC request contains `priority: undefined` and `scope: undefined` for both events. Relevant SDK source:

```js
bulkPush(type, inputs, options = {}) {
  const events = inputs.map((input) => {
    const baseMeta = input.additionalMetadata ?? options.additionalMetadata ?? {};
    // ...
    return {
      // ...
      additionalMetadata: ...,
      priority: input.priority, // options.priority ignored
      scope: input.scope,       // options.scope ignored
    };
  });
}
```

Compare with `push`, which reads all three fields directly from `options`.

**Additional context**
Workaround: copy `options.priority`, `options.scope`, and `options.additionalMetadata` onto every `EventWithMetadata` entry on the caller side before invoking `bulkPush`.

sushaan-k:
Hi! Picking up #3931.

`EventClient.bulkPush(type, inputs, options)` silently dropped `options.priority` and `options.scope`. The per-event mapping only read those fields from each input, and never fell back to the shared `options` argument. `additionalMetadata` already falls back to `options.additionalMetadata`, and `EventClient.push` reads all three fields directly from `options`, so the bulk path was the odd one out.

Fix mirrors the `additionalMetadata` pattern:

```ts
priority: input.priority ?? options.priority,
scope: input.scope ?? options.scope,
```

Per-event overrides still win, and the shared options apply to every event in the batch that doesn't set its own.

### Testing

Added a unit test that covers both paths: an event with no per-event override picks up the shared option, an event with a per-event override keeps its own.

```sh
cd sdks/typescript
pnpm exec jest src/clients/event/event-client.test.ts   # 6/6 pass
pnpm lint:check                                          # clean
pnpm exec tsc                                            # clean
```

Closes #3931


BloggerBust:
This fix looks good! It makes `bulkPush` consistent with `push` for `priority` and `scope`, while still allowing event values to override the shared options. The added test covers the important shared-default and event-override cases, and I also ran the tests and typechecker locally. Thanks!

BloggerBust:
@sushaan-k one release housekeeping request before merge: could you bump the TypeScript SDK package version in `sdks/typescript/package.json` to `1.23.0` and add a changelog entry for the `bulkPush` priority/scope fix in `sdks/typescript/CHANGELOG.md`?

PR #3933 can add its own separate changelog entry under the same version.

gregfurman:
Thanks @BloggerBust for the review and suggestions! Since the version of the ts SDK keeps changing, I'll update this to include a CHANGELOG entry under the `[Unreleased]` header.

gregfurman (approved):

PR #3960, opened 2026-05-19
Refactor: Pass `insertedAt` to OLAP payload reads

mrkaye97:
# Description

Passing an `InsertedAt` timestamp to all the OLAP payload reads, so that we can use it to create a partition key on read after offloading payloads

## Type of change


juliusgeo (approved):
LGTM!

PR #3979, opened 2026-05-21
feat(engine): index file-based payload offloads

mrkaye97:
# Description

This is the OSS work for index file-based payload offloads, with the basic idea being that instead of writing a key for each offload inline in the payloads table, we instead keep a much smaller table containing a pointer to an index file, which we can then read and use to look up the relevant payload.

I also removed the immediate offloads logic from the code here, since it's unused. And also cleaned up some other dead code.

The main changes to the OSS are just adding the index block tables and queries to read / write, and modifying the logic of the jobs significantly

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

<details id="ai-disclosure">

No AI was used here

</details>


cmd/hatchet-migrate/migrate/migrations/20260522014401_v1_0_113.sql
```diff
@@ -0,0 +1,31 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE EXTENSION IF NOT EXISTS btree_gist;
```
mrkaye97: IMPORTANT: I was assuming this was fine to add, but if not, we can do it another way. [Here](https://www.postgresql.org/docs/current/btree-gist.html) are the docs on this extension. It's a native extension and is supported by RDS, Azure PG, CloudSQL PG, Ubicloud, DO, Supabase, Neon, and PlanetScale although of course we can't guarantee that _every_ provider will have it...

grutt: can we fallback to a different index type if this doesnt exist? concerned about the oss

mrkaye97: this is all or nothing, so we can either go for it or do it a different way. I think this is probably the cleanest way, but it requires the extension

grutt: i suppose we'll fail on the migration on upgrade if its not available which is relatively tolerable, but i know we've tried hard to not have any dependencies.

seems like azure doesnt support this 🫠

mrkaye97: I think they do? 
Screenshot 2026-05-26 at 2 49 20 PM

I think it might ship with Postgres (>13), but I'm not completely sure how to verify that

cmd/hatchet-migrate/migrate/migrations/20260522014401_v1_0_113.sql
```diff
@@ -0,0 +1,32 @@
...
+        EXCLUDE USING GIST (payload_inserted_at_date WITH =, block_external_id_range WITH &&)
+);
+
+CREATE TABLE v1_payloads_olap_offloaded_block_index (
+    payload_inserted_at_date DATE NOT NULL,
+    block_external_id_range uuidrange NOT NULL,
+    index_file_key TEXT NOT NULL,
+    CONSTRAINT v1_payloads_olap_offloaded_block_index_date_range_excl
+        EXCLUDE USING GIST (payload_inserted_at_date WITH =, block_external_id_range WITH &&)
+);
```
abelanger5, line 25: I'm always a little nervous when seeing tables without a primary key...for example, some postgres replication tools won't work without (cough cough bucardo). Should the index file key be labeled with `PRIMARY KEY`?

pkg/config/loader/loader.go
```diff
@@ -162,6 +163,21 @@ func (c *ConfigLoader) InitDataLayer() (res *database.Layer, err error) {
...
+			if err != nil && err != pgx.ErrNoRows {
+				return fmt.Errorf("loading uuidrange oid: %w", err)
+			}
+			if err == nil {
+				conn.TypeMap().RegisterType(&pgtype.Type{
+					Name:  "uuidrange",
+					OID:   uuidrangeOID,
+					Codec: &pgtype.RangeCodec{ElementType: uuidType},
+				})
+			}
```
abelanger5, line 178: what happens in the case where we don't have a `uuidrange` type? user hits an unexpected pgx error later?

pkg/repository/olap.go
```diff
@@ -3086,29 +3060,80 @@ func (r *OLAPRepositoryImpl) readPayloads(ctx context.Context, tx sqlcv1.DBTX, t
...
+	if r.payloadStore.ExternalStoreEnabled() {
+		for _, opt := range opts {
+			if _, found := externalIdToPayload[opt.ExternalId]; found {
+				continue
+			}
+			if _, found := externalIdToRetrieveFromExternalOpt[opt.ExternalId]; found {
+				continue
+			}
+
+			indexFileKey, err := r.queries.GetOLAPOffloadedPayloadIndexBlock(ctx, r.pool, sqlcv1.GetOLAPOffloadedPayloadIndexBlockParams{
```
abelanger5: this looks like usage of `r.pool` inside of a possible `tx`?

abelanger5: at this point, we should definitely have a GHA review rule around this.

abelanger5: also: looks like a possible `n+1` query? anything we can do to avoid that?

pkg/repository/payloadstore.go
```diff
@@ -361,21 +321,81 @@ func (p *payloadStoreRepositoryImpl) retrieve(ctx context.Context, tx sqlcv1.DBT
...
+				InsertedAt: opt.InsertedAt,
+				TenantId:   opt.TenantId,
+				Type:       opt.Type,
+			}
+
+			if _, found := foundKeys[key]; found {
+				continue
+			}
+
+			indexFileKey, err := p.queries.GetOffloadedPayloadIndexBlock(ctx, p.pool, sqlcv1.GetOffloadedPayloadIndexBlockParams{
```
abelanger5: same problem potentially? use of `p.pool` inside a tx

pkg/repository/olap.go
mrkaye97: ugh yep, thanks 🤦 will fix this and the one below 

re: n+1 - I think it'd take a bigger refactor to fix, and was thinking that in most cases, N will be small, although we could refactor (maybe do this right after merging this one?)

pkg/config/loader/loader.go
mrkaye97: I think in practice, nobody will hit this except us since we're not exposing the payload offload stuff anywhere but our managed cloud, but yeah, in theory they'd hit an encoding error

cmd/hatchet-migrate/migrate/migrations/20260522014401_v1_0_113.sql
mrkaye97: I'll take the bait and push back here 😅 

The key can't be a PK, since it's non-unique: one day can have many blocks. We'd really need the exclusion constraint to be the PK, but if it's actually a PK and not an exclusion constraint like this, then we'd only be able to guarantee that there's no exact duplicate (using equality) date + range, and not that there are no two overlapping ranges for the same date, which is what this constraint gets us.

I agree in general about not having a PK being a code smell, but I think it's worth it for the guarantees we get here. Maybe I can add a comment on this?

pkg/repository/olap.go
mrkaye97: nvm, I had this backwards - there's another n+1 query (I think on the cloud side). I'll fix this one

cmd/hatchet-migrate/migrate/migrations/20260522014401_v1_0_113.sql
abelanger5: perhaps I'm misunderstanding `index_file_key` - doesn't it uniquely identify the index file in S3?

mrkaye97: Ooh, yes, that should be unique, but I don't think it really makes sense to have as the PK since we'll never do a lookup by it (and it's the "value" in the k-v sense). If we want to go that route, I think we'd be better off just having an autoincrementing integer PK, although we can also add a unique constraint on the `index_file_key`.

Sorry, was getting `key` here confused with the `key` column (a date) in the offload cutover offset table that I've been looking at every day

pkg/repository/sqlcv1/payload-store.sql
```diff
@@ -157,39 +157,42 @@ DELETE FROM v1_payload_cutover_job_offset
...
+    payload_inserted_at_date,
+    block_external_id_range,
+    index_file_key
+)
+VALUES (
+    @payloadInsertedAtDate::DATE,
+    uuidrange(@blockLowerExternalIdBound::UUID, @blockUpperExternalIdBound::UUID, '(]'),
+    @indexFileKey::TEXT
+)
+ON CONFLICT ON CONSTRAINT v1_payload_offloaded_block_index_date_range_excl DO NOTHING
```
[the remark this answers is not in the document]
mrkaye97: this is incorrect, just confirmed

abelanger5 (approved):

PR #3984, opened 2026-05-22
Fix: Payload external id duplicates from durable child spawning

mrkaye97:
# Description

Fixes a bug I introduced where child spawning would use the child's external id as the external id in the durable event log, which worked fine for the event log itself but broke the payload logic since we'd use the same external id for both the event log entry and the task input

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

<details id="ai-disclosure">

No AI use in this PR

</details>

### DIFF
--- a/pkg/repository/sqlcv1/durable_event_log.sql
+++ b/pkg/repository/sqlcv1/durable_event_log.sql
@@ -44,6 +44,8 @@ RETURNING v1_durable_event_log_file.*
 -- name: UpdateLogFile :one
 UPDATE v1_durable_event_log_file
 SET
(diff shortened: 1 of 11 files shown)


pkg/repository/sqlcv1/durable_event_log.sql
```diff
@@ -199,6 +203,7 @@ WITH inputs AS (
     SELECT
         i.tenant_id,
         i.external_id,
+        NULLIF(i.child_task_external_id, '00000000-0000-0000-0000-000000000000'::UUID),
```
grutt, line 206: this potentially problematic?

grutt (approved):

PR #3986, opened 2026-05-22
fix: Refactor worker cleanup logic

juliusgeo:
# Description

Refactors the worker cleanup logic so that it is more centralized around the worker. Adds a queue to record the worker id so that we don't need to rely on timing being good when sending the sigterm. Refactors the action runner cleanup so that it's all one function

Fixes # (issue)

## Type of change

## What's Changed

## Checklist

Changes have been:

- [ ] Linted and formatted

---

<details id="ai-disclosure">

writing repro tests
</details>


sdks/python/tests/unit/test_shutdown_ordering.py
```diff
@@ -0,0 +1,288 @@
...
+    # Let the task start and block in its executor thread.
+    await asyncio.sleep(0.05)
+    assert not process.killing, "should not be killing before event is set"
+
+    # Trigger shutdown via the event — the primary (non-signal) path.
+    stop_event.set()
+    await asyncio.wait_for(task, timeout=5.0)
+
+    assert process.killing, "_stop_action_loop should have set killing=True"
+    assert process.listener.cleanup_called, "listener.cleanup() should have been called"
```
mrkaye97: I'm not super convinced by this test, it looks sort of like we just stub in a fake action listener, set the `cleanup_called` and `killing` flags, and assert on those? I wonder if using a real action listener and then mocking some methods and using e.g. `assert_called` or `assert_called_once` on the mocks would be better? Not 100% sure here, just a feeling

sdks/python/tests/unit/test_shutdown_ordering.py
```diff
@@ -0,0 +1,288 @@
...
+    stub = _StubActionListener()
+    process.listener = stub  # type: ignore[assignment]
+
+    await process._stop_action_loop()
+    assert process.killing
+
+    # Second call must succeed silently.
+    await process._stop_action_loop()
+    # cleanup() is called exactly once (the second call short-circuits).
+    assert stub.cleanup_called
```
mrkaye97: this too - I'm not convinced this is really testing the behavior of the real listener as much as it tests the stubs we just created, although I could be wrong there!

sdks/python/tests/unit/test_shutdown_ordering.py
```diff
@@ -0,0 +1,288 @@
...
+    await asyncio.wait_for(task, timeout=5.0)
+
+    assert process.killing, "_stop_action_loop should have set killing=True"
+    assert process.listener.cleanup_called, "listener.cleanup() should have been called"
+
+
+# ---------------------------------------------------------------------------
+# Test 2: _stop_action_loop is idempotent (safe to call twice)
+# ---------------------------------------------------------------------------
+
```
mrkaye97: let's remove these sorts of comments

sdks/python/hatchet_sdk/worker/worker.py
```diff
@@ -725,6 +736,34 @@ async def _close(self) -> None:
...
 
+    async def _pause_task_assignment(self) -> None:
+        """Pause task assignment at the engine level.
+
+        Reads the worker_id that the listener subprocess deposited at startup,
+        then calls the REST API to mark the worker as paused so the engine
+        stops routing new work here while in-flight tasks finish.
+        """
+        try:
+            worker_id = await asyncio.get_event_loop().run_in_executor(
```
mrkaye97: sorry, I know I keep adding stuff 🤦 

we should use `await asyncio.to_thread(self._worker_id_queue.get, timeout=10)` here instead for simplicity IMO

mrkaye97 (approved):
lgtm, let's give it a shot! we'll also need to bump the patch version for this before merging

one other thing is we may or may not need the same changes in the `deprecated` listener - it's probably fine to just have it going forward and not need to support it there 🤷

sdks/python/hatchet_sdk/worker/worker.py
juliusgeo: Fixed that! Also bumped patch version

PR #3995, opened 2026-05-23
fix(otel): create per-item run_workflow span in batch run instrumenta…

### LINKED ISSUE #3994: [BUG] when we run a workflow in batch ( pass an array in .run() ) it only creates one parent span instead of one parent span per children in python and typescript SDK ( GO is good )
**Describe the issue**
[BUG] when we invoke a workflow in batch eg
```
export const resolveSubprocessorInfoWorkflow =
  hatchet.workflow<ResolveSubprocessorInfoInput>({
    name: "resolve-subprocessor-info",
    on: {
      event: "resolve-subprocessor-info",
    },
    inputValidator: resolveSubprocessorInfoInput,
  });

  resolveSubprocessorInfoWorkflow.run([{}, {}])
```

we get only one parent span this makes it hard to know which invocation starts and end where 

actual behaviour
Image

expected behaviour

Image

**Environment**
- SDK: [e.g. TypeScript v 1.22.3]
- Engine: [Cloud

---

<details id="ai-disclosure">

- **Details**: [e.g. re-creation steps, initial diagnosis]

</details>

Mdev303:
When invoking a workflow in batch via `.run([...])`, the TypeScript and Python OTel instrumentors created a single `hatchet.run_workflows` span and injected the same traceparent into every child's metadata. All N worker-side execution spans appeared as siblings under one batch parent, with no per-child trigger span.

Match the Go SDK structure: open a per-item `hatchet.run_workflow` PRODUCER span inside the batch span, and inject each child's traceparent from its own per-item span context.

behaviour now
image

Closes #3994

# Description

Fixes # (issue)

## Type of change

## What's Changed

- modified typescript and python instrumentor file in their SDK to generate spans for each children when workflow is invoked in bulk

## Checklist

Changes have been:

- [ ] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: claude code

</details>


promptless-for-oss:

Triggered by [hatchet-dev/hatchet#3995](https://github.com/hatchet-dev/hatchet/pull/3995)

This update documents the batch run span hierarchy for OpenTelemetry instrumentation. The existing OTel docs now clarify that `hatchet.run_workflows` is a batch parent span and include a new section explaining the three-level structure: batch parent → per-item `hatchet.run_workflow` trigger spans → worker execution spans.

Review: [Document batch run OTel span hierarchy](https://app.gopromptless.ai/suggestions/4d8f18ca-c494-415a-bc08-5e3bd20d3bec)

sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
```diff
@@ -640,6 +649,37 @@ def _wrap_bulk_push_event(
...
+            "instrumentor": "hatchet",
+            **{
+                f"hatchet.{k.value}": v
+                for k, v in attributes.items()
+                if v
+                and k not in self.config.otel.excluded_attributes
+                and v != "{}"
+                and v != "[]"
+            },
+        }
```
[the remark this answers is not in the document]
Mdev303: config.input is already typed str | None on WorkflowRunTriggerConfig, so it’s not a dict — no json.dumps needed.

valid on the if v filter, but this is the same pattern already used in _wrap_run_workflow right below — I just mirrored it in the helper. Of the falsy cases, only child_index=0 is actually a real divergence from the TS SDK (priority is 1–3 in practice, and sticky=False is the default we’d want to drop anyway). Happy to fix it in a follow-up that updates both call sites together so they stay consistent, rather than mixing it into this batch-span fix.

sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
```diff
@@ -640,6 +649,37 @@ def _wrap_bulk_push_event(
...
+            OTelAttribute.STICKY: config.options.sticky,
+            OTelAttribute.KEY: config.options.key,
+        }
+        return {
+            "instrumentor": "hatchet",
+            **{
+                f"hatchet.{k.value}": v
+                for k, v in attributes.items()
+                if v
+                and k not in self.config.otel.excluded_attributes
```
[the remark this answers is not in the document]
Mdev303: Not a real issue — OTelAttribute is a str-mixin enum (class OTelAttribute(str, Enum)), so members compare equal to their string values (OTelAttribute.NAMESPACE == "namespace" is True). On top of that, excluded_attributes is typed as list[OTelAttribute] on a Pydantic BaseSettings model, so raw strings from config/env get coerced into enum members at load time. The exclusion works either wa

sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
```diff
@@ -863,20 +903,40 @@ def _wrap_run_workflows(
...
+                            ),
                         ),
-                    ),
+                    )
                 )
-                for config in workflow_run_configs
-            ]
 
-            return wrapped(workflow_run_configs_with_meta)
+            try:
```
BloggerBust: I think this `try/finally` starts a bit too late. The item spans are created above, then `_build_run_workflow_attributes(...)`, `config.model_dump(...)`, and `_inject_traceparent_into_metadata(...)` run before we enter the block that ends the spans. If any of those raise partway through the loop, previously-created `item_spans` will not be ended. Could we wrap the setup in its own `try/except` and end any spans created so far before re-raising? That would mirror TypeScript SDK behavior, which already cleans up partially created item spans on setup failure.

sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
```diff
@@ -906,20 +966,40 @@ async def _wrap_async_run_workflows(
...
+                            ),
                         ),
-                    ),
+                    )
                 )
-                for config in workflow_run_configs
-            ]
 
-            return await wrapped(workflow_run_configs_with_meta)
+            try:
```
BloggerBust: Same issue here as in the sync wrapper. This `try/finally` starts after the setup loop has already created item spans. If the setup raises partway through, any item spans created so far will not be ended. Could we apply the same exception handling here as well, so the async wrapper mirrors the TypeScript cleanup behavior too?

BloggerBust (changes requested):
Thanks for the contribution, it is appreciated.

The span hierarchy looks like the right fix for #3994. Great work!

I left two inline comments on the Python sync/async wrappers. The main blocking concern is that both wrappers create item spans during setup before entering the `try/finally` that ends them, so setup failures can leak spans. I also think this needs some test coverage before merge, since this PR changes sensitive OTel span hierarchy/context propagation behavior in both SDKs.

Everything else looked reasonable. The fix appears to address the core of #3994.

Mdev303:
@BloggerBust Thanks for the review I fixed the try catch issue in python and added some tests

BloggerBust:
Thank you! I am happy with the fix and added test coverage.

cc @gregfurman

mrkaye97 (changes requested):
hey @Mdev303, thanks for the suggestion and the fix here! One important thing before we can merge: This introduces a potentially breaking change to downstream Opentelemetry collector stacks that people using the SDK may be using (i.e. DataDog, Honeycomb, and so on). We unfortunately can't confirm whether or not anyone is relying on the structure / nesting of the spans, and so it's not possible for us to know whether this will inconspicuously break dashboards for something using the SDK, for instance.

To mitigate this, I think we should add a new setting (env var). For example, adding something like this to the OTel config might work:

```python
individual_run_spans_for_bulk_run: bool = False
```

(note that, importantly, the default needs to be `False`).

Then, we'd need to gate this logic based on whether or not the flag is enabled.

Let me know if that makes sense, and thanks again!

Mdev303:
Hi @mrkaye97 thanks for reviewing the PR , just to be clear and as stated in the issue this issue only exists in the typescript and python SDK but the GO sdk already handled this correctly ( it creates one span per child in case of batch ) so the env flag would only do something in the python and TS SDKs but nothing for the GO one  it would be inconsistant , do you still want it ?

mrkaye97:
> Hi @mrkaye97 thanks for reviewing the PR , just to be clear and as stated in the issue this issue only exists in the typescript and python SDK but the GO sdk already handled this correctly ( it creates one span per child in case of batch ) so the env flag would only do something in the python and TS SDKs but nothing for the GO one it would be inconsistant , do you still want it ?

Yep!

Mdev303:
@mrkaye97 I made the changes and updated the tests let me know if everything looks good or not

sdks/python/hatchet_sdk/opentelemetry/instrumentor.py
```diff
@@ -863,20 +948,21 @@ def _wrap_run_workflows(
...
-                    options=TriggerWorkflowOptions(
-                        **config.options.model_dump(exclude={"additional_metadata"}),
-                        additional_metadata=_inject_traceparent_into_metadata(
-                            config.options.additional_metadata,
-                        ),
-                    ),
+            item_spans: list[ApiSpan] = []
+            try:
+                workflow_run_configs_with_meta = self._enhance_workflow_run_configs(
+                    workflow_run_configs, item_spans
```
mrkaye97: I'd very much prefer to not pass this `item_spans` list into the enhance method and then mutate it - IMO, this is pretty bug prone. Can we refactor this away?

BloggerBust (approved):
I re-validated the enabled / disabled behavior locally and all previous concerns have been addressed. This looks good to me @mrkaye97

mrkaye97 (commented):
Yep, looks great - we'll need patch version bumps and changelog entries for both SDKs, and then we can merge. Thanks for the contribution @Mdev303 !

Mdev303:
@mrkaye97 fixed the instrumentor py file and added changelogs + version bump for both SDKs

mrkaye97 (approved):

PR #4007, opened 2026-05-26
Skip cookie session creation for bearer requests

RitwijParmar:
## Summary
- skip DB-backed cookie session creation for new requests authenticated with an `Authorization: Bearer ...` header
- avoid emitting a `Set-Cookie` header for those bearer-only requests
- add integration coverage that verifies bearer requests do not create `UserSession` rows

## Why
This addresses #3913. Bearer-authenticated SDK/worker REST traffic can pass through the cookie session store with no existing browser cookie. Before this change, `Save` generated a new anonymous `UserSession` row and cookie for that request, which can grow the `UserSession` table at API-call volume in self-hosted deployments.

The guard is intentionally narrow: it only applies when the request is bearer-authenticated, the session is new, and there is no existing session ID. Existing cookie-backed browser sessions and explicit logout/delete behavior continue through the existing paths.

## Tests
- Not run locally: this environment does not have `go`/`gofmt` installed in PATH.
- Verified with `git diff --check`.

## AI disclosure
This PR was prepared with AI assistance from OpenAI Codex. I reviewed the issue, inspected the affected session-store code, made the patch, and added the regression test with human-in-the-loop review.

### DIFF
--- a/pkg/auth/cookie/sessionstore.go
+++ b/pkg/auth/cookie/sessionstore.go
@@ -215,6 +215,10 @@ func (store *UserSessionStore) Get(r *http.Request, name string) (*sessions.Sess
 func (store *UserSessionStore) Save(r *http.Request, w http.ResponseWriter, session *sessions.Session) error {
 	repo := store.repo
 
+	if isBearerAuthenticatedRequest(r) && session.IsNew && session.ID == "" {
+		return nil
+	}
+
 	if session.Options.MaxAge < 0 {
 		if session.ID != "" {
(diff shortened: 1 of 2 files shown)


gregfurman:
I'm closing this for a couple of reasons:

1. This issue was already been solved in https://github.com/hatchet-dev/hatchet/pull/3923
2. Changes were not tested (your environment doesn't have `go` installed?)
3. We do not allow AI usage for  `good first issues`.  See [AI_POLICY.md](https://github.com/hatchet-dev/hatchet/blob/main/AI_POLICY.md).

PR #4008, opened 2026-05-26
fix(engine): Durable DAG child output bug

mrkaye97:
# Description

Fixes a bug when spawning a DAG from durable tasks where we'd return the output of a single task in the DAG as the DAG's output, instead of returning the usual map from the task name to the task output. This is a bit of an unfortunate fix since it creates a new way of getting DAG outputs from the engine, since on the `WorkflowRunRef` we currently use the output events, but I think that it's probably the least confusing way to do this that will be backwards compatible.

Also added a test that spawns a bunch of DAGs and makes sure they all have all the correct keys returned

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

<details id="ai-disclosure">

No AI was used in this PR

</details>

### DIFF
--- a/pkg/repository/output.go
+++ b/pkg/repository/output.go
@@ -146,6 +146,8 @@ func ExtractOutputFromMatchData(data []byte) ([]byte, error) {
 	// it's confusing because in other places we use the task events on the SDK itself to aggregate
 	// the outputs of each child into a map, but here we do it on the engine. it'd be a good fixme for the future
 	// to consolidate the different ways we handle this kind of thing in different places to be more consistent.
+	// I (Matt) opted to do it this way in https://github.com/hatchet-dev/hatchet/pull/4008 to maintain
+	// backwards compatibility, and because it was the simplest bug fix for the issue we saw at the time.
 	var outer map[string]map[string][]json.RawMessage
 	if err := json.Unmarshal(data, &outer); err != nil {
 		return nil, fmt.Errorf("failed to unmarshal match data: %w", err)
--- a/pkg/repository/output.go
+++ b/pkg/repository/output.go
@@ -147,7 +147,22 @@ func ExtractOutputFromMatchData(data []byte) ([]byte, error) {
 	}
 
 	for _, keyMap := range outer {
-		for _, entries := range keyMap {
+		if len(keyMap) == 0 {
+			continue
+		}
+
+		if len(keyMap) == 1 {
+			if entries, ok := keyMap["output"]; ok && len(entries) > 0 {
+				var event TaskOutputEvent
+				if err := json.Unmarshal(entries[0], &event); err != nil {
+					return nil, fmt.Errorf("failed to unmarshal task output event from match data: %w", err)
+				}
+				return event.Output, nil
+			}
+		}
+
+		aggregated := make(map[string]json.RawMessage, len(keyMap))
+		for key, entries := range keyMap {
 			if len(entries) == 0 {
 				continue
 			}
@@ -157,8 +172,17 @@ func ExtractOutputFromMatchData(data []byte) ([]byte, error) {
 				return nil, fmt.Errorf("failed to unmarshal task output event from match data: %w", err)
 			}
 
-			return event.Output, nil
+			if len(event.Output) > 0 {
+				aggregated[key] = json.RawMessage(event.Output)
+			}
 		}
+
+		result, err := json.Marshal(aggregated)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal aggregated DAG output: %w", err)
+		}
+
+		return result, nil
 	}
 
 	return nil, fmt.Errorf("no entries found in match data")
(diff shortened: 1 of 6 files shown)


promptless-for-oss:

Triggered by [PR #4008](https://github.com/hatchet-dev/hatchet/pull/4008)

Added clarification to the child-spawning documentation explaining that when you spawn a DAG workflow and wait for its result, the output is returned as a dictionary keyed by task name (e.g., `{"extract": {...}, "transform": {...}, "load": {...}}`).

Review: [Clarify DAG child output format](https://app.gopromptless.ai/suggestions/c9a0e8d6-3ebc-4267-bce5-313759056ae6)

pkg/repository/output.go
```diff
@@ -141,13 +141,35 @@ func newTaskEventFromBytes(b []byte) (*TaskOutputEvent, error) {
...
 	}
 
 	for _, keyMap := range outer {
-		for _, entries := range keyMap {
+		if len(keyMap) == 0 {
+			continue
+		}
+
+		if len(keyMap) == 1 {
+			if entries, ok := keyMap["output"]; ok && len(entries) > 0 {
```
gregfurman, line 166: So if we get a map with a single kv entry `{ "output": [...] }` then we should return early. Does the aggregation loop below not handle this case? Or is this a special scenario that needs accounting for -- in which case perhaps some documentation could be helpful!

gregfurman (approved):

gregfurman (changes requested):
Just saw the CI failed. Is this related or just a flake?

```
FAILED examples/durable/test_durable.py::test_dag_spawn_returns_full_output - hatchet_sdk.exceptions.FailedTaskRunExceptionGroup: Workflow run 1584e75d-2775-4201-948e-68d0f9444247 failed.

--- Exception 1 ---
KeyError: 'dag_child_1'
Traceback (most recent call last):
  File "/home/runner/work/hatchet/hatchet/sdks/python/hatchet_sdk/worker/runner/runner.py", line 269, in inner_callback
    output = task.result()
  File "/home/runner/work/hatchet/hatchet/sdks/python/hatchet_sdk/worker/runner/runner.py", line 361, in async_wrapped_action_func
    return await task.aio_call(ctx, dependencies)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/hatchet/hatchet/sdks/python/hatchet_sdk/runnables/task.py", line 462, in aio_call
    return await self._fn(workflow_input, cast(Context, ctx), **dependencies)  # type: ignore
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/hatchet/hatchet/sdks/python/examples/durable/worker.py", line 103, in durable_spawn_many_dags
    child_index=result["dag_child_1"]["child_index"],
                ~~~~~~^^^^^^^^^^^^^^^
KeyError: 'dag_child_
```

mrkaye97:
> Just saw the CI failed. Is this related or just a flake?
> 
> ```
> FAILED examples/durable/test_durable.py::test_dag_spawn_returns_full_output - hatchet_sdk.exceptions.FailedTaskRunExceptionGroup: Workflow run 1584e75d-2775-4201-948e-68d0f9444247 failed.
> 
> --- Exception 1 ---
> KeyError: 'dag_child_1'
> Traceback (most recent call last):
>   File "/home/runner/work/hatchet/hatchet/sdks/python/hatchet_sdk/worker/runner/runner.py", line 269, in inner_callback
>     output = task.result()
>   File "/home/runner/work/hatchet/hatchet/sdks/python/hatchet_sdk/worker/runner/runner.py", line 361, in async_wrapped_action_func
>     return await task.aio_call(ctx, dependencies)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/runner/work/hatchet/hatchet/sdks/python/hatchet_sdk/runnables/task.py", line 462, in aio_call
>     return await self._fn(workflow_input, cast(Context, ctx), **dependencies)  # type: ignore
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/runner/work/hatchet/hatchet/sdks/python/examples/durable/worker.py", line 103, in durable_spawn_many_dags
>     child_index=result["dag_child_1"]["child_index"],
>                 ~~~~~~^^^^^^^^^^^^^^^
> KeyError: 'dag_child_
> ```

Was this on the `old_engine_new_sdk` test? If yes, then yeah this is expected to fail on the old engine, since we don't have the fix yet so it's reproducing the bug

pkg/repository/output.go
mrkaye97: I don't think so (I'll need to double check) because I think the issue here is that when there's only a single task, there's a special case where we want to return the output _not keyed_ by the task name, but in every other case (DAGs), we want each task's output keyed by its name. I can add a comment!

gregfurman (approved):

PR #4012, opened 2026-05-26
fix: sleep bug for anything over 59s

### LINKED ISSUE #4017: [BUG] GoLang sdk is limited to sleeps below 60 seconds.
**Describe the issue**
Using a sleep for more than 59s causes unexpected durations. The bug is in the way the postgres function works and with the way go converts a duration into a string.

I chose the easiest approach to always use ms rather than convert between as you would also lose accuracy if you did an 1hr + 10m for example.

I havent written a unit test as there are none in this package, happy to add a test but not sure the best approach as I havent spent long in the code base so pointers are welcome :)
```sql
  CREATE OR REPLACE FUNCTION convert_duration_to_interval(duration text) RETURNS interval AS $$
  DECLARE

  CREATE OR REPLACE FUNCTION convert_duration_to_interval(duration text) RETURNS interval AS $$
  DECLARE
      num_value INT;
  BEGIN
      num_value := substring(duration from '^\d+');  -- extracts leading digits only → 1

      RETURN CASE
          WHEN duration LIKE '%ms' THEN ...
          WHEN duration LIKE '%s'  THEN make_interval(secs => num_value)   -- ← MATCHES THIS
          WHEN duration LIKE '%m'  THEN make_interval(mins => num_value)
          WHEN duration LIKE '%h'  THEN make_interval(hours => num_value)
          ...
      END;
  END;
```
"1h0m0s" ends with s → matches %s → make_interval(secs => 1) = 1 second instead of 1 hour.
So SleepFor(time.Hour) sleeps for 1 second, SleepFor(2time.Hour) sleeps for 2 seconds, SleepFor(30time.Minute) sleeps for 30 seconds, etc.

**Environment**
- SDK: Golang
- Engine: Self hosted but I imagine applies to both. 

**Expected behavior**
Allow sleeps over 60s

```go
	hc.NewStandaloneDurableTask(
		"say-hello",
		func(ctx hatchet.DurableContext, req struct {
			Name string `json:"name"`
		}) (any, error) {
			fmt.Printf("hello %s", req.Name)

			if _, err := ctx.SleepFor(time.Hour); err != nil{
				return nil, err
			}


			return nil, nil
		},
	**)**
```

darren-west:
# Description

Using a sleep for more than 59s causes unexpected durations. The bug is in the way the postgres function works and with the way go converts a duration into a string. 

I chose the easiest approach to always use ms rather than convert between as you would also lose accuracy if you did an 1hr + 10m for example.

I havent written a unit test as there are none in this package, happy to add a test but not sure the best approach as I havent spent long in the code base so pointers are welcome :) 

```sql
  CREATE OR REPLACE FUNCTION convert_duration_to_interval(duration text) RETURNS interval AS $$
  DECLARE

  CREATE OR REPLACE FUNCTION convert_duration_to_interval(duration text) RETURNS interval AS $$
  DECLARE
      num_value INT;
  BEGIN
      num_value := substring(duration from '^\d+');  -- extracts leading digits only → 1

      RETURN CASE
          WHEN duration LIKE '%ms' THEN ...
          WHEN duration LIKE '%s'  THEN make_interval(secs => num_value)   -- ← MATCHES THIS
          WHEN duration LIKE '%m'  THEN make_interval(mins => num_value)
          WHEN duration LIKE '%h'  THEN make_interval(hours => num_value)
          ...
      END;
  END;
```
- "1h0m0s" ends with s → matches %s → make_interval(secs => 1) = 1 second instead of 1 hour.
- So SleepFor(time.Hour) sleeps for 1 second, SleepFor(2*time.Hour) sleeps for 2 seconds, SleepFor(30*time.Minute) sleeps for 30 seconds, etc.

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

<details id="ai-disclosure">

- **Details**: I used it to diagnose the issue and then wrote the fix manually, using Claude code. 

</details>


BloggerBust (changes requested):
Hey Darren, thank you for finding and fixing this bug. I reproduced the issue with a local hatchet instance and verified your fix resolves it.

Can you please add a small unit test for the serialization format? I think this is your first contribution to Hatchet, so I thought I would offer you some code here to help.

Please add this as `pkg/worker/condition/condition_test.go`:

```go
//go:build !e2e && !load && !rampup && !integration

package condition

import (
	"testing"
	"time"

	"github.com/stretchr/testify/assert"

	contracts "github.com/hatchet-dev/hatchet/internal/services/shared/proto/v1"
)

func TestSleepConditionToPB_SleepForFormat(t *testing.T) {
	cases := []struct {
		duration time.Duration
		want     string
	}{
		{0, "0ms"},
		{500 * time.Millisecond, "500ms"},
		{59 * time.Second, "59000ms"},
		{60 * time.Second, "60000ms"},
		{61 * time.Second, "61000ms"},
		{30 * time.Minute, "1800000ms"},
		{time.Hour, "3600000ms"},
		{time.Hour + 10*time.Minute, "4200000ms"},
	}

	for _, tc := range cases {
		t.Run(tc.duration.String(), func(t *testing.T) {
			c := SleepCondition(tc.duration)
			pb := c.ToPB(contracts.Action_CREATE)

			assert.Len(t, pb.SleepConditions, 1)
			assert.Equal(t, tc.want, pb.SleepConditions[0].SleepFor)
		})
	}
}
```

You can run it locally with:

```sh
go test -v ./pkg/worker/condition/...
```

Once that test is added and passes, I will be happy with it.

Small non-blocker for next time: it is usually easier to open PRs from a feature branch rather than your fork’s `main` branch. No need to change anything for this PR, but using a branch like `fix/sleep-duration-ms` makes future contributions easier for you since your fork won't diverge from upstream.

darren-west:
Hi @BloggerBust thats great thanks, all pushed.

BloggerBust:
Thanks for the quick turnaround!

LGTM: cc @gregfurman  for merge approval and the `accepted` label on #4017.

mrkaye97 (approved):
lgtm, thanks for the fix!

PR #4015, opened 2026-05-27
feat(ts-sdk): add durable content step checkpoinitng

### LINKED ISSUE #3992: [FEAT] Add ctx.step() for dynamic, checkpointed side effects inside durable tasks

Inside a `durableTask`, every non-deterministic or I/O operation (DB call, LLM request, HTTP fetch, etc.) has to be wrapped in its own registered task and invoked via `ctx.runChild(...)` to be checkpointed in the durable event log. This has three downsides:

1. **Boilerplate & registration overhead** — every tiny side effect needs a separately defined and registered task in the worker init. Forgetting to register one is a common "task not found" footgun.
2. **UI noise** — the dashboard shows every one of these micro-tasks, drowning out the actual workflow steps a human cares about.
3. **Cost** — Hatchet bills per task, so wrapping fine-grained DB reads or RNG calls as full tasks gets expensive fast.

A lightweight `ctx.step(name, fn)` (or `ctx.checkpoint(...)`) primitive available inside durable tasks that:

- Checkpoints its result in the durable event log just like `runChild`, so retries skip already-executed steps.
- Is created **inline / dynamically** — no pre-registration with the worker.
- Does **not** show up as its own row in the dashboard (or shows collapsed under its parent task).
- Is **not billed as a separate task**.

Conceptually similar to inline activities in other durable runtimes (Temporal's `SideEffect` / inline activity pattern, Inngest's `step.run`, Restate's `ctx.run`).

```ts
const user = await ctx.step("load-user", () => db.users.findById(id));
const score = await ctx.step("score", () => llm.classify(user.bio));
```

- **Current approach:** register a shared task per side effect and call via `ctx.runChild`. Works but causes the issues above.
- **Skipping durability** for cheap calls (just calling the DB directly inside the durable task). Breaks determinism guarantees on replay.
- **Batching** multiple side effects into one coarse task. Hurts observability and forces awkward grouping that doesn't match the natural step boundaries.

---

<details id="ai-disclosure">

- **Details**: Claude (Opus 4.7) was used to draft and structure the issue from my notes; problem framing and proposed API are mine.

</details>

codeafridi:
fixes #3992 

# Description

Adds `ctx.step(name, fn)` and `ctx.checkpoint(name, fn)` to the TypeScript `DurableContext`, backed by the existing durable memo event path. This lets durable tasks checkpoint dynamic side effects without registering separate child tasks.

Fixes #3992

## Type of change

## What's Changed

- [x] Added `DurableContext.step()` for named, checkpointed side effects.
- [x] Added `DurableContext.checkpoint()` as an alias for `step()`.
- [x] Added type coverage for checkpointed object and primitive results.

## Checklist

Changes have been:

- [x] Linted and formatted

## Testing

- `./node_modules/.bin/tsc --noEmit`
- `./node_modules/.bin/prettier "src/v1/client/worker/context.ts" "src/v1/examples/simple/types.test-d.ts" --check`
- `./node_modules/.bin/jest src/v1/examples/simple/types.test.ts --runInBand`

---

### DIFF
--- a/sdks/typescript/src/v1/client/worker/context.ts
+++ b/sdks/typescript/src/v1/client/worker/context.ts
@@ -22,7 +22,7 @@ import WorkflowRunRef from '@hatchet/util/workflow-run-ref';
 import { Conditions, Render } from '@hatchet/v1/conditions';
 import { conditionsToPb } from '@hatchet/v1/conditions/transformer';
 import { CreateWorkflowDurableTaskOpts, CreateWorkflowTaskOpts } from '@hatchet/v1/task';
-import { JsonObject, OutputType } from '@hatchet/v1/types';
+import { JsonObject, JsonValue, OutputType } from '@hatchet/v1/types';
 import { Action as ConditionAction } from '@hatchet/protoc/v1/shared/condition';
 import { HatchetClient } from '@hatchet/v1';
 import { applyNamespace } from '@hatchet/util/apply-namespace';
@@ -44,6 +44,7 @@ import { waitForPreEviction } from './deprecated/pre-eviction';
 type TriggerData = Record<string, Record<string, any>>;
 
 type ChildRunOpts = RunOpts & { key?: string; sticky?: boolean };
+type StepResult = Exclude<JsonValue, undefined>;
 
 export interface SleepResult {
   /** The sleep duration in milliseconds. */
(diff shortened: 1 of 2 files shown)


promptless-for-oss:
## 📝 Documentation Update

I've created a documentation suggestion for the new `DurableContext.step()` and `checkpoint()` methods:

**Changes:**
- **API Reference** (`reference/typescript/Context.mdx`): Added method entries with parameter tables and return types
- **Conceptual Guide** (`v1/durable-tasks.mdx`): Added "Step checkpointing (TypeScript)" section with usage examples and best practices

[**Review the documentation suggestion →**](https://app.gopromptless.ai/suggestions/12532fc9-638f-4672-a381-909774202e15)

---
<sub>🤖 Generated by [Promptless](https://gopromptless.ai)</sub>

mrkaye97:
Hi @codeafridi - thanks for the contribution! I'm going to close this per the comment I left [here](https://github.com/hatchet-dev/hatchet/issues/3992#issuecomment-4625177072). 

One other note on the implementation: It probably would make sense to export `memo` directly instead of adding these wrappers, but I'm not sure that we want to commit to supporting this yet, as there are many open questions on how the implementation should behave

PR #4016, opened 2026-05-27
fix(sdks/go): filters not being serialised

### LINKED ISSUE #4018: [BUG] GoLang sdk filters do not work
**Describe the issue**
You cannot use filters in the GoLang sdk as they are not sent to the server. 

**Environment**
- SDK: GoLang
- Engine: I imagine both but tested on standalonme

**Expected behavior**
Filters to work within the GoLang sdk. 

```go
	hc.NewStandaloneTask(
		"say-hello",
		func(ctx hatchet.Context, req struct {
			Name string `json:"name"`
		}) (any, error) {
			fmt.Printf("hello %s", req.Name)

			return nil, nil
		},
		hatchet.WithFilters(types.DefaultFilter{
			Expression: "input.name in paylod.friends",
			Scope:      "friends",
			Payload:    map[string]any{"friends": []string{"Alice", "Bob", "Charlie"}},
		}),
	)

```

darren-west:
# Description

Filters do not work in the go sdk. An example of a filter might be:

```go
	hc.NewStandaloneTask(
		"say-hello",
		func(ctx hatchet.Context, req struct {
			Name string `json:"name"`
		}) (any, error) {
			fmt.Printf("hello %s", req.Name)

			return nil, nil
		},
		hatchet.WithFilters(types.DefaultFilter{
			Expression: "input.name in paylod.friends",
			Scope:      "friends",
			Payload:    map[string]any{"friends": []string{"Alice", "Bob", "Charlie"}},
		}),
	)

```

## Type of change

## What's Changed

- The filters are added as a workflow option and converted to protobuf. 

## Checklist

Changes have been:

- [x] Tested manually on my machine
- [x] Linted and formatted

I propose we add testing in for the go sdk as it is lacking.


BloggerBust (changes requested):
Hey @darren-west, nice find. It is also really nice to see another contribution from you, thank you!

I was able to reproduce the issue and it is a legitimate bug. I have reviewed your code and your fix looks right to me.

Before we merge this, can you please add some test coverage?

The most useful regression test would check that `DefaultFilters` is populated in the protobuf request produced by the declaration `Dump()`. Ideally the test would start from `WithFilters(...)`. At the very least, it would be great to have a focused `Dump()` serialization test.

I validated a temporary test locally with:

```sh
go test -v -run TestDump ./sdks/go/internal/
```

darren-west:
@BloggerBust Sorry I missed this but I have now pushed a test. Because of the way the workflow is setup its difficult to test using `WithFilters()`.

I think ideally it would be worth refactoring a few bits to enable the testing of the exported functions so the public api is the main focus being tested but perhaps that would be worth pulling out to a separate PR.

BloggerBust:
Thank you @darren-west!

@gregfurman, this looks good to me.

mrkaye97:
Hi @darren-west, thanks for the improvement here! One small note is I think `WithFilters` should be called `WithDefaultFilters` for consistency, but otherwise this looks great

sdks/go/workflow.go
```diff
@@ -340,9 +341,9 @@ func WithEvents(events ...string) TaskOption {
 	}
 }
 
-// WithFilters sets default filters for event-triggered tasks.
-func WithFilters(filters ...types.DefaultFilter) TaskOption {
-	return func(config *taskConfig) {
+// WithFilters sets default filters for event-triggered workflows or standalone tasks.
+func WithFilters(filters ...types.DefaultFilter) WorkflowOption {
```
gregfurman:
```suggestion
// WithDefaultFilters sets default filters for event-triggered workflows or standalone tasks.
func WithDefaultFilters(filters ...types.DefaultFilter) WorkflowOption {
```

sdks/go/examples/on-event/main.go
```diff
@@ -57,7 +57,7 @@ func LowerWithFilter(client *hatchet.Client) *hatchet.StandaloneTask {
 	return client.NewStandaloneTask(
 		"lower", accessFilterPayload,
 		hatchet.WithWorkflowEvents(SimpleEvent),
-		hatchet.WithFilters(types.DefaultFilter{
+		hatchet.WithDefaultFilters(types.DefaultFilter{
```
gregfurman, line 60: Ahh good catch! Sorry for breaking your changes 😅

gregfurman (approved):
LGTM!

PR #4019, opened 2026-05-27
fix: go use durable listener for async results

grutt:
# Description

Async spawn methods (RunNoWait) did not correctly subscribe to durable state for durable children.

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

Added test cases, verified failing before patch.

<details id="ai-disclosure">

- **Details**: Used cursor to verify parity with python impl and iterate on tests.

</details>

### DIFF
--- a/pkg/client/admin.go
+++ b/pkg/client/admin.go
@@ -36,6 +36,74 @@ type ChildWorkflowOpts struct {
 	DesiredWorkerLabels map[string]*types.DesiredWorkerLabel
 }
 
+// NewChildWorkflowTriggerRequest builds the trigger request used to start a child workflow.
+func NewChildWorkflowTriggerRequest(workflowName string, input interface{}, opts *ChildWorkflowOpts, sharedMeta map[string]string) (*v1contracts.TriggerWorkflowRequest, error) {
+	if opts == nil {
+		opts = &ChildWorkflowOpts{}
+	}
+
+	inputBytes, err := json.Marshal(input)
+	if err != nil {
+		return nil, fmt.Errorf("could not marshal input: %w", err)
+	}
+
+	if opts.ChildIndex < math.MinInt32 || opts.ChildIndex > math.MaxInt32 {
+		return nil, fmt.Errorf("child index out of range")
+	}
+	childIndex := int32(opts.ChildIndex) // nolint:gosec
+
+	request := &v1contracts.TriggerWorkflowRequest{
+		Name:                    workflowName,
+		Input:                   string(inputBytes),
+		ParentId:                &opts.ParentId,
+		ParentTaskRunExternalId: &opts.ParentTaskRunId,
+		ChildIndex:              &childIndex,
+		ChildKey:                opts.ChildKey,
+		DesiredWorkerId:         opts.DesiredWorkerId,
+	}
+
+	additionalMetadata := mergeAdditionalMetadata(sharedMeta, opts.AdditionalMetadata)
+	if additionalMetadata != nil {
+		if err := WithRunMetadata(additionalMetadata)(request); err != nil {
+			return nil, fmt.Errorf("could not apply run metadata: %w", err)
+		}
+	}
+
+	if opts.Priority != nil {
+		if err := WithPriority(*opts.Priority)(request); err != nil {
+			return nil, fmt.Errorf("could not apply run priority: %w", err)
+		}
+	}
+
+	if opts.DesiredWorkerLabels != nil {
+		if err := WithDesiredWorkerLabels(opts.DesiredWorkerLabels)(request); err != nil {
+			return nil, fmt.Errorf("could not apply desired worker labels: %w", err)
+		}
+	}
+
+	return request, nil
+}
+
+func mergeAdditionalMetadata(sharedMeta map[string]string, opt *map[string]string) map[string]string {
+	if sharedMeta == nil && opt == nil {
+		return nil
+	}
+
+	additionalMeta := make(map[string]string)
+
+	for key, value := range sharedMeta {
+		additionalMeta[key] = value
+	}
+
+	if opt != nil {
+		for key, value := range *opt {
+			additionalMeta[key] = value
+		}
+	}
+
+	return additionalMeta
+}
+
 type WorkflowRun struct {
 	Name    string
 	Input   interface{}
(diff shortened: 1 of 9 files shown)


pkg/client/durable_task_listener.go
```diff
@@ -464,6 +472,94 @@ func (l *DurableTaskListener) removePendingCallback(key PendingCallbackKey) {
...
+func (l *DurableTaskListener) SendTriggerRunsRequest(
+	ctx context.Context,
+	taskExternalID string,
+	invocationCount int32,
+	triggerOpts []*v1.TriggerWorkflowRequest,
+) ([]TriggerRunAckEntry, error) {
+	ackKey := PendingAckKey{TaskID: taskExternalID, SignalKey: int64(invocationCount)}
+	ackCh := l.AddPendingEventAck(ackKey)
+
+	if l.l != nil {
```
mrkaye97: do we need the nil check here? 😅

sdks/go/client.go
```diff
@@ -915,17 +922,37 @@ func (c *Client) runWorkflowInternal(ctx context.Context, otelCtx context.Contex
...
+		durableRef, handled, durableErr := runDurableChildWorkflow(ctx, durableWorkflowName, input, runOpts)
+		if handled {
+			if durableErr != nil {
+				span.SetStatus(codes.Error, durableErr.Error())
+				span.RecordError(durableErr)
+				return nil, durableErr
+			}
+			span.SetAttributes(attribute.String(hatchetotel.AttrChildWorkflowRunID, durableRef.RunId))
+			return durableRef, nil
+		}
```
mrkaye97, line 948: will this run every child workflow as a durable child? not too familiar with the `hCtx, ok := ...` bit, but it looks like the `ok` will be `true` regardless of whether or not the parent is durable?

sdks/go/client.go
```diff
@@ -962,27 +994,47 @@ func (c *Client) RunMany(ctx context.Context, workflowName string, inputs []RunM
 	)
 	defer span.End()
 
+	if _, ok := originalCtx.(Context); ok {
+		durableWorkflowName := workflowName
+		if ns := c.legacyClient.Namespace(); ns != "" && !strings.HasPrefix(durableWorkflowName, ns) {
+			durableWorkflowName = fmt.Sprintf("%s%s", ns, durableWorkflowName)
+		}
+
+		if durableRefs, handled, err := runManyDurableChildWorkflows(originalCtx, otelCtx, durableWorkflowName, inputs); handled {
```
mrkaye97, line 1003: same here maybe? again not too sure

sdks/go/durable_child.go
```diff
@@ -0,0 +1,295 @@
...
+	entries, err := listener.SendTriggerRunsRequest(
+		hCtx.GetContext(),
+		hCtx.StepRunId(),
+		invocationCount,
+		[]*v1.TriggerWorkflowRequest{trigger},
+	)
+	if err != nil {
+		return nil, true, fmt.Errorf("failed to spawn durable child workflow: %w", err)
+	}
+	if len(entries) == 0 {
```
mrkaye97, line 67: might want an assertion here that `len(entries) == 1`

sdks/go/durable_child.go
```diff
@@ -0,0 +1,295 @@
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to wait for durable child workflow: %w", err)
+	}
+
+	if len(payloadBytes) == 0 {
+		return map[string]any{}, nil
+	}
+
+	var payload any
```
mrkaye97: at the very least, I think we can probably use `map[string]any` here? since we know it has to be a map I think?

sdks/go/durable_child.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	additionalMetadata, err := marshalAdditionalMetadata(runOpts.AdditionalMetadata)
+	if err != nil {
+		return nil, err
+	}
+	var priority *int32
+	if runOpts.Priority != nil {
+		priority = &[]int32{int32(*runOpts.Priority)}[0]
+	}
```
mrkaye97: a lot of this looks like code we must have somewhere else already - curious if we can reuse something existing

sdks/go/durable_child.go
```diff
@@ -0,0 +1,295 @@
...
+
+	metadataBytes, err := json.Marshal(metadata)
+	if err != nil {
+		return nil, fmt.Errorf("could not marshal additional metadata: %w", err)
+	}
+
+	metadataString := string(metadataBytes)
+	return &metadataString, nil
+}
```
mrkaye97: same with this, do we not have this somewhere else?

sdks/go/durable_child.go
```diff
@@ -0,0 +1,295 @@
...
+
+func desiredWorkerLabelsToProto(labels map[string]*DesiredWorkerLabel) map[string]*v1.DesiredWorkerLabels {
```
mrkaye97: and this, I swear I'd implemented this

pkg/client/durable_task_listener.go
grutt: consistent inconsistency

sdks/go/durable_child.go
grutt: gc

sdks/go/client.go
grutt: ill update the name so its a little more clear but this whole ctx check pattern isnt totally intuitive to me. 

but this is correct, basically we do all the durable checks in the run... method here and short circuit if its dispatched, else we run as the old version

pkg/client/admin.go
```diff
@@ -36,6 +36,74 @@ type ChildWorkflowOpts struct {
...
+	}
+
+	inputBytes, err := json.Marshal(input)
+	if err != nil {
+		return nil, fmt.Errorf("could not marshal input: %w", err)
+	}
+
+	if opts.ChildIndex < math.MinInt32 || opts.ChildIndex > math.MaxInt32 {
+		return nil, fmt.Errorf("child index out of range")
+	}
```
mrkaye97, line 52: is this a real risk...?

grutt: pre-existing check

mrkaye97 (approved):

pkg/client/admin.go
mrkaye97: 🤦

PR #4029, opened 2026-05-28
feat: simplify local dashboard development workflow

### LINKED ISSUE #3005: [DEVEX] Locally serve API and frontend via `hatchet-dashboard`

Running Hatchet's frontend locally requires spinning up `hatchet-server` and `frontend/app` as separate processes, plus a Caddy reverse proxy to route `app.dev.hatchet-tools.com` to `localhost`. 

This is unnecessarily complex when the `hatchet-dashboard` binary already serves both the API and frontend together.

Use the `hatchet-dashboard` binary to serve both the API and frontend directly over localhost, eliminating the need for Caddy, hostname aliasing, and managing separate (and tightly coupled) processes.

N/A

**Additional context**

See discussion https://github.com/hatchet-dev/hatchet/pull/2972#discussion_r2791212620

dhimanAbhi:
# Description
Adds a new hatchet-dashboard runtime and updates the local development workflow to launch the frontend Vite dev server, API server, and engine together through a single process entrypoint.

Developers can now run `task start-dev` which starts:
- frontend Vite dev server
- API server
- engine

directly on localhost without requiring Caddy or local hostname aliasing.

This change simplifies the local development setup by replacing the previous Caddy-based reverse proxy workflow with a unified dashboard binary.

Fixes #3005

## Type of change

## What's Changed
- Added cmd/hatchet-dashboard
- Added hack/dev/start-dashboard.sh
- Updated task start-dev to use the dashboard runtime
- Added support for launching the frontend Vite dev server from Go
- Simplified local development workflow by removing dependency on Caddy during development

## Checklist

Changes have been:

- [ ] Linted and formatted

---

<details id="ai-disclosure">

</details>


Taskfile.yaml
```diff
@@ -130,9 +132,7 @@ tasks:
     deps:
       - task: goose-migrate
       - task: start-db
-      - task: start-api
-      - task: start-engine
-      - task: start-frontend
```
gregfurman, line 135: This is basically what hatchet lite does already, except we build the static files and serve them via that go `staticfileserver` instead of by running the server as a seperate process. Why not just use hatchet-lite here instead? 🤔

cmd/hatchet-dashboard/main.go
```diff
@@ -0,0 +1,120 @@
...
+	}()
+
+	// vite server
+	go func() {
+		var cmd *exec.Cmd
+		if runtime.GOOS == "windows" {
+			cmd = exec.Command("cmd", "/c", "npm", "run", "dev")
+		} else {
+			cmd = exec.Command("npm", "run", "dev")
+		}
```
gregfurman: Unsure we want a hard dependency here on NPM / `os` calls 😬

gregfurman (commented):
Hey! Can you clarify how this changes behaviour from what we already do with Hatchet lite? I'm wondering if we could make changes to the taskfile without any additional dependencies here to remove the dep on Caddy.

Taskfile.yaml
dhimanAbhi: That makes sense. My reasoning for not using hatchet-lite directly was mainly around the frontend development experience.

`hatchet-lite` currently serves built static assets, whereas during development we already rely on the Vite dev server (npm run dev) for HMR and faster iteration. The idea here was essentially to simplify the existing local development workflow (task start-frontend + API + engine) into a single entrypoint without requiring Caddy/hostname aliasing.

cmd/hatchet-dashboard/main.go
dhimanAbhi: That’s fair. I agree the current implementation probably introduces too much coupling around spawning npm/os processes from Go.

I think a cleaner approach would be to keep frontend startup within the existing start-frontend task and avoid having the dashboard runtime manage frontend processes directly.

PR #4045, opened 2026-05-29
feat(engine): idempotency

### LINKED ISSUE #3673: [FEAT] User-configurable idempotency keys

A common request is idempotency keys during enqueue time, potentially supporting some kind of expiration or debouncing mechanism. 

While durable tasks spawning child workflows are implicitly and explicitly idempotency (via the `childKey`), it's desirable to allow this behavior to be configurable for regular enqueueing. It's possible that we would also extend this to event enqueueing. As a result, idempotency keys should be as generic as possible (given that we already use idempotency keys internally to de-duplicate fired schedules, this is a requirement anyway). 

There is some existing work in #2950 which we will take a close look at and likely build on top of.

mrkaye97:
# Description

Adds support for user-facing idempotency configuration on tasks and workflows, to enable e.g. duplicate event pushes to not trigger duplicate tasks. Idempotency is configured on the task / workflow, and then collisions raise an error (which can be handled) on the SDKs when they occur on the trigger path. On the event path, the runs that would've been triggered are dropped.

Fixes #3673

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

<details id="ai-disclosure">

- **Details**: Used Claude Code for generating the other three SDKs after implementing Python, and for wiring up some API logic

</details>


internal/services/admin/v1/server.go
```diff
@@ -641,55 +656,70 @@ func (a *AdminServiceImpl) ingest(ctx context.Context, tenantId uuid.UUID, opts
...
+
+		if err != nil {
+			return nil, fmt.Errorf("could not trigger workflows: %w", err)
+		}
+
+		if a.tw != nil {
+			if signalErr := a.tw.SignalCreated(ctx, tenantId, tasks, dags); signalErr != nil {
+				a.l.Error().Ctx(ctx).Err(signalErr).Msg("failed to signal created tasks and DAGs")
+			}
+		}
```
abelanger5: IIRC this is sending the tasks and DAGs to the OLAP controller, so if the triggerWriter isn't enabled, we'll lose those tasks on the OLAP side. 

Perhaps we should consider always enabling the triggerWriter?

internal/services/admin/v1/server.go
```diff
@@ -641,55 +656,70 @@ func (a *AdminServiceImpl) ingest(ctx context.Context, tenantId uuid.UUID, opts
...
 
-	msg, err := tasktypes.TriggerTaskMessage(
-		tenantId,
-		optsToSend...,
-	)
+	// important: if there's an idempotency key, we can't fall back to the MQ as we'd lose
+	// any tx-safety we get if we did, so we call `TriggerFromWorkflowNames` directly for those,
+	// and only fall back to the MQ for opts without idempotency keys
+	optsWithIdempotencyKeys := make([]*v1.WorkflowNameTriggerOpts, 0, len(optsToSend))
+	optsWithoutIdempotencyKeys := make([]*v1.WorkflowNameTriggerOpts, 0, len(optsToSend))
```
abelanger5: it might be a little cleaner to just use the triggerWriter for the whole batch if we have a single idempotency key, otherwise we have a bunch of partial states to deal with and send to the user? like a completed transaction with a bunch of partial rabbitmq writes?

internal/services/admin/v1/server.go
```diff
@@ -587,7 +602,7 @@ func (a *AdminServiceImpl) generateExternalIds(ctx context.Context, tenantId uui
 	return a.repo.Triggers().PopulateExternalIdsForWorkflow(ctx, tenantId, opts)
 }
```
abelanger5, line 603: I think the answer is no, but to put it on your radar: given the recent changes to durable tasks and now this PR, are we able to get rid of this `PopulateExternalIdsForWorkflow`? It's a major bottleneck in spawning a child workflow and the idempotency work is meant to replace it if we can change the behavior of regular tasks.

internal/services/admin/server_v1.go
```diff
@@ -286,53 +302,86 @@ func (i *AdminServiceImpl) ingest(ctx context.Context, tenantId uuid.UUID, opts
...
-	if err != nil {
-		return fmt.Errorf("could not create event task: %w", err)
+	if len(optsWithIdempotencyKeys) > 0 {
+		tasks, dags, collisions, err := i.repov1.Triggers().TriggerFromWorkflowNames(ctx, tenantId, optsWithIdempotencyKeys)
+
+		if err != nil {
+			return nil, fmt.Errorf("could not trigger workflows with idempotency keys: %w", err)
+		}
+
+		if i.tw != nil {
```
abelanger5: same comment here on the separate triggerWriter

internal/services/ticker/schedule_workflow_v1.go
```diff
@@ -2,48 +2,40 @@ package ticker
...
 
-	var pgErr *pgconn.PgError
-	// if we get a unique violation, it means we tried to create a duplicate idempotency key, which means this
-	// run has already been processed, so we should just return
-	if err != nil && errors.As(err, &pgErr) && pgErr.Code == pgerrcode.UniqueViolation {
-		t.l.Info().Ctx(ctx).Msgf("idempotency key for scheduled workflow %s already exists, skipping", scheduledWorkflowId)
-		return nil
-	} else if err != nil {
-		return fmt.Errorf("could not create idempotency key: %w", err)
+	claimed, err := t.repov1.Idempotency().ClaimKey(ctx, tenantId, scheduledWorkflowId.String(), expiresAt, scheduledWorkflowId)
```
abelanger5: we might want to differentiate between idempotency keys claimed internally vs externally? for example, `hatchet_internal_` prefix and prevent users from setting this prefix on an idempotency key? it would be really confusing for a user to hit a collision here.

pkg/repository/trigger.go
```diff
@@ -594,21 +639,122 @@ type createCoreUserEventOpts struct {
...
+
+	for _, tuple := range triggerCandidateTuples {
+		if tuple.idempotency == nil {
+			tuples = append(tuples, tuple)
+		} else {
+			externalIdToTuple[tuple.externalId] = tuple
+			key, err := r.evalIdempotencyKey(tuple)
+			if err != nil {
+				return nil, nil, nil, err
+			}
```
abelanger5, line 718: looks like failure to eval the idempotency key drops the whole batch (and isn't recoverable, so we might infinitely retry). shouldn't we be writing this to the CEL failure table and continuing to make progress with the rest of the batch?

internal/services/admin/v1/server.go
mrkaye97: Yeah that's a good idea, I'll look into that

internal/services/admin/v1/server.go
mrkaye97: Yeah good point, I was kind of struggling with this because it feels bad, but I definitely prefer the all or nothing way, so I'll go for that

internal/services/admin/v1/server.go
mrkaye97: Yeah I think not yet? Need to remove caching on the non-durable child spawn path first?

internal/services/ticker/schedule_workflow_v1.go
mrkaye97: In this case I don't think it's possible to hit a collision since the key being the external id means you'd need to reuse the same external id somehow? I think prefixing is a fine idea though, I can do that - agreed it'd be confusing if this did happen

pkg/repository/trigger.go
mrkaye97: Hah, yeah, I forgot about the CEL failure table, will do

pkg/repository/sqlcv1/olap.sql
```diff
@@ -1860,8 +1860,8 @@ WHERE
     )
     AND (
         sqlc.narg('eventIds')::UUID[] IS NULL OR
-        EXISTS (
-            SELECT 1
+        (e.id, e.seen_at) IN (
+            SELECT event_id, event_seen_at
             FROM v1_event_lookup_table_olap elt
```
mrkaye97, line 1883: drive by: this filter did nothing (same as below)

pkg/repository/sqlcv1/tasks.sql
```diff
@@ -1364,7 +1364,9 @@ SELECT
     run_external_id,
     event_id,
     event_seen_at,
-    filter_id
+    CASE WHEN filter_id = '00000000-0000-0000-0000-000000000000'::uuid THEN NULL
+        ELSE filter_id
```
mrkaye97, line 1419: drive by: preventing nil filter ids

pkg/repository/olap.go
```diff
@@ -2694,7 +2700,7 @@ func (r *OLAPRepositoryImpl) GetEventWithPayload(ctx context.Context, externalId
 			EventKey:                event.Key,
 			EventPayload:            payload,
 			EventAdditionalMetadata: event.AdditionalMetadata,
-			EventScope:              event.Scope.String,
+			EventScope:              eventScope,
 			QueuedCount:             queuedCount,
 			RunningCount:            runningCount,
```
mrkaye97, line 2826: drive by: these were ending up as empty strings in the ui

sdks/typescript/src/v1/client/admin.ts
```diff
@@ -22,6 +31,48 @@ import { batch } from '@hatchet/util/batch';
...
+    if (!binData) return '';
+    const status = RpcStatus.decode(binData);
+    for (const detail of status.details) {
+      if (detail.typeUrl.includes('IdempotencyCollisionError')) {
+        return IdempotencyCollisionErrorProto.decode(detail.value).existingRunExternalId ?? '';
+      }
+    }
+  } catch {
+    // ignore decoding errors
+  }
```
mrkaye97, line 118: @grutt do you have any idea how to do this in a less insane way? there's a library in python, but for Ruby + TS I needed to vendor this `RpcStatus` proto and use it here

promptless-for-oss:

Triggered by [PR #4045](https://github.com/hatchet-dev/hatchet/pull/4045)

Created comprehensive documentation for the new idempotency feature including a conceptual guide explaining when to use idempotency, how it works, TTL configuration guidance, collision handling with SDK examples for all four languages (Python, TypeScript, Go, Ruby), and best practices for building effective idempotency expressions.

Review: [Document idempotency feature](https://app.gopromptless.ai/suggestions/46989edd-010b-46d6-b46d-d227ff88fb81?org_id=org_39xXGTrVLocyuD3SpA2P4PdSmX2)

internal/services/admin/v1/admin.go
```diff
@@ -149,31 +150,33 @@ func NewAdminService(fs ...AdminServiceOpt) (AdminService, error) {
...
+	slots := 0
 	if opts.grpcTriggersEnabled {
-		pubBuffer = msgqueue.NewMQPubBuffer(opts.mq)
-
-		tw = trigger.NewTriggerWriter(opts.mq, opts.repo, opts.l, pubBuffer, opts.grpcTriggerSlots, opts.promGate)
+		slots = opts.grpcTriggerSlots
 	}
 
+	pubBuffer = msgqueue.NewMQPubBuffer(opts.mq)
+	tw := trigger.NewTriggerWriter(opts.mq, opts.repo, opts.l, pubBuffer, slots, opts.promGate)
```
mrkaye97: todo: check if these changes are correct - was a weird merge (looks like we also have duped pub buffers now

pkg/repository/sqlcv1/idempotency-keys.sql
```diff
@@ -19,55 +7,57 @@ WHERE
...
-), to_update AS (
-    SELECT
-        ck.tenant_id,
-        ck.expires_at,
-        ck.key,
-        ic.claimed_by_external_id
-    FROM candidate_keys ck
-    JOIN incoming_claims ic ON (ck.key, ck.key_index) = (ic.key, ic.claim_index)
-    WHERE ck.tenant_id = @tenantId::UUID
     FOR UPDATE SKIP LOCKED
```
abelanger5, line 31: `order by` for consistent ordering/avoiding deadlocks?

pkg/repository/idempotency.go
```diff
@@ -2,19 +2,28 @@ package repository
...
 	"github.com/hatchet-dev/hatchet/pkg/repository/sqlcv1"
 )
 
 type WasSuccessfullyClaimed bool
 type IdempotencyKey string
 
+type ClaimIdempotencyKeysOpt struct {
+	Key                 string
+	ClaimedByExternalId uuid.UUID
+	ExpiresAt           pgtype.Timestamptz
```
abelanger5: validation on these opts?

internal/services/controllers/task/trigger/trigger.go
abelanger5, line 1: I saw the comment on idempotency collisions when triggering from events means we drop the triggered task, which makes sense -- but should we link that existing task to the event somehow in the OLAP database? Can imagine that this would be tricky from a debugging perspective.

sql/schema/v1-olap.sql
abelanger5, line 1: should we add the ability to do a reverse search by idempotency key to find a specific task run (or set of task runs)? I'd imagine that would be extremely useful

internal/services/admin/admin.go
```diff
@@ -147,31 +148,33 @@ func NewAdminService(fs ...AdminServiceOpt) (AdminService, error) {
...
 
+	slots := 0
 	if opts.grpcTriggersEnabled {
-		pubBuffer = msgqueue.NewMQPubBuffer(opts.mqv1)
-
-		tw = trigger.NewTriggerWriter(opts.mqv1, opts.repov1, opts.l, pubBuffer, opts.grpcTriggerSlots, opts.promGate)
+		slots = opts.grpcTriggerSlots
 	}
 
+	pubBuffer = msgqueue.NewMQPubBuffer(opts.mqv1)
```
abelanger5, line 158: this looks like a bad merge?

internal/services/admin/server_v1.go
abelanger5, line 1: Not seeing any documentation on how this works with bulk triggering, and admittedly the behavior is a little confusing to me: why does this correspond to returning a gRPC error? The expected behavior for me would be returning the external id of the existing run along with an indication that this was an idempotency key collision. I would also expect the entire batch to be returned successfully, which doesn't seem compatible with raising gRPC errors.

abelanger5 (changes requested):
Overall this looks really good, some questions around API signature and usability from a UI perspective.

sql/schema/v1-olap.sql
mrkaye97: sure, I guess we could tie the key to the `v1_task` -> `v1_runs_olap` to allow for searching?

internal/services/admin/server_v1.go
mrkaye97: I think this is a bit complicated.... If we want to do this in a non-breaking way, we need to return two lists of ids I think, one for the successful triggers, and one for the idempotency violations. But a lot of places on the SDKs rely on ordering / make assumptions that the length of the list of run ids returned by the engine on bulk trigger matches the number of ids / potentially the ordering in the opts, which I think makes this a little tricky.

Maybe I can add a new protobuf message that represents a partial success or something... I'll check on that

mrkaye97: Okay so one thing that seems to work is raising an error that includes both the successful runs and the failures, and then you can catch it and do whatever you like. I think this is maybe the best option, but open to ideas here

internal/services/controllers/task/trigger/trigger.go
mrkaye97: Okay so I did think about this a bit, I'm not sure there's a great way to go. For the tasks it's easier because you'd get the error on trigger, but for events, I guess we could either add an idempotency key to the events table (maybe this would work?), create a new table to store collisions, or actually write the rows into `v1_event_to_run`, which I think doesn't make sense since that'd make it look like multiple events triggered the same run.

Any thoughts there?

mrkaye97:
@abelanger5 responded with a couple questions, and think I got all the other feedback fixed

internal/services/admin/server_v1.go
abelanger5: that sounds reasonable to me, though a little odd

internal/services/controllers/task/trigger/trigger.go
abelanger5: I think we'd need an array of idempotency keys on the events table, but yeah I could see that working!

abelanger5 (approved):

internal/services/controllers/task/trigger/trigger.go
mrkaye97: I think maybe we should not do this until someone asks, since I just foresee us making the wrong call here and needing to change it in the future...

PR #4006, opened 2026-05-25
fix: gate managed worker pages

AVtheking:
# Description

Adds a shared gate for managed worker pages so self-hosted deployments and cloud tenants without the managed worker feature enabled see an explanatory fallback instead of rendering pages that call cloud-only APIs.

Fixes #3896

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation change (pure documentation change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking changes to code which doesn't change any behaviour)
- [ ] CI (any automation pipeline changes)
- [ ] Chore (changes which are not directly related to any business logic)
- [ ] Test changes (add, refactor, improve or change a test)
- [ ] This change requires a documentation update

## What's Changed

- [x] Added a shared `ManagedWorkersGate` for managed worker routes.
- [x] Added fallback messaging for self-hosted deployments where managed workers are unavailable.
- [x] Added fallback messaging for cloud tenants where the managed worker feature flag is disabled.
- [x] Wrapped the managed worker list, create, demo template, and detail pages with the gate.

## Checklist

Changes have been:

- [x] Tested (unit, integration, or manually with steps specified)
- [x] Linted and formatted
- [ ] Documented (where applicable)
- [ ] Added to CHANGELOG (where applicable) -- see [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)

## Testing

- Ran `task fmt-app`
- Checked IDE lints for edited files
- Checked that the managed workers route on a self-hosted deployment shows the unavailable state.
- `pnpm --dir frontend/app run typecheck` fails because `cypress/support/seeded-users.generated` is missing in existing Cypress files, unrelated to this change.

## Screenshots
Screenshot 2026-05-26 at 8 45 23 PM

---

Screenshot 2026-05-26 at 8 45 44 PM

<details id="ai-disclosure">
<summary><b>🤖 AI Disclosure</b></summary>

- [x] _I acknowledge that an LLM was used in the creation of this Pull Request, in accordance with Hatchet's [AI_POLICY.md](./AI_POLICY.md)._

- **Details**: Used autocomplete in Cursor and drafting PR description.

</details>

PR #4075, opened 2026-06-02
fix: block /managed-workers routes in self-hosted Lite mode (#3896)

### LINKED ISSUE #3896: [BUG] [FRONTEND] Invalid access of `/managed-workers` in self-hosted mode
**Describe the issue**

We shouldn't be able to access `/managed-workers` in self-hosted version of the Hatchet app. Instead, when we do, we are just met with an infinite loading screen and the life spammed out of a page that DNE.

Also, subpages still exist so we can attempt to add a managed worker via `/managed-workers/create` 🤷 

**Environment**
- Engine: Hatchet Lite v0.86.8

**Expected behavior**

Return a `404` or a dialogue saying this isn't supported in self-hosted mode.

Build the vite app, then launch hatchet lite i.e

```sh
cd frontend/app && pnpm install && npm run build
LITE_STATIC_ASSET_DIR=./frontend/app/dist LITE_FRONTEND_PORT=5173 task start-lite
```

Then, after seeding etc., accessing `https://app.dev.hatchet-tools.com/tenants/707d0855-80ab-4e1f-a156-f1c4546cbf52/managed-workers` should give you the following:

Image

**Additional context**
Anything else that might help us understand or reproduce the problem — e.g. specific page URL, logged-in vs logged-out, related docs or workflows.

N/A

biswajeetdev:
## What

In self-hosted (Lite) deployments, navigating directly to `/managed-workers` or `/managed-workers/create` shows an infinite loading screen instead of a 404 or an unsupported message.

## Why it happens

The managed-workers page component fires cloud-only queries (billing state, managed-compute costs, payment methods) that never resolve when the cloud metadata endpoint returns an error/404 — which is always the case in Lite mode. The sidebar already hides the "Managed Compute" nav link in Lite mode via the `managedWorkerEnabled` feature flag (which is never set when cloud is disabled), but the routes themselves had no guard, so direct URL access bypassed that check entirely.

## What changed

Added a `requireCloudOrControlPlane()` loader helper in `router.tsx` that checks `isCloudEnabled || isControlPlaneEnabled` (same two signals the rest of the router uses for cloud detection). If neither is true, it throws `notFound()`, which renders the existing `NotFound` component via the nearest `notFoundComponent` handler.

The loader is attached to all four managed-workers routes:
- `managed-workers` (index)
- `managed-workers/create`
- `managed-workers/demo-template`
- `managed-workers/$managedWorker`

This is a single-file, 23-line change — no new components, no new abstractions.

## How to test

1. Run Hatchet in self-hosted (Lite) mode (no cloud metadata endpoint configured).
2. Navigate directly to `http://localhost:<port>/tenants/<tenant-id>/managed-workers`.
3. **Before this fix**: infinite spinner.
4. **After this fix**: 404 page renders immediately.

For cloud/hosted deployments: behavior is unchanged — the loader resolves immediately since `isCloudEnabled` is true.

Fixes #3896

### DIFF
--- a/frontend/app/src/router.tsx
+++ b/frontend/app/src/router.tsx
@@ -17,6 +17,7 @@ import {
   createRoute,
   createRouter,
   lazyRouteComponent,
+  notFound,
   redirect,
 } from '@tanstack/react-router';
 import { Outlet } from '@tanstack/react-router';
@@ -198,6 +199,24 @@ const v1RedirectRoute = createRoute({
     throw redirect({ to: appRoutes.authenticatedRoute.to });
   },
 });
+
+/**
+ * Throws a notFound response when the request originates from a self-hosted
+ * (Lite) deployment that has neither cloud nor control-plane support.
+ * Managed-compute routes are a cloud-only feature; visiting them in Lite mode
+ * produces an infinite loading screen because the billing/cloud queries never
+ * resolve (see GitHub issue hatchet-dev/hatchet#3896).
+ */
+async function requireCloudOrControlPlane() {
+  const [{ isCloudEnabled }, { isControlPlaneEnabled }] = await Promise.all([
+    queryClient.fetchQuery(getCloudMetadataQuery),
+    fetchControlPlaneStatus(),
+  ]);
+
+  if (!isCloudEnabled && !isControlPlaneEnabled) {
+    throw notFound();
+  }
+}
 
 async function getOrganizationIdForTenantInRouter(tenantId: string) {
   const [{ isCloudEnabled }, { isControlPlaneEnabled }] = await Promise.all([
@@ -438,6 +457,7 @@ const tenantWorkerRoute = createRoute({
 const tenantManagedWorkersRoute = createRoute({
   getParentRoute: () => tenantRoute,
   path: 'managed-workers',
+  loader: () => requireCloudOrControlPlane(),
   component: lazyRouteComponent(
     () => import('./pages/main/v1/managed-workers/index.tsx'),
     'default',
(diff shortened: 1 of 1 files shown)


gregfurman (changes requested):
Thanks for the contribution here! This is a really neat solution to the gating problem however, we actually wanted accessing this page to render a messages on whether this is enabled or not. See https://github.com/hatchet-dev/hatchet/pull/4006

Perhaps we retain this utility function `requireCloudOrControlPlane`, place it in a util package/dir (unsure if we already have one somewhere in the `frontend/app/src`), and allow us to gate pages where necessary instead of applying this to the `managed-workers` endpoints within this PR. Wdyt?

gregfurman:
👋 @biswajeetdev Do you still have capacity to address my review?

biswajeetdev:
Apologies for the long silence @gregfurman — that's on me, and thanks for the nudge back in June.

Yes, still interested. Re-reading your review, I want to check I've got the shape right before I push anything, because I think you're asking for something narrower than what I opened with:

1. Keep `requireCloudOrControlPlane`, but lift it out of `router.tsx` into a shared util so it can gate any page, not just managed-workers.
2. **Drop** the managed-workers route wiring from this PR — that's the part you don't want here.
3. Where it is applied, render an explanatory "this is a cloud-only feature" message rather than my `throw notFound()`, which is the behaviour #4006 established.

If that's right, this PR shrinks to essentially just the extracted utility, and the actual gating lands separately. Two things I'd want your call on:

- **Where should it live?** I didn't find an obvious existing util home under `frontend/app/src` — closest candidates looked like `lib/` or `utils/`. Happy to follow whatever convention you'd prefer rather than invent a new directory.
- **Is a util with no callers worth merging on its own?** If #4006 already covers the message-rendering path, this PR may have been overtaken and closing it might be cleaner than reshaping it. I'd genuinely rather close it than leave you carrying a stale PR — no ego about it.

Let me know which way you'd like to go and I'll turn it around.

PR #4077, opened 2026-06-03
fix(client): reconnect durable and workflow run listeners after stream loss

igor-kupczynski:
# Description

The v0 durable/workflow-run listeners could permanently lose or strand their gRPC streams across stream loss, especially during graceful server drains that surface as `io.EOF`. This PR hardens both shared listeners so registered handlers keep the stream alive through EOF, reconnects are coalesced, listener startup does not create duplicate streams, and reconnect attempts do not block normal send/close paths behind `clientMu`.

Fixes # (issue)

<details>
<summary><b>Ported vs. new (for reviewers)</b></summary>

**Ported into `DurableEventsListener` from the `WorkflowRunsListener` hardening in #2934**

- `singleflight`-coalesced reconnect (`reconnectGroup` + `retryListen` / `doRetryListen`, mirroring `retrySubscribe` / `doRetrySubscribe`).
- `generation` counter and `getClientSnapshot`, so `Recv` runs on a snapshot and `clientMu` is never held while blocking.
- `retrySend` comparing generation before/after a failed send and triggering a reconnect instead of retrying on the dead stream.
- Consecutive-error cap and the `codes.Unavailable` 1s backoff in the listen loop.
- nil-safe `Close`.

**New in this PR (applied to both listeners)**

- Lazy, restartable listen loop: `listening` state behind `listenMu`, `ensureListening` / `startListening` / `stopListening`, and `listen()` split out of `Listen()`. The loop starts on the first `AddSignal` / `AddWorkflowRun` and can restart after it exits, instead of leaving a dead listener.
- Internal listen-loop cleanup closes only the stream it owns and clears the shared listener reference only if the stored instance still matches, without terminally closing reusable listener objects.
- The new client is assigned only after subscription replay succeeds, so a failed reconnect can't leave a half-initialized client.
- Generation is tracked inside the listen loop, so an out-of-band reconnect is picked up without forcing another reconnect.
- EOF reconnects when handlers are still registered, so a rolling engine deploy does not strand an already-parked waiter.
- Initial listener startup marks `listening` synchronously before launching the goroutine, avoiding cold-start double-subscribe/leaked stream races.
- Reconnect loops no longer hold `clientMu` across sleeps, stream construction, or subscription replay; the mutex is only held for the final client/generation swap.
- Coalesced reconnects run on a stable background context after caller-context preflight, so one cancelled caller cannot poison the shared reconnect.
- `Workflow.Result()`: when `AddWorkflowRun` kept failing, the retry loop never advanced because `retries` was never incremented, so registration could spin indefinitely; this PR increments `retries` on each failed attempt.

</details>

## Behavioral changes / reviewer notes

- **`io.EOF` handling:** `io.EOF` is terminal for the listen loop only when there are no registered handlers left or the listener is shutting down. If handlers are still registered, EOF triggers reconnect and subscription replay instead of leaving waiters stranded on a dead stream.
- **Durable vs. workflow-run handlers:** durable event handlers are one-shot—they are removed after a successful dispatch. Workflow-run handlers stay registered until explicit removal (e.g. result delivery or cleanup), so they keep the stream subscribed across EOF/reconnect until the run is finished.
- **`AddSignal` / abandoned waiters:** there is no `RemoveSignal` API. A signal waiter that is abandoned without cancellation remains registered and therefore keeps EOF-driven reconnects active. That matches the durable one-shot waiter model (each `AddSignal` corresponds to an active wait); callers that need to drop a waiter should cancel the associated context rather than leaving it registered.

## Type of change

## What's Changed

- [x] Brought `DurableEventsListener` to reconnect parity with `WorkflowRunsListener` (#2934).
- [x] Made both listeners' listen loops lazy and restartable so a dropped stream recovers instead of dying.
- [x] Assigned the reconnected client only after subscription replay succeeds; made `Close` nil-safe.
- [x] Reconnect on `io.EOF` when handlers are still registered.
- [x] Restored shared-listener cleanup on top-level listen exit without terminally closing reusable listener objects.
- [x] Prevented cold-start duplicate streams by marking listeners active before launching the receive goroutine.
- [x] Avoided holding `clientMu` across reconnect backoff/replay work.
- [x] Fixed `Workflow.Result()` infinite retry loop when `AddWorkflowRun` persistently fails (`retries` was never incremented).

## Checklist

Changes have been:

- [x] Tested (unit tests added in `durable_listener_test.go` and `listener_test.go`)
- [x] Linted and formatted

## Testing

`go test ./pkg/client` -- covers broken-stream reconnect, EOF reconnect with registered handlers, listener restart after loop exit, immediate first `Add*` after `getXxx` without opening a second stream, reconnects that do not block client snapshots, the workflow result retry-counter fix, and internal stream cleanup that keeps cached listener pointers reusable.

---

<details id="ai-disclosure">

- **Details**: Cursor was used to help implement the listener reconnect hardening, add related tests, respond to review feedback, and draft/update this PR description.

</details>

### DIFF
--- a/pkg/client/durable_listener.go
+++ b/pkg/client/durable_listener.go
@@ -73,6 +73,14 @@ func (r *subscribeClientImpl) getDurableEventsListener(
 	}
 
 	r.durableEventsListener = w
+	if !w.startListening() {
+		if closeErr := w.Close(); closeErr != nil {
+			r.l.Error().Ctx(ctx).Err(closeErr).Msg("failed to close durable events listener")
+		}
+
+		r.durableEventsListener = nil
+		return nil, errListenerClosed
+	}
 
 	go func() {
 		defer func() {
@@ -88,8 +96,9 @@ func (r *subscribeClientImpl) getDurableEventsListener(
 			}
 			r.durableEventsListenerMu.Unlock()
 		}()
+		defer w.stopListening()
 
-		err := w.Listen(ctx)
+		err := w.listen(ctx)
 
 		if err != nil {
 			r.l.Error().Ctx(ctx).Err(err).Msg("failed to listen for durable events")
--- a/pkg/client/durable_listener.go
+++ b/pkg/client/durable_listener.go
@@ -34,6 +34,7 @@ type DurableEventsListener struct {
 	reconnectGroup singleflight.Group
 	listenMu       sync.Mutex
 	listening      bool
+	closed         bool
 
 	l *zerolog.Logger
 
@@ -99,6 +100,10 @@ func (r *subscribeClientImpl) getDurableEventsListener(
 }
 
 func (w *DurableEventsListener) retryListen(ctx context.Context) error {
+	if w.isClosed() {
+		return errListenerClosed
+	}
+
 	_, err, _ := w.reconnectGroup.Do("reconnect", func() (interface{}, error) {
 		return nil, w.doRetryListen(ctx)
 	})
@@ -108,6 +113,10 @@ func (w *DurableEventsListener) retryListen(ctx context.Context) error {
 func (w *DurableEventsListener) doRetryListen(ctx context.Context) error {
 	w.clientMu.Lock()
 	defer w.clientMu.Unlock()
+
+	if w.isClosed() {
+		return errListenerClosed
+	}
 
 	retries := 0
 
@@ -148,6 +157,10 @@ func (w *DurableEventsListener) doRetryListen(ctx context.Context) error {
 			continue
 		}
 
+		if w.isClosed() {
+			return errListenerClosed
+		}
+
 		w.client = client
 		w.generation++
 		return nil
@@ -168,11 +181,17 @@ func (w *DurableEventsListener) isListening() bool {
 	return w.listening
 }
 
+func (w *DurableEventsListener) isClosed() bool {
+	w.listenMu.Lock()
+	defer w.listenMu.Unlock()
+	return w.closed
+}
+
 func (w *DurableEventsListener) startListening() bool {
 	w.listenMu.Lock()
 	defer w.listenMu.Unlock()
 
-	if w.listening {
+	if w.listening || w.closed {
 		return false
 	}
 
@@ -187,6 +206,10 @@ func (w *DurableEventsListener) stopListening() {
 }
 
 func (w *DurableEventsListener) ensureListening(ctx context.Context) error {
+	if w.isClosed() {
+		return errListenerClosed
+	}
+
 	if w.isListening() {
 		return nil
 	}
@@ -196,6 +219,10 @@ func (w *DurableEventsListener) ensureListening(ctx context.Context) error {
 	}
 
 	if !w.startListening() {
+		if w.isClosed() {
+			return errListenerClosed
+		}
+
 		return nil
 	}
 
@@ -220,6 +247,10 @@ func (l *DurableEventsListener) AddSignal(
 	signalKey string,
 	handler DurableEventHandler,
 ) error {
+	if l.isClosed() {
+		return errListenerClosed
+	}
+
 	t := listenTuple{
 		taskId:    taskId,
 		signalKey: signalKey,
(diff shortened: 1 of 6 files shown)


pkg/client/listener.go
```diff
@@ -136,21 +141,26 @@ func (r *subscribeClientImpl) getWorkflowRunsListener(
...
-			if err != nil {
-				r.l.Error().Err(err).Msg("failed to close workflow run events listener")
-			}
-
 			r.workflowRunListenerMu.Lock()
-			r.workflowRunListener = nil
+			if r.workflowRunListener == w {
+				r.workflowRunListener = nil
+			}
 			r.workflowRunListenerMu.Unlock()
```
igor-kupczynski, line 150: The goroutine now only clears the shared `workflowRunListener` reference. It deliberately does not call  `Close()`, because `Close()` is terminal.

pkg/client/listener.go
```diff
@@ -314,30 +428,60 @@ func (l *WorkflowRunsListener) retrySend(workflowRunId string) error {
...
-	client, _ := l.getClientSnapshot()
+	client, generation := l.getClientSnapshot()
+	if client == nil {
+		return fmt.Errorf("client is not connected")
+	}
+	defer func() {
+		if closeErr := client.CloseSend(); closeErr != nil {
+			l.l.Warn().Err(closeErr).Msg("failed to close workflow run stream after listen exit")
+		}
+	}()
```
igor-kupczynski, line 452: The receive loop closes the exact stream it was reading from, instead of closing `l.client`. This avoids accidentally closing a newer stream if another goroutine restarted the listener after this loop exited.

pkg/client/listener.go
```diff
@@ -362,6 +506,14 @@ func (l *WorkflowRunsListener) Listen(ctx context.Context) error {
 }
 
 func (l *WorkflowRunsListener) Close() error {
+	l.listenMu.Lock()
+	l.closed = true
+	l.listenMu.Unlock()
+
+	return l.closeStream()
+}
```
igor-kupczynski, line 514: `Close()` remains the shutdown path: it marks the listener closed and then closes the current stream. Internal stream cleanup uses `closeStream()` so listener objects can still be reused/restarted.

pkg/client/workflow.go
```diff
@@ -123,6 +123,8 @@ func (c *Workflow) Result() (*WorkflowResult, error) {
 
 			break
 		}
+
+		retries++
```
igor-kupczynski, line 127: Small drive-by fix: this retry loop previously never incremented `retries` after a failed `AddWorkflowRun`, so a permanently failing listener registration could spin forever instead of respecting `DefaultActionListenerRetryCount`.

pkg/client/listener.go
```diff
@@ -314,30 +428,60 @@ func (l *WorkflowRunsListener) retrySend(workflowRunId string) error {
...
+		}
+	}()
 
 	for {
 		event, err := client.Recv()
 
 		if err != nil {
-			if errors.Is(err, io.EOF) || status.Code(err) == codes.Canceled {
+			switch {
+			case errors.Is(err, io.EOF):
```
igor-kupczynski, line 459: ⚠️  Behavior change: `io.EOF` is only treated as terminal when there are no registered handlers or the listener is shutting down. If handlers remain, we reconnect and replay subscriptions.

pkg/client/durable_listener.go
```diff
@@ -166,79 +292,159 @@ func (l *DurableEventsListener) AddSignal(
...
+
+	client, generation := l.getClientSnapshot()
+	if client == nil {
+		return fmt.Errorf("client is not connected")
+	}
+	defer func() {
+		if closeErr := client.CloseSend(); closeErr != nil {
+			l.l.Warn().Err(closeErr).Msg("failed to close durable event stream after listen exit")
+		}
+	}()
```
igor-kupczynski, line 364: Durable listener now mirrors the workflow listener reconnect model: `Recv()` uses a client/generation snapshot, reconnects replay registered durable-event handlers, and the loop closes only the stream it owns on exit so concurrent restarts do not get their newer stream closed.

abelanger5 (approved):

sdks/go/e2e/durable_test.go
```diff
@@ -114,6 +114,8 @@ func TestDurableSleepCancelReplay(t *testing.T) {
 	})
 	require.NoError(t, err)
 
+	pollUntilRunStatus(t, ctx, sharedClient, ref.RunId, string(rest.V1TaskStatusRUNNING))
```
igor-kupczynski, line 117: This is a drive by fix, I've noticed a flaky test in a related area, and wanted to double check if this is not related to the core change. (Looks like it's just a race, other assertions in the file poll first, and this was not).

PR #4085, opened 2026-06-04
fix(ci): Rework release pipeline for tags

gregfurman:
# Description

Fixes a race condition where the `pre-release.yaml` pipeline has not completed before the `release.yaml` is triggered -- causing flakes on `docker pull` for the intermediate images we want to promote to `latest`

Refactors to include more matrix usage for image build jobs, since there is a lot of repetition.

## Type of change

## What's Changed

- Fixes `release.yaml` to only run after successful completion of `pre-release.yaml` workflow.
- Refactors `release.yaml` and `pre-release.yaml` to use `matrix` strategy to keep things DRY.

## Checklist

Changes have been:

- [ ] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Claude web app to automate each job to leverage matrix approach in `pre-release.yml`.

</details>

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -10,9 +10,34 @@ on:
         required: true
 name: Release
 jobs:
+  verify:
+      runs-on: ubuntu-latest
+      if: ${{ github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success' }}
+      outputs:
+        tag: ${{ steps.tag_name.outputs.tag }}
+        should_publish: ${{ steps.skip.outputs.should_publish }}
+      steps:
+        - id: tag_name
+          run: |
+            if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+              echo "tag=${{ github.event.inputs.tag }}" >> "$GITHUB_OUTPUT"
+            else
+              echo "tag=${{ github.event.workflow_run.head_branch }}" >> "$GITHUB_OUTPUT"
+            fi
+
+        - id: skip
+          run: |
+            tag="${{ steps.tag_name.outputs.tag }}"
+            if echo "$tag" | grep -qE '^v[0-9]+\.[0-9]+\.[0-9]+$'; then
+              echo "should_publish=true" >> "$GITHUB_OUTPUT"
+            else
+              echo "should_publish=false" >> "$GITHUB_OUTPUT"
+            fi
+
   load:
-    if: ${{ github.event.workflow_run.conclusion == 'success' }}
     runs-on: ubicloud-standard-8
+    needs: [verify]
+    if: needs.verify.outputs.should_publish == 'true'
     timeout-minutes: 30
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
BloggerBust: I think this tag extraction is broken for the new `workflow_run`. In a workflow triggered by `workflow_run`, `github.ref` points to the default branch rather than the tag that triggered the upstream workflow.

See: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run

That means this may resolve to something like `refs/heads/main` instead of `v0.88.0`. If that happens, the Docker promotion steps would try to pull and push tags like `refs/heads/main-amd64`.

I think the tag needs to come from the `workflow_run` event instead of `github.ref`, likely `github.event.workflow_run.head_branch`.

.github/workflows/release.yaml
```diff
@@ -11,6 +11,7 @@ on:
 name: Release
 jobs:
   load:
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
BloggerBust: I think this `if` condition breaks the manual `workflow_dispatch`. For manual runs there is no `workflow_run` event, so `github.event.workflow_run.conclusion == 'success'` should evaluate false and skip the job.

This applies to both `load` and `push-hatchet-server`. Should this allow either manual dispatch or successful prerelease completion?

```yaml
if: ${{ github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success' }}
```

.github/workflows/release.yaml
BloggerBust, line 50: For the new `workflow_run` path, should this checkout use the pre-release run’s commit?

Without an explicit `ref`, `actions/checkout` uses the SHA for the event that triggered this workflow. For `workflow_run`, that appears to be the default branch context, not necessarily the tag commit that `pre-release.yaml` ran against.

I think this should probably check out: `github.event.workflow_run.head_sha`.

.github/workflows/release.yaml
```diff
@@ -1,8 +1,8 @@
 on:
-  push:
-    tags:
-      - "v*" # Trigger on version tags like v0.73.10
-      - "!v*-alpha*" # Do not trigger for alpha releases
```
BloggerBust, line 5: The old `release.yaml` trigger explicitly excluded `v*-alpha*` tags. Since `pre-release.yaml` still runs on `v*`, does this change mean successful alpha prereleases will now trigger `release.yaml` and promote alpha images to `latest`?

BloggerBust (changes requested):
Hey Greg, thanks for working on this. The `workflow_run` sequencing looks right for fixing the release race, but I found a few workflow issues that I think should be addressed before merge.

I am happy to re-review after updates.

.github/workflows/release.yaml
gregfurman: Ahh good catch. Let me rework this so that both of these jobs require a single job that outputs the tag correctly

.github/workflows/release.yaml
gregfurman: Apparently github.event.workflow_run.head_branch is the solution to this: https://github.com/orgs/community/discussions/27124#discussioncomment-3254715

.github/workflows/release.yaml
gregfurman: Should be addressed. There's a `skip` step that indicates whether the rest of the jobs can continue running (which is `false` if the tag is not semver)

.github/workflows/release.yaml
gregfurman: Should be addressed now! Uses the tag from the `verify` job

BloggerBust:
Thank you Greg, all issues have been resolved.

@grutt  - This looks good to me now.

mrkaye97 (approved):

PR #4089, opened 2026-06-05
fix: Fail fast when creating partitions concurrently with analyze

juliusgeo:
# Description

We have run into issues where partition creation queries will timeout because they are ran at the same time that an ANALYZE is running (due to the analyzeTask and partitionTask running on similar intervals). This PR changes it so that during partition creation we will fail fast if a lock is held on the relevant table, so that it can be retried 15 mins later.

Fixes # (issue)

## Type of change

## What's Changed

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

Claude code, generating tests and research

</details>


pkg/repository/olap.go
```diff
@@ -350,6 +350,38 @@ func newOLAPRepository(shared *sharedRepository, olapRetentionPeriod time.Durati
...
 }
 
+// runPartitionDDLWithLockTimeout wraps fn in a transaction with a short lock_timeout so that DDL
+// fails fast with ErrPartitionLockConflict instead of blocking if ANALYZE (or another DDL) holds a
+// conflicting ShareUpdateExclusiveLock on the parent table.
+
+// Only CREATE TABLE / ALTER TABLE ATTACH PARTITION may be passed in fn. DETACH PARTITION
+// CONCURRENTLY cannot run inside a transaction and must use a raw connection instead.
+func runPartitionDDLWithLockTimeout(ctx context.Context, pool *pgxpool.Pool, fn func(tx pgx.Tx) error) error {
+	tx, err := pool.Begin(ctx)
```
mrkaye97: I think we'll probably want to use the `sqlchelpers.PrepareTx` here? then we can `defer rollback()`

pkg/repository/olap.go
```diff
@@ -474,14 +491,34 @@ func (r *OLAPRepositoryImpl) UpdateTablePartitions(ctx context.Context) error {
...
 		}
 
+		// releaseConn resets lock_timeout to 0 before returning the connection to the pool so
+		// the setting doesn't bleed into subsequent users of the same ddlPool connection.
+		releaseConn := func() {
+			resetCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+			defer cancel()
+			if _, resetErr := conn.Exec(resetCtx, "SET lock_timeout = 0"); resetErr != nil {
+				r.l.Error().Err(resetErr).Msg("failed to reset lock_timeout on DDL connection")
+			}
```
mrkaye97, line 495: I think this is related to an open issue / PR?

pkg/repository/olap_partition_test.go
```diff
@@ -229,3 +251,248 @@ func TestOLAPUpdateTablePartitions_LockContention(t *testing.T) {
...
+// The migration creates today's partitions, so CreateOLAPPartitions(today) is always a no-op.
+// CreateOLAPPartitions(tomorrow) runs on r.pool and tries to ATTACH PARTITION — the lock blocks it.
+func TestOLAPUpdateTablePartitions_FailsFastDuringAnalyze(t *testing.T) {
+	pool, cleanup := setupPostgresWithMigration(t)
+	defer cleanup()
+
+	ctx := context.Background()
+
+	// ddlPool is separate from pool so DDL connections and lock-holder connections never compete
+	// for pool slots.
```
mrkaye97: I'm not too sure, but I'm not convinced these tests are worth maintaining - a regression here is far from catastrophic and it's just more surface area to maintain if we want to change things. WDYT?

pkg/repository/task.go
```diff
@@ -24,6 +25,18 @@ import (
 	"github.com/hatchet-dev/hatchet/pkg/telemetry"
 )
 
+// ErrPartitionLockConflict is returned when a partition DDL operation cannot acquire the necessary
+// table lock because another operation (e.g. ANALYZE) is holding a conflicting lock. The caller
+// should treat this as a transient condition and retry later.
```
mrkaye97: let's remove comments like these, the string on the line below this is more than enough imo

pkg/repository/task.go
```diff
@@ -24,6 +25,18 @@ import (
...
+var ErrPartitionLockConflict = errors.New("partition DDL could not acquire lock due to concurrent table operation")
+
+// isLockNotAvailable reports whether err is a PostgreSQL lock_not_available error (SQLSTATE 55P03),
+// which is raised when a lock_timeout fires.
+func isLockNotAvailable(err error) bool {
+	var pgErr *pgconn.PgError
+	return errors.As(err, &pgErr) && pgErr.Code == "55P03"
```
mrkaye97: We should use `pgerrcode.LockNotAvailable` here I think

mrkaye97 (commented):
Left a couple comments!

pkg/repository/olap_partition_test.go
juliusgeo: That's fair, I mainly wanted to add them in to validate that there aren't issues with setting the lock timeout with the ddlPool--but now that that is validated I'm fine with removing them.

pkg/repository/olap.go
juliusgeo: Hmm, not able to find one that seems related--do you have a link?

mrkaye97: I think this! https://github.com/hatchet-dev/hatchet/issues/3898

pkg/repository/task.go
juliusgeo: fixed!

pkg/repository/olap.go
juliusgeo: Yep, that makes sense--fixed!

pkg/repository/olap.go
juliusgeo: I think we should be fine here because we're using LOCAL

pkg/repository/task.go
juliusgeo: fixed!

mrkaye97 (approved):

PR #4105, opened 2026-06-06
fix(api): add cron job to clean up UserSession

NirajNair:
# Description

Introduces cron scheduler in the API layer and a cron job that runs every hour deleting expired and orphaned `UserSession` rows addressing the unbounded table growth issue reported in #3913.

Based on the ~1 row/sec insertion rate reported in the issue, approx. 3600 rows accumulate per hour - hence choosing 1-hour cron interval that balances timely cleanup.

As per this [comment](https://github.com/hatchet-dev/hatchet/pull/3923#issuecomment-4462998355), `UserSession` rows are deleted when:
- `expiresAt < NOW()`  OR 
- `userId = NULL AND createdAt < NOW() - 24 hours`

Follow up fix for issue #3913

## Type of change

## What's Changed

- [x] Added a cron scheduler and a job in API layer that runs every 1 hour to delete expired and orphaned UserSession

## Checklist

- [x] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: I've used Pi coding agent (model: Minimax M2.7) to understand repo structure, code patterns, compare and align implementation with existing cron job patterns, verify naming conventions, and generate tests.

</details>


pkg/repository/user_session_test.go
```diff
@@ -0,0 +1,251 @@
...
+	var count int
+	err := pool.QueryRow(ctx(t), `SELECT COUNT(*) FROM "UserSession" WHERE "id" = ANY($1::uuid[])`, sessionIds).Scan(&count)
+	require.NoError(t, err)
+
+	return count
+}
+
+// TestCleanupUserSessions verifies the cleanup logic handles both conditions:
+// 1. Expired sessions (expiresAt < NOW())
+// 2. Unauthenticated sessions (userId IS NULL) older than 24 hours
```
gregfurman: Can we remove all of this inline documentation?

pkg/repository/user_session_test.go
```diff
@@ -0,0 +1,251 @@
...
+		ddlPool: pool,
+		l:       &logger,
+		queries: sqlcv1.New(),
+	}
+	return &userSessionRepository{
+		sharedRepository: shared,
+	}
+}
+
+func createTestUser(t *testing.T, pool *pgxpool.Pool) uuid.UUID {
```
gregfurman, line 31: nit Shouldn't these helper functions include a `t.Helper()`?

api/v1/server/cron/cleanup.go
gregfurman, line 1: Q: I'm wondering if this implementation should instead be in `internal/services/controllers/retention` instead of at the `api` layer -- especially since this would allow cleanups to respect a users data retention settings. I'm not sold on either approach so would be curious to hear your thoughts!

cc @abelanger5

pkg/repository/user_session.go
```diff
@@ -37,6 +37,8 @@ type UserSessionRepository interface {
 	Update(ctx context.Context, sessionId uuid.UUID, opts *UpdateSessionOpts) (*sqlcv1.UserSession, error)
 	Delete(ctx context.Context, sessionId uuid.UUID) (*sqlcv1.UserSession, error)
 	GetById(ctx context.Context, sessionId uuid.UUID) (*sqlcv1.UserSession, error)
+
+	CleanupUserSessions(ctx context.Context) (int, error)
```
gregfurman: Unsure we really need the no. of sessions cleaned up as a return type. Is this just for logging?

pkg/repository/user_session.go
```diff
@@ -165,3 +167,23 @@ func (r *userSessionRepository) GetById(ctx context.Context, sessionId uuid.UUID
...
+
+func (r *userSessionRepository) CleanupUserSessions(ctx context.Context) (int, error) {
+	const batchSize int32 = 1000
+	var totalDeleted int
+
+	for {
+		result, err := r.queries.CleanupUserSessions(ctx, r.pool, batchSize)
+		if err != nil {
+			return totalDeleted, err
+		}
```
gregfurman, line 183: If no sessions have expired (and no data is found), we're going to be returning a false positive `pgx.ErrNoRows` type -- which would be logged as an error etc.

Think we should be guarding against this with a `errors.Is(err, pgx.ErrNoRows)` check that returns a `nil` error if true.

api/v1/server/cron/cleanup.go
```diff
@@ -0,0 +1,25 @@
...
+		if err != nil {
+			if deletedCount > 0 {
+				c.l.Err(err).Ctx(ctx).Int("deleted_count", deletedCount).Msg("user sessions cleanup cron job failed partially")
+			} else {
+				c.l.Err(err).Ctx(ctx).Msg("user sessions cleanup cron job failed")
+			}
+			return
+		}
+
+		c.l.Info().Ctx(ctx).Int("deleted_count", deletedCount).Msg("user sessions cleanup cron job completed")
```
gregfurman: IMO we don't need to log the success case -- especially at the INFO level.

api/v1/server/cron/scheduler.go
```diff
@@ -0,0 +1,62 @@
...
+
+	return &CronScheduler{
+		s:      s,
+		l:      c.Logger,
+		repoV1: c.V1,
+	}, nil
+}
+
+func (c *CronScheduler) Start() (func() error, error) {
+	ctx, cancel := context.WithCancel(context.Background())
```
gregfurman: nit: we could have a `var err error` defined with a `defer` e.g.

```golang
var err error

defer func() {
   if err != nil {
      cancel()
   }
}()
```

That way, we automatically cancel the operation if err an error is returned.

gregfurman: super-nit: Although we don't do this anywhere else, could be a nice opportunity to use a `context.WithCancelCause` [function](https://pkg.go.dev/context#WithCancelCause) and pass the err as the reason 👀

pkg/repository/user_session_test.go
```diff
@@ -0,0 +1,251 @@
...
+		if tc.shouldDelete {
+			assert.False(t, exists, "session '%s' should be deleted from DB", tc.name)
+		} else {
+			assert.True(t, exists, "session '%s' should exist in DB", tc.name)
+		}
+	}
+}
+
+// TestCleanupUserSessions_MultiBatch verifies cleanup works correctly when
+// there are more rows than the batch size (1000), forcing multiple batches.
```
gregfurman: Also these comments. They really just pollute the diff

pkg/repository/sqlcv1/users.sql
```diff
@@ -154,3 +154,19 @@ DELETE FROM
...
+      "expiresAt" < NOW()
+      OR (
+          "userId" IS NULL
+          AND "createdAt" < NOW() - INTERVAL '24 hours'
+      )
+   ORDER BY "createdAt" ASC
+   LIMIT @batchSize::int
+)
+DELETE FROM "UserSession"
+WHERE "id" IN (SELECT "id" FROM sessions_to_delete);
```
gregfurman: Why can we not just do a `DELETE FROM`? Why do we need to collect the sessions beforehand?

gregfurman: Why can we not just do a `DELETE FROM` with the above `WHERE` clause? Is there a reason we need to collect the sessions beforehand?

gregfurman (changes requested):

pkg/repository/sqlcv1/users.sql
NirajNair: Cleanup cron deletes rows in batches of 1000 and for that we need LIMIT clause but DELETE (similar to UPDATE) does not support LIMIT clause ([PostgresSQL Docs](https://www.postgresql.org/docs/current/sql-delete.html?utm_source=chatgpt.com#:~:text=While%20there%20is%20no%20LIMIT%20clause%20for%20DELETE%2C%20it%20is%20possible%20to%20get%20a%20similar%20effect%20using%20the%20same%20method%20described%20in%20the%20documentation%20of%20UPDATE%3A)). I believe ORDER BY is not supported either as I couldn't find it in the synopsis of both operations.

PostgresSQL docs recommend to use CTE for batch deletes/updates. Also, this query pattern already exists in the repo, for eg. in `CleanupV1QueueItem` query.

One improvement that we can do:
change:
`DELETE FROM "UserSession"
WHERE "id" IN (SELECT "id" FROM sessions_to_delete);`

to:
`DELETE FROM "UserSession" AS us
USING sessions_to_delete AS s
WHERE us."id" = s."id";`

NirajNair: https://github.com/hatchet-dev/hatchet/pull/4105/changes/BASE..44f19ad5a5aa079d1a6a423faac2527eea7d0d4d#r3379251867

api/v1/server/cron/cleanup.go
NirajNair: Sure. 

Q - Should we log deleted count in L16? If not then we can remove the `int` return type mentioned in this [comment](https://github.com/hatchet-dev/hatchet/pull/4105/changes/BASE..44f19ad5a5aa079d1a6a423faac2527eea7d0d4d#r3374337073)

pkg/repository/user_session.go
NirajNair: Yes, its just for logging.

gregfurman: If it's just for logging then I think it's safe to just replace it with a single `error` type.

pkg/repository/user_session_test.go
NirajNair: yes, will add it.

pkg/repository/sqlcv1/users.sql
gregfurman: Duplicate sorry!

pkg/repository/sqlcv1/users.sql
```diff
@@ -154,3 +154,19 @@ DELETE FROM
...
+   SELECT "id"
+   FROM "UserSession"
+   WHERE
+      "expiresAt" < NOW()
+      OR (
+          "userId" IS NULL
+          AND "createdAt" < NOW() - INTERVAL '24 hours'
+      )
+   ORDER BY "createdAt" ASC
+   LIMIT @batchSize::int
```
gregfurman: Q: Wonder if we should be using a `FOR UPDATE SKIP LOCKED` here to prevent data races 🤔 Or is it fine since we don't expect to be concurrently writing to the same rows only reading?

pkg/repository/sqlcv1/users.sql
gregfurman: Thanks for clarifying! Yeah the logic does make sense. I suppose I'm just wondering if this batching is really solving the problem that we think it is.

Like, my main concern is that we're only ever really expect a very large cleanup call on first run, since users may have accumulated millions of rows. Unless someone is _really_ abusing new session creation, I wouldn't expect anywhere close to this number following the previous fix + introduction of cleanups.

From the original issue:

> In our self-hosted Hatchet deployment, the UserSession table in the Hatchet Postgres instance has grown to 2.6M rows (~1 GB) over ~30 days. 
>```
> SELECT count(*) FILTER (WHERE "userId" IS NULL), count(*) FILTER (WHERE "userId" IS NOT NULL)
> FROM "UserSession";
> -- 2,604,487 anonymous   |   9 with userId (4 real users)
> ```

Perhaps better put: if someone has accumulated millions of rows (worst case), I'm unsure a batch size of 1000 is going to cut it. Otherwise, I see the average and best cases as being very similar in magnitude --that is, 1000 could be a HUGE overshoot when we only ever expect a couple of dozen rows.

If we assume this is the average case, that'll mean 2600 individual calls being made within a loop that doesn't really have backoff or jittering between each operation.

Like, at this point it's probably cheaper to just re-create the entire table or maybe just accept that the first call is going to be expensive.

Maybe it could be worthwhile to include a DB migration that recreates the table with valid session keys instead of relying on the cron approach exclusively.

@abelanger5 would love your input here!

gregfurman (changes requested):

pkg/repository/sqlcv1/users.sql
NirajNair: I don't think it will cause any data races but using `FOR UPDATE SKIP LOCKED` will make the job efficient by letting other API instances `SELECT` rows that are not already locked by some other instance.

pkg/repository/sqlcv1/users.sql
gregfurman: OK chatted internally. Think this approach as-is is fine. However, we should add something to the release notes/changelog (in `CHANGELOG.md`) about the drop invalid approach since there are currently users with a large backlog of undeleted rows.

Do you have capacity to add a section to the CHANGELOG.md on this? I'm thinking something like:

````md
## [Unreleased]

### Upgrade Notes

* This update includes an additional cleanup job for the `UserSessions` table to correctly remove expired/invalid user sessions from the table -- addressing unbounded growth. However, we recommend running the following query against your Postgres instance to ensure a bulk-cleanup:

```sql
-- creates a tmp_UserSessions from only valid UserSession rows, 
-- drops the whole UserSessions table, renames the tmp_UserSessions
-- to UserSessions.
```

````

(`Unreleased` may not exist... if it doesn't we should add it)

File: api/v1/server/cron/cleanup.go
Comment (gregfurman) [reply api/v1/server/cron/cleanup.go:1]: I've given this some more thought and I think we should be adding this as a cleanup to the `internal/services/controllers/retention` instead. Wdyt? @NirajNair

**Review by gregfurman — CHANGES_REQUESTED** [review]
Hey @NirajNair Have left 2 more comments. I think this should move to the `internal/services/controllers/retention` since that means we can control it via the user-configured data retention settings. Also, we should definitely be adding something to the `CHANGELOG.md` on this.

Looking forward to getting this merged 🙌

File: pkg/repository/sqlcv1/users.sql (hunk 8 above)
Comment (NirajNair) [reply pkg/repository/sqlcv1/users.sql]: Sure, will update CHANGELOG.md.

Quick Q: should we also have an `Added` section under `Unreleased`?

File: api/v1/server/cron/cleanup.go
Comment (NirajNair) [reply api/v1/server/cron/cleanup.go:1]: I agree because I think `internal/services/controllers/retention` already has a cron scheduler and the tasks that it runs looks like maintenance tasks, similar to this one. Also, we won't be introducing a new scheduler in the API layer at all if we move.

File: internal/services/controllers/retention/controller.go [hunk 10]
```diff
@@ -141,44 +144,50 @@ func New(fs ...RetentionControllerOpt) (*RetentionControllerImpl, error) {
...
+		userSessionRetention: opts.userSessionRetention,
 	}, nil
 }
 
 func (rc *RetentionControllerImpl) Start() (func() error, error) {
 	rc.l.Debug().Msg("starting retention controller")
 
-	ctx, cancel := context.WithCancel(context.Background())
+	var err error
+	ctx, cancel := context.WithCancelCause(context.Background())
```
gregfurman, line 165: praise: Beautiful 😎

internal/services/controllers/retention/controller.go
```diff
@@ -187,15 +196,27 @@ func (rc *RetentionControllerImpl) Start() (func() error, error) {
...
 
 		if err != nil {
-			cancel()
 			return nil, fmt.Errorf("could not set up runCleanupOldWorkers: %w", err)
 		}
 	}
 
+	if rc.userSessionRetention {
+		_, err = rc.s.NewJob(
+			gocron.DurationJob(1*time.Hour),
```
gregfurman: nit: the other definitions have this duration value as a separate variable, which IMO is more readable i.e

```go
userSessionInterval := 1*time.Hour

_, err = rc.s.NewJob(
	gocron.DurationJob(userSessionInterval),
```

CHANGELOG.md
```diff
@@ -1,3 +1,32 @@
...
+DROP TABLE "UserSession";
+
+ALTER TABLE tmp_UserSession RENAME TO "UserSession";
+
+COMMIT;
+```
+
+### Added
+-  Scheduled user session cleanup via retention controller by @NirajNair in [#4105](https://github.com/hatchet-dev/hatchet/pull/4105)
+
```
gregfurman:
```suggestion
```

Sorry the changelog stuff is quite cryptic. This section will automatically be appended on when we generate the release notes (see https://github.com/hatchet-dev/hatchet/releases/tag/v0.89.0). I'm more concerned with the human-written prose beforehand (which you added above 🙏).

CHANGELOG.md
```diff
@@ -1,3 +1,32 @@
+## [Unreleased]
+
+### Upgrade Notes
+
+* This update includes an additional cleanup job for the `UserSession` table to correctly remove expired/invalid user sessions from the table -- addressing unbounded growth. However, we recommend running the following query against your Postgres instance to ensure a bulk-cleanup:
+
+```sql
```
gregfurman, line 7: @mrkaye97 Can you give this a look-over? Seems fine to me but would like to make sure!

pkg/repository/user_session_test.go
```diff
@@ -0,0 +1,217 @@
...
+	pool, cleanup := setupPostgresWithMigration(t)
+	defer cleanup()
+
+	testUserId := createTestUser(t, pool)
+	repo := createUserSessionRepository(pool)
+
+	err := repo.CleanupUserSessions(ctx(t))
+	require.NoError(t, err)
+
+	testCases := []struct {
```
gregfurman, line 73: praise: Loving the usage of testing tables

gregfurman (commented):
This is looking great! Left some non-blocking comments. Only actionable thing I _really_ care about from your side is removing that `Added` section in the `CHANGELOG.md` (see my comment).

Otherwise, just need an extra pair of eyes on the advertised SQL query and then this looks good to 🚢

CHANGELOG.md
NirajNair: I'm sorry I wasn't aware of that. Will remove it.

NirajNair:
I've removed `Added` from CHANGELOG.md and created a new variable `userSessionInterval` in controller.go as suggested.

gregfurman (approved):
LGTM! Thanks for addressing my comments 😄

PR #4117, opened 2026-06-08
fix(ci): Incorrect engine labeling

gregfurman:
# Description

The negation in the `.github/labeler.yml` has caused introduced a false positive that labels anything not matching the negation as `engine`. This PR aims to fix this by ensuring we correctly skip the `.goreleaser.yml` file.

## Type of change

## What's Changed

- Fixes negation to ensure we only label as `engine` when the following conditions are BOTH true:
   - Any file within the following is changed `["cmd/**", "pkg/**", "internal/**", "api/**",  "sql/**"]`
   - At least one changed file is NOT `cmd/hatchet-cli/.goreleaser.yaml`

## Checklist

Changes have been:

- [ ] Linted and formatted

## Testing
- Tested on my local fork -- where just a change to `sdks/python` was made and labeled correctly in PR https://github.com/gregfurman/hatchet/pull/289

---

<details id="ai-disclosure">

</details>

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -27,7 +27,6 @@ github_actions:
 
 engine:
   - changed-files:
-      - all-globs-to-all-files: "!cmd/hatchet-cli/.goreleaser.yaml"
       # TODO: Disambiguating between components is non-trivial,
       # For now, we rely on a mixture of files changed + title scoping.
       - any-glob-to-any-file:
@@ -36,6 +35,7 @@ engine:
           - "internal/**"
           - "api/**"
           - "sql/**"
+      - all-globs-to-any-file: "!cmd/hatchet-cli/.goreleaser.yaml"
 
 dashboard:
   - changed-files:
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -26,6 +26,10 @@ github_actions:
         - ".github/labeler.yml"
 
 engine:
+  # ensures that ALL of the provided options must match for the label to be applied
+- all:
+  # NOTE: Linting fails in vscode due to the JSON schema at https://www.schemastore.org/pull-request-labeler.json
+  # not being applied correctly via the redhat YAML extension. See https://github.com/actions/labeler/issues/746
   - changed-files:
       # TODO: Disambiguating between components is non-trivial,
       # For now, we rely on a mixture of files changed + title scoping.
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -35,7 +35,8 @@ engine:
           - "internal/**"
           - "api/**"
           - "sql/**"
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -36,6 +35,8 @@ engine:
           - "internal/**"
           - "api/**"
           - "sql/**"
+      - all-globs-to-all-files:
+        - "!cmd/hatchet-cli/.goreleaser.yml"
```
igor-kupczynski, line 43: I could be misunderstanding this, but the default here would be `any`   ([docs](https://github.com/actions/labeler/tree/main#:~:text=base%20option%20is-,provided,-without%20a%20top)), so:
- A PR that only touches `sdks/python/**` can still get engine, because `all-globs-to-all-files`: `'!cmd/hatchet-cli/.goreleaser.yml'` is satisfied even when no engine paths changed.
- A PR that only changes `cmd/hatchet-cli/.goreleaser.yml` can still get engine, because `cmd/**` matches via `any-glob-to-any-file`.

We'd probably want `all`, e.g.

```
engine:
  - all:   # <--- here 
    - changed-files:
      - any-glob-to-any-file:
          - "cmd/**"
          - "pkg/**"
          - "internal/**"
          - "api/**"
          - "sql/**"
      - all-globs-to-all-files:
        - "!cmd/hatchet-cli/.goreleaser.yml"
```

(In the docs linked about they have an example titled `# Add 'source' label to any change to src files within the source dir EXCEPT for the docs sub-folder`).

WDYT?

gregfurman: > If a base option is provided without a top-level key, then it will default to `any`. More specifically, the following two configurations are equivalent:

Ahhh I missed this. What ridiculous syntax... thanks for the good catch!

igor-kupczynski: Agree it's ridiculous

igor-kupczynski (approved):
LGTM 🚢🇮🇹

gregfurman:
I think GH actions is having a moment? Doesn't seem to be running actions against my branch, instead seemingly checking out `9423e2a` (see https://github.com/hatchet-dev/hatchet/actions/runs/27200617845/job/80303400908?pr=4117#step:2:95)

igor-kupczynski:
The error now seems to be:
```
golangci-lint............................................................Failed
- hook id: golangci-lint
- exit code: 3

Error: can't load config: the Go language version (go1.25) used to build golangci-lint is lower than the targeted Go version (1.26)
The command is terminated due to an error: can't load config: the Go language version (go1.25) used to build golangci-lint is lower than the targeted Go version (1.26)
```

Which is probably unrelated.

gregfurman:
I think this is a GH actions problem. Seems the incorrect `github.ref` is being passed in 🤷

PR #4132, opened 2026-06-09
fix(engine): exclude evicted runtimes from timeouts

### LINKED ISSUE #4128: [BUG] Durable suspended (evicted) task wrongly timed-out → false TIMED_OUT + duplicate run
**Describe the issue**

A durable (suspended/evicted) runtime is excluded from the **reassign** loop but **not** from the **timeout** loop. If a durable task parks longer than its `step_timeout`, it is failed and retried while the original may still be restored — leading to a **false TIMED_OUT** and a **duplicate run** of the same step.

This is correctness finding **F3** of an audit shared with @abelanger5 by email on 2026-06-09. Filed as a public issue at maintainer request (advisory route reserved for F1/F2 cross-tenant). Reference advisory: GHSA-r7w5-4j6c-jxxc.

**Environment**
- SDK: any (server-side bug, SDK-agnostic)
- Engine: Self-hosted, commit `f64a6a9` (HEAD at time of review). Same v1 engine path on Cloud.

**Expected behavior**

An evicted durable task should be skipped by `ListTasksToTimeout` while it is parked, mirroring `ListTasksToReassign`'s `evicted_at IS NULL` guard. Only after the durable event fires and the runtime is restored should the timeout countdown resume.

**Actual behavior**

The evicted row stays in `ListTasksToTimeout`'s result set (no `evicted_at IS NULL` guard) and `timeout_at` is not nulled by `EvictTask`. Once parked beyond `step_timeout`:
1. `ListTasksToTimeout` matches the parked row.
2. `FailTaskAppFailure` bumps `retry_count`, `releaseTasks` deletes the runtime.
3. A fresh attempt is queued — the *second* run.
4. When the durable event eventually fires, `RestoreEvictedTasks` filters `WHERE r.evicted_at IS NOT NULL` and finds nothing → silent no-op.

Net effect: status is `TIMED_OUT` even though the step was simply parked, and the side effects of the step are executed twice.

Cross-checked across four queries (file:line at commit `f64a6a9`):

- `pkg/repository/sqlcv1/tasks.sql:1013-1031` — `EvictTask` sets `evicted_at = NOW(), worker_id = NULL` and (per its own comment, `:1015-1016`) skips rows whose timeout already passed — but it leaves `timeout_at` unchanged on the rows it does evict.
- `pkg/repository/sqlcv1/tasks.sql:390-392` — `ListTasksToTimeout` selects on `timeout_at <= NOW()` with **no `evicted_at IS NULL` guard**.
- `pkg/repository/sqlcv1/tasks.sql:431` — `ListTasksToReassign` **does** have `AND runtime.evicted_at IS NULL` — the safe sibling that shows the intended invariant the timeout query omits.
- `pkg/repository/sqlcv1/tasks-overwrite.go:858-862` — `RestoreEvictedTasks` filters `WHERE r.evicted_at IS NOT NULL`; once the timeout loop's `releaseTasks` deletes the runtime, restore silently no-ops.
- `pkg/worker/context.go:527` — `RefreshTimeout` is only callable by the running worker; nothing refreshes `timeout_at` while parked.

**Confidence:** confirmed at code level across the four queries; the exact wall-clock interleaving (parked > `step_timeout`) wants a reproduction test — happy to write one if useful.

**Suggested fix**

Either of these (option A is the minimal patch):

A. Add `AND runtime.evicted_at IS NULL` to `ListTasksToTimeout`'s `expired_runtimes` CTE — mirrors the `ListTasksToReassign` invariant and is a one-line change.

B. Null `timeout_at` in `EvictTask` and restore it on `RestoreEvictedTasks` — slightly more state to maintain but preserves the meaning of `timeout_at` as "absolute deadline".

I'd lean A: it's the smaller change, it makes the two control-loop queries symmetric, and it leaves `timeout_at` semantics untouched.

**Additional context**

This was found via manual line-by-line reading of the v1 engine (no SAST, no Snyk) using a **proof by contrast** method — each finding is cross-checked against a sibling path in the same repo that does the right thing. Here the sibling that gets it right is `ListTasksToReassign`'s `evicted_at IS NULL` clause — the very pattern missing in `ListTasksToTimeout`.

Reported via Orionix · m-be290108 on GitHub · @orionix_ai · fiverr.com/mathieuballot

---

<details id="ai-disclosure">

- **Details**: Claude (Anthropic) was used as a research and writing assistant during the manual code audit. All file:line citations, the proof-by-contrast reasoning, and the suggested fix were verified by direct reading of the cited source in this repo at commit `f64a6a9`. No code was generated by the LLM; the assistant was used to structure findings and draft prose.

</details>

### LINKED ISSUE #4130: [BUG] Concurrent timeout & reassign loops: ListTasksToReassign lacks FOR UPDATE SKIP LOCKED
**Describe the issue**

`timeoutTaskOperations` and `reassignTaskOperations` are independent per-tenant pools that run **concurrently** and can both target the same expired+stale runtime. `ListTasksToReassign` has **no `FOR UPDATE`/`SKIP LOCKED`** locking clause, whereas the sibling `ListTasksToTimeout` does. The `retry_count` guard prevents a double increment, so this is **not** a lost-update — but the losing path still runs `releaseTasks` and can emit a duplicate monitoring event for one logical expiry, and the missing `FOR UPDATE` on reassign is an ordering smell that may surface as duplicate work or noisy events under load.

This is correctness finding **F5** of an audit shared with @abelanger5 by email on 2026-06-09. Filed as a public issue at maintainer request. Reference advisory (cross-tenant findings F1/F2): GHSA-r7w5-4j6c-jxxc. Companion correctness issues: #4128 (F3), #4129 (F4).

**Environment**
- SDK: any (server-side bug, SDK-agnostic)
- Engine: Self-hosted, commit `f64a6a9` (HEAD at time of review). Same v1 engine path on Cloud.

**Expected behavior**

The two control loops should not be able to pick the same row simultaneously. `ListTasksToReassign` should mirror `ListTasksToTimeout`'s `FOR UPDATE SKIP LOCKED` locking, so whichever loop wins the race processes the row and the other skips.

**Actual behavior**

`ListTasksToReassign` selects without any locking clause. Under concurrent execution of the two control loops on an expired-and-stale runtime, both can read the row, both reach `FailTask*`, and both call `releaseTasks` / emit monitoring events for the same logical expiry. The `retry_count` guard prevents the second increment from corrupting state, but the path still runs to its end and emits the duplicate event.

File:line refs at commit `f64a6a9`:

- `internal/services/controllers/task/controller.go:234`, `:252` — `timeoutTaskOperations` and `reassignTaskOperations` are independent per-tenant pools, run concurrently.
- `pkg/repository/sqlcv1/tasks.sql:397` — `ListTasksToTimeout` has `FOR UPDATE SKIP LOCKED` (the safe sibling).
- `pkg/repository/sqlcv1/tasks.sql:431` — `ListTasksToReassign` (the `tasks_on_inactive_workers` CTE) has **no** `FOR UPDATE`/`SKIP LOCKED` — the gap.
- `pkg/repository/task.go:853`, `:889` — both paths funnel into `FailTask*`, which does `retry_count + 1`.
- `pkg/repository/sqlcv1/tasks.sql:291`, `:354` — the `retry_count` guard that prevents the double increment but does not prevent the duplicate-work / duplicate-event run.

**Confidence:** **to confirm** with a forced-interleave test invoking `ProcessTaskTimeouts` and `ProcessTaskReassignments` on the same expired+stale runtime — happy to write one. The code-level asymmetry between the two queries is clear; the runtime impact depends on how often the two loops contend on the same row in practice.

**Suggested fix**

Add `FOR UPDATE SKIP LOCKED` to `ListTasksToReassign`'s `tasks_on_inactive_workers` CTE. This makes the two control-loop queries symmetric and lets PostgreSQL serialise the contention at the row level — the same pattern the timeout query already uses.

**Additional context**

Found via proof by contrast against the sibling query in the same file (`ListTasksToTimeout` at `tasks.sql:397`). The two queries are otherwise structurally similar; the missing locking clause is the asymmetry.

Reported via Orionix · m-be290108 on GitHub · @orionix_ai · fiverr.com/mathieuballot

---

<details id="ai-disclosure">

- **Details**: Claude (Anthropic) was used as a research and writing assistant during the manual code audit. All file:line citations, the proof-by-contrast reasoning, and the suggested fix were verified by direct reading of the cited source in this repo at commit `f64a6a9`. No code was generated by the LLM; the assistant was used to structure findings and draft prose.

</details>

BloggerBust:
# Description

This fixes two task runtime issues:

1. `ListTasksToTimeout` now excludes evicted runtimes, preventing evicted durable tasks from being wrongly timed out.
2. `ListTasksToReassign` now uses `FOR UPDATE OF runtime SKIP LOCKED`, preventing timeout and reassign loops from selecting the same runtime row concurrently.

A regression test was added for the timeout behavior to confirm that evicted runtimes are excluded while non-evicted expired runtimes are still returned.

Fixes #4128
Fixes #4130

## Type of change

* [x] Bug fix (non-breaking change which fixes an issue)
* [x] Test changes (add, refactor, improve or change a test)

## What's Changed

* [x] Exclude evicted runtimes from `ListTasksToTimeout`.
* [x] Add `FOR UPDATE OF runtime SKIP LOCKED` to `ListTasksToReassign`.
* [x] Add a repository regression test confirming evicted runtimes are not returned by `ListTasksToTimeout`.
* [x] Update the generated SQLC query text in `tasks.sql.go` to match `tasks.sql`.

## Checklist

Changes have been:

* [x] Tested (unit, integration, or manually with steps specified)
* [x] Linted and formatted

## Testing

Validated with:

```sh
go test -run TestListTasksToTimeout_ExcludesEvictedRuntimes ./pkg/repository/
go test ./pkg/repository/
go build ./pkg/repository/sqlcv1/...
go build ./pkg/repository/...
go build ./internal/services/controllers/task/...
go vet ./pkg/repository/...
```

---

<details id="ai-disclosure">

* [x] *I acknowledge that an LLM was used in the creation of this Pull Request, in accordance with Hatchet's [AI_POLICY.md](./AI_POLICY.md).*

* **Details**: Claude Code was used to help investigate the reported task runtime correctness issues. It also recommended minimal SQL changes which I agreed with and used. For quickness, I had claude draft the regression test. I reviewed the changes and validated them locally before submitting.

</details>

### DIFF
--- a/pkg/repository/sqlcv1/tasks.sql
+++ b/pkg/repository/sqlcv1/tasks.sql
@@ -390,6 +390,8 @@ WITH expired_runtimes AS (
     WHERE
         tenant_id = @tenantId::uuid
         AND timeout_at <= NOW()
+        -- evicted tasks are not eligible for timeout
+        AND evicted_at IS NULL
     ORDER BY
         task_id, task_inserted_at, retry_count
     LIMIT
@@ -431,6 +433,7 @@ WITH tasks_on_inactive_workers AS (
         AND runtime.evicted_at IS NULL
     LIMIT
         COALESCE(sqlc.narg('limit')::integer, 1000)
+    FOR UPDATE OF runtime SKIP LOCKED
 )
 SELECT
     v1_task.id,
--- a/pkg/repository/task_timeout_eviction_test.go
+++ b/pkg/repository/task_timeout_eviction_test.go
@@ -1,0 +1,86 @@
+//go:build !e2e && !load && !rampup && !integration
+
+package repository
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/hatchet-dev/hatchet/pkg/repository/sqlcv1"
+)
+
+// TestListTasksToTimeout_ExcludesEvictedRuntimes verifies that evicted
+// tasks are not selected for timeout processing.
+func TestListTasksToTimeout_ExcludesEvictedRuntimes(t *testing.T) {
+	pool, cleanup := setupPostgresWithMigration(t)
+	defer cleanup()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	defer cancel()
+
+	tenantID := uuid.New()
+	workerID := uuid.New()
+	stepID := uuid.New()
+	workflowID := uuid.New()
+	workflowVersionID := uuid.New()
+	workflowRunID := uuid.New()
+
+	now := time.Now().UTC()
+	pastTimeout := now.Add(-5 * time.Minute)
+
+	_, err := pool.Exec(ctx, `
+		INSERT INTO "Tenant" ("id", "name", "slug", "createdAt", "updatedAt")
+		VALUES ($1, 'test-tenant', 'test-tenant', NOW(), NOW())
+	`, tenantID)
+	require.NoError(t, err)
+
+	_, err = pool.Exec(ctx, `
+		INSERT INTO "Worker" ("id", "tenantId", "name", "lastHeartbeatAt", "createdAt", "updatedAt")
+		VALUES ($1, $2, 'test-worker', NOW(), NOW(), NOW())
+	`, workerID, tenantID)
+	require.NoError(t, err)
+
+	// Insert two v1_task rows: one for the evicted runtime, one for the non-evicted runtime.
+	// Use OVERRIDING SYSTEM VALUE to set the identity column.
+	_, err = pool.Exec(ctx, `
+		INSERT INTO v1_task (id, inserted_at, tenant_id, queue, action_id, step_id, step_readable_id,
+			workflow_id, workflow_version_id, workflow_run_id, schedule_timeout,
+			step_timeout, sticky, external_id, display_name, input, step_index)
+		OVERRIDING SYSTEM VALUE
+		VALUES
+			(1, $1, $2, 'default', 'test-action', $3, 'test-step', $4, $5, $6, '5m', '30s', 'NONE', $7, 'test-task-1', '{}', 0),
+			(2, $1, $2, 'default', 'test-action', $3, 'test-step', $4, $5, $6, '5m', '30s', 'NONE', $8, 'test-task-2', '{}', 1)
+	`, now, tenantID, stepID, workflowID, workflowVersionID, workflowRunID, uuid.New(), uuid.New())
+	require.NoError(t, err)
+
+	// Insert two v1_task_runtime rows, both with timeout_at in the past:
+	// - Task 1: evicted and should NOT be returned
+	// - Task 2: not evicted and SHOULD be returned
+	_, err = pool.Exec(ctx, `
+		INSERT INTO v1_task_runtime (task_id, task_inserted_at, retry_count, worker_id, tenant_id, timeout_at, evicted_at)
+		VALUES
+			(1, $1, 0, $2, $3, $4, $5),
+			(2, $1, 0, $2, $3, $4, NULL)
+	`, now, workerID, tenantID, pastTimeout, now.Add(-3*time.Minute))
+	require.NoError(t, err)
+
+	queries := sqlcv1.New()
+
+	results, err := queries.ListTasksToTimeout(ctx, pool, sqlcv1.ListTasksToTimeoutParams{
+		Tenantid: tenantID,
+		Limit:    pgtype.Int4{Int32: 100, Valid: true},
+	})
+	require.NoError(t, err)
+
+	// Only the non-evicted runtime (task_id=2) should be returned
+	assert.Len(t, results, 1, "should return exactly one task (the non-evicted one)")
+	if len(results) == 1 {
+		assert.Equal(t, int64(2), results[0].ID, "returned task should be the non-evicted one (id=2)")
+	}
+}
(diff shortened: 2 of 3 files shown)


pkg/repository/task_timeout_eviction_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/hatchet-dev/hatchet/pkg/repository/sqlcv1"
+)
+
+// TestListTasksToTimeout_ExcludesEvictedRuntimes verifies that evicted
+// tasks are not selected for timeout processing.
+func TestListTasksToTimeout_ExcludesEvictedRuntimes(t *testing.T) {
```
gregfurman, line 20: suggestion: An e2e test could be nice here since the manual SQL approach is kinda difficult to follow.

pkg/repository/sqlcv1/tasks.sql
gregfurman, line 1: Should we include this in a new entry for `CHANGELOG.md` or is it not _really_ user-facing? 🤔

gregfurman (approved):
LGTM. Verified your test fails on `main` but passes with your new SQL changes added. Great stuff! 🙌

pkg/repository/task_timeout_eviction_test.go
BloggerBust: Thanks for the feedback Greg. I was concerned that an e2e test involving timeouts like this might be fragile and risk breaking CI intermittently. The repro requires the task’s execution timeout to expire while the runtime is evicted, so there are a lot of moving parts. The test coverage I did add directly validates the logic changed in this PR. However, I am happy to investigate a reliable e2e strategy if you think the added coverage is important.

pkg/repository/sqlcv1/tasks.sql
BloggerBust: That is a good question. I thought about this and I think it  is worth keeping in the generated CHANGELOG.md. While the change is internal, the bug affects durable task behavior. That seems useful to include under Fixed.

pkg/repository/task_timeout_eviction_test.go
gregfurman: Maybe we add a follow up issue? Then happy to resolve!

PR #4137, opened 2026-06-10
fix(repository): key failureOpts by taskId in failTasksTx to fix wrong error message after dedup

### LINKED ISSUE #4131: [BUG] failTasksTx attaches the wrong error message after dedup/reorder
**Describe the issue**

`failTasksTx` (`pkg/repository/task.go`) deduplicates and reorders the incoming `tasks` slice but then indexes `failureOpts` by the *position of the released task* — not by `taskId`. On a mixed-error `FailTasks` batch, the resulting `NewFailedTaskOutputEvent` can attach the **wrong error message** to a task. No state corruption — just misleading failure diagnostics in run history.

This is correctness finding **F6** of an audit shared with @abelanger5 by email on 2026-06-09. Filed as a public issue at maintainer request. Reference advisory (cross-tenant findings F1/F2): GHSA-r7w5-4j6c-jxxc. Companion correctness issues: #4128 (F3), #4129 (F4), #4130 (F5).

**Environment**
- SDK: any (server-side bug, SDK-agnostic)
- Engine: Self-hosted, commit `f64a6a9` (HEAD at time of review). Same v1 engine path on Cloud.

**Expected behavior**

When `FailTasks` is called with a batch where different tasks have different error messages, each task's `FailedTaskOutputEvent` should carry **its own** error message — i.e. `failureOpts` should be looked up by a stable key (`taskId` or `taskId+retryCount`), not by position in the post-dedup slice.

**Actual behavior**

The slice is deduped and reordered before being indexed positionally:

1. `pkg/repository/task.go:841` — `tasks = uniqueSet(tasks)` dedups and reorders the slice relative to the caller-ordered `failureOpts`.
2. `pkg/repository/task.go:915` — `releasedTasks` derives from the (already deduped/reordered) `tasks`.
3. `pkg/repository/task.go:923-924` — `NewFailedTaskOutputEvent(releasedTask, failureOpts[i].ErrorMessage)` indexes `failureOpts` by the position of `releasedTasks` — not by `taskId`.

So on a mixed-error batch the error text can be mismatched: task A's emitted event carries task B's error message.

File:line refs at commit `f64a6a9`:

- `pkg/repository/task.go:841` — `tasks = uniqueSet(tasks)` (the dedup/reorder).
- `pkg/repository/task.go:915` — `releasedTasks` derived from `tasks` post-dedup.
- `pkg/repository/task.go:923-924` — `NewFailedTaskOutputEvent(releasedTask, failureOpts[i].ErrorMessage)` — positional index against the post-dedup ordering.

Conceptual repro:

1. Call `FailTasks` with a batch like `[{T1, "errA"}, {T2, "errB"}, {T1, "errA"}]` (duplicate of T1, distinct from T2).
2. `uniqueSet` returns `[T1, T2]` (or `[T2, T1]` — order depends on the dedup implementation, which is exactly the symptom).
3. `releasedTasks[0]` is whichever of T1/T2 came out first; the event is built with `failureOpts[0].ErrorMessage` — `"errA"`.
4. If `uniqueSet` reordered to `[T2, T1]`, T2's event carries T1's error message.

**Suggested fix**

Key `failureOpts` by `taskId+retryCount` (a `map[failureOptKey]string`) and look up by `releasedTask.ID + releasedTask.RetryCount`, not by the slice position. One-block change in `failTasksTx`.

**Impact**

Low. Misleading failure diagnostics in run history (operators see the wrong error attributed to a failed task); no state corruption, no double execution, no retry-count drift.

**Additional context**

Found via proof by contrast: the surrounding code carefully keys other per-task lookups (e.g. `releasedTasks` membership) by ID — only this one event-emission path uses a positional index against a slice that was just deduped.

Reported via Orionix · m-be290108 on GitHub · @orionix_ai · fiverr.com/mathieuballot

---

<details id="ai-disclosure">

- **Details**: Claude (Anthropic) was used as a research and writing assistant during the manual code audit. All file:line citations, the proof-by-contrast reasoning, and the suggested fix were verified by direct reading of the cited source in this repo at commit `f64a6a9`. No code was generated by the LLM; the assistant was used to structure findings and draft prose.

</details>

burnerlee:
# Description

Fixes https://github.com/hatchet-dev/hatchet/issues/4131

`failTasksTx` built `tasks[]` parallel to `failureOpts[]` (position `i` matched), then called `uniqueSet(tasks)` which deduplicates and may reorder the slice. After that, `releasedTasks` was derived from the deduped `tasks`, but the event emission at line 940 still indexed `failureOpts[i]` by position — now misaligned after dedup/reorder.

On a mixed-error batch (e.g. `[{T1,"errA"},{T2,"errB"},{T1,"errA"}]`), `uniqueSet` returns `[T1,T2]` but if order shifts, T2's event gets T1's error message.

**Fix**: build `errorMessageByTaskKey map[string]string` keyed by `"taskId:retryCount"` before `uniqueSet`, then look up by `releasedTask.ID:releasedTask.RetryCount` at event emission — stable regardless of dedup ordering.

## Type of change

## What's Changed

- [x] In `failTasksTx` (`pkg/repository/task.go`): build `errorMessageByTaskKey` map from `failureOpts` before `uniqueSet` call, keyed by `taskId:retryCount`
- [x] Replace positional `failureOpts[i].ErrorMessage` lookup with map lookup by `releasedTask.ID` + `releasedTask.RetryCount`

## Checklist

Changes have been:

- [x] Tested (manually verified compile; no existing unit tests for this path)
- [x] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Claude Code (claude-sonnet-4-6) used to identify the bug location, derive context, and generate the fix. The fix logic and file:line references were verified by direct code reading.

</details>

### DIFF
--- a/pkg/repository/task.go
+++ b/pkg/repository/task.go
@@ -830,6 +830,7 @@ func (r *TaskRepositoryImpl) failTasksTx(ctx context.Context, tx sqlcv1.DBTX, te
 	defer span.End()
 
 	tasks := make([]TaskIdInsertedAtRetryCount, len(failureOpts))
+	errorMessageByTaskKey := make(map[string]string, len(failureOpts))
 	appFailureTaskIds := make([]int64, 0)
 	appFailureTaskInsertedAts := make([]pgtype.Timestamptz, 0)
 	appFailureTaskRetryCounts := make([]int32, 0)
@@ -841,6 +842,7 @@ func (r *TaskRepositoryImpl) failTasksTx(ctx context.Context, tx sqlcv1.DBTX, te
 
 	for i, failureOpt := range failureOpts {
 		tasks[i] = *failureOpt.TaskIdInsertedAtRetryCount
+		errorMessageByTaskKey[fmt.Sprintf("%d:%d", failureOpt.Id, failureOpt.RetryCount)] = failureOpt.ErrorMessage
 
 		if failureOpt.IsAppError {
 			appFailureTaskIds = append(appFailureTaskIds, failureOpt.Id)
@@ -937,7 +939,8 @@ func (r *TaskRepositoryImpl) failTasksTx(ctx context.Context, tx sqlcv1.DBTX, te
 	outputs := make([][]byte, len(releasedTasks))
 
 	for i, releasedTask := range releasedTasks {
-		out := NewFailedTaskOutputEvent(releasedTask, failureOpts[i].ErrorMessage).Bytes()
+		errMsg := errorMessageByTaskKey[fmt.Sprintf("%d:%d", releasedTask.ID, releasedTask.RetryCount)]
+		out := NewFailedTaskOutputEvent(releasedTask, errMsg).Bytes()
 
 		outputs[i] = out
 	}


burnerlee:
hello @m-be290108 requesting a review. TIA!

m-be290108:
Hi @burnerlee — thanks for taking this on so quickly!

I read the diff against what I described in #4131:

- `errorMessageByTaskKey` keyed by `"taskId:retryCount"` matches the spec @BloggerBust pointed out, and the lookup at line 942 is stable regardless of `uniqueSet` reordering. The correctness fix is exactly right.
- The "Tested (manually verified compile)" checklist item is the one piece I'd push back on. The bug only surfaces on mixed-error batches where `uniqueSet` changes the order — which is precisely the case a unit test would catch. `pkg/repository/task_partition_test.go` already wires up `setupPostgresWithMigration` and would be a natural place to add one: feeding `[{T1,"errA"}, {T2,"errB"}, {T1,"errA"}]` and asserting the emitted event for T2 carries `"errB"` (not `"errA"`) would lock the invariant going forward.

Happy to leave the test decision to the maintainers, but flagging since this is exactly the kind of bug that comes back without a regression test.

LGTM on the fix itself.

burnerlee:
> Hi @burnerlee — thanks for taking this on so quickly!
> 
> I read the diff against what I described in #4131:
> 
> * `errorMessageByTaskKey` keyed by `"taskId:retryCount"` matches the spec @BloggerBust pointed out, and the lookup at line 942 is stable regardless of `uniqueSet` reordering. The correctness fix is exactly right.
> * The "Tested (manually verified compile)" checklist item is the one piece I'd push back on. The bug only surfaces on mixed-error batches where `uniqueSet` changes the order — which is precisely the case a unit test would catch. `pkg/repository/task_partition_test.go` already wires up `setupPostgresWithMigration` and would be a natural place to add one: feeding `[{T1,"errA"}, {T2,"errB"}, {T1,"errA"}]` and asserting the emitted event for T2 carries `"errB"` (not `"errA"`) would lock the invariant going forward.
> 
> Happy to leave the test decision to the maintainers, but flagging since this is exactly the kind of bug that comes back without a regression test.
> 
> LGTM on the fix itself.

Thank you!

pkg/repository/task.go
```diff
@@ -937,7 +939,8 @@ func (r *TaskRepositoryImpl) failTasksTx(ctx context.Context, tx sqlcv1.DBTX, te
 	outputs := make([][]byte, len(releasedTasks))
 
 	for i, releasedTask := range releasedTasks {
-		out := NewFailedTaskOutputEvent(releasedTask, failureOpts[i].ErrorMessage).Bytes()
+		errMsg := errorMessageByTaskKey[fmt.Sprintf("%d:%d", releasedTask.ID, releasedTask.RetryCount)]
```
gregfurman, line 942: Just confirming that this key will always exist here? Otherwise this would panic.

gregfurman (approved):
LGTM! Unfortunate that we don't have an easy way of extracting a unique run id when the same task is retried. But this seems like a reasonable workaround 👍

pkg/repository/task.go
burnerlee: Yes — the key will always exist. `releasedTasks` is derived from `tasks`, which is built directly from `failureOpts` using the same `Id` and `RetryCount` fields (lines 843, 3780). So every `releasedTask.ID:releasedTask.RetryCount` was inserted into `errorMessageByTaskKey` during the same loop. Also worth noting: a missing key in a Go map returns the zero value (`""`) rather than panicking, so even in an unexpected edge case this would just emit an empty error message — same behaviour as before for tasks with no error message set.

m-be290108:
@Gregfurman — happy to chime in since I wrote the review.

By construction, yes: the lookup key is guaranteed to be present. The `errorMessageByTaskKey` map is populated in the same `for i, failureOpt := range failureOpts` loop at line 845, with key `fmt.Sprintf("%d:%d", failureOpt.Id, failureOpt.RetryCount)`. Every `releasedTask` downstream descends from one of these `failureOpts` (via `tasks` → `uniqueSet(tasks)` → `releasedTasks`), so `releasedTask.ID:releasedTask.RetryCount` will always match a key that was inserted.

That said, defensive coding wouldn't hurt — `errMsg, ok := ...` with a fallback to an empty string would make a missing-key scenario explicit if invariants ever drift in a future refactor. Worth a future cleanup, not blocking this PR imo.

Edge case worth noting: if two `failureOpts` ever share the same `Id+RetryCount` but different `ErrorMessage`, last-write-wins on the map. Doesn't matter for the bug being fixed (the original code already collapsed them via positional indexing), but worth knowing if upstream callers ever start sending duplicated `failureOpts` intentionally.

PR #4139, opened 2026-06-10
fix(engine): drain grpc streams and bound graceful shutdown

igor-kupczynski:
# Description

On every engine rollout, `GracefulStop()` blocks indefinitely on long-lived subscriber streams (`SubscribeToWorkflowEvents`, `SubscribeToWorkflowRuns`, `ListenForDurableEvent`, `DurableTask`) — unlike worker `ListenV2` streams, nothing closes them server-side. The pod is then SIGKILLed at the end of the grace period mid-stream, producing bursts of broken pipe / `UNAVAILABLE` errors on workers and clients (and error-promoted shutdown logs from the cleanup module), even though no work is lost.

This PR makes engine shutdown deterministic:

- A new `streams.Registry` tracks cancel functions for long-lived subscriber/durable streams; the gRPC teardown cancels all sessions first, so `GracefulStop` can complete.
- `GracefulStop` is bounded by a new `runtime.grpcShutdownTimeout` config (`SERVER_GRPC_SHUTDOWN_TIMEOUT`, default `10s`), falling back to a hard `Stop()` with an error log.
- The health server is moved from second to the end of the teardown (only telemetry shutdown and the server/database disconnects follow it), so `/ready` and `/live` reflect the not-ready state for effectively the entire teardown — this is what makes load balancer health checks against `/ready` meaningful during shutdown.
- The cleanup module's `lines` channel is now buffered per its own comment: previously the logger goroutine only started draining once the promotion deadline fired, so `Run()` stalled for the full `TimeLimit` (9s) on *every* shutdown before executing any cleanup fn, and always error-promoted the logs. With the stall fixed, `TimeLimit` is purely a log-promotion threshold; it is raised from 9s to 15s so it only trips when the 10s gRPC drain bound is exhausted and the rest of cleanup is slow.
- On the worker side, `retrySubscribe` no longer sleeps before the first retry attempt, so workers resubscribe immediately when the engine hangs up their stream during a drain.

## Shutdown behavior: before vs after

Before — `GracefulStop` waits forever on subscriber streams, and the kill is what "closes" them:

```mermaid
sequenceDiagram
    participant K8s as kubelet
    participant Engine as EnginePod
    participant Worker as WorkerSDK
    participant Sub as RunSubscriber
    K8s->>Engine: SIGTERM
    Engine->>Engine: health server torn down almost first
    Engine->>Worker: ListenV2 closed (fin)
    Engine->>Engine: GracefulStop waits on subscriber streams
    Note over Engine,Sub: nothing closes these streams server-side,<br/>they end only when the client hangs up
    K8s->>Engine: SIGKILL at end of grace period
    Engine--xWorker: connection severed (broken pipe)
    Engine--xSub: stream severed (UNAVAILABLE)
    Note over Engine: remaining cleanup fns never run
```

After — streams are hung up first, so `GracefulStop` completes and the pod exits cleanly:

```mermaid
sequenceDiagram
    participant K8s as kubelet
    participant Engine as EnginePod
    participant Worker as WorkerSDK
    participant Sub as RunSubscriber
    K8s->>Engine: SIGTERM
    Engine->>Engine: readiness flips to 503 (health server stays up)
    Engine->>Engine: CancelStreamSessions on both dispatchers
    Engine->>Sub: subscriber streams end promptly (cancelled)
    Engine->>Worker: ListenV2 closed (fin)
    Worker->>Worker: reconnects to a healthy replica
    Engine->>Engine: GracefulStop completes (10s bound, hard Stop fallback)
    Engine->>Engine: remaining cleanup, health server shut down at the end
    Engine->>K8s: clean exit, no SIGKILL
```

## Type of change

## What's Changed

- [x] `internal/services/shared/streams`: new session registry; closed after `CancelAll()` so late registrations are cancelled immediately (+ unit tests)
- [x] `internal/services/dispatcher` + `internal/services/dispatcher/v1`: register stream sessions in the five long-lived handlers; `CancelStreamSessions()` on both services. `DurableTask` is restructured so a recv-only goroutine hands messages to the handler goroutine over a channel: cancellation hangs up idle streams, processing is serialized with the deferred stream cleanup (no re-registration after cleanup), and unexpected `Recv` errors are returned to the client again (+ unit tests)
- [x] `internal/services/grpc/server.go`: bounded `GracefulStop` with `WithShutdownTimeout` opt
- [x] `pkg/config/server/server.go`: `runtime.grpcShutdownTimeout` (default 10s)
- [x] `cmd/hatchet-engine/engine/run.go`: cancel sessions at top of gRPC teardown (v0 + v1 paths); health cleanup moved to the end
- [x] `pkg/client/dispatcher.go`: `retrySubscribe` sleeps after a failed attempt instead of before the first one
- [x] `pkg/cleanup/cleanup.go`: buffer the `lines` channel (fixes a `TimeLimit`-long stall before any cleanup fn ran), wait for the logger to drain before returning, TimeLimit 9s → 15s (+ unit test)

## Checklist

Changes have been:

- [x] Linted and formatted

## Testing

- New unit tests for the stream session registry and the cleanup module (`go test -race ./internal/services/shared/streams/... ./pkg/cleanup/...`)
- New unit tests for the `DurableTask` stream lifecycle: session cancel unblocks a pending `Recv` and returns nil, recv errors are surfaced, EOF returns nil (`go test -race ./internal/services/dispatcher/v1/...`)
- Existing dispatcher tests pass (`go test ./internal/services/dispatcher/...`)
- `go build` + `go vet` clean on all touched packages
- Manual verification planned: SIGTERM a local engine with an active worker + run subscription, confirm exit within ~10s and prompt stream termination

---

<details id="ai-disclosure">

- **Details**: Implementation and PR description written with Cursor (Fable 5), based on a shutdown-path audit; reviewed by a human before merge.

</details>

### DIFF
--- a/internal/services/dispatcher/dispatcher.go
+++ b/internal/services/dispatcher/dispatcher.go
@@ -14,6 +14,7 @@ import (
 	"github.com/hatchet-dev/hatchet/internal/msgqueue"
 	"github.com/hatchet-dev/hatchet/internal/services/dispatcher/contracts"
 	"github.com/hatchet-dev/hatchet/internal/services/shared/recoveryutils"
+	"github.com/hatchet-dev/hatchet/internal/services/shared/streams"
 	tasktypes "github.com/hatchet-dev/hatchet/internal/services/shared/tasktypes/v1"
 	"github.com/hatchet-dev/hatchet/internal/syncx"
 	"github.com/hatchet-dev/hatchet/pkg/analytics"
@@ -48,13 +49,21 @@ type DispatcherImpl struct {
 	workflowRunBufferSize               int
 	streamEventBufferTimeout            time.Duration
 
-	dispatcherId uuid.UUID
-	workers      *workers
-	a            *hatcheterrors.Wrapped
+	dispatcherId   uuid.UUID
+	workers        *workers
+	a              *hatcheterrors.Wrapped
+	streamSessions *streams.Registry
 
 	durableCallbackFn func(taskExternalId uuid.UUID, invocationCount int32, branchId, nodeId int64, payload []byte) error
 	analytics         analytics.Analytics
 	version           string
+}
+
+// CancelStreamSessions hangs up all registered long-lived subscriber streams. It is
+// called during shutdown before GracefulStop, which would otherwise block on them
+// until the process is killed.
+func (d *DispatcherImpl) CancelStreamSessions() {
+	d.streamSessions.CancelAll()
 }
 
 var ErrWorkerNotFound = fmt.Errorf("worker not found")
@@ -274,6 +283,7 @@ func New(fs ...DispatcherOpt) (*DispatcherImpl, error) {
 		repov1:                              opts.repov1,
 		dispatcherId:                        opts.dispatcherId,
 		workers:                             &workers{},
+		streamSessions:                      streams.NewRegistry(),
 		s:                                   s,
 		a:                                   a,
 		cache:                               opts.cache,
--- a/internal/services/dispatcher/server_v1.go
+++ b/internal/services/dispatcher/server_v1.go
@@ -321,6 +321,9 @@ func (s *DispatcherImpl) subscribeToWorkflowRunsV1(server contracts.Dispatcher_S
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	deregister := s.streamSessions.Register(cancel)
+	defer deregister()
+
 	wg := sync.WaitGroup{}
 	sendMu := sync.Mutex{}
 	ringIndex := 0
@@ -874,6 +877,9 @@ func (s *DispatcherImpl) subscribeToWorkflowEventsByWorkflowRunIdV1(workflowRunI
 	ctx, cancel := context.WithCancel(stream.Context())
 	defer cancel()
 
+	deregister := s.streamSessions.Register(cancel)
+	defer deregister()
+
 	retries := 0
 	foundWorkflowRun := false
 
@@ -1059,6 +1065,9 @@ func (s *DispatcherImpl) subscribeToWorkflowEventsByAdditionalMetaV1(key string,
 
 	ctx, cancel := context.WithCancel(stream.Context())
 	defer cancel()
+
+	deregister := s.streamSessions.Register(cancel)
+	defer deregister()
 
 	wg := sync.WaitGroup{}
 
--- a/internal/services/grpc/server.go
+++ b/internal/services/grpc/server.go
@@ -66,6 +66,8 @@ type Server struct {
 	otelCollector otelcol.OTelCollector
 	tls           *tls.Config
 	insecure      bool
+
+	shutdownTimeout time.Duration
 }
 
 type ServerOpt func(*ServerOpts)
@@ -85,6 +87,8 @@ type ServerOpts struct {
 	otelCollector otelcol.OTelCollector
 	tls           *tls.Config
 	insecure      bool
+
+	shutdownTimeout time.Duration
 }
 
 func defaultServerOpts() *ServerOpts {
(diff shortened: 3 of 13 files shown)


pkg/cleanup/cleanup.go
```diff
@@ -22,7 +22,7 @@ type Cleanup struct {
 func New(logger *zerolog.Logger) Cleanup {
 	return Cleanup{
 		Fns:       []CleanupFn{},
-		TimeLimit: time.Second * 9,
+		TimeLimit: time.Second * 25,
```
igor-kupczynski: 9 seconds seem a bit tight, let's bump it to 25

internal/services/grpc/server.go
```diff
@@ -366,7 +385,22 @@ func (s *Server) startGRPC() (func() error, error) {
 	}()
 
 	cleanup := func() error {
-		grpcServer.GracefulStop()
```
igor-kupczynski, line 369: This is one of the key changes, `GracefulStop()` blocks new connections, but doesn't stop in-flight connections, just waits until they finish (at least this is how I read the docs):
```
// GracefulStop stops the gRPC server gracefully. It stops the server from
// accepting new connections and RPCs and blocks until all the pending RPCs are
// finished.
```

[src](https://pkg.go.dev/google.golang.org/grpc#Server.GracefulStop)

internal/services/dispatcher/server_v1.go
```diff
@@ -321,6 +321,9 @@ func (s *DispatcherImpl) subscribeToWorkflowRunsV1(server contracts.Dispatcher_S
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	deregister := s.streamSessions.Register(cancel)
+	defer deregister()
```
igor-kupczynski, line 325: This registers (and deregisters) the streams on the server. ==> so we can send them cancelation when the server is being shut down.

internal/services/dispatcher/dispatcher.go
```diff
@@ -48,15 +49,23 @@ type DispatcherImpl struct {
...
 	durableCallbackFn func(taskExternalId uuid.UUID, invocationCount int32, branchId, nodeId int64, payload []byte) error
 	analytics         analytics.Analytics
 	version           string
 }
 
+// CancelStreamSessions hangs up all registered long-lived subscriber streams. It is
+// called during shutdown before GracefulStop, which would otherwise block on them
+// until the process is killed.
+func (d *DispatcherImpl) CancelStreamSessions() {
+	d.streamSessions.CancelAll()
```
grutt, line 66: we're erroring on the sdks for this?

internal/services/grpc/server.go
grutt: this makes sense to me

grutt: one other question here is should/can we reject connections at the load balancer? i suppose there's a case where a retry consistently will hit a terminating pod?

internal/services/dispatcher/dispatcher.go
igor-kupczynski: I don't think the SDKs would error on this, as this path closes the streams cleanly. Cancelling a session makes the handler return nil, so I think the client sees a normal end-of-stream (OK status / EOF) and resubscribe.

The error case only remains if a stream somehow doesn't drain within the 10s grpcShutdownTimeout,.

internal/services/grpc/server.go
igor-kupczynski: This is up to the NLB config. The NLB health checks should target `/ready` with e.g. 5s interval  2x failure threshold.

igor-kupczynski:
@juliusgeo @grutt  I've pushed another commit with a small change of behavior and unit-tests; I think this is good to merge. I'll ping you on a different channel with a "repro".

The tests seem to be flaky and unrelated to this change. We need to run fable in a loop to fix them 🙈

grutt (approved):

PR #4142, opened 2026-06-10
fix(sdks/go): register standalone task cron and event triggers

### LINKED ISSUE #4127: [BUG] Go SDK: `WithCron` option for standalone tasks appears to be a no-op
**Describe the issue**

The Go SDK option `WithCron` is [documented](https://pkg.go.dev/github.com/hatchet-dev/hatchet@v0.88.8/sdks/go#WithCron) as:

> WithCron configures standalone tasks to run on a cron schedule. Only applicable to standalone tasks, not workflow tasks.

However, this option appears not to set any cron schedules when applied to [`(*Client).NewStandaloneTask`](https://pkg.go.dev/github.com/hatchet-dev/hatchet@v0.88.8/sdks/go#Client.NewStandaloneTask):

```go
c.NewStandaloneTask("foo", /* ... */, hatchet.WithCron("0 0 * * *"))
```

**Environment**
- SDK: Go v0.88.8
- Engine: any

**Expected behavior**

Passing `WithCron` to `NewStandaloneTask` creates a cron schedule for the task.

Claude Code helped me produce an example. Running it gives:

```
> go run ./hatchet_withcron.go
{"level":"debug","service":"client","time":"2026-06-09T10:47:27.774984-04:00","message":"connecting to localhost:7077 without TLS"}
{"level":"info","service":"client","time":"2026-06-09T10:47:27.775071-04:00","message":"gzip compression enabled for gRPC client"}
=== WithCron (TaskOption) ===
name:"with-cron" tasks:{readable_id:"with-cron" action:"with-cron:with-cron" conditions:{} slot_requests:{key:"default" value:1}}
=== WithWorkflowCron (WorkflowOption) ===
name:"with-workflow-cron" cron_triggers:"* * * * *" tasks:{readable_id:"with-workflow-cron" action:"with-workflow-cron:with-workflow-cron" conditions:{} slot_requests:{key:"default" value:1}} cron_input:"{}"
```

Contents of `hatchet_withcron.go`:

```go
package main

import (
	"encoding/base64"
	"fmt"
	"os"

	v0 "github.com/hatchet-dev/hatchet/pkg/client"
	hatchet "github.com/hatchet-dev/hatchet/sdks/go"
)

// offlineToken is an unsigned JWT carrying just the claims the SDK's config
// loader reads. Claims are only base64-decoded (no signature check, no dial),
// so the whole example runs offline with no engine and no real credentials.
func offlineToken() string {
	claims := base64.RawURLEncoding.EncodeToString([]byte(
		`{"server_url":"http://localhost:8888","grpc_broadcast_address":"localhost:7077","sub":"00000000-0000-0000-0000-000000000000","exp":4102444800}`,
	))
	return "e30." + claims + ".unsigned"
}

func noop(ctx hatchet.Context, _ any) (any, error) { return nil, nil }

func main() {
	_ = os.Setenv("HATCHET_CLIENT_TLS_STRATEGY", "none")

	c, err := hatchet.NewClient(v0.WithToken(offlineToken()))
	if err != nil {
		panic(err)
	}

	withCron := c.NewStandaloneTask("with-cron", noop, hatchet.WithCron("* * * * *"))
	withWorkflowCron := c.NewStandaloneTask("with-workflow-cron", noop, hatchet.WithWorkflowCron("* * * * *"))

	// Dump() is the exact CreateWorkflowVersionRequest sent at registration.
	// Printing the whole request shows the broken case carries no cron anywhere,
	// not just in a field we chose to look at.
	a, _, _, _ := withCron.Dump()
	b, _, _, _ := withWorkflowCron.Dump()

	fmt.Println("=== WithCron (TaskOption) ===")
	fmt.Println(a)
	fmt.Println("=== WithWorkflowCron (WorkflowOption) ===")
	fmt.Println(b)
}
```

**Additional context**

N/A

---

<details id="ai-disclosure">

- **Details**:
  - I tried to use `WithCron` and found that no cron schedule was created.
  - I asked Claude Code why it thought this was the case.
  - Claude Code read through my code, and the Hatchet SDK code, and concluded that the `WithCron` option wasn't doing anything.
  - I asked Claude Code to produce a minimal, standalone example. It produced the example above.
  - I ran the example and confirmed `.Dump()`ing  the result of `NewStandaloneTask` with `WithCron` produces output that doesn't include the cron schedule anywhere.

</details>

BloggerBust:
# Description

Fixes #4127

`WithCron` and `WithEvents` were accepted by `NewStandaloneTask`, but the triggers were not included in the workflow registration request. Standalone tasks are implemented as single-task workflows, and triggers need to be registered at the workflow level.

This PR registers standalone task cron and event triggers as workflow-level triggers before creating the underlying workflow. The same fix is applied to `NewStandaloneDurableTask`.

## Type of change

* [x] Bug fix (non-breaking change which fixes an issue)
* [x] Test changes (add, refactor, improve or change a test)

## What's Changed

* [x] Register `WithCron` from standalone task options as workflow-level cron triggers.
* [x] Register `WithEvents` from standalone task options as workflow-level event triggers.
* [x] Apply the same behavior to `NewStandaloneTask` and `NewStandaloneDurableTask`.
* [x] Add Go SDK tests covering standalone task cron and event trigger registration output.

## Checklist

Changes have been:

* [x] Tested (unit, integration, or manually with steps specified)
* [x] Linted and formatted

## Testing

```sh
gofmt -w sdks/go/client.go sdks/go/standalone_triggers_test.go
go test -run "TestNewStandalone" ./sdks/go/
go test ./sdks/go/
go build ./sdks/go/...
go vet ./sdks/go/...
```

---

<details id="ai-disclosure">

* [x] *I acknowledge that an LLM was used in the creation of this Pull Request, in accordance with Hatchet's [AI_POLICY.md](./AI_POLICY.md).*

* **Details**: Claude Code was used to investigate the bug and advise on a minimal fix. I validated the fix locally before submitting.

</details>

### DIFF
--- a/sdks/go/standalone_triggers_test.go
+++ b/sdks/go/standalone_triggers_test.go
@@ -1,0 +1,82 @@
+//go:build !e2e && !load && !rampup && !integration
+
+package hatchet
+
+import (
+	"testing"
+
+	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/assert"
+
+	v0Client "github.com/hatchet-dev/hatchet/pkg/client"
+	cloudrest "github.com/hatchet-dev/hatchet/pkg/client/cloud/rest"
+	"github.com/hatchet-dev/hatchet/pkg/client/rest"
+)
+
+// stubV0Client implements v0Client.Client with minimal stubs so that
+// NewStandaloneTask can build a declaration without a live server.
+type stubV0Client struct{}
+
+func (s *stubV0Client) Admin() v0Client.AdminClient              { return nil }
+func (s *stubV0Client) Cron() v0Client.CronClient                { return nil }
+func (s *stubV0Client) Schedule() v0Client.ScheduleClient        { return nil }
+func (s *stubV0Client) Dispatcher() v0Client.DispatcherClient    { return nil }
+func (s *stubV0Client) Event() v0Client.EventClient              { return nil }
+func (s *stubV0Client) Subscribe() v0Client.SubscribeClient      { return nil }
+func (s *stubV0Client) API() *rest.ClientWithResponses           { return nil }
+func (s *stubV0Client) CloudAPI() *cloudrest.ClientWithResponses { return nil }
+func (s *stubV0Client) Logger() *zerolog.Logger                  { l := zerolog.Nop(); return &l }
+func (s *stubV0Client) TenantId() string                         { return "00000000-0000-0000-0000-000000000000" }
+func (s *stubV0Client) Namespace() string                        { return "" }
+func (s *stubV0Client) CloudRegisterID() *string                 { return nil }
+func (s *stubV0Client) RunnableActions() []string                { return nil }
+
+// newTestClient returns a Client backed by a stub, suitable for
+// offline tests with no server
+func newTestClient() *Client {
+	return &Client{legacyClient: &stubV0Client{}}
+}
+
+// sampleTaskFn is a minimal function matching the standalone task signature.
+func sampleTaskFn(_ Context, _ any) (any, error) { return nil, nil }
+
+// sampleDurableFn is a minimal function matching the standalone durable task signature.
+func sampleDurableFn(_ DurableContext, _ any) (any, error) { return nil, nil }
+
+// WithCron and WithEvents on standalone tasks should appear in the registration request.
+
+func TestNewStandaloneTask_WithCron(t *testing.T) {
+	c := newTestClient()
+	task := c.NewStandaloneTask("cron-task", sampleTaskFn, WithCron("*/5 * * * *"))
+	req, _, _, _ := task.Dump()
+	assert.Equal(t, []string{"*/5 * * * *"}, req.CronTriggers)
+}
+
+func TestNewStandaloneTask_WithEvents(t *testing.T) {
+	c := newTestClient()
+	task := c.NewStandaloneTask("event-task", sampleTaskFn, WithEvents("user:created"))
+	req, _, _, _ := task.Dump()
+	assert.Equal(t, []string{"user:created"}, req.EventTriggers)
+}
+
+func TestNewStandaloneDurableTask_WithCron(t *testing.T) {
+	c := newTestClient()
+	task := c.NewStandaloneDurableTask("durable-cron", sampleDurableFn, WithCron("0 0 * * *"))
+	req, _, _, _ := task.Dump()
+	assert.Equal(t, []string{"0 0 * * *"}, req.CronTriggers)
+}
+
+func TestNewStandaloneTask_MultipleCrons(t *testing.T) {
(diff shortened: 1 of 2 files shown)


sdks/go/standalone_triggers_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/stretchr/testify/assert"
+
+	v0Client "github.com/hatchet-dev/hatchet/pkg/client"
+	cloudrest "github.com/hatchet-dev/hatchet/pkg/client/cloud/rest"
+	"github.com/hatchet-dev/hatchet/pkg/client/rest"
+)
+
+// stubV0Client implements v0Client.Client with minimal stubs so that
+// NewStandaloneTask can build a declaration without a live server.
+type stubV0Client struct{}
```
gregfurman, line 18: suggestion: Maybe this should move to a separate `stubs/client.go`?

sdks/go/standalone_triggers_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+// sampleTaskFn is a minimal function matching the standalone task signature.
+func sampleTaskFn(_ Context, _ any) (any, error) { return nil, nil }
+
+// sampleDurableFn is a minimal function matching the standalone durable task signature.
+func sampleDurableFn(_ DurableContext, _ any) (any, error) { return nil, nil }
+
+// WithCron and WithEvents on standalone tasks should appear in the registration request.
+
+func TestNewStandaloneTask_WithCron(t *testing.T) {
```
gregfurman: nit(non-blocking): these tests could be converted into testing tables quite nicely 👀

sdks/go/standalone_triggers_test.go
gregfurman, line 1: suggestion(non-blocking): Wdyt of adding a test to `sdks/go/examples/cron` to test these workflows e2e? There are a lot standalone cron workflows there that should've been flagged as broken if we had some tests 😬

gregfurman (approved):
LGTM. I think with some more pro-active testing on the examples this would've been caught a lot sooner but that is probably out-of-scope for this PR.

sdks/go/standalone_triggers_test.go
BloggerBust: That makes sense if we start reusing the stub. Since it is only used by this test file right now, my preference is to keep it local and refactor it out later if it gets reused.

sdks/go/standalone_triggers_test.go
BloggerBust: Thanks for this feedback. I agree, that your suggested level of coverage would have caught this sooner. My preference is to defer the extended coverage  to a follow-up PR. That way the scope of this PR stays focused on the SDK registration fix.

gregfurman (approved):

PR #4153, opened 2026-06-11
feat: v2 grpc protos

grutt:
# Description

Long lived feature branch for new V2 GRPC wire protocols. Simplifies API surface, clarifies naming consistency.

## Type of change

## What's Changed

- [ ] **New v2 proto contracts** under `api-contracts/v2/` defining five services with consistent naming:
  - `WorkerService` — `Register`, `Session` (single bidi stream replacing the v1 register+listen split), `UpsertLabels`, `Deregister`
  - `TaskService` — `TriggerTasks`, `CancelTaskRuns`, `ReplayTaskRuns`, `GetTaskRun`, `SubscribeTaskRuns`
  - `DurableTaskService` — `Session`, `Branch`
  - `EventService` — `PutEvents`, `ReplayEvents`
  - `LogService` — `PutLogs`, `PutStreamEvents`
  - Shared messages extracted into `v2/shared/` (`conditions.proto`, `task_config.proto`)
- [ ] **ConnectRPC codegen** — added `protoc-gen-connect-go` to `hack/proto/proto.sh`, generating `internal/services/shared/proto/v2` (`.pb.go`) and `v2connect` handler/client stubs
- [ ] **v2 service scaffolding** in `internal/services/v2/` — one implementation per service built on a shared `service` struct with functional options (`WithRepository`, `WithMessageQueue`, `WithLogger`, `WithAnalytics`); all RPCs currently return `CodeUnimplemented` as placeholders for follow-up PRs
- [ ] **Connect server wiring** — the gRPC `Server` now optionally starts a second HTTP server serving the v2 Connect handlers (`startConnect`), with new `WithWorkerV2`/`WithTaskV2`/`WithDurableTaskV2`/`WithEventV2`/`WithLogV2`/`WithConnectPort` options and combined cleanup
- [ ] **Connect auth middleware** — `ConnectAuthN` interceptor mirroring `GRPCAuthN` (bearer token → tenant lookup, analytics source/tenant context keys, OTel tenant attribute) for unary and streaming handlers
- [ ] **Config** — new `runtime.grpcV2Port` / `SERVER_GRPC_V2_PORT` (default `7071`) and engine startup wiring in `cmd/hatchet-engine/engine/run.go`

## Checklist

Todo:
- [ ] Realize Engine Service
- [ ] Typescript SDK
- [ ] Python SDK
- [ ] Go SDK
- [ ] Ruby SDK

Changes have been:

- [ ] Linted and formatted

<details id="ai-disclosure">

</details>


api-contracts/v2/shared/conditions.proto
```diff
@@ -0,0 +1,57 @@
...
+message UserEventMatchCondition {
+    BaseMatchCondition base = 1;
+    string event_key = 2; // (required) the user event key to match on
+    optional string event_scope = 3; // (optional) a scope for the user event condition, used for subsetting candidate events
+    optional google.protobuf.Timestamp consider_events_since = 4; // (optional) only match events created after this time
+}
+
+// TaskConditions gate the creation, queueing, skipping or cancellation of task
+// runs within a DAG.
+message TaskConditions {
```
grutt, line 46: Drop?

api-contracts/v2/shared/task_config.proto
```diff
@@ -0,0 +1,79 @@
+syntax = "proto3";
+
+option go_package = "github.com/hatchet-dev/hatchet/internal/services/shared/proto/v2";
+
+package v2;
+
+import "google/protobuf/duration.proto";
+
+enum RateLimitDuration {
```
mnafees: The naming has always been confusing to me as a user. I think we should call this as `RateLimitUnit`

api-contracts/v2/shared/task_config.proto
```diff
@@ -0,0 +1,79 @@
...
+// units. Limits are upserted when the task is registered: static keys declare
+// their total via limit, dynamic keys via limit_values_expression.
+message RateLimit {
+    oneof key {
+        string static_key = 1; // a static key for the rate limit
+        string key_expression = 2; // a CEL expression for determining the rate limit key
+    }
+    optional int32 limit = 3; // (optional) the total units available per window for a static key
+    optional int32 units = 4; // (optional) the number of units this task run consumes, default 1
+    optional string units_expression = 5; // (optional) a CEL expression for determining the number of units consumed
```
mnafees: It will be clearer if we called these `count` instead of `units`

api-contracts/v2/shared/task_config.proto
```diff
@@ -0,0 +1,79 @@
...
+    optional string limit_values_expression = 6; // (optional) a CEL expression for determining the total amount of rate limit units for a dynamic key
+    optional RateLimitDuration duration = 7; // (optional) the rate limit window, default MINUTE
```
mnafees: This can then be `RateLimitUnit unit`

api-contracts/v2/shared/task_config.proto
```diff
@@ -0,0 +1,79 @@
...
+    CONCURRENCY_LIMIT_STRATEGY_UNSPECIFIED = 0;
+    CONCURRENCY_LIMIT_STRATEGY_CANCEL_IN_PROGRESS = 1;
+    CONCURRENCY_LIMIT_STRATEGY_GROUP_ROUND_ROBIN = 2;
+    CONCURRENCY_LIMIT_STRATEGY_CANCEL_NEWEST = 3;
+}
+
+message Concurrency {
+    string expression = 1; // (required) a CEL expression producing the concurrency group key
+    optional int32 max_runs = 2; // (optional) the maximum number of concurrent runs per group key, default 1
+    optional ConcurrencyLimitStrategy limit_strategy = 3; // (optional) the strategy when the limit is reached, default CANCEL_IN_PROGRESS
```
mnafees, line 47: note: we need to be very clear about what the default strat is to the end user

api-contracts/v2/shared/task_config.proto
mnafees: Instead of 2 `optional`s can we use [`oneof`](https://protobuf.dev/programming-guides/proto3/#oneof)?

api-contracts/v2/logs.proto
```diff
@@ -0,0 +1,53 @@
...
+// LogService ingests log lines and stream events emitted by task runs. Both
+// RPCs are bulk-first; a single put is a one-element batch.
+service LogService {
+    rpc PutLogs(PutLogsRequest) returns (PutLogsResponse) {}
+
+    rpc PutStreamEvents(PutStreamEventsRequest) returns (PutStreamEventsResponse) {}
+}
+
+enum LogLevel {
+    LOG_LEVEL_UNSPECIFIED = 0;
```
mnafees, line 15: I am wondering if we need the `unspecified` enum val here rather than defaulting to something else since in the application code we might simply default to one of the levels?

igor-kupczynski (commented):
I really like the new shape, it feels "designed" and consistent, rather that incremental growth.

api-contracts/v2/shared/conditions.proto
```diff
@@ -0,0 +1,57 @@
...
+
+package v2;
+
+import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
+
+// ConditionAction is the action taken on the task run when a condition (or its
+// OR group) matches.
+enum ConditionAction {
+    CONDITION_ACTION_UNSPECIFIED = 0;
```
mrkaye97, line 13: I've been curious about this - why do we need `UNSPECIFIED` in a bunch of places? I find this to be kind of an odd pattern

api-contracts/v2/shared/task_config.proto
mrkaye97: +1

api-contracts/v2/shared/task_config.proto
mrkaye97: +1

api-contracts/v2/shared/task_config.proto
```diff
@@ -0,0 +1,79 @@
...
+    }
+    optional int32 limit = 3; // (optional) the total units available per window for a static key
+    optional int32 units = 4; // (optional) the number of units this task run consumes, default 1
+    optional string units_expression = 5; // (optional) a CEL expression for determining the number of units consumed
+    optional string limit_values_expression = 6; // (optional) a CEL expression for determining the total amount of rate limit units for a dynamic key
+    optional RateLimitDuration duration = 7; // (optional) the rate limit window, default MINUTE
+}
+
+enum ConcurrencyLimitStrategy {
+    CONCURRENCY_LIMIT_STRATEGY_UNSPECIFIED = 0;
```
mrkaye97, line 38: same here

api-contracts/v2/shared/task_config.proto
```diff
@@ -0,0 +1,79 @@
...
+message Concurrency {
+    string expression = 1; // (required) a CEL expression producing the concurrency group key
+    optional int32 max_runs = 2; // (optional) the maximum number of concurrent runs per group key, default 1
+    optional ConcurrencyLimitStrategy limit_strategy = 3; // (optional) the strategy when the limit is reached, default CANCEL_IN_PROGRESS
+}
+
+// WorkerPinning keeps a run and its child runs on the same worker, e.g. to
+// reuse expensive local state across tasks.
+enum WorkerPinning {
+    WORKER_PINNING_UNSPECIFIED = 0;
```
mrkaye97, line 53: same here

api-contracts/v2/shared/task_config.proto
```diff
@@ -0,0 +1,79 @@
...
+}
+
+message RetryPolicy {
+    int32 max_retries = 1; // (optional) the number of retries for the task run, default 0
+    optional float backoff_factor = 2; // (optional) the exponential backoff factor between retries
+    optional google.protobuf.Duration backoff_max = 3; // (optional) the maximum backoff between retries
+}
+
+enum WorkerLabelComparator {
+    WORKER_LABEL_COMPARATOR_UNSPECIFIED = 0;
```
mrkaye97, line 65: same here (gonna stop leaving these comments now)

api-contracts/v2/durable_task.proto
```diff
@@ -0,0 +1,184 @@
...
+//
+// The invocation_count on requests is a monotonically increasing count that
+// uniquely identifies an "attempt" at running a durable task. Each time the
+// task is started it gets a new invocation count, incremented by one from the
+// previous invocation. This lets the server and the worker differentiate
+// between attempts of the same task running in different places, preventing
+// race conditions from duplication, and allows older invocations to be evicted
+// cleanly.
+service DurableTaskService {
+    rpc Session(stream DurableSessionRequest) returns (stream DurableSessionResponse) {}
```
mrkaye97, line 24: did we rename this?

api-contracts/v2/durable_task.proto
```diff
@@ -0,0 +1,184 @@
...
+message WaitForAck {
+    EventLogEntryRef ref = 1;
+}
+
+// EntryCompleted notifies the worker that an awaited event log entry has a
+// result.
+message EntryCompleted {
+    EventLogEntryRef ref = 1;
+    bytes payload = 2;
+}
```
mrkaye97: missing the new fields from the other day (is_failure and error_message), and also, I kind of liked the verbose name for this `DurableTaskEventLogEntryCompletedResponse` even if it's long

api-contracts/v2/events.proto
```diff
@@ -0,0 +1,50 @@
...
+    // Both RPCs are bulk-first; a single event is a one-element batch.
+    rpc PutEvents(PutEventsRequest) returns (PutEventsResponse) {}
+
+    rpc ReplayEvents(ReplayEventsRequest) returns (ReplayEventsResponse) {}
+}
+
+message EventInput {
+    string key = 1; // (required) the key for the event
+    bytes payload = 2; // (optional) the payload for the event
+    google.protobuf.Timestamp occurred_at = 3; // (optional) when the event was generated, defaults to receipt time
```
mrkaye97: we use `seen_at` internally, so we should try to be consistent IMO

api-contracts/v2/events.proto
```diff
@@ -0,0 +1,50 @@
...
+    optional bytes extra = 4; // (optional) extra user metadata for the event
```
mrkaye97, line 22: are we renaming `additional_metadata`? IMO, this is a big change, and we need to think about it

api-contracts/v2/events.proto
```diff
@@ -0,0 +1,50 @@
...
+    optional int32 priority = 5; // (optional) the priority for runs triggered by this event
+    optional string scope = 6; // (optional) the filter scope, used for subsetting candidate filters at evaluation time
+}
+
+message PutEventsRequest {
+    repeated EventInput events = 1;
+}
+
+message Event {
+    string event_external_id = 1; // the external id of the event
```
mrkaye97, line 32: IMO, this should be `id`. We have a lot of icky prefixing around, and this is a good opportunity to get rid of it. It's obvious that this id is external since it's being sent back to the user, and it pertains to the event, since it's on the `Event` message

api-contracts/v2/shared/task_config.proto
grutt: we've called the number of "consumption" units so we'd need to change that too

alternatives here are Period or Window?

api-contracts/v2/events.proto
```diff
@@ -0,0 +1,50 @@
...
+    optional bytes extra = 5;
+    optional string scope = 6;
+}
+
+message PutEventsResponse {
+    repeated Event events = 1; // the created events, in request order
+}
+
+message ReplayEventsRequest {
+    repeated string event_external_ids = 1; // (required) the external ids of the events to replay
```
mrkaye97, line 45: same here: `ids`?

api-contracts/v2/task.proto
```diff
@@ -0,0 +1,183 @@
...
+    TASK_RUN_STATUS_FAILED = 4;
+    TASK_RUN_STATUS_CANCELLED = 5;
+    TASK_RUN_STATUS_EVICTED = 6;
+}
+
+message DefaultFilter {
+    string expression = 1; // (required) the CEL expression for the filter
+    string scope = 2; // (required) the scope of the filter
+    optional bytes payload = 3; // (optional) a JSON payload for the filter
+}
```
mrkaye97, line 50: While we're at it, we should add `DefaultAdditionalMetadata` too. We have a hack for that on the Py SDK right now, but should make it a "real" thing here

api-contracts/v2/task.proto
```diff
@@ -0,0 +1,183 @@
...
+
+// TaskDefinition declares a named, versioned task. A task either stands alone
+// or participates in a DAG via parents/conditions on tasks registered by the
+// same worker. Definitions are registered through the worker session
+// (RegisterWorker in worker.proto).
+message TaskDefinition {
```
mrkaye97: I dislike this for the same reason I regret that we called `v1_task` that - this really should just be `Task`, and everything else should be `TaskRun`

api-contracts/v2/task.proto
```diff
@@ -0,0 +1,183 @@
...
+    optional string description = 3; // (optional) a human-readable description
+    optional string version = 4; // (optional) the task version
+    optional google.protobuf.Duration execution_timeout = 5; // (optional) the maximum run duration once started
+    optional google.protobuf.Duration schedule_timeout = 6; // (optional) the maximum time a run may wait to be scheduled
+    optional RetryPolicy retry_policy = 7; // (optional) retry behavior, default no retries
+    repeated RateLimit rate_limits = 8; // (optional) the rate limits for the task
+    repeated Concurrency concurrency = 9; // (optional) the concurrency limits for the task
+    optional TaskConditions conditions = 10; // (optional) conditions gating run creation
+    repeated string parents = 11; // (optional) names of parent tasks in the same registration; if empty, this is a root task
+    optional WorkerPinning worker_pinning = 12; // (optional) keeps runs of this task and its children on the same worker
```
mrkaye97, line 67: is this the new name of sticky?

api-contracts/v2/task.proto
```diff
@@ -0,0 +1,183 @@
...
+    optional int32 default_priority = 13; // (optional) the default priority for runs of this task
+    map<string, int32> slot_requests = 14; // (optional) slot requests (slot_type -> units)
+    map<string, DesiredWorkerLabel> desired_worker_labels = 15; // (optional) desired worker label affinities
+    optional bytes input_schema = 16; // (optional) the JSON schema for the task input
+    bool is_durable = 17; // (optional) whether the task is durable
+    repeated string event_triggers = 18; // (optional) event keys that trigger runs of this task
+    repeated string cron_triggers = 19; // (optional) cron expressions that trigger runs of this task
+    repeated DefaultFilter default_filters = 20; // (optional) default filters applied to event triggers
+}
```
mrkaye97, line 85: `default_additional_metadata` here, probably

api-contracts/v2/task.proto
```diff
@@ -0,0 +1,183 @@
...
+
+// RegisteredTask is returned in RegisterWorkerAck (worker.proto), keyed by
+// task name.
+message RegisteredTask {
+    string task_id = 1; // the stable id of the task
```
mrkaye97, line 90: this is going to be _extremely_ confusing, since `v1_task` has an `id`, but presumably this corresponds to something closer to the current `Workflow` id

api-contracts/v2/task.proto
grutt: yes

api-contracts/v2/events.proto
grutt: yeah i had a note somewhere else on this, but ive always hated the name... its only existed because we had a "metadata" object on everything already and we rushed it

agreed it may be too large of a blast radius though

api-contracts/v2/task.proto
grutt: yeah im into this

api-contracts/v2/task.proto
```diff
@@ -0,0 +1,183 @@
...
+        ExtraFilter extra = 2; // subscribe to all runs matching an extra key/value
+    }
+}
+
+message ExtraFilter {
+    string key = 1;
+    string value = 2;
+}
+
+message TaskRunEvent {
```
mrkaye97, line 187: this is for stream events right? if yes, also very confusing naming (I think this is mostly a relic of the current code)

api-contracts/v2/events.proto
grutt: yeah i think i agree here

api-contracts/v2/events.proto
grutt: +1

api-contracts/v2/worker.proto
```diff
@@ -0,0 +1,193 @@
...
+    // returns a worker_id. It does not make the worker available for
+    // assignments; the worker is assignable only while a session is attached.
+    // Registrations that are never attached expire after a server-side TTL.
+    rpc Register(RegisterWorkerRequest) returns (RegisterWorkerResponse) {}
+
+    // Session must begin with an AttachWorker message referencing a registered
+    // worker_id. Attaching with an unknown or expired worker_id fails the
+    // stream with NOT_FOUND, in which case the worker should Register again
+    // before reattaching.
+    rpc Session(stream WorkerSessionRequest) returns (stream WorkerSessionResponse) {}
```
mrkaye97, line 37: what's a session?

api-contracts/v2/worker.proto
```diff
@@ -0,0 +1,193 @@
...
+// definitions in a single request; the worker can run the actions of the
+// tasks it registers. Registration is idempotent: re-registering with changed
+// task definitions returns the same worker_id with updated task versions.
+message RegisterWorkerRequest {
+    string name = 1; // (required) the name of the worker
+    repeated TaskDefinition tasks = 2; // (required) the task definitions this worker serves; tasks may reference each other as parents
+    map<string, int32> slot_config = 3; // (required) slot capacity for this worker (slot_type -> units)
+    map<string, WorkerLabel> labels = 4; // (optional) worker labels (i.e. state or other metadata)
+    optional RuntimeInfo runtime_info = 5; // (optional) information about the worker's runtime environment
+    optional string webhook_id = 6; // (optional) the id of the webhook the worker is associated with
```
mrkaye97: probably don't need webhook worker support anymore?

grutt: gc!

api-contracts/v2/worker.proto
```diff
@@ -0,0 +1,193 @@
...
+        ReleaseSlot release_slot = 6;
+    }
+}
+
+message AttachWorkerAck {
+    string worker_id = 1;
+}
+
+// Assignment instructs the worker to run a task.
+message Assignment {
```
mrkaye97: This is sort of confusing, because an "assignment" IMO is really an engine-side thing, where the engine has mapped a task run to a worker, so I think this naming is a bit odd. This looks to be to be what the old `AssignedAction` was?

grutt: perhaps TaskInvocation to start adopting the durable term more broadly?

mrkaye97: ugh yeah, I'm wary of overloading terms like this, but I do agree that'd be reasonable. let's think about it some more - really this basically "all the data we need to be able to run the task on the worker", so some term that encapsulates that would be ideal

grutt: i mean ideally its not overloaded imo, i think this should ALWAYS represent a single instruction to invoke the function

grutt: we just needed the count more strictly for durable

api-contracts/v2/task.proto
grutt: good catch, i didnt actually look too closely at this yet because im not sure what our plan is with new streams...

api-contracts/v2/logs.proto
```diff
@@ -0,0 +1,53 @@
...
+package v2;
+
+import "google/protobuf/timestamp.proto";
+
+// LogService ingests log lines and stream events emitted by task runs. Both
+// RPCs are bulk-first; a single put is a one-element batch.
+service LogService {
+    rpc PutLogs(PutLogsRequest) returns (PutLogsResponse) {}
+
+    rpc PutStreamEvents(PutStreamEventsRequest) returns (PutStreamEventsResponse) {}
```
abelanger5: doesn't really make sense to put stream events in the `LogService`, perhaps this should be its own service entirely? I think a dedicated `LogService` does make sense, if we'd like to add query and streaming log RPCs in the future

api-contracts/v2/worker.proto
```diff
@@ -0,0 +1,193 @@
...
+    optional string extra = 5;
+}
+
+// RegisterWorkerRequest establishes the worker and registers its task
+// definitions in a single request; the worker can run the actions of the
+// tasks it registers. Registration is idempotent: re-registering with changed
+// task definitions returns the same worker_id with updated task versions.
+message RegisterWorkerRequest {
+    string name = 1; // (required) the name of the worker
+    repeated TaskDefinition tasks = 2; // (required) the task definitions this worker serves; tasks may reference each other as parents
```
abelanger5: I think static registration of `TaskDefinition` limits our options: would much rather see a separate RPC for registering task definitions, and potentially a separate RPC for registering a list of tasks which a worker can run (so it can be updated) -- could be convinced the latter is a bad idea, but dynamically registering task definitions feels important

api-contracts/v2/worker.proto
```diff
@@ -0,0 +1,193 @@
...
+// server responds with AttachWorkerAck.
+message AttachWorker {
+    string worker_id = 1; // (required) the worker_id from RegisterWorkerResponse
+}
+
+message Heartbeat {
+    google.protobuf.Timestamp sent_at = 1; // (required) the time the worker sent the heartbeat
+}
+
+// AssignmentAck acknowledges receipt of an assignment before the worker starts
```
abelanger5: should `AssignmentAck` be blocking from an SDK perspective? will significantly increase latency if so.

api-contracts/v2/worker.proto
```diff
@@ -0,0 +1,193 @@
...
+message AttachWorkerAck {
+    string worker_id = 1;
+}
+
+// Assignment instructs the worker to run a task.
+message Assignment {
+    string task_run_external_id = 1; // the external id of the task run
+    string task_name = 2; // the name of the task
+    string action = 3; // the action id to run
+    bytes input = 4; // the input payload for the run
```
abelanger5, line 173: if we're breaking things, I'm wondering if perhaps we want to support a shared `Payload` type which can embed either a `bytes` message, a payload reference (via the `v1_payload` external id), and in the future could reference a set of codecs to support encryption / decryption.

api-contracts/v2/events.proto
abelanger5: `tags` or `labels` would be a better name

api-contracts/v2/task.proto
```diff
@@ -0,0 +1,183 @@
...
+    optional bytes payload = 3; // (optional) a JSON payload for the filter
+}
+
+// TaskDefinition declares a named, versioned task. A task either stands alone
+// or participates in a DAG via parents/conditions on tasks registered by the
+// same worker. Definitions are registered through the worker session
+// (RegisterWorker in worker.proto).
+message TaskDefinition {
+    string name = 1; // (required) the unique task name
+    string action = 2; // (required) the action id workers run for this task
```
abelanger5: do we want to consider renaming `action` to `queue`? it's what it maps to, would be nice to remove references to action from an SDK perspective

api-contracts/v2/task.proto
```diff
@@ -0,0 +1,183 @@
...
+    optional string description = 3; // (optional) a human-readable description
+    optional string version = 4; // (optional) the task version
+    optional google.protobuf.Duration execution_timeout = 5; // (optional) the maximum run duration once started
+    optional google.protobuf.Duration schedule_timeout = 6; // (optional) the maximum time a run may wait to be scheduled
+    optional RetryPolicy retry_policy = 7; // (optional) retry behavior, default no retries
+    repeated RateLimit rate_limits = 8; // (optional) the rate limits for the task
+    repeated Concurrency concurrency = 9; // (optional) the concurrency limits for the task
+    optional TaskConditions conditions = 10; // (optional) conditions gating run creation
+    repeated string parents = 11; // (optional) names of parent tasks in the same registration; if empty, this is a root task
```
abelanger5, line 66: what if I want to register a task as both a standalone task and a task which is part of a DAG with parents?

I do think supporting a `DAG` rpc to register DAGs would make sense, it's explicit. there's also a use-case for versioning DAGs separately from tasks which we should probably think about

abelanger5 (commented):
One more thing to consider: `namespace` as part of the API definitions rather than a client-only thing

api-contracts/v2/shared/task_config.proto
grutt: bumping this thread -- id vote for window

mrkaye97: window is fine with me

api-contracts/v2/worker.proto
grutt: no i think we optimistically run, ill clarify this in the docstring

api-contracts/v2/logs.proto
grutt: yes

api-contracts/v2/task.proto
grutt: yeah so i was thinking we add a "workflow_id" which can be the same as the task id (the root parent) else its the root parent all the way down the stack

grutt: shoot we're talking about the task not the task run

mrkaye97: my point exactly 😅

api-contracts/v2/task.proto
grutt: yeah i like this

api-contracts/v2/task.proto
grutt: thinking ahead... will this eventually evaluate a cel

PR #4156, opened 2026-06-12
fix: gate loadtest emission on registration

igor-kupczynski:
# Description

Context: this test failed on over 50% of PRs recently, this is an attempt at fixing it. It's automated, as in it comes from an agent, but it was reviewed by a human before submitting. We will only gain adopting it :) Bonus: it also improves error messages on failures.

Fixes the `load-online-migrate` flake by gating loadtest event emission on confirmed workflow registration instead of a blind worker-start sleep. This turns slow or broken registration against the N-1 schema into an explicit registration failure and prevents pre-registration events from being counted as phantom losses.

The PR also improves diagnostics for the online-migration job: engine startup now polls the engine-only gRPC port (`7077`) with process liveness checks, the loadtest uses `--registrationTimeout 20s`, and failure logs are uploaded as artifacts. The 20s timeout intentionally stays below the workflow's 30s pre-migration sleep so `PutWorkflow` must succeed on the N-1 schema rather than after the migration.

This restores the documented `WorkerDelay` behavior too: the delayed-worker case now emits while the registered worker waits to start, so queue buildup is actually exercised again. The mid-test migration is now anchored after confirmed registration rather than process start; on a healthy engine this should be sub-second, but the stress window is deliberately tied to registration readiness.

Fixes #

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] CI (any automation pipeline changes)
- [x] Test changes (add, refactor, improve or change a test)

## What's Changed

- [x] Added a registration gate and `--registrationTimeout` for `cmd/hatchet-loadtest` before event emission starts.
- [x] Mirrored the registration gate in the `rampup` loadtest variant.
- [x] Added unit coverage for registration success, timeout, and registration-error propagation.
- [x] Replaced blind `load-online-migrate` engine startup sleep with readiness/liveness checks and failure artifacts.

## Checklist

Changes have been:

- [x] Tested (unit, integration, or manually with steps specified)
- [x] Linted and formatted
- [ ] Documented (where applicable)
- [ ] Added to CHANGELOG (where applicable) -- see [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)

## Testing

- `go test ./cmd/hatchet-loadtest/... -run TestWaitForRegistration -count=1`
- `go test -c -tags load -o <tmp>/loadtest.test ./cmd/hatchet-loadtest`
- `go test -c -tags rampup -o <tmp>/rampup.test ./cmd/hatchet-loadtest/rampup`
[...]

PR #4159, opened 2026-06-12
fix: guard changelog docs sync

igor-kupczynski:
# Description

Fixes generated changelog drift that caused the `test / generate` check to fail after the TypeScript SDK changelog was updated without the corresponding docs changelog.

This keeps `task sync-changelog` as the public command, moves its implementation into a shared script, and wires that script into pre-commit so SDK-only changelog changes are checked during PR CI.

Related: #4156

## Type of change

## What's Changed

- Added `hack/docs/sync-changelog.sh` as the shared changelog docs generator.
- Updated `task sync-changelog` to delegate to the shared script.
- Added a local pre-commit hook so changelog docs are regenerated before CI merge checks.
- Regenerated the missing TypeScript docs changelog entry for `v1.24.0`.

## Checklist

Changes have been:

- [x] Linted and formatted

## Testing

- `task sync-changelog`
- `pre-commit run sync-changelog --all-files`
- `pre-commit run check-yaml --files .pre-commit-config.yaml Taskfile.yaml`
- `git diff --check`

---

<details id="ai-disclosure">

- **Details**: Used Cursor to investigate the GitHub Actions failure, implement the shared changelog sync guard, run focused verification, and draft this PR description.

</details>


igor-kupczynski:
@gregfurman What do you think about this one? It extracts the changelog gen to a common script, and then also runs it from the pre-commit hook. This should avoid stale changelogs going forward

frontend/docs/pages/reference/changelog/typescript.mdx
```diff
@@ -1,5 +1,11 @@
 {/* AUTOGENERATED — do not edit. Run `task sync-changelog` to regenerate from sdks/typescript/CHANGELOG.md */}
 
+## v1.24.0 - 2026-06-11
+
+### Added
+
+- Added a `getDetails` method to `hatchet.runs` to retrieve task details.
+
```
igor-kupczynski, line 14: One time sync

gregfurman: Duplicate of https://github.com/hatchet-dev/hatchet/pull/4160 -- but the diff is identical so shouldn't cause any issues if we merge the auto generated one

hack/docs/sync-changelog.sh
```diff
@@ -0,0 +1,45 @@
...
+          print ""
+        }
+        blank=0
+        print line
+        previous_heading=(line ~ /^#+ /)
+        next
+      }
+      !blank { print; blank=1 }
+      { previous_heading=0 }
+    '
```
gregfurman: This is to ensure the original generated changelogs are correctly formatted, right?

FYI, the `generate-docs` task already applies formatting to the generated docs file via `fmt-docs` -- which is why I never bothered to include any formatting/prettifying here.

Taskfile.yaml
```diff
@@ -77,24 +77,7 @@ tasks:
...
-    vars:
-      CHANGELOGS:
-        map:
-          platform: 'CHANGELOG.md'
-          python: 'sdks/python/CHANGELOG.md'
-          typescript: 'sdks/typescript/CHANGELOG.md'
-          ruby: 'sdks/ruby/src/CHANGELOG.md'
-    cmd:
-        for: { var: CHANGELOGS }
-        cmd: >-
```
gregfurman, line 89: Why not leave the matrix here and pass in the src and dest to the `sync-changelog.sh` e.g

```yaml
cmd:
    for: { var: CHANGELOGS }
    cmd: bash ./hack/docs/sync-changelog.sh  {{.ITEM}} {{.KEY}} 
```

.pre-commit-config.yaml
```diff
@@ -1,4 +1,12 @@
 repos:
+  - repo: local
+    hooks:
+      - id: sync-changelog
+        name: sync changelog docs
+        entry: bash hack/docs/sync-changelog.sh
+        language: system
+        pass_filenames: false
+        files: ^(CHANGELOG\.md|sdks/(python/CHANGELOG\.md|typescript/CHANGELOG\.md|ruby/src/CHANGELOG\.md)|frontend/docs/pages/reference/changelog/.*\.mdx|hack/docs/sync-changelog\.sh|Taskfile\.yaml)$
```
gregfurman: If we successfully update an mdx, this will trigger again due to `frontend/docs/pages/reference/changelog/.*` -- is this desirable?

Why not a simpler pattern that just checks if any of the `CHANGELOG.md` files are updated:

```suggestion
        files: CHANGELOG\.md$
```

.pre-commit-config.yaml
```diff
@@ -1,4 +1,12 @@
 repos:
+  - repo: local
+    hooks:
+      - id: sync-changelog
+        name: sync changelog docs
+        entry: bash hack/docs/sync-changelog.sh
```
gregfurman: nit: we could just pass in the direct `task sync-changelog` task here as well

frontend/docs/pages/reference/changelog/typescript.mdx
igor-kupczynski: #4160 is closed though

.pre-commit-config.yaml
igor-kupczynski: My thinking here was that a contributor may not have go-task installed, but maybe that's overblown? `lint.yaml` will likely need to install it though

igor-kupczynski:
Thanks @gregfurman, great feedback!

.pre-commit-config.yaml
igor-kupczynski: OK, for now I'm keeping it as a script, let me know what do you think

hack/docs/sync-changelog.sh
igor-kupczynski: Coming back to this one: dropping the formatting entirely is what's breaking CI now, so I've added a minimal blank-line normalization to the script (not full prettier). Reasoning:

The generated mdx files have two consumers with conflicting requirements:

1. The docs workflow runs `prettier --list-different` over `pages/**/*.mdx`, so the committed files must be prettier-clean.
2. The pre-commit hook requires the committed files to match the script output byte-for-byte. In the `lint` job there's no `fmt-docs` running after the hook, so "prettier will fix it later" doesn't apply on that path. That's the current failure: `sdks/typescript/CHANGELOG.md` has a double blank line, the script copies it verbatim, prettier had collapsed it in the committed mdx, and the hook flags the file as modified on every run.

Options I considered:

1. Run actual prettier in the script works, but pulls node/pnpm into every context where pre-commit runs (including the Go-centric lint job), same concern as the go-task discussion.
2. Just fix the stray blank line in the source changelog could work today, but the same failure comes back the next time someone writes a double blank line in a changelog, which is the manual discipline this PR is trying to remove.
3. Add the generated mdx to `.prettierignore` and commit verbatim output. It makes the script the sole formatter, but lets whatever whitespace is in the source changelogs leak into the docs files.
4. Minimal normalization in the script (chosen): collapse runs of blank lines and drop trailing ones. One line of awk, no new dependencies, and it's exactly the delta between raw extraction and what prettier produces for this content. I verified the output is byte-identical to the committed mdx for all three SDK changelogs, so the hook is now idempotent on both the lint and generate paths.

.pre-commit-config.yaml
```diff
@@ -1,4 +1,24 @@
...
+        language: system
+        pass_filenames: false
+        files: ^sdks/python/CHANGELOG\.md$
+      - id: sync-typescript-changelog
+        name: sync typescript changelog docs
+        entry: bash hack/docs/sync-changelog.sh sdks/typescript/CHANGELOG.md frontend/docs/pages/reference/changelog/typescript.mdx
+        language: system
+        pass_filenames: false
+        files: ^sdks/typescript/CHANGELOG\.md$
+      - id: sync-ruby-changelog
```
gregfurman, line 16: Should the root level `CHANGELOG.md` for the Hatchet platform also be included?

igor-kupczynski: gc, the root level is missing; however adding it also re-surfaced the script needs the minimal normalization from the earlier thread:
- Blank lines: root CHANGELOG.md has many double blank lines between sections
- Spaces: two bullet lines in source have double spaces that prettier collapses (Fix  layout → Fix layout).

SDK mdx files were already idempotent; platform hadn’t been regenerated through the normalized path though.

igor-kupczynski:
Let me take a step back here. @gregfurman, my intention with this PR was to make it harder to commit a CHANGELOG with an invalid format using a simple pre-commit hook. However, given the discussion we have here, it's more complex and controversial than I anticipated.

In your opinion: does it make sense to do it in the first place? Do you have a better alternative in mind, to avoid CHANGELOG related build failures?

gregfurman (approved):
Have given this another look and am happy for it to go in. Think guarding via pre commit hooks makes sense.

igor-kupczynski:
Merging! I have not addressed your final comment, but I'm happy to fast follow depending on the path we want to pick https://github.com/hatchet-dev/hatchet/pull/4159#discussion_r3412044230

PR #4163, opened 2026-06-12
fix(dashboard): persist sidebar scroll position across navigation

chandan-1427:
## Description
When navigating between routes using the sidebar (especially Settings items), 
the sidebar scroll position resets to the top instead of maintaining its 
previous position.

## Steps to Reproduce
1. Open the dashboard with enough sidebar items to require scrolling
2. Scroll down in the sidebar to the Settings section
3. Click any Settings nav item (General, API Tokens, etc.)
4. Observe sidebar jumps back to top

## Expected
Sidebar maintains its scroll position after navigation

## Root Cause
The sidebar scroll container has no scroll position persistence. On re-render 
triggered by route change, the scroll position resets to 0.

## Fix
- Add `scrollPosition` and `setScrollPosition` to `useSidebar` hook using 
  existing `useLocalStorageState` pattern (consistent with how `collapsed` 
  and `expandedWidth` are persisted)
- Restore scroll position on mount via `useEffect`
- Save scroll position on scroll with 100ms debounce to avoid excessive 
  localStorage writes

## Files Changed
- `frontend/app/src/components/hooks/use-sidebar.tsx`
- `frontend/app/src/components/v1/nav/side-nav.tsx`

- [x] _I acknowledge that an LLM was used in the creation of this Pull Request, 
in accordance with Hatchet's AI_POLICY.md._
- **Details**: Claude was used to help identify root cause and suggest fixes.

</details>

### DIFF
--- a/frontend/app/src/components/hooks/use-sidebar.tsx
+++ b/frontend/app/src/components/hooks/use-sidebar.tsx
@@ -14,7 +14,6 @@ type SidebarState = 'open' | 'closed';
 const SIDEBAR_WIDTH_LEGACY_KEY = 'v1SidebarWidth';
 const SIDEBAR_WIDTH_EXPANDED_KEY = 'v1SidebarWidthExpanded';
 const SIDEBAR_COLLAPSED_KEY = 'v1SidebarCollapsed';
-const SIDEBAR_SCROLL_KEY = 'v1SidebarScroll';
 
 // Tailwind's `md` breakpoint (px). Used to decide when the v1 sidebar is a persistent column.
 const WIDE_BREAKPOINT_PX = 768;
@@ -45,8 +44,6 @@ type SidebarProviderState = {
   toggleCollapsed: () => void;
   expandedWidth: number;
   setExpandedWidth: (width: number) => void;
-  scrollPosition: number;
-  setScrollPosition: (pos: number) => void;
 };
 
 const initialState: SidebarProviderState = {
@@ -60,8 +57,6 @@ const initialState: SidebarProviderState = {
   toggleCollapsed: () => null,
   expandedWidth: DEFAULT_EXPANDED_SIDEBAR_WIDTH,
   setExpandedWidth: () => null,
-  scrollPosition: 0,
-  setScrollPosition: () => null,
 };
 
 const SidebarProviderContext =
@@ -121,11 +116,6 @@ export function SidebarProvider({
     setCollapsed(!collapsed);
   }, [collapsed, setCollapsed]);
 
-  const [scrollPosition, setScrollPosition] = useLocalStorageState(
-    SIDEBAR_SCROLL_KEY,
-    0,
-  );
-
   useEffect(() => {
     const handleResize = () => {
       setIsWide(window.innerWidth >= WIDE_BREAKPOINT_PX);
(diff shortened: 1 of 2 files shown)


frontend/app/src/components/v1/nav/side-nav.tsx
```diff
@@ -252,6 +256,12 @@ export function SideNav({ className, navItems: navSections }: SideNavProps) {
     setIsResizing(false);
   }, [setStoredCollapsed, storedCollapsed]);
 
+  useEffect(() => {
```
BloggerBust: I think this restoration effect should avoid depending on `scrollPosition`. Since `scrollPosition` is updated by the scroll handler, this effect runs after every debounced scroll save. It also sets `scrollTop` programmatically, and can trigger another scroll event. The value may stabilize, but it creates unnecessary save -> render -> restore -> scroll-event churn.

Could we restore the scroll position on mount or after navigation instead of every time the saved scroll position changes? Also, `useLayoutEffect` would be better for restoring scroll position because it runs before paint and avoids a visible jump from the top back to the saved position.

frontend/app/src/components/v1/nav/side-nav.tsx
```diff
@@ -455,7 +465,16 @@ export function SideNav({ className, navItems: navSections }: SideNavProps) {
           <>
             {/* Scrollable navigation area (keep scrollbar flush to sidebar edge) */}
             <div
+              ref={scrollRef}
               data-cy="v1-sidebar-scroll"
+              onScroll={() => {
+                if (scrollSaveTimer.current) {
+                  clearTimeout(scrollSaveTimer.current);
+                }
+                scrollSaveTimer.current = setTimeout(() => {
```
BloggerBust: Can we add cleanup for this timer on unmount? that would avoid calling `setScrollPosition` after unmount if a debounce is still pending.

frontend/app/src/components/hooks/use-sidebar.tsx
```diff
@@ -116,6 +121,11 @@ export function SidebarProvider({
     setCollapsed(!collapsed);
   }, [collapsed, setCollapsed]);
 
+  const [scrollPosition, setScrollPosition] = useLocalStorageState(
```
BloggerBust: Do we need localStorage for this? The sidebar appears to stay mounted across these route changes, so a ref-only approach may be enough to preserve scroll position during navigation. That would avoid React state updates and localStorage writes while the user scrolls.

BloggerBust (changes requested):
Thanks for working on this @chandan-1427,

I left a couple of comments. The main issue is that restoring whenever `scrollPosition` changes can create unnecessary churn because `scrollPosition` is updated by the scroll handler itself. Otherwise, your fix looks correct.

chandan-1427:
> Thanks for working on this @chandan-1427,
> 
> I left a couple of comments. The main issue is that restoring whenever `scrollPosition` changes can create unnecessary churn because `scrollPosition` is updated by the scroll handler itself. Otherwise, your fix looks correct.

Thanks for the detailed review @BloggerBust , all your points are valid. The scroll position is naturally preserved by the DOM and no explicit save/restore is needed at all. I'll verify this is the case and if so, remove the scrollPosition/setScrollPosition state and localStorage key entirely, along with the restore effect and the debounced save. If there's a remount edge case where a ref fallback is needed, I'll make sure to add proper timer cleanup too.

BloggerBust (approved):
I validated this locally and the nav bar does not jump back up to the top. I also confirmed locally that on main it does. Fix LGTM.

PR #4164, opened 2026-06-12
fix: durable satisfied callback order guarantees

grutt:
# Description

Fixes # (issue)

## Type of change

## What's Changed

- [x] Adds a satisfied order column to the durable event log tables which represents the order which we've seen callbacks
- [x] Adds buffered checks on sdks to "park" non-continguous callbacks to ensure promise resolution determinism on restore

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Verify the bug, and port sdks from typescript "primary"

</details>


pkg/client/durable_ordered_release.go
```diff
@@ -0,0 +1,277 @@
...
+import (
+	"fmt"
+	"time"
+
+	v1 "github.com/hatchet-dev/hatchet/internal/services/shared/proto/v1"
+)
+
+const (
+	defaultParkTimeout = 5 * time.Second
+	defaultGapTimeout  = 60 * time.Second
```
grutt: I think these could be shorter

cmd/hatchet-migrate/migrate/migrations/20260611160000_v1_0_117.sql
```diff
@@ -0,0 +1,15 @@
...
+ALTER TABLE v1_durable_event_log_entry ADD COLUMN IF NOT EXISTS satisfied_order BIGINT;
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+ALTER TABLE v1_durable_event_log_entry DROP COLUMN IF EXISTS satisfied_order;
+-- +goose StatementEnd
+-- +goose StatementBegin
+ALTER TABLE v1_durable_event_log_file DROP COLUMN IF EXISTS latest_satisfied_order;
+-- +goose StatementEnd
```
mrkaye97: small bit of cleanup (and we need to rename this migration too btw)

```suggestion
-- +goose Up
-- +goose StatementBegin
ALTER TABLE v1_durable_event_log_file ADD COLUMN latest_satisfied_order BIGINT NOT NULL DEFAULT 0;
ALTER TABLE v1_durable_event_log_entry ADD COLUMN satisfied_order BIGINT;
-- +goose StatementEnd

-- +goose Down
-- +goose StatementBegin
ALTER TABLE v1_durable_event_log_entry DROP COLUMN satisfied_order;
ALTER TABLE v1_durable_event_log_file DROP COLUMN latest_satisfied_order;
-- +goose StatementEnd
```

cmd/hatchet-migrate/migrate/migrations/20260611160000_v1_0_117.sql
```diff
@@ -0,0 +1,15 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE v1_durable_event_log_file ADD COLUMN IF NOT EXISTS latest_satisfied_order BIGINT NOT NULL DEFAULT 0;
```
mrkaye97: what's the backwards-compat story here for existing callbacks btw? or for a durable task where it starts, it's waiting, then we deploy and run the migration but we're still midway through?

internal/services/dispatcher/server_v1.go
```diff
@@ -538,12 +539,16 @@ func (d *DispatcherServiceImpl) deliverSatisfiedEntries(taskExternalId string, r
...
+			// NOTE: memo entries are never stamped with a satisfied_order: their results
+			// travel synchronously on the MemoAck, so an ordered release would stall the
+			// completion sequence on entries the worker never awaits.
 			if err := d.DeliverDurableEventLogEntryCompletion(
 				taskExtId,
 				result.MemoResult.InvocationCount,
 				result.MemoResult.BranchId,
 				result.MemoResult.NodeId,
 				result.MemoResult.ResultPayload,
+				nil,
```
mrkaye97: I'm a little confused by this? feels like memo should be exactly the same as everything else? I guess I understand how the `memoack` on the worker might bypass existing logic, but it feels like a bug waiting to happen

cmd/hatchet-migrate/migrate/migrations/20260611160000_v1_0_117.sql
```diff
@@ -0,0 +1,15 @@
...
+-- +goose StatementEnd
+-- +goose StatementBegin
+ALTER TABLE v1_durable_event_log_entry ADD COLUMN IF NOT EXISTS satisfied_order BIGINT;
```
mrkaye97: this feels like it should be non-null going forward, although backfilling isn't really tractible

pkg/client/durable_ordered_release.go
mrkaye97: I don't love the naming here, I don't think it's very self explanatory - any other ideas for these? gap makes more sense to me than park and gate sweep

pkg/repository/sqlcv1/durable_event_log.sql
```diff
@@ -82,7 +83,35 @@ WHERE durable_task_id = @durableTaskId::BIGINT
   AND node_id = @nodeId::BIGINT;
 
 
+-- name: ListDurableEventLogEntriesBeforeNode :many
+SELECT *
+FROM v1_durable_event_log_entry
+WHERE durable_task_id = @durableTaskId::BIGINT
+  AND durable_task_inserted_at = @durableTaskInsertedAt::TIMESTAMPTZ
+  AND tenant_id = @tenantId::UUID
+  AND node_id < @nodeId::BIGINT
```
mrkaye97, line 92: should this also consider which branch we're on? since if we have e.g. two branches like this:

```
branch_id, node_id
1, 1
1, 2
1, 3
1, 4
2, 2
2, 3
2, 4
```

and we run this query for `@nodeId < 3` e.g., we'll get both branches first and second nodes, which may not be what we want?

pkg/repository/sqlcv1/durable_event_log.sql
```diff
@@ -82,7 +83,35 @@ WHERE durable_task_id = @durableTaskId::BIGINT
...
+-- acquisition order consistent across concurrent transactions.
+WITH inputs AS (
+    SELECT
+        UNNEST(@durableTaskIds::BIGINT[]) AS durable_task_id,
+        UNNEST(@durableTaskInsertedAts::TIMESTAMPTZ[]) AS durable_task_inserted_at
+)
+SELECT lf.durable_task_id
+FROM v1_durable_event_log_file lf
+JOIN inputs i ON (lf.durable_task_id, lf.durable_task_inserted_at) = (i.durable_task_id, i.durable_task_inserted_at)
+FOR UPDATE OF lf
```
mrkaye97: shouldn't we use `GetAndLockLogFile` for this instead? what's the case where we need to lock many log files at once?

pkg/repository/sqlcv1/durable_event_log.sql
```diff
@@ -82,7 +83,35 @@ WHERE durable_task_id = @durableTaskId::BIGINT
...
+;
```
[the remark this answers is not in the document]
mrkaye97: +1

pkg/repository/sqlcv1/durable_event_log.sql
```diff
@@ -91,19 +120,44 @@ WITH inputs AS (
...
         UNNEST(@childTaskErrorMessages::TEXT[]) AS child_task_error_message
+), to_stamp AS (
+    SELECT
+        e.durable_task_id,
+        e.durable_task_inserted_at,
+        e.branch_id,
+        e.node_id,
+        ROW_NUMBER() OVER (
+            PARTITION BY e.durable_task_id, e.durable_task_inserted_at
+            ORDER BY e.node_id, e.branch_id
```
mrkaye97: do we want to order by `(node, branch)` or `(branch, node)`?

pkg/repository/sqlcv1/durable_event_log.sql
```diff
@@ -91,19 +120,44 @@ WITH inputs AS (
...
+    FROM v1_durable_event_log_entry e
+    JOIN inputs i ON (e.durable_task_id, e.durable_task_inserted_at, e.branch_id, e.node_id) = (i.durable_task_id, i.durable_task_inserted_at, i.branch_id, i.node_id)
+    WHERE e.satisfied_order IS NULL
 ), updated AS (
     UPDATE v1_durable_event_log_entry
     SET
         is_satisfied = true,
-        satisfied_at = COALESCE(satisfied_at, NOW()),
+        satisfied_at = COALESCE(v1_durable_event_log_entry.satisfied_at, NOW()),
+        satisfied_order = COALESCE(v1_durable_event_log_entry.satisfied_order, slf.latest_satisfied_order + ts.stamp_offset),
```
mrkaye97: not sure I understand this offset thing, what's the goal here? seems like we should always just be adding 1?

pkg/repository/durable_events.go
```diff
@@ -1623,6 +1638,36 @@ func (r *durableEventsRepository) handleWaitFor(ctx context.Context, tx sqlcv1.D
...
 
+func (r *durableEventsRepository) latestSatisfiedOrderBeforeBranchPoint(ctx context.Context, tx sqlcv1.DBTX, tenantId uuid.UUID, nodeId, branchId int64, task *sqlcv1.FlattenExternalIdsRow, nextBranchIdToBranchPoint map[int64]*sqlcv1.V1DurableEventLogBranchPoint) (int64, error) {
+	entries, err := r.queries.ListDurableEventLogEntriesBeforeNode(ctx, tx, sqlcv1.ListDurableEventLogEntriesBeforeNodeParams{
+		Durabletaskid:         task.ID,
+		Durabletaskinsertedat: task.InsertedAt,
+		Tenantid:              tenantId,
+		Nodeid:                nodeId,
+	})
+	if err != nil {
+		return 0, fmt.Errorf("failed to list durable event log entries before branch point: %w", err)
```
mrkaye97, line 1719: this is kind of related to the previous comment on the query itself, but maybe we should be filtering for the branch id in the query, and that'd simplify this a lot?

pkg/repository/durable_events.go
```diff
@@ -1690,9 +1735,21 @@ func (r *durableEventsRepository) HandleBranch(ctx context.Context, tenantId uui
...
+	if err != nil {
+		return nil, err
+	}
+
 	logFile, err = r.queries.UpdateLogFile(ctx, tx, sqlcv1.UpdateLogFileParams{
-		BranchId:              sqlchelpers.ToBigInt(&newBranchId),
-		NodeId:                sqlchelpers.ToBigInt(&zero),
+		BranchId:             sqlchelpers.ToBigInt(&newBranchId),
+		NodeId:               sqlchelpers.ToBigInt(&zero),
+		LatestSatisfiedOrder: sqlchelpers.ToBigInt(&latestSatisfiedOrder),
```
mrkaye97, line 1821: is this right? I think we maybe should be resetting this to zero when the node is reset to zero?

pkg/repository/match.go
```diff
@@ -818,6 +819,22 @@ func (m *sharedRepository) processEventMatches(ctx context.Context, tx sqlcv1.DB
...
+		err = m.queries.LockDurableEventLogFiles(ctx, tx, sqlcv1.LockDurableEventLogFilesParams{
+			Durabletaskids:         lockTaskIds,
+			Durabletaskinsertedats: lockTaskInsertedAts,
+		})
+
+		if err != nil {
+			return nil, fmt.Errorf("failed to lock durable event log files: %w", err)
+		}
+	}
+
```
mrkaye97: is the risk here that something else could be writing `latestSatisfiedOrder` to a lower value?

pkg/repository/match.go
```diff
@@ -858,6 +875,11 @@ func (m *sharedRepository) processEventMatches(ctx context.Context, tx sqlcv1.DB
 
 		initialEntry.InvocationCount = cb.InvocationCount
 
+		if cb.SatisfiedOrder.Valid {
```
mrkaye97: Can we just use pointers instead of `pgtype.Int8`..... another refactor for this incoming, I hate the implicit-nullable stuff in these `pgtype` types

pkg/repository/match.go
```diff
@@ -1526,3 +1548,41 @@ func (m *sharedRepository) userEventCondition(orGroupId uuid.UUID, readableDataK
 func getDurableSleepEventKey(sleepId int64) string {
 	return fmt.Sprintf("sleep-%d", sleepId)
 }
+
+func sortedUniqueLogFileRefs(taskIds []int64, insertedAts []pgtype.Timestamptz) ([]int64, []pgtype.Timestamptz) {
```
mrkaye97: makes me a bit nervous assuming these ids and inserted ats are in the same order, are the same length, etc. - maybe better to pass a `[]IdInsertedAt` in here instead?

sdks/python/hatchet_sdk/contracts/v1/shared/condition_pb2_grpc.py
```diff
@@ -4,7 +4,7 @@
 import warnings
 
 
-GRPC_GENERATED_VERSION = '1.76.0'
+GRPC_GENERATED_VERSION = '1.81.1'
```
mrkaye97: we should revert this I think - in theory running `./generate.sh` wouldn't cause this

sdks/python/hatchet_sdk/clients/listeners/durable_event_listener.py
```diff
@@ -515,6 +530,130 @@ async def _handle_response(self, response: DurableTaskResponse) -> None:
...
+        if order <= gate.released:
+            # re-delivery of an already-released completion (e.g. after
+            # reconnect): bypass the gate
+            self._deliver_completion(completed_key, result)
+            return
+
+        gate.held[order] = (completed_key, result)
+        self._pump_gate(gate)
+
+    def _pump_gate(self, gate: OrderedReleaseGate) -> None:
```
mrkaye97: naming

cmd/hatchet-migrate/migrate/migrations/20260611160000_v1_0_117.sql
grutt: legacy should be nil and we should be okay, as we'll increment foward

cmd/hatchet-migrate/migrate/migrations/20260611160000_v1_0_117.sql
grutt: yeah exactly. i agree fwiw

sdks/python/examples/durable_eviction/test_durable_eviction.py
```diff
@@ -444,3 +445,25 @@ async def test_restore_idempotency(hatchet: Hatchet) -> None:
...
+
+    assert result["status"] == "completed"
+    assert result["a"] == {
+        "first": {"label": "a-first"},
+        "second": {"label": "a-second"},
+    }
+    assert result["b"] == {
+        "first": {"label": "b-first"},
+        "second": {"label": "b-second"},
+    }
```
mrkaye97, line 469: this is a legit regression test - fails reliably on main, passes on the branch, and I think reproduces the exact failure case

mrkaye97:
@grutt we need this one 😅

grutt:
oops!

mrkaye97 (approved):

PR #4192, opened 2026-06-15
feat(engine): byte-aware task-completion flushing to bound bulk-write memory

igor-kupczynski:
# Description

A completion burst of tasks carrying large outputs can drive a single `CreateTaskEvents` bulk insert to inline up to ~1GB of payloads into one bind parameter, which OOMs Postgres backends (`SQLSTATE 53200`, "out of memory") and surfaces to workers as gRPC failures. This was reproduced locally against `postgres:18` with a per-backend address-space ulimit: a 10k-child fanout × 1MB outputs OOM'd 4/4 runs.

Root cause on the engine side: the MQ sub-buffer (`internal/msgqueue/mq_sub_buffer.go`) batches task-completed messages purely **by count** (`SUB_BUFFER_SIZE`) with no awareness of payload size, so per-statement memory scales with `batch_count × payload_size × copies × concurrent_backends`.

This PR makes completion flushes **byte-aware**: a flush stops draining once accumulated payload bytes reach a configurable cap, bounding the size of each bulk DB write. A local experiment confirmed the theory — capping the sub-buffer kept the same workload OOM-free (per-statement bind params ~50MB instead of up to ~1GB). It complements the count cap (the primary limiter, default lowered to 10 on `main`) with a size-aware guard for large-payload bursts.

Fixes # (no issue)

## Type of change

## What's Changed

- [x] `internal/msgqueue/mq_sub_buffer.go`: add a `maxFlushBytes` knob (`WithMaxFlushBytes`) plumbed through `MQSubBuffer` -> `newMsgIDBuffer` -> `msgIdBuffer`. The flush drains via a byte-aware `drainForFlush()` that stops once accumulated payload bytes reach the cap, while always taking at least one message so an oversized message still drains (falls back to the count-only `drainN` when the cap is disabled).
- [x] `pkg/config/server/server.go`: new runtime config `TaskCompletionMaxFlushBytes` (env `SERVER_TASK_COMPLETION_MAX_FLUSH_BYTES`, **default 64MB**). The default is an intentionally loose backstop -- there was previously no byte limit at all -- well above normal flush sizes but below the ~126MB per-backend allocation that triggered the OOM; it can be tightened once we have production data.
- [x] `internal/services/controllers/task/controller.go` + `cmd/hatchet-engine/engine/run.go`: thread the config into the task-completion buffer via `WithCompletionMaxFlushBytes`.
- [x] Tests for byte-bounded flushing and the oversized-message guarantee.

Behavior: tiny-payload workloads still flush by count first (unchanged); fat-payload completion bursts are now additionally bounded by bytes. A value `<= 0` disables the byte cap.

> [!NOTE]
> Rebased/merged on top of `main`, which independently refactored the buffers (shared `bufferCore` + generic `drainN`), lowered the default `SUB_BUFFER_SIZE` 1000 -> 10, and already fixed the earlier global-vs-instance drain bug. This PR reconciles the byte cap into that new structure, so the latent-bug fix is no longer part of the diff.

## Checklist

Changes have been:

- [x] Linted and formatted

## Testing

- `go test ./internal/msgqueue/` passes, including new `TestFlushRespectsMaxBytes` (observed per-flush sizes `[1024, 4096x9, 3072]`, none exceed the 4096-byte test cap) and `TestFlushTakesAtLeastOneOversizedMessage` (each oversized message drains one-per-flush), alongside `main`'s new buffer tests.
- `go build` / `go vet` clean for the affected packages.

---

<details id="ai-disclosure">

- **Details**: Implemented with the Cursor agent (Claude); used for writing the byte-aware flush logic, config wiring, and the unit tests, following a human-reviewed plan grounded in a local OOM reproducer.

</details>


pkg/config/server/server.go
```diff
@@ -287,6 +287,16 @@ type ConfigFileRuntime struct {
 	// ReplayEnabled controls whether the server enables replay for tasks
 	ReplayEnabled bool `mapstructure:"replayEnabled" json:"replayEnabled,omitempty" default:"true"`
 
+	// TaskCompletionMaxFlushBytes caps the total payload bytes drained into a
+	// single task-completion flush (one bulk DB write), bounding per-statement
+	// memory so a burst of large task outputs can't blow up a Postgres backend.
+	// The completion buffer still flushes by count first; whichever cap is hit
+	// first wins. The default is a loose backstop (64 MiB) -- well above normal
```
mnafees: nit: PR description mentioned MB but here we are using MiB, whereas I don't think it should matter much in the app context but in the Postgres context I'd double check

igor-kupczynski: Good callout, I've clarified in the doc. FWIIW this value never research postgres; it is a limit enforced by the go code.

igor-kupczynski:
Closing this. After #4073 (default sub/pub buffer size 1000 -> 10, "flush when full") this byte cap can't engage at the default and so can't explain the OOMs it was written for.

Walking the code: the completion path is one-output-per-message end to end -- the dispatcher sends one `task-completed` message per task (`CompletedTaskMessage` -> a single `CompletedTaskPayload.Output`), and the sub-buffer drains up to `bufferSize` *messages* into one `CompleteTasks` bulk write. With output capped at ~4MB (`GRPCMaxMsgSize` = 4194304) and `bufferSize` = 10, a completion flush tops out at ~40MB -- under the 64MB cap, so `drainForFlush` would never trigger it.

The ~1GB bind parameter came from the old `bufferSize` = 1000 regime (1000 x ~1MB), which is what the customer logs and the local repro (https://github.com/hatchet-dev/hatchet-workstation/pull/29) were run against -- both predate #4073. So the hypothesis was right for the old engine but is obsolete once #4073 ships.

This change is only useful paired with *raising* the message buffer count (so the byte cap becomes the real safety limit for large payloads while small payloads still get a high count). As a standalone no-op at buffer=10 it's not worth carrying.

Separately, the current alerts (HDX) point at concurrency-strategy queries on us-east-1-shard-4, a different path that neither #4073 nor this PR touches; that's where the investigation is moving. Will reopen / revisit if we decide to bump the buffer count.

PR #4196, opened 2026-06-15
fix(python, ts): durable listener hanging from race condition

mrkaye97:
### Fixed

- Fixes a race condition in the durable event listener where an engine restart can cause the listener to get stuck because of the existence of two separate request queues that are not kept in sync, causing requests to hang indefinitely after being added to the old queue in some cases.

## Type of change

Changes have been:

- [x] Linted and formatted

<details id="ai-disclosure">

No AI use in this PR

</details>


mnafees (approved):

juliusgeo (approved):
LGTM!

mnafees (approved):

PR #3775, opened 2026-04-28
Fix: Add cron job to deactivate stale step concurrency configs 

mrkaye97:
# Description

Adds a cron job to deactivate stale records in `v1_step_concurrency` to keep concurrency loops more performant / less CPU hogging

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

PR #4202, opened 2026-06-15
feat(engine): cron to run stale concurrency cleanup

### LINKED ISSUE #3823: [BUG] Self hosted hatchet engine consistently corrupts its own database after around a week of being in use
We have a constant problem with the self-hosted version of the Hatchet engine, where it completely breaks after around 4–7 days of active use. When this happens, the Hatchet engine stops properly assigning tasks to workers, stops updating any workflow statuses, spams "context deadline exceeded" errors in the logs, and overloads the database. Disconnecting workers or restarting the engine does not help — the only solution is to completely reset the RabbitMQ and PostgreSQL databases and start from a clean slate. Also, creating a new Hatchet instance using the old database, without any workers connected and with an empty RabbitMQ state, still reproduces this issue.

We have two instances of the Hatchet engine in use: production and testing. The production one has only one tenant and a relatively static set of workers and workflows. The testing one has multiple tenants and a much bigger set of workers and workflows, separated with a namespace mechanism. This issue only happens on the testing instance.

**Environment**
- SDK: TypeScript v1.10.3
- Engine: Self hosted v0.74.14
- DB db.m8g.2xlarge - 8CPU 32 MEM with work_mem set to 64 mb

We have 4 different "corrupted" databases on hand - please suggest what should be investigated there.

Top database load while hatchet engine tries to use corrupted database:
Image

We have complete logs for last couple days, but its difficult to find a specific point of interest  - a lot of different errors basically start to appear at different times. Sample of logs while hatchet in engine in broken state:
Image

mrkaye97:
# Description

Bringing this back into the oss so it runs everywhere to clean up stale step concurrencies to prevent them from piling up and causing increased latency

Fixes https://github.com/hatchet-dev/hatchet/issues/3823

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

<details id="ai-disclosure">

- **Details**: Used Claude Code to "restore" the previous implementation from https://github.com/hatchet-dev/hatchet/pull/3775

</details>


grutt (approved):

gregfurman:
@mrkaye97 This addresses https://github.com/hatchet-dev/hatchet/issues/3823 right? 

Once merged, I'll follow up there.

mrkaye97:
> @mrkaye97 This addresses #3823 right?

yep, thanks for flagging - I'll link it

Issue #4210, opened 2026-06-16
[FEAT] hatchet-sdk (python, maybe other) should log that worker has successfully started


StefanBRas:
**Is your feature request related to a problem? Please describe.**
When starting a worker with the python sdk it will print
```
[INFO]  🪓 -- 2026-06-16 10:28:44,903 - ------------------------------------------
[INFO]  🪓 -- 2026-06-16 10:28:44,904 - STARTING HATCHET...
[INFO]  🪓 -- 2026-06-16 10:28:44,910 - starting runner...
```
And then nothing more. It then looks like it's stuck trying to start the runner

**Describe the solution you'd like**
It should print something like `Runner successfully started, waiting for tasks`

**Describe alternatives you've considered**
Remembering that it means it has actually started and telling other people that as well

**Additional context**

It's a minor thing but it has just been bugging me for a while haha.
It's mostly annoying that for an infrastructure team that does not work directly with hatchet, the logs makes it look like it couldn't start correctly.

I can make the PR if you agree with it

---

<details id="ai-disclosure">
<summary><b>🤖 AI Disclosure</b></summary>

- [ ] _I acknowledge that an LLM was used in the creation of this Issue, in accordance with Hatchet's [AI_POLICY.md](./AI_POLICY.md)._

No AI

</details>

mrkaye97:
Hey @StefanBRas - agreed! I'll add this, I'm working on a couple other small fixes now too

mrkaye97:
should be fixed!

PR #4219, opened 2026-06-16
fix(python): startup logging, retry behavior on sending step action events

mrkaye97:
# Description

### Fixed

- Properly suppresses gRPC fork support log lines on startup when fork support is disabled
- Improves logging around worker startup and shutdown
- Improves retry logic around sending step action events to the engine to better handle transient failures and avoid losing events

Fixes # (issue)

#4210 

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

<details id="ai-disclosure">

No AI use

</details>


sdks/python/hatchet_sdk/worker/worker.py
```diff
@@ -487,8 +487,8 @@ async def _aio_start(self) -> None:
         main_pid = os.getpid()
 
         logger.info("------------------------------------------")
-        logger.info("STARTING HATCHET...")
-        logger.debug(f"worker runtime starting on PID: {main_pid}")
+        logger.info("starting hatchet...")
+        logger.debug(f"worker starting on PID: {main_pid}")
```
igor-kupczynski, line 491: Nit, but the docs need an update as well; they show the "old" messages:

Image

sdks/python/hatchet_sdk/config.py
```diff
@@ -183,6 +184,17 @@ class ClientConfig(BaseSettings):
     force_shutdown_on_shutdown_signal: bool = False
     tenacity: TenacityConfig = TenacityConfig()
 
+    @model_validator(mode="after")
+    def warn_deprecated_fields(self) -> "ClientConfig":
+        if "grpc_enable_fork_support" in self.model_fields_set:
```
igor-kupczynski: Not an expert here, but we have this code some lines above:
```py
    model_config = create_settings_config(
        env_prefix="HATCHET_CLIENT_",
    )
```

so I think `HATCHET_CLIENT_GRPC_ENABLE_FORK_SUPPORT` env will result in `"grpc_enable_fork_support" in self.model_fields_set` so the warning will be emitted even on the happy path

sdks/python/hatchet_sdk/clients/dispatcher/dispatcher.py
```diff
@@ -58,6 +56,13 @@ def _get_or_create_client(self) -> DispatcherStub:
 
         return self.client
 
+    def _get_or_create_aio_client(self) -> DispatcherStub:
```
igor-kupczynski, line 59: There's one more "old" path here: https://github.com/hatchet-dev/hatchet/blob/main/sdks/python/hatchet_sdk/clients/dispatcher/dispatcher.py#L264-L266

igor-kupczynski (commented):
This makes sense to me, the improvements in env setup/logging and the changes to retry behavior. I'd consider splitting them, but the PR is still small and reasonable so that's not a strong opinion.

I've raised a few nit comments.

sdks/python/hatchet_sdk/worker/worker.py
mrkaye97: good catch, thanks!

sdks/python/hatchet_sdk/clients/dispatcher/dispatcher.py
mrkaye97: thanks!

sdks/python/hatchet_sdk/config.py
mrkaye97: oh good call, let me test this out again 👀

mrkaye97: okay yeah it does, I missed it because I'm swallowing deprecation warnings. hmm.... will figure out a way around this

mrkaye97: maybe I'll just remove it 🤔

igor-kupczynski (approved):
Thanks for addressing!

PR #4232, opened 2026-06-18
fix(sdks/python): send aio_put_stream events with grpc.aio

### LINKED ISSUE #4230: [BUG] Python SDK aio_put_stream can delay task starts when many streams run at once
**Describe the issue**

We saw this in our company development setup. We are using Hatchet 0.84 with long lived async Python workers.

Our workers stream LLM output to clients as many small SSE chunks. Each chunk calls `ctx.aio_put_stream(chunk)`. This means `PutStreamEvent` runs at high frequency inside each task run, and across many concurrent task runs.

`ctx.aio_put_stream` currently sends each stream event by calling the synchronous `put_stream` path in a thread:

```py
await asyncio.to_thread(self.put_stream, data)
```

`put_stream` then sends `PutStreamEvent` with the synchronous gRPC stub.

When many tasks stream at the same time, those calls can keep the shared thread pool busy. The runner also uses that thread pool to read assigned actions from its queue. When the pool is busy, new task runs can take a long time to start.

In our development setup, we saw warnings like this:

```text
THE TIME TO START THE TASK RUN IS TOO LONG, THE EVENT LOOP MAY BE BLOCKED
```

**Environment**

- Hatchet: 0.84
- SDK: Python SDK
- Worker type: long lived async Python workers
- Workload: LLM streaming, many small chunks per run, many concurrent runs

**Expected behavior**

`ctx.aio_put_stream` should not use the shared thread pool for each stream event.

It should send `PutStreamEvent` with `grpc.aio`, the async gRPC client. This keeps stream sending async and avoids making task start depend on thread pool availability.

The stream index should still be assigned before sending the event, so stream chunks keep their existing order.

**Code to reproduce, logs, or screenshots**

After seeing this in our development setup, I tried to reproduce it with self contained local checks.

First, I ran a microbenchmark with a local fake gRPC server. It compares the current path with an async gRPC path:

```py
await asyncio.to_thread(sync_stub.PutStreamEvent, request)
await aio_stub.PutStreamEvent(request)
```

The async gRPC path avoided filling the shared thread pool.

Second, I ran a local Hatchet engine worker test. It started 200 concurrent runs, and each run sent 175 stream chunks with `ctx.aio_put_stream`.

In that test:

- Current path: 185 task start warnings
- Async gRPC path: 0 task start warnings
- Stream chunks still arrived in order

---

<details id="ai-disclosure">

- **Details**: I used Claude Code and Codex to work on this.

</details>

juanluisdb:
# Description

Fixes #4230.

This changes `ctx.aio_put_stream` so it sends `PutStreamEvent` with a lazy `grpc.aio` events stub.

Before this change, `ctx.aio_put_stream` called the synchronous `put_stream` path through `asyncio.to_thread`. Under high streaming load, those synchronous gRPC calls can keep the shared thread pool busy. The runner also uses that pool to read assigned actions from its queue, so task starts can be delayed.

The stream index is still assigned before the RPC is awaited. This keeps the existing stream ordering behavior.

## Type of change

## What's Changed

- [x] Add a lazy async events stub to the Python events client.
- [x] Send `aio_put_stream` events with `grpc.aio` instead of `asyncio.to_thread`.
- [x] Share stream request construction between the sync and async stream paths.
- [x] Add unit coverage for async stream retry behavior and stream index ordering.

## Checklist

Changes have been:

- [x] Linted and formatted

## Testing

I ran these checks locally with the project virtualenv:

```text
pytest tests/unit/test_aio_put_stream.py -q -p no:cacheprovider
pytest tests/unit -q -p no:cacheprovider
ruff check .
mypy --config-file=pyproject.toml hatchet_sdk/clients/events.py hatchet_sdk/context/context.py tests/unit/test_aio_put_stream.py
black hatchet_sdk/clients/events.py hatchet_sdk/context/context.py tests/unit/test_aio_put_stream.py --check --diff --workers 1
pydoclint .
```

<details id="ai-disclosure">

- **Details**: I used Claude Code and Codex to work on this.

</details>


sdks/python/hatchet_sdk/context/context.py
```diff
@@ -544,7 +544,16 @@ async def aio_put_stream(self, data: str | bytes) -> None:
         :param data: The data to send to the Hatchet API. Can be a string or bytes.
         :return: None
         """
-        await asyncio.to_thread(self.put_stream, data)
+        try:
+            ix = self._increment_stream_index()
```
mrkaye97, line 548: so there's one very important implementation detail here that I need to think about it a bit, which is that this is not concurrency-safe. We might need to wrap this in a lock to make sure that works properly, I'll think about this a bit 🤔

sdks/python/hatchet_sdk/clients/events.py
```diff
@@ -289,29 +302,57 @@ def log(
...
+        async def put_stream_event(
+            request: PutStreamEventRequest,
+            metadata: tuple[tuple[str, str]],
+        ) -> PutStreamEventResponse:
+            return cast(
+                PutStreamEventResponse,
+                await aio_events_service_client.PutStreamEvent(  # type: ignore[misc]
+                    request, metadata=metadata
+                ),
+            )
```
mrkaye97: maybe we can factor this out so we don't re-create it on every call? small optimization

sdks/python/hatchet_sdk/context/context.py
juanluisdb: ok, i see, the same counter is used by both `put_stream` and `aio_put_stream` right?

sdks/python/hatchet_sdk/clients/events.py
juanluisdb: makes sense. Updated this so  `aio_stream` is not recreated per call, while keeping the aio stub lazy and the retry behavior the same.

sdks/python/hatchet_sdk/context/context.py
mrkaye97: Sorry for just circling back! But yes, it's the same counter

mrkaye97: Anyways, I think this is probably okay - I struggle to see a case where you'd put stream events concurrently are care about ordering

mrkaye97:
Hey @juanluisdb - I sort of commandeered your PR to expand the fixes to a couple more problematic places 😅 I'm going to keep working on it a bit, and hoping to get it merged tomorrow or Weds after some eyes from someone else on the team too. Thanks for the fix here!

juanluisdb:
> Hey @juanluisdb - I sort of commandeered your PR to expand the fixes to a couple more problematic places 😅 I'm going to keep working on it a bit, and hoping to get it merged tomorrow or Weds after some eyes from someone else on the team too. Thanks for the fix here!

Sounds cool, thanks! @mrkaye97

juliusgeo (approved):

Issue #4228, opened 2026-06-18
Go SDK Retries


igor-kupczynski:
How the Go SDK retries today, and what I propose to change. This first part documents the **current state** only; proposed changes to follow.

## Architecture

The v1 SDK (`sdks/go/`) is a thin wrapper over the legacy v0 client (`pkg/client/`). It adds no transport logic of its own — feature clients call the legacy REST client or the gRPC sub-clients directly, so all retry behavior lives in `pkg/client/`.

We have these categories of clients:

- **gRPC unary calls** — request/response RPCs.
- **gRPC streams and listeners** — long-lived subscriptions that Recv in a loop.
- **REST / HTTP** — the generated REST client.

Retry is configured in two places: the gRPC dial + interceptors in `pkg/client/client.go`, and the REST client, also built in `client.go` on the default `http.Client` with a request hook that adds the `Authorization: Bearer <token>` header and nothing else.

### 1. gRPC unary: single shared policy

All unary RPCs flow through one chained interceptor installed at dial time:

- Max 5 attempts.
- Exponential backoff with jitter (5s base, 10% jitter).
- 30s per-retry timeout.
- Retried codes: `ResourceExhausted`, `DeadlineExceeded`, `FailedPrecondition`, `Internal`, `Unavailable`.
- Globally disable-able via `NoGrpcRetry`.

This covers `WorkflowService`, `AdminService` v1, `Dispatcher`, `V1Dispatcher`, and `EventsService`. The SDK's `Admin()` holds both the v0 `WorkflowService` and v1 `AdminService` stubs, and both wrap the same retry conn — so every admin RPC the SDK issues is retried. This layer is uniform.

Note: a second, retry-less gRPC dialer exists at `pkg/client/v1/grpc-client.go`, but it is **not** an SDK path. It is `hatchet-api`'s internal REST→gRPC proxy client, which forwards inbound REST calls to the engine's gRPC admin service. We consider it **out of scope** for this project.

### 2. gRPC streams and listeners: each is different

The retries here can mean two things:

[...]

PR #4240, opened 2026-06-19
feat(sdks/go): add retry foundation for REST reads

igor-kupczynski:
# Description

Adds the first scoped Go SDK retry change from #4228: shared retry helpers, REST retries for bodyless reads, gRPC retry construction cleanup with full-jitter backoff, and retry controls to disable SDK client retries.

Refs #4228

## Type of change

## What's Changed

- Adds shared retry/backoff helpers to GoSDK (in `pkg/client/retry/`)
- REST
  - The main thing to review is `pkg/client/retry/rest.go` --> `func (d *RestDoer) Do(req *http.Request) (*http.Response, error)`
  - The rest client uses this _doer_
  - The retry behavior: 5 attempts with jitter: `0-250ms, 0-500ms, 0-1s, 0-2s`
  - Eligible for retry: bodyless `GET`/`HEAD`  (it's not possibly to retry a request with body as that body can be consumed only once; we'd need to manually manage save/cacheing it and forgo splicing)
- gRPC
  - The main thing to review is `pkg/client/retry/grpc.go` --> `func GRPCDialOptions(l *zerolog.Logger, enabled bool) []grpc.DialOption`
  - Most of the gRPC retry behavior stays the same for this PR, the only changes are:
  - Change: the retry now uses jitter: `0-5s, 0-10s, 0-20s, 0-40s`
  - Change: `FailedPrecondition` won't be retried (see below)

### REST body replay

Bodied REST requests (`POST`, `PUT`, `PATCH`, and other methods with a body) are excluded from retry scope. In Go, an `http.Request` body is a one-shot `io.ReadCloser` unless `GetBody` is populated or the retry layer buffers and reconstructs the body. The generated REST clients do not set `req.GetBody`, so retrying bodied calls would require body replay plumbing and per-endpoint idempotency review. This PR limits REST retries to bodyless `GET`/`HEAD` reads.

### FailedPrecondition audit

Server-side audit of current `codes.FailedPrecondition` producers:

- The only production server return is `DispatcherImpl.Heartbeat` in `internal/services/dispatcher/server.go`, when a worker has a prior `ListenV2` listener session but `IsActive` is false.
- The only user-facing Go SDK path that observes it is the worker background heartbeat loop in `pkg/client/dispatcher.go`.
- Retrying that error does not reopen the listener stream; it only blocks the heartbeat goroutine behind gRPC retry backoff while listener reconnect remains driven separately by `ListenV2.Recv` errors.

Dropping `FailedPrecondition` from retryable gRPC codes is therefore a low-risk semantic cleanup: it is a rejected heartbeat / inactive-listener signal, not a transient transport fault.

### Duplicate gRPC write execution

This PR does **not** reduce duplicate-write risk from gRPC retries. The unary interceptor still retries all unary RPCs, including writes, with no method-level classification. That behavior change is deferred to a later semantic retry-policy PR.

### Retry controls

- `HATCHET_CLIENT_NO_RETRY` disables both REST read retries and gRPC SDK client retries.
- `HATCHET_CLIENT_NO_GRPC_RETRY` remains the existing gRPC-only retry control and continues to disable only the gRPC retry interceptor.

## Checklist

Changes have been:

- [x] Linted and formatted

## Testing

Latest verification:

- `task fmt-go`
- `task fmt-docs`
- `go test ./pkg/client/retry ./pkg/client ./pkg/config/client`
- pre-commit hooks on committed changes, including `golangci-lint`

## Related

This PR intentionally does not add gRPC method-level retry classification, REST write retries, or stream/listener reconnect normalization. Those are planned follow-ups from #4228.

---

<details id="ai-disclosure">

- **Details**: Cursor agents were used to review the implementation, refine docs copy, identify simplification opportunities, and prepare this PR description. The final changes were checked with focused Go tests, docs formatting, and pre-commit hooks.

</details>

### DIFF
--- a/pkg/client/retry/backoff.go
+++ b/pkg/client/retry/backoff.go
@@ -10,10 +10,10 @@ import (
 )
 
 const (
-	// gRPC retry timing keeps the existing 5s exponential shape, but uses
-	// full jitter so clients spread retry pressure across the whole delay window.
-	// Consecutive retry delay windows are 0-5s, 0-10s, 0-20s, 0-40s,
-	// then capped at 0-80s.
+	// gRPC retry backoff uses full jitter over exponential windows so clients
+	// spread retry pressure. After the initial RPC fails, middleware invokes
+	// backoff with attempt=1 for the first retry; delay windows are 0-5s,
+	// 0-10s, 0-20s, 0-40s, then capped at 0-80s.
 	grpcMaxRetries    = 5
 	grpcBaseDelay     = 5 * time.Second
 	grpcBackoffFactor = 2.0
--- a/pkg/client/retry/backoff.go
+++ b/pkg/client/retry/backoff.go
@@ -1,0 +1,125 @@
+package retry
+
+import (
+	"context"
+	"math"
+	"math/rand/v2"
+	"net/http"
+	"strconv"
+	"time"
+)
+
+const (
+	// gRPC retry timing keeps the existing 5s exponential shape, but uses
+	// full jitter so clients spread retry pressure across the whole delay window.
+	// Consecutive retry delay windows are 0-5s, 0-10s, 0-20s, 0-40s,
+	// then capped at 0-80s.
+	grpcMaxRetries    = 5
+	grpcBaseDelay     = 5 * time.Second
+	grpcBackoffFactor = 2.0
+	grpcMaxDelay      = 80 * time.Second
+
+	// REST retry timing defaults keep SDK calls responsive while still covering
+	// short gateway and rate-limit blips. With 5 total attempts, retry delay
+	// windows are 0-250ms, 0-500ms, 0-1s, and 0-2s, unless a valid Retry-After
+	// header supplies the delay for an HTTP 429 response.
+	restMaxAttempts       = 5
+	restBaseDelay         = 250 * time.Millisecond
+	restBackoffFactor     = 2.0
+	restMaxDelay          = 5 * time.Second
+	restMaxRetryAfter     = 5 * time.Second
+	restPerAttemptTimeout = 30 * time.Second
+
+	// restDiscardDrainLimitBytes is a heuristic cap, not a protocol threshold.
+	// Draining small intermediate retry responses lets net/http reuse the TCP
+	// connection after Close. The cap is large enough for typical API error
+	// bodies, but small enough that an unexpectedly large or slow discarded body
+	// does not delay the next retry for long.
+	restDiscardDrainLimitBytes = 64 * 1024
+)
+
+type clock struct {
+	now    func() time.Time
+	sleep  func(context.Context, time.Duration) error
+	jitter func(int64) int64
+}
+
+func defaultClock() clock {
+	return clock{
+		now:    time.Now,
+		sleep:  sleepContext,
+		jitter: rand.Int64N,
+	}
+}
+
+func sleepContext(ctx context.Context, d time.Duration) error {
+	if d <= 0 {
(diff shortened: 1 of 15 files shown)


abelanger5 (approved):

pkg/client/retry/rest.go
```diff
@@ -0,0 +1,202 @@
...
+
+		delay := d.restBackoffDelay(attempt, resp)
+		if err := d.clock.sleep(ctx, delay); err != nil {
+			return nil, err
+		}
+	}
+
+	if lastTransportErr != nil {
+		if ctx.Err() != nil {
+			return nil, ctx.Err()
```
chandan-1427, line 127: Hey @igor-kupczynski  , I'm just a contributor reading your issue #4228 and exploring your PRs. Noticed that if restMaxAttempts = 0 the Do method returns (nil, nil) the loop never runs, lastTransportErr is never set, and the final return nil, ctx.Err() returns nil. Don't know if this is worth mentioning but just did.

pkg/client/retry/backoff.go
```diff
@@ -0,0 +1,125 @@
...
+	"math/rand/v2"
+	"net/http"
+	"strconv"
+	"time"
+)
+
+const (
+	// gRPC retry timing keeps the existing 5s exponential shape, but uses
+	// full jitter so clients spread retry pressure across the whole delay window.
+	// Consecutive retry delay windows are 0-5s, 0-10s, 0-20s, 0-40s,
```
mnafees: nit: I have a suspicion about the 0-5s window comment here. Are we sure this is the case as the grpc library depends on attempt > 0 to actually retry?

mnafees (approved):
LGTM to me overall, thanks for the work!

pkg/client/retry/rest.go
igor-kupczynski: Hey @chandan-1427 thanks for chiming in! Good catch. Today this is not reachable because `restMaxAttempts` is an internal constant set to `5`, but I agree relying on that invariant implicitly is a footgun if we later make retry attempts configurable.

I’m going to make the invariant explicit in the internal REST retry constructor/config path: validate that max attempts is at least `1` and error otherwise. This keeps the public SDK surface unchanged while avoiding the possible `(nil, nil)` return if the internal config ever drifts.

pkg/client/retry/backoff.go
igor-kupczynski: good catch @mnafees the middleware calls the backoff function first with `attempt=1`, so as written this produces 0-10s, not the documented 0-5s. I think I'll normalize the middleware attempt to zero-based.

pkg/client/retry/rest.go
igor-kupczynski: Fix in https://github.com/hatchet-dev/hatchet/pull/4240/commits/e97b6dff060acd05f71d57742d26f8757818ef8a

pkg/client/retry/backoff.go
igor-kupczynski: https://github.com/hatchet-dev/hatchet/pull/4240/changes/9fb8bce8de17e20f1c9ee57398cbc9b115d245dd

pkg/client/retry/rest.go
chandan-1427: > Fix in [e97b6df](https://github.com/hatchet-dev/hatchet/commit/e97b6dff060acd05f71d57742d26f8757818ef8a)

thanks for the context

PR #4246, opened 2026-06-20
fix(python-sdk): handle empty proto bytes in durable memo ack on replay

### LINKED ISSUE #4235: [BUG] DurableContext._aio_memo (Python SDK v1.33.9) crashes on durable-task replay
**Describe the issue**

`DurableContext._aio_memo` (Python SDK v1.33.9) crashes on durable-task replay when the engine returns a memo ack with `memo_already_existed=True` but no committed result payload. The unset proto field deserializes to empty bytes `b''` rather than `None`, slips past the SDK's `is not None` guard, and is passed to `TypeAdapter.validate_json(b'')`, raising:

```
pydantic_core.ValidationError: 1 validation error for <YourMemoModel>
  Invalid JSON: EOF while parsing a value at line 1 column 0 [type=json_invalid, input_value=b'', input_type=bytes]
```

Root cause is two spots:

1. `clients/listeners/durable_event_listener.py` (~line 408) builds the ack with `memo_result_payload=memo_ack.memo_result_payload`, reading the proto field directly. Since `memo_result_payload` is `optional bytes` in `v1/dispatcher.proto`, an **unset** field returns proto3's default `b''` (not `None`) — no `HasField` check. (The sibling `DurableTaskEventLogEntryResult.from_proto` in the same file does this correctly via a truthiness check `if ... proto.payload:`.)

2. `context/context.py` (~lines 1125–1136) then guards only against `None`:
   ```python
   if ack.memo_already_existed and ack.memo_result_payload is None:
       logger.warning("memo key found ... but no data was returned. rerunning ...")
   if ack.memo_already_existed and ack.memo_result_payload is not None:   # b'' passes here
       result = adapter.validate_json(ack.memo_result_payload, ...)       # validate_json(b'') -> crash
   else:
       result = await fn(*args, **kwargs)
   ```
   With `b''`, the recompute branch is skipped and `validate_json(b'')` raises.

This is triggered when a worker is evicted/SIGTERM'd in the window after the memo node is registered (`send_event(MemoEvent(..., result=None))`) but before `send_memo_completed_notification` durably commits — e.g. a rolling deploy while a durable task is suspended in `aio_wait_for_event`. On replay the engine reports the node as existing with an empty payload, and every replay re-crashes identically.

**Environment**
- SDK: Python `hatchet-sdk` v1.33.9 (pydantic 2.13, Python 3.12)
- Engine: Cloud

**Expected behavior**

An existing memo node with no committed result payload should be treated the same as `None` — i.e. fall through to the recompute branch (the warning log already anticipates this case: "memo key found in durable storage but no data was returned. rerunning the function to recompute the value") — rather than attempting `validate_json(b'')` and raising a `ValidationError`.

Minimal, no engine needed — demonstrates the proto-mapping + guard gap directly:

```python
from hatchet_sdk.contracts.v1.dispatcher_pb2 import DurableTaskEventMemoAckResponse
from pydantic import TypeAdapter, BaseModel

resp = DurableTaskEventMemoAckResponse(memo_already_existed=True)  # payload left unset
assert resp.HasField("memo_result_payload") is False
assert resp.memo_result_payload == b""        # proto3 default, NOT None  <-- the trap

class MyMemo(BaseModel):
    job_id: str

payload = resp.memo_result_payload            # what listener line ~408 stores in the ack
assert (payload is None) is False             # the context.py guard misses it
TypeAdapter(MyMemo).validate_json(payload)    # ValidationError: Invalid JSON: EOF ... input_value=b''
```

Production crash traceback (replay after a rolling worker deploy):

```
File ".../hatchet_sdk/context/context.py", line 1132, in _aio_memo
    result = adapter.validate_json(serialized_result, context=HATCHET_PYDANTIC_SENTINEL)
File ".../pydantic/type_adapter.py", line 492, in validate_json
    return self.validator.validate_json(...)
pydantic_core._pydantic_core.ValidationError: 1 validation error for MyMemo
  Invalid JSON: EOF while parsing a value at line 1 column 0 [type=json_invalid, input_value=b'', input_type=bytes]
```

Correlated worker shutdown log right before the stranding deploy:
```
received signal SIGTERM...
gracefully exiting runner...
evicted N waiting durable run(s) during shutdown
```

**Suggested fix**

Respect proto field presence in `durable_event_listener.py` so an unset payload maps to `None`:

```python
memo_result_payload=(
    memo_ack.memo_result_payload
    if memo_ack.HasField("memo_result_payload")
    else None
),
```

(Or, defensively, in `context.py` treat a falsy payload as recompute: `if ack.memo_already_existed and not ack.memo_result_payload:`.)

**Additional context**

Even with the crash fixed, the recovery path re-runs `fn()`. For a non-idempotent `fn` that performs an external side effect (e.g. submitting a billed third-party job and returning its id), recompute re-runs the side effect — the original side effect already happened but its id was lost with the uncommitted payload, and it's unrecoverable from inside `_aio_memo`. The "compute once" guarantee only holds for idempotent `fn`s. Worth documenting that memoized functions should be idempotent, or offering an idempotency hook.

---

<details id="ai-disclosure">

- **Details**: Claude (Anthropic, via Claude Code) was used for the full investigation — diagnosing the proto/guard root cause from production logs, constructing the minimal no-engine reproduction, and drafting this issue. Findings were verified by a human against the installed SDK source.

</details>

chandan-1427:
## Description

Fixes #4235

When a worker is evicted (SIGTERM/rolling deploy) in the window between memo node registration and result commit, the engine returns `memo_already_existed=True` with an unset payload. Proto3 deserializes unset `bytes` fields as `b''` rather than `None`, which slipped past the `is not None` guard in `context.py` and caused `validate_json(b'')` to raise a Pydantic `ValidationError` on every subsequent replay — permanently breaking the durable task run.

## Type of change

## What's Changed

- `durable_event_listener.py`: map `b''` to `None` using `or None` pattern when building the memo ack, consistent with existing `error_message` handling at line 132
- `context.py`: replace `is not None` with a truthiness check so an empty payload correctly falls through to the recompute branch instead of crashing

## Checklist

Changes have been:
- [x] Tested (manually reproduced the crash using the repro script from #4235, verified fix resolves it)
- [ ] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Used Claude for root cause analysis and reproduction script validation. Code changes written and verified manually.

</details>


mrkaye97:
Hi @chandan-1427, thanks for the fix! But I'm curious: how are you hitting this scenario? Memo isn't exported right now, so it shouldn't be getting used except for internally

chandan-1427:
Hey @mrkaye97, great question. Users don't call `_aio_memo` directly — but these public durable context methods call it internally:

- `ctx.aio_now()` → calls `_aio_memo` directly
- `ctx.aio_sleep_until()` → calls `aio_now()` on every invocation
- `ctx.aio_wait_for_event(lookback_window=...)` → calls `aio_now()` when `lookback_window` is provided

So the trigger path is:
1. A durable task calls any of the above methods
2. `_aio_memo` registers the memo node with the engine
3. Worker receives SIGTERM (e.g. rolling deploy) before the result is committed
4. On replay, the same method is called again
5. Engine returns `memo_already_existed=True` with an unset payload (`b''`)
6. The `is not None` guard passes `b''` to `validate_json(b'')` → crash on every replay

mrkaye97:
Makes sense, thanks! Can you write a changelog entry and bump the patch version of the library? Then we can merge

chandan-1427:
> Makes sense, thanks! Can you write a changelog entry and bump the patch version of the library? Then we can merge

sure

chandan-1427:
> Makes sense, thanks! Can you write a changelog entry and bump the patch version of the library? Then we can merge

done

mrkaye97:
awesome, thanks! looks like some linting is needed? you can probably `task generate-all` to fix it

chandan-1427:
> awesome, thanks! looks like some linting is needed? you can probably `task generate-all` to fix it

yeah linting has some issues with me, let me solve it

mrkaye97 (approved):

chandan-1427:
> awesome, thanks! looks like some linting is needed? you can probably `task generate-all` to fix it

Sorry for the back and forth! Linting is sorted now

PR #4257, opened 2026-06-22
feat(sdks/go): stream reconnect and lifecycle for listeners

igor-kupczynski:
# Description

Stream/listener half of #4228, building on the retry primitives from #4240 (merged).

The legacy Go SDK (`pkg/client`) keeps four long-lived gRPC streams: workflow-run subscriptions, durable event subscriptions, the worker action stream, and the metadata event stream. Each had its own reconnect logic layered on top of interceptor retry, with different behaviors, e.g.
- `Workflow.Result()` could hang forever if the shared listener died
- the worker action listener gave up after 5 transient `Recv` failures (which could killed the worker on a flaky network)
- a caller cancelling its context could tear down a listener shared with other callers

This PR replaces that with one explicit app-level reconnect model (full-jitter backoff, 1s base / 30s cap) shared by every stream.

## Behavior changes

- Sync paths (initial subscribe, `AddWorkflowRun`, `AddSignal`, sends): bounded reconnect — 5 attempts, then fail.
- Background listen loops: reconnect indefinitely on transient errors; permanent errors and 10 consecutive no-progress failures stop the loop and surface to callers.
- `Workflow.Result()` fails fast when subscribing fails, and waiters are unblocked if the listener dies permanently.
- The worker action listener reconnects until closed instead of dying after 5 transient failures (V2→V1 `Unimplemented` fallback preserved).
- Reconnects run on the stream's lifecycle context, so caller ctx cancellation never kills a shared listener. Interceptor stream retry is disabled on the RPCs the app-level loop now owns.

## Review guide

Four shared components are the main part to look at and then listeners are thin wrappers over them:

1. `stream_classify.go`: four-verdict error classifier used by both recv and reconnect paths.
2. `reconnecting_stream.go` -->`reconnectingStream[C]`: singleflighted connects, generation counter for stale clients, bounded sync connect, lifecycle ctx owned by `Close()`.
3. `stream_listen.go`: the single flat listen loop (backoff sleeps, no-progress cap).
4. `handler_registry.go` --> `handlerRegistry[K,E]`: RWMutex over plain maps, registration-id-guarded removal, `failAll` on permanent death.

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

## Testing

- `go build ./... && go test ./pkg/client/... ./pkg/worker/... -count=1 -race`
- Unit tests per shared component: classifier, reconnecting stream (coalescing, generations, bounded sync connect), listen loop (no-progress cap, unbounded transient reconnect, EOF policy), handler registry (stale removal, `failAll`, races).
- Per-listener behavior tests: bounded sync subscribe/send, worker V2→V1 fallback and survival past 5 transient failures, `Workflow.Result()` fail-fast and permanent-death notification, handler-error isolation.

---

<details id="ai-disclosure">

- **Details**: Cursor (Claude) used for implementation, test coverage, review follow-ups, and PR description drafting.

</details>


pkg/client/retry/stream.go
```diff
@@ -0,0 +1,82 @@
...
+
+// SleepStreamBackoff waits for the stream reconnect backoff delay or until ctx is cancelled.
+func SleepStreamBackoff(ctx context.Context, attempt int) error {
+	return Sleep(ctx, StreamBackoffDelay(attempt))
+}
+
+// ClassifyStreamError maps a stream error to a reconnect decision.
+func ClassifyStreamError(ctx context.Context, err error) StreamDecision {
+	if err == nil {
+		return StreamDecisionStop
```
igor-kupczynski, line 58: this is defensive, err should be non-nill; add a comment

pkg/client/reconnecting_stream.go
```diff
@@ -0,0 +1,266 @@
...
+		}
+
+		lastErr = err
+		s.l.Warn().Err(err).Str("stream", s.name).Int("attempt", attempt+1).Msg("stream send failed")
+
+		if _, genAfter, _ := s.snapshot(); genAfter != gen {
+			continue
+		}
+
+		if rerr := s.connectSync(ctx); rerr != nil {
```
mnafees: Should we honour some sort of deadline on the context so that this is not blocked for too long? The reason for the question is that I see we try `StreamSyncMaxAttempts` here and then `StreamSyncMaxAttempts` in `connectSync` as well which seems to stack up to a minute or so's delay unless that is going to expected behavior?

pkg/client/listener.go
```diff
@@ -125,462 +130,195 @@ func (r *subscribeClientImpl) getWorkflowRunsListener(
...
-	if err := w.retrySubscribe(ctx); err != nil {
+	lifecycle := l.stream.lifecycleContext()
+	if err := l.ensureListening(lifecycle); err != nil {
 		return err
 	}
 
-	if !w.startListening() {
-		if w.isClosed() {
-			return errListenerClosed
+	remove := l.reg.store(workflowRunId, sessionId, handler, onError)
```
mnafees, line 179: We store the handler before calling `retrySend` so if the background loop happens to reconnect in that gap, `replayHandlers` will already send a subscribe for this key, and then `retrySend` sends it again. We need to make sure for the same run ID the subscribe is idempotent on the server side it seems.

pkg/client/listener.go
```diff
@@ -125,462 +130,195 @@ func (r *subscribeClientImpl) getWorkflowRunsListener(
...
-
-// retrySubscribe coalesces concurrent reconnection attempts via singleflight.
-// Multiple goroutines calling this concurrently will share a single reconnection attempt.
-func (w *WorkflowRunsListener) retrySubscribe(ctx context.Context) error {
-	if w.isClosed() {
-		return errListenerClosed
 	}
 
-	if err := ctx.Err(); err != nil {
+	if err := l.ensureListening(lifecycle); err != nil {
```
mnafees, line 191: nit: Calling this after a successful send is basically always a no-op because the gate is already active, right?

pkg/client/retry/stream.go
```diff
@@ -0,0 +1,85 @@
...
+		return StreamDecisionNoProgress
+	}
+
+	switch st.Code() {
+	case codes.Canceled:
+		return StreamDecisionStop
+	case codes.Unauthenticated, codes.PermissionDenied, codes.InvalidArgument,
+		codes.FailedPrecondition, codes.NotFound, codes.Unimplemented:
+		return StreamDecisionStop
+	case codes.Unavailable, codes.Internal, codes.DeadlineExceeded, codes.ResourceExhausted:
```
mnafees, line 80: Quick question on putting `DeadlineExceeded` in the retry bucket. On the sync paths the `ctx.Err()` checks catch a real client-side deadline, but if the server returns a `DeadlineExceeded` status while our context is still live, we'll keep reconnecting on it. Is that intended, or should it count as no-progress instead?

pkg/client/reconnecting_stream.go
igor-kupczynski: Good catch. You are correct that the send path and the reconnect path each being "bounded at 5" compounds (up to ~25 connect attempts and roughly a minute-plus of backoff worst case). BTW this mirrors what the legacy retrySend → doRetrySubscribe on main already did, but that's not a reason to keep it.

I will flatten retrySend to make exactly one connectOnce attempt per failed send, with a single backoff per attempt.

igor-kupczynski: ffe8fc9

pkg/client/listener.go
igor-kupczynski: Interesting, so I think we are fine. Here's my thinking:
- Duplicate subscribes for the same run ID on a stream are idempotent because `subscribeToWorkflowRunsV1` funnels every subscribe into `workflowRunAcks.addWorkflowRun`, which is a map write (`acks[id]` <- `false`), and `sendEvent` only sends while the entry exists, deleting it after.
- the `store`->`replay`->`retrySend` race produces an extra wire message

One caveat: delivery is at-least-once: a re-subscribe after the completion was already acked re-delivers the finished event via the poller. That behavior predates this PR, and afaict it's on purpose as handlers are duplicate-tolerant (`Result()` takes the first value off a buffered channel).

pkg/client/listener.go
igor-kupczynski: Yeah, so after a successful send this is indeed a no-op.

It's there for the loop dying between the first `ensureListening` and the send completing, as without it, the handler would be registered and subscribed on the wire, but nothing would be receiving.

There's the `TestWorkflowRunsListenerRestartsAfterListenExits` to test for this behavior.

pkg/client/retry/stream.go
igor-kupczynski: Intended, here's my thinking:

* The classifier checks `ctx.Err()` first, so a real client-side deadline stops the loop.
* What lands in the retry bucket is only a DEADLINE_EXCEEDED status sent by the server and my working assumption is that on a long-lived stream that usually means a network blip cut an otherwise healthy connection --> so that's a transient failure we want to reconnect through.
* OTOH if we counted it as no-progress, a worker would "die" as a result.

WDYT?

mnafees (approved):

igor-kupczynski:
Looking at the `test / integration` failure:

```
=== RUN   TestScheduler_PoolIntegration_RemovingTenantStopsSnapshots
    scheduler_integration_test.go:386: unexpected snapshot for removed tenant 6684eb77-ca86-4b99-85aa-007fc3e76933
--- FAIL: TestScheduler_PoolIntegration_RemovingTenantStopsSnapshots (1.56s)
FAIL	github.com/hatchet-dev/hatchet/pkg/scheduling/v1	7.624s
```

- The `integration` job failed on `TestScheduler_PoolIntegration_RemovingTenantStopsSnapshots` in `pkg/scheduling/v1`. This is a timing-sensitive test that removes a tenant, sleeps 50ms for async loop cleanup, then asserts no further snapshots arrive; it looks like an in-flight snapshot for the removed tenant raced past the teardown window.
- this PR touches only `pkg/client/*` (Go SDK listener/stream refactor), which has no dependency link to the scheduler package.


I'll see if I can get an agent to harden this test in a separate PR.

**edit** the test passed on a re-run
**edit2** flaky test fix: https://github.com/hatchet-dev/hatchet/pull/4385

PR #4258, opened 2026-06-22
feat: core concurrency rewrite to use outbox + in-memory approach

abelanger5:
# Description

Modifies the concurrency implementation to use an in-memory index for ordering slots along with a transactional outbox for processing incremental updates to the index. Fixes a number of known concurrency performance issues; in particular, group round robin requiring a large sequential scan with many queued items.

Also fixes a priority sorting / ordering bug on the cancel in progress / cancel newest strategies. These strategies should now be working as expected. 

## Type of change

## What's Changed

- [X] More performant concurrency queues

## Checklist

Changes have been:

- [X] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Claude used to glue together parts of the implementation and write tests; most core logic written by hand with the exception of the rollback log, which was implemented by Claude. 

</details>


pkg/config/loader/loader.go
```diff
@@ -730,8 +731,25 @@ func createControllerLayer(dc *database.Layer, cf *server.ServerConfigFile, vers
 
 	promGate := prometheus.NewGate(dc.V1.TenantEntitlement(), cf.Prometheus.TenantScoped, &l)
 
+	// Dedicated pool for the concurrency outbox. It must be separate from the main pool: the
+	// outbox holds a transaction open while flushing (which itself calls UpdateConcurrencySlots
+	// and needs another connection), so sharing the main pool risks deadlock under saturation.
```
abelanger5: This comment is no longer accurate, we probably could use the same pool -- need to audit this.

pkg/scheduling/v1/concurrency/strategy.go
```diff
@@ -0,0 +1,748 @@
...
+)
+
+type ConcurrencyStrategy struct {
+	outbox pgoutbox.Outbox
+	repo   repository.ConcurrencyRepository
+	// TODO(memory): buildIndex reads every queued slot (and its schedule timeout) into memory. Empty
+	// sub-queues are pruned after each batch (see pruneEmpty), so idle keys don't accumulate, but a
+	// large live backlog still grows this without limit - we eventually need a bounded or evicting
+	// strategy (e.g. spill to disk/db) for very high key cardinality. Not critical yet.
+	subQueues map[string]*subQueue
```
abelanger5: to discuss: should we accept this risk? don't love the idea of committing unbounded memory growth, but I still think this is a net win and the worst-case scenario of the previous implementation had massive memory issues as well. 

we can't just reject work after a certain amount of items because we need to keep processing items off the WAL, so if we ignore just the inserts the state of the index would be incorrect. we also wouldn't be able to queue in a correct way. 

I've considered a swap-based approach which would let us swap / spillover to disk for each subqueue, but it's a lot of overhead and filling up disk is arguably just as bad.

cmd/hatchet-migrate/migrate/migrations/20260619000000_v1_0_120.sql
```diff
@@ -0,0 +1,70 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION after_v1_concurrency_slot_insert_outbox_function()
+RETURNS trigger AS $$
+BEGIN
+    INSERT INTO outbox.messages (topic, payload)
```
mrkaye97, line 6: is this a trigger that's writing into a schema that's owned by a third-party (albeit ours) library.....?

abelanger5: yes - there's no way to remove this from the trigger path without a major refactor, because the triggers decide when slots get inserted or deleted

pkg/scheduling/v1/concurrency/sub_queue.go
```diff
@@ -0,0 +1,44 @@
+package concurrency
+
+// subQueue represents the queue for a specific concurrency key
+type subQueue struct {
```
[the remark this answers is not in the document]
abelanger5: it's not immediately obvious that a subQueue maps to a concurrency key, so going to keep this

pkg/scheduling/v1/concurrency.go
```diff
@@ -63,9 +67,21 @@ func newConcurrencyManager(conf *sharedConfig, tenantId uuid.UUID, strategy *sql
...
+	ctx, cancel := context.WithCancel(context.Background())
+
+	// we only use the new in-memory index for non-parent strategies
+	var concurrencyStrategy *concurrency.ConcurrencyStrategy
+	if !strategy.ParentStrategyID.Valid &&
+		(strategy.Strategy == sqlcv1.V1ConcurrencyStrategyGROUPROUNDROBIN ||
+			strategy.Strategy == sqlcv1.V1ConcurrencyStrategyCANCELINPROGRESS ||
+			strategy.Strategy == sqlcv1.V1ConcurrencyStrategyCANCELNEWEST) {
+		concurrencyStrategy = concurrency.NewConcurrencyStrategy(ctx, repo, strategy, conf.outbox, &l)
+	}
```
abelanger5: Let's add a flag like `SERVER_CONCURRENCY_IN_MEMORY_INDEX_ENABLED` which gets propagated through, we can also remove these `strategy.Strategy` checks since this is implemented for all 3

cmd/hatchet-migrate/migrate/migrations/20260619000000_v1_0_120.sql
mrkaye97: :blood-sob:

pkg/scheduling/v1/concurrency/index.go
```diff
@@ -0,0 +1,300 @@
...
+		q.timedOutQueue = newHeap(timedOutCompare, timedOutSetIndex, timedOutGetIndex)
+	}
+
+	return q
+}
+
+func (q *inMemorySlotIndex) insert(s slot) {
+	// drop any existing entry for this task so re-inserts (e.g. retries) don't duplicate.
+	q.removeFromPriority(s.taskId)
+	q.removeFromTimedOut(s.taskId)
```
mrkaye97, line 136: I wonder if it'd be better to do nothing / reject the new insert? seems like the remove + re-insert is a lot of work for not much benefit?

pkg/scheduling/v1/concurrency/index.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+}
+
+func (q *inMemorySlotIndex) delete(taskId int64) (slot, bool) {
+	i, ok := q.priorityIndex(taskId)
+	if !ok {
+		return slot{}, false
+	}
+	s := q.priorityQueue.values[i]
+	q.priorityQueue.delete(i)
```
mrkaye97, line 150: are we worried about these being concurrency-safe? e.g. what happens if we call `delete` concurrently in two different places - is it safe to rely on integer indexing in that case?

pkg/scheduling/v1/concurrency/heap.go
```diff
@@ -0,0 +1,207 @@
...
+			h.insert(e.value)
+		}
+	}
+	h.journal = h.journal[:0]
+}
+
+func (h *heap[T]) insert(value T) {
+	h.values = append(h.values, value)
+	h.setIndex(value, len(h.values)-1)
+	h.up(len(h.values) - 1)
```
mrkaye97, line 84: re: concurrency - maybe a mutex here would help (same for on delete, etc.)?

pkg/scheduling/v1/concurrency/heap.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	minVal := h.values[0]
+	h.delete(0)
+	return minVal
+}
+
+func (h *heap[T]) delete(i int) {
+	if i < 0 || i >= len(h.values) {
+		panic("heap: delete: index out of range")
+	}
```
mrkaye97, line 137: feels like we should error here so we can handle it gracefully instead of panicking

pkg/scheduling/v1/concurrency/index.go
```diff
@@ -0,0 +1,300 @@
...
+	case a > b:
+		return 1
+	default:
+		return 0
+	}
+}
+
+func cmp32(a, b int32) int {
+	switch {
+	case a < b:
```
mrkaye97: this is crying out for generics.... feels like we shouldn't need both of these

pkg/scheduling/v1/concurrency/heap.go
mrkaye97, line 1: just as an aside, I'm a bit surprised there isn't a heap / priority queue library we can pull off the shelf?

pkg/repository/scheduler_concurrency.go
```diff
@@ -799,3 +839,188 @@ func (c *ConcurrencyRepositoryImpl) DeactivateStaleStepConcurrency(ctx context.C
...
+	defer rollback()
+
+	offset := 0
+
+	for {
+		slots, err := c.queries.ListConcurrencySlotsForIndexing(ctx, tx, sqlcv1.ListConcurrencySlotsForIndexingParams{
+			Tenantid:   tenantId,
+			Strategyid: strategyId,
+			Offset:     int32(offset),
+			Limit:      10000,
```
mrkaye97: I'm a bit nervous that limit + offset pagination won't scale well here for large numbers of rows in `v1_concurrency_slot`. Since we order by tuples of `v1_concurrency_slot_query_idx` anyways, maybe we can paginate by them too?

WDYT?

pkg/scheduling/v1/concurrency/strategy.go
```diff
@@ -0,0 +1,737 @@
...
+	// loaded at build time would otherwise sit unqueued until a new WAL message happened to touch its
+	// sub-queue. This must happen before we process any WAL messages.
+	initialResult, err := c.runInitialQueueing(ctx)
+
+	if err != nil {
+		return nil, err
+	}
+
+	for {
+		msgs, err := c.outbox.ProcessMessages(ctx, c.topic)
```
mrkaye97, line 219: so is the basic idea here that we first build the index on boot (or whenever the strategy is registered?) and then we periodically flush messages from the outbox, which inserts new stuff off the "wal" into the queues?

pkg/scheduling/v1/concurrency/strategy.go
```diff
@@ -0,0 +1,737 @@
...
+// preempt a running slot when a higher-priority slot is waiting.
+func (c *ConcurrencyStrategy) decide() decideFn {
+	switch c.strategy.Strategy {
+	case sqlcv1.V1ConcurrencyStrategyCANCELINPROGRESS:
+		return decideCancelInProgress
+	case sqlcv1.V1ConcurrencyStrategyCANCELNEWEST:
+		return decideCancelNewest
+	default:
+		return decideGroupRoundRobin
+	}
```
mrkaye97, line 509: small nit, but I'd much prefer to have a case for round robin, and have the default panic

pkg/scheduling/v1/concurrency/strategy.go
```diff
@@ -0,0 +1,737 @@
...
+// applyWAL replays a sub-queue's WAL messages into its running/queued indexes, bringing the index in
+// sync with the database. It returns the superseded/stale slots that must be cancelled with
+// CONCURRENCY_LIMIT: a retry's older slot (the incoming message has a higher retry count) or a stale
+// incoming slot for a task already present at an equal-or-higher retry count.
+func applyWAL(sq *subQueue, msgs []walMessage) []slot {
+	superseded := make([]slot, 0)
+
+	for _, msg := range msgs {
+		switch msg.Operation {
+		case "INSERT":
```
mrkaye97, line 595: maybe an enum for this?

mrkaye97 (commented):
I left some questions / comments - I still need to read through the tests

pkg/scheduling/v1/concurrency/index.go
abelanger5: It's not concurrency-safe and isn't designed to be -- everything from the caller perspective is serialized (we read out of the channels and update the index serially), so adding a mutex is unnecessary overhead

pkg/scheduling/v1/concurrency/heap.go
abelanger5: See note on concurrency above

pkg/scheduling/v1/concurrency/heap.go
abelanger5: There's `container/heap` and some of the typing work is based on https://github.com/jba/heap (documented in the file) but there's nothing with the concept of a commit log so we'd need to re-implement anyway. I also think this is a considerable DX improvement over `container/heap` particularly with the generics support.

pkg/scheduling/v1/concurrency/strategy.go
abelanger5: Yep pretty much. The WAL keeps the index up to date and correct.

pkg/repository/scheduler_concurrency.go
abelanger5: You mean like with a `<=` and `>=` instead of a limit / offset?

abelanger5: That should be doable

pkg/scheduling/v1/concurrency/index.go
abelanger5: Yeah I can switch to `cmp.Compare`

pkg/repository/scheduler_concurrency.go
mrkaye97: yeah, or just `>` + `LIMIT` I think is good? then we can keep a consistent chunk size but pagination should be a lot faster, at least I think (I don't remember why we switched in the payload job to using the <= instead of limit, might've been a similar performance thing, could be worth double checking what we do there since paginating through those tables doesn't get worse asymptotically)

pkg/scheduling/v1/concurrency/strategy.go
mrkaye97: Gotcha - I was wondering as I was reading through if maybe we should use some heuristic to determine how much stuff we should read into memory at any given time, since for group round robin (don't think it applies to the others) there's not a ton of value in just naively holding lots of queue items in memory, vs. feeding them through on closer to an "as-needed" basis? not sure if you thought about that at all, but I was thinking maybe only allowing `maxRuns * 2` items to be held in memory in the heap at any given time for strategies using round robin could limit memory pressure a lot. seems like for the cancel strategies it's less of an issue since there's significantly less risk of building up a large queue in memory

abelanger5: Yeah I was thinking about different strategies here too. For what you proposed, in the worst-case we still see memory pressure with a huge number of keys, since we'd need to store `maxRuns * 2` for each key, but it's a good suggestion.

It's also tricky to coordinate new `INSERT` statements coming in from the WAL. In this strategy, every new `INSERT` would require an additional database lookup if we're in an "offloaded" state for that key. 

Given that in the worst-case, we'd still be filling up memory (for the other strategies as well), perhaps we should think of an index structure with more flexibility, perhaps something like Postgres indexes whose entries are backed by 8kb pages on disk, with a fixed-size shared buffer for pages. "Disk" in this case could be the local disk, or it could be the Postgres database using a temp table.

pkg/scheduling/v1/concurrency/index.go
abelanger5: there's a very critical assumption that slots are unique by taskId, I think this is ok? it's an `o(1)` lookup using the map entry anyway, so should be low overhead

pkg/scheduling/v1/concurrency/heap.go
abelanger5: just modified this to `return`

pkg/scheduling/v1/concurrency/strategy.go
abelanger5: fixed 👍

pkg/scheduling/v1/concurrency/index.go
abelanger5: fixed 👍

cmd/hatchet-migrate/migrate/migrations/20260619000000_v1_0_120.sql
grutt: i suppose we have to be extra careful on migrations of the library schema eh?

sql/schema/v1-core.sql
```diff
@@ -2404,3 +2404,63 @@ CREATE TABLE tenant_entitlement (
...
+            'priority', dr.priority,
+            'taskId', dr.task_id,
+            'taskInsertedAt', dr.task_inserted_at,
+            'taskRetryCount', dr.task_retry_count,
+            'scheduleTimeoutAtMs', (EXTRACT(EPOCH FROM dr.schedule_timeout_at) * 1000)::bigint
+        )
+    FROM deleted_rows dr
+    JOIN v1_step_concurrency sc ON sc.id = dr.strategy_id
+    WHERE sc.parent_strategy_id IS NULL
+      AND sc.strategy IN ('GROUP_ROUND_ROBIN', 'CANCEL_IN_PROGRESS', 'CANCEL_NEWEST');
```
grutt: is there any way to make this ref the enum or something? concerned we'll forget to keep this in sync with strategies

cmd/hatchet-migrate/migrate/migrations/20260619000000_v1_0_120.sql
mrkaye97: yeah... fwiw, I think this is a very risky move. marginally better because we own the library, but still, _especially_ with this being in the OSS where maybe people could be on different versions? idk

pkg/repository/scheduler_concurrency.go
```diff
@@ -799,3 +839,192 @@ func (c *ConcurrencyRepositoryImpl) DeactivateStaleStepConcurrency(ctx context.C
 func (c *ConcurrencyRepositoryImpl) ListTenantsWithManyStepConcurrencies(ctx context.Context, threshold int64) ([]*sqlcv1.ListTenantsWithManyStepConcurrenciesRow, error) {
 	return c.queries.ListTenantsWithManyStepConcurrencies(ctx, c.pool, threshold)
 }
+
+func (c *ConcurrencyRepositoryImpl) ReadConcurrencySlotsForIndexing(ctx context.Context, tenantId uuid.UUID, strategyId int64, writeCh chan<- *sqlcv1.ListConcurrencySlotsForIndexingRow) error {
+	// we don't want to hold the transaction open if we're blocked on the write channel, so we use this
+	// context to escape the <- writeCh loop and close the tx
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
```
grutt, line 847: intentionally 5m?

grutt (commented):
Seems chill

abelanger5 (commented):
Two items before merge:
1. Ensuring that we have proper tracing on the new strategy paths so we can easily track down regressions / performance issues with the new strategy
2. Figuring out a mechanism so that when the outbox is disabled, we don't accumulate messages (we likely want a silent/no-op flushing mechanism). It would be great to hook into a list of topics that we're publishing to on pgoutbox.

cmd/hatchet-migrate/migrate/migrations/20260619000000_v1_0_120.sql
abelanger5: the `pgoutbox` version is tagged and pinned and we'll fail pretty loudly in tests if we can't write into it. I don't think it's that risky?

pkg/repository/scheduler_concurrency.go
abelanger5: pushed this 👍

sql/schema/v1-core.sql
abelanger5: removed this, not needed

pkg/repository/scheduler_concurrency.go
abelanger5: yeah; I figured if we're holding the tx open for longer we're going to start seeing issues

mrkaye97 (approved):

PR #4262, opened 2026-06-23
feat: Tenant/user/organization token tagging

juliusgeo:
# Description

Frontend changes to add support for user groups and tenant tagging.

image
image
image

Fixes # (issue)

## Type of change

## What's Changed

## Checklist

Changes have been:

- [ ] Linted and formatted

---

<details id="ai-disclosure">

</details>


api-contracts/openapi/components/schemas/tenant.yaml
```diff
@@ -204,10 +204,14 @@ TenantMember:
     tenant:
       $ref: "#/Tenant"
       description: The tenant associated with this tenant member.
+    explicit:
```
mrkaye97: we should try to use `is` prefixing for these types of booleans I think, e.g. `isExplicit`

api-contracts/openapi/components/schemas/tenant.yaml
```diff
@@ -204,10 +204,14 @@ TenantMember:
...
+      type: boolean
+      description: Whether this membership was explicitly granted (as opposed to synced via user-group tags). Only explicit members can have their role edited or be removed.
   required:
     - metadata
     - user
     - role
+    - explicit
```
mrkaye97: this is a breaking change if we make it required so I think we need to make it optional

internal/services/admin/contracts/workflows.pb.go
mrkaye97, line 1: Can we revert all of these regenerations? looks like they're just bumping a protoc version

juliusgeo: yep, fixed

api-contracts/openapi/components/schemas/tenant.yaml
juliusgeo: I renamed it to `manually_added` before I saw this which I think should work--I can rename it to `is_manually_added` but would prefer not to do another rename 🫠

api-contracts/openapi/components/schemas/tenant.yaml
juliusgeo: made it optional!

frontend/app/src/components/modals/add-org-member-to-tenant-modal.tsx
```diff
@@ -0,0 +1,150 @@
...
+    (tenantMembersQuery.data?.rows ?? []).map(
+      (m: TenantMember) => m.user.email,
+    ),
+  );
+
+  const members = (orgQuery.data?.members ?? []).filter(
+    (m) => !tenantMemberEmails.has(m.email),
+  );
+
+  const addMutation = useMutation({
```
mrkaye97, line 60: it's a bit of a personal preference, but I've been liking wrapping these mutations in callbacks, so below this you'd have:

```ts
const addOrgTenantMembers = useCallback((organizationId: string, tenantId: string) => addMutation.mutate(...), [])
```

so that you can easily call `addOrgTenantMembers(...)` and we sort of abstract away some of the react query magic to make it look a little more like "normal typescript". make of that what you will though, not too wedded to it, although I do prefer it (I also like exporting things like `isPending` from the hook directly)

frontend/app/src/pages/organizations/$organization/components/create-user-group-modal.tsx
```diff
@@ -0,0 +1,119 @@
...
+              value={role}
+              onValueChange={(v) => setRole(v as TenantMemberRoleType)}
+              disabled={createMutation.isPending}
+            >
+              <SelectTrigger id="group-role">
+                <SelectValue />
+              </SelectTrigger>
+              <SelectContent>
+                <SelectItem value={TenantMemberRoleType.MEMBER}>
+                  MEMBER
```
mrkaye97, line 97: can we render the enum values directly here instead of duplicating? e.g.:

```ts
{Object.keys(TenantMemberRoleType).map(t => ( 
                <SelectItem value={t}>
                 {t}
                </SelectItem>
))}
```

(not sure if this is how you map over an enum)

frontend/app/src/components/modals/add-org-member-to-tenant-modal.tsx
juliusgeo: fixed!

mrkaye97 (approved):
just to confirm - this works fine in the OSS when there's no tags table or API stuff right?

PR #4266, opened 2026-06-23
feat(engine): dags-as-durable-tasks

mrkaye97:
# Description

Initial implementation of the dag operator, which basically creates an internal durable task that processes the dag via child spawning. The idea here is we basically run a loop inside the operator that performs all the operations of a normal durable task, and that "durable task" handles all the durable operations of the DAG - spawning children, sleeping, etc.

I tried to make this as minimally invasive as possible and also backwards compatible. One note is we sort of "trick" the OLAP side of the house into thinking that there's a "real" DAG being run when there isn't, so that we don't really need any changes on that side. 

We're using the tenant entitlements to enable the operator, but once there's a workflow version that's registered to use the operator, it's using it forever (this allows for incremental rollouts to only new workflow versions and certain tenants, but also makes it a bit harder to roll back).

I'll try to leave comments in important places in the PR. I also wrote a lot of tests (some unit tests on the Go side and some E2E tests in Python) that should confirm most of the behavior is working 

to do:

- [x] render preview of dag even if nothing has run yet in the dash
- [x] make sure all the existing apis work (check everything that returns dags)
- [x] write more tests for all the different cases of conditions, skipping behavior, cancelling behavior, and concurrency
- [x] make sure that a dag that's registered with the flag enabled still works as expected after the flag is disabled, and vice versa
- [x] figure out what happens to running dags when this ships
- [x] fix the dashboard / runs list api's pagination when we're listing runs that include dag orchestrators (since they and their children get collapsed)
- [x] add link from task to orchestrator run in the dashboard like we have for dags
- [x] trace view is broken since it tries to fetch the dag run by its external id
- [ ] Make sure the `workflow_run_id` for the dag children correspond's to the durable task external id
- [x] There's a bug (need to reproduce) where if the first task in a dag has a `wait_for` (sleep), we get a bunch of "could not send task to worker'
- [x] figure out how to enable replaying from a specific place in the DAG
- [x] test that on-failure tasks still work properly
- [x] Look into the size of the buffered channels to figure out if we need a larger buffer, what happens when we exceed the buffer size, etc.

Fixes # (issue)

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

<details id="ai-disclosure">

- **Details**: Used Claude Code (mostly Fable when I had credits) to do code review, find and fix bugs, write tests and examples, and convert the original implementation with lots of hacks on the OLAP side to the new one where we "trick" the OLAP side into thinking there's a real DAG running

</details>


cmd/hatchet-migrate/migrate/migrations/20260708133531_v1_0_126.sql
```diff
@@ -2,6 +2,11 @@
 -- +goose StatementBegin
 ALTER TABLE "Worker" ADD COLUMN "operatorId" UUID;
 
+ALTER TABLE "Step" ADD COLUMN "isDagOrchestrator" BOOLEAN NOT NULL DEFAULT false;
+
+ALTER TABLE v1_task ADD COLUMN is_dag_orchestrator BOOLEAN NOT NULL DEFAULT false;
+ALTER TABLE v1_tasks_olap ADD COLUMN is_dag_orchestrator BOOLEAN NOT NULL DEFAULT false;
```
mrkaye97: note to self: needs a cloud migration

frontend/app/src/pages/main/v1/workflow-runs-v1/components/v1/task-runs-columns.tsx
```diff
@@ -112,7 +112,7 @@ export const columns: (
       <DataTableColumnHeader column={column} title={TaskRunColumn.taskName} />
     ),
     cell: ({ row }) => {
-      if (row.getCanExpand()) {
+      if (row.original.type === 'DAG') {
```
mrkaye97, line 188: this lets us open the workflow run detail page for an operator DAG before any of the tasks have actually been spawned

frontend/app/src/pages/main/v1/workflow-runs-v1/$run/index.tsx
```diff
@@ -303,8 +303,11 @@ function ExpandedWorkflowRun({ id }: { id: string }) {
   );
 
   const durableTaskIds = useMemo(
-    () => taskRuns.filter((t) => t.isDurable).map((t) => t.taskExternalId),
-    [taskRuns],
+    () =>
+      taskRuns
+        .filter((t) => t.isDurable && t.taskExternalId !== id)
```
mrkaye97, line 352: there's a bunch of stuff like this - basically removing the orchestrator from the ids

internal/services/controllers/olap/engine_spans.go
```diff
@@ -404,6 +405,9 @@ func (tc *OLAPControllerImpl) synthesizeEngineSpans(ctx context.Context, tenantI
 	batchStartedTimes := make(map[taskRetryKey]time.Time)
 
 	for _, e := range events {
+		if e.isDagOrchestrator {
+			continue
+		}
```
mrkaye97, line 410: don't want to render spans for the orchestrator task since it messes up nesting and is "internal" / not user facing

internal/services/dispatcher/dispatcher.go
```diff
@@ -809,41 +813,61 @@ func (d *DispatcherImpl) populateTaskData(
...
+				d.l.Warn().Ctx(ctx).Err(err).Msg("failed to unmarshal input")
+				continue
+			}
+		}
 
-				if err != nil {
-					d.l.Warn().Ctx(ctx).Err(err).Msg("failed to unmarshal input")
-					continue
+		if len(currInput.DagParentTaskRunIds) > 0 {
+			dagParentOutputs, err := d.repov1.Tasks().GetDagParentOutputs(ctx, tenantId, currInput.DagParentTaskRunIds)
```
mrkaye97, line 871: not sure what to do about this - seems very suboptimal to need a db lookup here, will circle back

internal/services/dispatcher/server.go
```diff
@@ -1195,33 +1195,81 @@ func (s *DispatcherImpl) subscribeToWorkflowRunsV1(server contracts.Dispatcher_S
...
+								TaskRunExternalId: event.TaskExternalId.String(),
+								TaskName:          stepName,
+								JobRunId:          event.TaskExternalId.String(),
+								Output:            &out,
+							})
+						}
+						continue
+					}
+				} else {
+					childEvents, err := s.repov1.Tasks().ListDurableOrchestratorChildOutputEvents(ctx, tenantId, event.TaskExternalId)
```
mrkaye97, line 1255: same here - really not ideal to need the db lookup, will circle back

internal/services/dispatcher/server_v1.go
```diff
@@ -1221,3 +1221,126 @@ func (d *DispatcherServiceImpl) DeliverDurableEventLogEntryCompletion(taskExtern
...
+	}
+
+	if err := d.mq.SendMessage(ctx, msgqueue.TASK_PROCESSING_QUEUE, msg); err != nil {
+		return fmt.Errorf("failed to send replay message for child task %s: %w", childExternalId, err)
+	}
+
+	return nil
+}
+
+func (d *DispatcherServiceImpl) TriggerDAGStep(ctx context.Context, tenantId uuid.UUID, req *operator.DAGStepTriggerRequest) (*operator.DAGStepTriggerResult, error) {
```
mrkaye97, line 1582: this is one of the important methods, since it's what gets called by the operator to trigger tasks

internal/services/dispatcher/server_v1.go
```diff
@@ -1221,3 +1221,126 @@ func (d *DispatcherServiceImpl) DeliverDurableEventLogEntryCompletion(taskExtern
...
+			AdditionalMetadata:   task.AdditionalMetadata,
+			ParentExternalId:     &task.ExternalID,
+			ParentTaskId:         &task.ID,
+			ParentTaskInsertedAt: &task.InsertedAt.Time,
+			ChildIndex:           &childIndex,
+			DagParentTaskRunIds:  req.DagParentTaskRunIds,
+			IsSkipped:            req.IsSkipped,
+			IsCancelled:          req.IsCancelled,
+			DesiredWorkerLabels:  req.DesiredWorkerLabels,
+			WorkflowRunId:        &orchestratorWorkflowRunId,
```
mrkaye97, line 1615: important stuff here:

1. we basically pass most things along (e.g. the metadata, the input, etc.)
2. we tie the workflow run id manually to the orchestrator's workflow run id

pkg/operator/dagoperator/operator.go
```diff
@@ -218,50 +197,190 @@ func (d *DAGOperator) run(ctx context.Context, action *contracts.AssignedAction)
...
+		}
+		return d.fail(action, fmt.Errorf("dag failed: %w", dagErr))
+	}
+
+	output := make(map[string]json.RawMessage, len(tasks))
+	for _, t := range tasks {
+		if t.output != nil {
+			b, err := json.Marshal(t.output)
+			if err == nil {
+				output[t.readableId] = json.RawMessage(b)
```
mrkaye97, line 364: this results in a DAG output that maps step ids to outputs, similar to what we see in the dash and over the API

pkg/repository/sqlcv1/olap-overwrite.go
```diff
@@ -132,6 +132,16 @@ WITH filtered AS (
...
+
+		-- fixme: this might be really slow
+		AND NOT EXISTS (
+			SELECT 1
+			FROM v1_tasks_olap t
+			WHERE
+				t.external_id = r.parent_task_external_id
+				AND COALESCE(t.is_dag_orchestrator, FALSE)
+				AND t.tenant_id = r.tenant_id
+		)
```
mrkaye97: so, this makes me very nervous, but I'm not sure how else to do this. Basically, we need to remove any dag _children_ from the runs list query, since they'll be double counted (we should only include dag parents, dags, and standalone tasks)

pkg/repository/sqlcv1/olap-overwrite.go
```diff
@@ -169,50 +179,71 @@ func (q *Queries) CountWorkflowRuns(ctx context.Context, db DBTX, arg CountWorkf
...
+		AND NOT EXISTS (
+			SELECT 1
+			FROM v1_tasks_olap t
 			WHERE
-				lt.tenant_id = $1::uuid
-				AND lt.external_id = $11::UUID
+				t.external_id = r.parent_task_external_id
+				AND COALESCE(t.is_dag_orchestrator, FALSE)
+				AND t.tenant_id = r.tenant_id
 		)
```
mrkaye97: same here

pkg/operator/dagoperator/dag.go
```diff
@@ -2,100 +2,575 @@ package dagoperator
...
+
+func isDagCancelledErr(err error) bool {
+	var e *dagCancelledError
+	return errors.As(err, &e)
+}
+
 type dag struct {
-	requestCh chan<- *v1contracts.DurableTaskRequest
+	requestCh   chan<- *v1contracts.DurableTaskRequest
+	matchRepo   repository.MatchRepository
```
abelanger5: I think pulling out the `EvaluateBoolExpr` into a subpackage with essentially no dependencies (particularly no DB dependencies) would make it much easier to test this file.

This also seems like a good candidate for table-test unit testing with a set of comprehensive examples; in particular, we'd likely want code coverage on all of the feature paths (wait conditions, etc).

abelanger5: I'm aware that building out testing for this file would require the test to sort of act like its own mini workflow engine, since it needs to be able to receive and send durable task req/responses over the channels...I'm assuming that hardcoding when we send data over these channels for each test is simple enough that this is doable, but I could also see a world where it's complex enough that we don't want to do this

pkg/operator/dagoperator/dag.go
```diff
@@ -2,100 +2,575 @@ package dagoperator
...
 		case <-ctx.Done():
-			return
-		case resp := <-responseCh:
-			dag.taskConsumer(resp)
+			return ctx.Err()
+		case resp, ok := <-responseCh:
+			if !ok {
+				return fmt.Errorf("durable task session closed")
+			}
+			d.taskConsumer(resp)
```
abelanger5: q: did we ever merge that "stable callback ordering" PR? That's required for this to work

pkg/operator/dagoperator/dag.go
```diff
@@ -2,100 +2,575 @@ package dagoperator
...
+	readableId   string
+	index        int32
+	parents      []*task
+	isCompleted  bool
+	isFailed     bool
+	isCancelled  bool
+	isTriggered  bool
+	isSkipped    bool
+	errorMessage string
+	output       map[string]interface{}
```
abelanger5, line 118: I don't think this is tractable, I think we need to find a way to not keep the output of the task in memory for longer than necessary. Perhaps we can invert the expression evaluation to evaluate the output when the parent completes instead of evaluating it on the child.

pkg/operator/dagoperator/dag.go
```diff
@@ -2,100 +2,575 @@ package dagoperator
...
-		areParentsSatisfied := true
+func (d *dag) taskEmitter(ctx context.Context) error {
+	if d.err != nil {
+		return nil
+	}
 
-		for _, parent := range task.parents {
-			if !parent.isCompleted {
-				areParentsSatisfied = false
+	for _, t := range d.tasks {
```
abelanger5: I wonder if it's more beneficial to keep a list of `pendingTasks` versus `finalizedTasks` so that we only loop through `pendingTasks` which could theoretically still emit an event

pkg/operator/dagoperator/dag.go
```diff
@@ -2,100 +2,575 @@ package dagoperator
...
+		}
+
+		var parentTaskRunIds []uuid.UUID
+		for _, p := range d.tasks {
+			if p.isCompleted && !p.isFailed && p.workflowRunExternalId != nil {
+				parentTaskRunIds = append(parentTaskRunIds, *p.workflowRunExternalId)
+			}
+		}
+
+		result, err := d.triggerStep(ctx, t.actionId, t.workflowName, t.index, parentTaskRunIds, skip, cancelled)
```
abelanger5: does this need to be synchronous / blocking the loop, or can we add to a waitgroup and evaluate at the bottom of `taskEmitter`?

abelanger5: Perhaps this wouldn't get us anything though, because we need to guard the event log which I imagine we touch when we call this method, so we serialize it anyway

pkg/operator/dagoperator/dag.go
abelanger5, line 1: nit: a quick pass adding spans/traces throughout this file would be awesome

pkg/repository/sqlcv1/olap-overwrite.go
abelanger5: probably a broader conversation, but should we not be treating the OLAP side of things as static, so we just insert data as if we're inserting DAGs?

internal/services/admin/v1/server.go
```diff
@@ -767,6 +798,32 @@ func (a *AdminServiceImpl) PutWorkflow(ctx context.Context, req *contracts.Creat
...
+		return nil
+	}
+
+	config, err := json.Marshal(struct{}{})
+
+	if err != nil {
+		return fmt.Errorf("could not marshal DAG operator config: %w", err)
+	}
+
+	_, err = a.repo.Operators().CreateOperator(ctx, tenantId, v1.CreateOperatorOpts{
```
abelanger5, line 952: it looks like we only call this on `PutWorkflow`. Should we be calling this from other places, or is the thinking that this only really enables when a user restarts their workers?

pkg/repository/sqlcv1/olap-overwrite.go
mrkaye97: ah, I didn't even think about that... 🤔 it would make all the API surface stuff a lot easier

mrkaye97: I guess the negative is we'd still need `v1_dags_olap`, `v1_dag_to_task_olap`, etc.?

pkg/operator/dagoperator/dag.go
mrkaye97: not yet, but yeah, agreed - there are some open comments on that one still

internal/services/admin/v1/server.go
mrkaye97: I was thinking it was fine to only do this here since the `isDagOperator` flag is tied to the `WorkflowVersion`, so until someone registers a new workflow version with a dag, we'd never end up needing to call this, which I think is fine. We could also call it on startup I guess?

pkg/repository/sqlcv1/olap-overwrite.go
abelanger5: yeah - I'm not sure it's a long-term fix, but short-term it would make our lives easier (obviously just moves complexity somewhere else)

mrkaye97: lemme see if I can get it to work. maybe when we eventually do V2, we can do that second push for OLAP since we potentially won't need to support as much old API surface

pkg/operator/dagoperator/dag.go
```diff
@@ -2,100 +2,894 @@ package dagoperator
...
+		if !ready {
+			stillPending = append(stillPending, t)
+			continue
+		}
 
-			for _, condition := range task.conditions {
-				areConditionsSatisfied = areConditionsSatisfied && condition.isSatisfied
+		cancelled := false
+		for _, p := range t.parents {
+			if p.isCancelled || p.isFailed {
```
juliusgeo, line 284: could this be combined with the loop above?

pkg/operator/dagoperator/dag.go
```diff
@@ -2,100 +2,894 @@ package dagoperator
...
+		}
 
-				// emit a request for the condition if it's not triggered
-				// TODO: emit a request for the condition
-				// if !condition.isTriggered {
+		skip := false
 
-				// }
+		if !cancelled {
+			if allParentsSkipped(t) {
```
juliusgeo, line 293: in general, a lot of the logic here seems like it would be duplicated many times over the total execution path. Would it be possible to turn these into booleans on the child task itself, and make the parents responsible for updating them? Then we wouldn't have to do all this iteration. Not sure if it would make a difference, but that's the main thing that jumps out to me here

pkg/repository/durable_events.go
```diff
@@ -86,6 +86,12 @@ type IngestTriggerRunsEntry struct {
 	WorkflowRunExternalId uuid.UUID
 	IsSatisfied           bool
 	AlreadyExisted        bool
+
+	ChildNeedsReplay bool
+
+	// ReExecuted is true when the entry was newly created this invocation, i.e. the child
+	// actually runs rather than being satisfied from a cached log entry.
+	ReExecuted bool
```
juliusgeo, line 94: what's the difference between this and "AlreadyExisted"?

pkg/repository/workflow.go
```diff
@@ -24,6 +24,16 @@ import (
 
 var ErrDagParentNotFound = errors.New("dag parent not found")
 
+type DagNode struct {
+	// include both of these to make it easier to render the dag in other places
+	ParentReadableIds []string
+	ChildReadableIds  []string
+
+	// todo: expand this to include conditions and other stuff so we can render dags from json?
```
juliusgeo, line 32: should this be addressed?

mrkaye97: not yet I don't think - this will be useful for "dynamic dags" in the future though, where someone could post a json to us somehow and we'd execute that

pkg/repository/durable_events.go
mrkaye97: oh yeah good catch, I think we can probably simplify with `!AlreadyExisted`

pkg/operator/dagoperator/dag.go
mrkaye97: I agree there's a lot of duplication - re: booleans on the task itself, not sure what you mean there, mind giving an example?

internal/services/dispatcher/dispatcher.go
mrkaye97: @juliusgeo we should probably address this in a follow-up

internal/services/dispatcher/server.go
mrkaye97: @juliusgeo this one too

juliusgeo (approved):

PR #4281, opened 2026-06-25
fix: on-demand lease manager for cold start concurrency queue

mnafees:
# Description

This PR aims to fix the cold start latency concurrency workflows can face which increases the latency between QUEUED -> ASSIGNED_TO_WORKER status progression.

## Type of change

## What's Changed

- [x] Adds method to acquire on demand concurrency lease for cold starts of queues
- [x] Adds unit test for the behavior
- [x] Removes redundant RMQ message type and mechanism

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

</details>


internal/services/scheduler/v1/scheduler.go
```diff
@@ -384,19 +382,6 @@ func (s *Scheduler) handleNewQueue(ctx context.Context, msg *msgqueue.Message) e
 	return nil
 }
 
-func (s *Scheduler) handleNewConcurrencyStrategy(ctx context.Context, msg *msgqueue.Message) error {
```
abelanger5, line 387: Why are we removing this? It feels like this is a different case that we should also handle: a completely new concurrency strategy being registered from `PutWorkflow`. We likely are not propagating this message yet, but we probably should.

pkg/scheduling/v1/lease_manager.go
```diff
@@ -418,52 +418,67 @@ func (l *LeaseManager) acquireConcurrencyLeases(ctx context.Context) error {
...
+// acquireConcurrencyStrategyOnDemand acquires a lease for a single concurrency strategy on-demand
+// (the cold-start path). It returns the strategy when a ConcurrencyManager should be created for it,
+// or nil when the lease is already held here or is owned by another scheduler.
+//
+// Unlike the periodic lease loop, this takes a BLOCKING lock and returns the strategy to the caller
+// rather than delivering it over the (non-blocking, lossy) lease channel. Callers run this off the
+// hot path in their own goroutine, so blocking is fine -- and it guarantees the acquisition is not
+// silently dropped under contention (the previous TryLock + non-blocking-send approach would no-op
+// when several cold strategies raced each other or the periodic poll, falling back to the 5s poll).
+func (l *LeaseManager) acquireConcurrencyStrategyOnDemand(ctx context.Context, strategyId int64) (*sqlcv1.V1StepConcurrency, error) {
```
abelanger5: I'm not sure why this needs to use a different path than the `notifyNewQueue` mechanism. I think we can rethink the non-blocking write to the `concurrencyLeasesCh` to potentially make it blocking, if that's what we're worried about.

From what I can tell there are two important things here:
1. That we call `notify` after we instantiate the concurrency strategy (which we should probably do anyway)
2. That we call `notifyNewConcurrencyStrategy` when a message arrives with a strategy ID which isn't currently in our leases.

internal/services/scheduler/v1/scheduler.go
mnafees: Restored the handler. In terms of the `PutWorkflow` change - I think my changes already address this when the very first task arrives with concurrency? So IMO the publishing of this RMQ message from `PutWorkflow` is only proactive optimization?

pkg/scheduling/v1/tenant_manager.go
```diff
@@ -279,7 +284,11 @@ func (t *tenantManager) addConcurrencyStrategy(strategy *sqlcv1.V1StepConcurrenc
 		}
 	}
 
-	t.concurrencyStrategies = append(t.concurrencyStrategies, newConcurrencyManager(t.cf, t.tenantId, strategy, t.concurrencyResultsCh, t.concurrencyAdvisoryLock, t.concurrencyParentAdvisoryLock))
+	c := newConcurrencyManager(t.cf, t.tenantId, strategy, t.concurrencyResultsCh, t.concurrencyAdvisoryLock, t.concurrencyParentAdvisoryLock)
+	t.concurrencyStrategies = append(t.concurrencyStrategies, c)
+
+	// notify the new manager so it schedules its waiting task now instead of waiting for its first tick.
+	c.notify(ctx)
```
abelanger5: nit: we should consider doing this on concurrency manager instantiation instead? I think the manager has a notifier channel as well, on instantiation we can just put a message into that channel so that it runs immediately after startup? having to remember to notify each time we call `newConcurrencyManager` is poor design I think

pkg/scheduling/v1/tenant_manager.go
```diff
@@ -304,6 +313,9 @@ func (t *tenantManager) notifyConcurrency(ctx context.Context, strategyIds []int
 			continue
 		}
 
+		// mark this strategy as already managed so we don't try to acquire a lease for it below
+		delete(strategyIdsMap, c.strategy.ID)
```
abelanger5: are we sure about this? I see immediately above this we're doing:

```go
if _, ok := strategyIdsMap[c.strategy.ID]; !ok {
		continue
}
```

So it seems like we might be doing more processing as a result of deleting this entry. Perhaps we need two copies of the map?

pkg/scheduling/v1/tenant_manager.go
```diff
@@ -263,13 +263,18 @@ func (t *tenantManager) setConcurrencyStrategies(strategies []*sqlcv1.V1StepConc
 	}
 
 	for _, strategy := range strategiesSet {
-		newArr = append(newArr, newConcurrencyManager(t.cf, t.tenantId, strategy, t.concurrencyResultsCh, t.concurrencyAdvisoryLock, t.concurrencyParentAdvisoryLock))
+		c := newConcurrencyManager(t.cf, t.tenantId, strategy, t.concurrencyResultsCh, t.concurrencyAdvisoryLock, t.concurrencyParentAdvisoryLock)
+		newArr = append(newArr, c)
+
+		// notify the new manager so it picks up already-queued work now instead of waiting for its first
+		// tick (matters on startup/rebalance). notify() is a non-blocking send, so it's fine under the lock.
+		c.notify(ctx)
```
abelanger5: same comment as above/below

internal/services/controllers/task/controller.go
```diff
@@ -42,30 +42,30 @@ type TasksController interface {
...
-	replayEnabled       bool
-	analyzeCronInterval time.Duration
-	signaler            *signal.OLAPSignaler
-	tw                  *trigger.TriggerWriter
-	promGate            *prometheus.Gate
+	replayEnabled                            bool
+	analyzeCronInterval                      time.Duration
+	signaler                                 *signal.OLAPSignaler
+	tw                                       *trigger.TriggerWriter
+	promGate                                 *prometheus.Gate
```
abelanger5: nit: is this linted properly?

abelanger5 (approved):

PR #4286, opened 2026-06-26
fix: add last active column to step concurrency table

mnafees:
# Description

This PR adds a new `last_active` column to `v1_step_concurrency` to mirror `v1_queue`

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

</details>


cmd/hatchet-migrate/migrate/migrations/20260626125722_v1_0_119.sql
```diff
@@ -0,0 +1,181 @@
...
+    WITH inactive_strategies AS (
+        SELECT
+            strategy.*
+        FROM
+            new_table cs
+        JOIN
+            v1_step_concurrency strategy ON strategy.workflow_id = cs.workflow_id AND strategy.workflow_version_id = cs.workflow_version_id AND strategy.id = cs.strategy_id
+        WHERE
+            strategy.is_active = FALSE
+            OR strategy.last_active < NOW() - INTERVAL '1 hour'
```
mnafees: We also want to update the v1_step_concurrency when the last_active has been stale for more than an hour - hence a very hot strat will only update the last_active every hour rather than every time.

cmd/hatchet-migrate/migrate/migrations/20260626125722_v1_0_119.sql
```diff
@@ -0,0 +1,181 @@
+-- +goose Up
+-- +goose StatementBegin
+
+ALTER TABLE v1_step_concurrency
+    ADD COLUMN last_active TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP;
```
mrkaye97: I think this should be `timestamptz` - using `timestamp(3)` in other places was a mistake

pkg/repository/sqlcv1/concurrency.sql
```diff
@@ -802,6 +802,9 @@ WITH tenant_step_concurrencies AS (
     FROM v1_step_concurrency sc
     WHERE sc.tenant_id = @tenantId::UUID
         AND sc.is_active = TRUE
+        -- only deactivate strategies idle for a full day, so a strategy used periodically stays warm
+        -- and never pays a cold start (mirrors v1_queue's 1-day activity window)
+        AND sc.last_active < NOW() - INTERVAL '1 day'
```
mrkaye97: I wonder if this is a long enough lookback 🤔 do these operate at the key level, or the strategy level? (I don't remember)

I'm nervous that if someone has a customer that doesn't have any runs for a while, then they'll see cold starts for them again (seems to not be the case if it's at a strategy level though but just would be good to confirm)

sql/schema/v1-core.sql
```diff
@@ -907,12 +915,13 @@ BEGIN
             v1_step_concurrency strategy ON strategy.workflow_id = cs.workflow_id AND strategy.workflow_version_id = cs.workflow_version_id AND strategy.id = cs.strategy_id
         WHERE
             strategy.is_active = FALSE
+            OR strategy.last_active_at < NOW() - INTERVAL '1 hour'
```
mrkaye97, line 912: do we want this here? not 100% sure, just wanted to check

abelanger5: I think this is reasonable FWIW, prevents dead tuple accumulation

mrkaye97 (approved):

PR #4303, opened 2026-06-29
fix: early return if context is already cancelled

grutt:
# Description

Avoid rollback attempts when the transaction context has already been canceled, reducing noisy rollback failure logs and stack traces during context cancellation paths.

## Type of change

## What's Changed

- [x] Return early from `DeferRollback` when the provided context is already done.

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Cursor/GPT-5.5 assisted with drafting the PR description from the branch diff.

</details>

### DIFF
--- a/pkg/repository/sqlchelpers/tx.go
+++ b/pkg/repository/sqlchelpers/tx.go
@@ -144,6 +144,11 @@ func AcquireConnectionWithStatementTimeout(ctx context.Context, pool *pgxpool.Po
 }
 
 func DeferRollback(ctx context.Context, l *zerolog.Logger, rollback func(context.Context) error) {
+
+	if ctx.Done() != nil {
+		return
+	}
+
 	if err := rollback(ctx); err != nil {
 		if !errors.Is(err, pgx.ErrTxClosed) {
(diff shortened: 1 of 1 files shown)


pkg/repository/sqlchelpers/tx.go
```diff
@@ -144,6 +144,11 @@ func AcquireConnectionWithStatementTimeout(ctx context.Context, pool *pgxpool.Po
 }
 
 func DeferRollback(ctx context.Context, l *zerolog.Logger, rollback func(context.Context) error) {
+
+	if ctx.Done() != nil {
```
mrkaye97: is this a tri-boolean? (true/false/nil)? or do we treat `nil` as `false` and `true` as itself? in any case, I think we should make this:
```suggestion
	if ctx.Done() != nil && ctx.Done() {
```

grutt: gotta love go. gc

mrkaye97 (approved):

PR #4308, opened 2026-06-29
chore: drop external payload not found errors

grutt:
# Description

Adds a terminal external payload not found error and teaches RabbitMQ pre-ack handling to drop those messages instead of retrying indefinitely. This prevents missing S3 offloaded payload index files from generating noisy retry storms, while keeping other S3 failures retryable.

The permanent drop log is enriched with structured fields like `permanent_pre_ack_reason=external_payload_not_found`, `external_payload_kind`, and `external_payload_key` for filtering and alerting.

Fixes #

## Type of change

## What's Changed

- [x] Add a shared `ErrExternalPayloadNotFound` typed error for terminal external payload misses.
- [x] Treat external payload not found errors as permanent RabbitMQ pre-ack failures.
- [x] Add structured permanent drop reason fields to RabbitMQ logs.
- [x] Map S3 `NoSuchKey` index file fetch errors to the terminal payload error.
- [x] Add focused tests for permanent pre-ack classification and S3 index file error classification.

## Checklist

Changes have been:

- [x] Linted and formatted

## Testing

- `go test ./internal/msgqueue/rabbitmq -run 'TestIsPermanentPreAckError' -count=1`
- `go test ./pkg/repository -run '^$' -count=1`
- `GOWORK=<temp local workspace> go test github.com/hatchet-dev/hatchet-cloud/internal/repository -run 'Test(FetchWithFallback|IndexFileFetchError)' -count=1`

---

<details id="ai-disclosure">

- **Details**: Cursor agent was used to inspect the retry/error paths, implement the targeted code changes, add focused tests, and run validation commands.

</details>

### DIFF
--- a/internal/msgqueue/rabbitmq/rabbitmq.go
+++ b/internal/msgqueue/rabbitmq/rabbitmq.go
@@ -25,6 +25,7 @@ import (
 	"github.com/hatchet-dev/hatchet/internal/queueutils"
 	"github.com/hatchet-dev/hatchet/pkg/logger"
 	"github.com/hatchet-dev/hatchet/pkg/random"
+	"github.com/hatchet-dev/hatchet/pkg/repository"
 	"github.com/hatchet-dev/hatchet/pkg/telemetry"
 )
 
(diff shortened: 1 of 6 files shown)


internal/msgqueue/rabbitmq/rabbitmq.go
```diff
@@ -912,28 +915,49 @@ func (t *MessageQueueImpl) subscribe(
 }
 
 func isPermanentPreAckError(err error) bool {
+	return permanentPreAckErrorReason(err) != ""
+}
+
+func permanentPreAckErrorReason(err error) string {
```
abelanger5: I don't think this can work at the msgqueue layer, because messages are buffered (on both the pub and sub side). So consumers will need to shed these messages. 

If we wanted to only nack part of the batch we'd need need to support individual acks for each message in the buffer, which we don't currently. But even if we did this, there's no way of getting around the pub buffer (which is why the pub buffer should be used sparingly)

PR #4324, opened 2026-07-01
feat: expose separate olap and core retention config

grutt:
# Description

Adds separate configuration for core and OLAP partition retention so partition cleanup windows can be tuned independently from the tenant data retention default. When either new setting is unset, it falls back to `SERVER_LIMITS_DEFAULT_TENANT_RETENTION_PERIOD`.

Fixes # (issue)

## Type of change

## What's Changed

- [x] Added `SERVER_LIMITS_CORE_PARTITION_RETENTION` and `SERVER_LIMITS_OLAP_PARTITION_RETENTION`.
- [x] Wired core partition cleanup to use the core retention override with fallback to tenant retention.
- [x] Wired OLAP partition cleanup to use the OLAP retention override with fallback to tenant retention.
- [x] Documented the new configuration variables.

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Cursor agent was used to inspect the existing retention wiring, implement the config changes, update docs, and run focused validation.

</details>

### DIFF
--- a/frontend/docs/pages/self-hosting/configuration-options.mdx
+++ b/frontend/docs/pages/self-hosting/configuration-options.mdx
@@ -203,6 +203,8 @@ Variables marked with ⚠️ are conditionally required when specific features a
 | Variable                                        | Description                     | Default Value |
 | ----------------------------------------------- | ------------------------------- | ------------- |
 | `SERVER_LIMITS_DEFAULT_TENANT_RETENTION_PERIOD` | Default tenant retention period | `720h`        |
+| `SERVER_LIMITS_CORE_PARTITION_RETENTION`        | Core partition retention period | Tenant default |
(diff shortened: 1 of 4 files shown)


mrkaye97 (commented):
I don't love this idea, mostly because even though the databases are split, the services they back are tightly coupled. Just as one example, if we e.g. had two-day OLAP retention but one-day core retention, you'd be able to see a task from two days ago in the dashboard, but you wouldn't be able to replay it. Is the main value prop here the opposite case - to have longer core db retention than OLAP?

mrkaye97 (approved):
discussed offline ✅ sounds like we need to separate these to allow for different instances sharing a db to have different retention config and not compete for partition create / drop

Issue #3680, opened 2026-04-22
[FEAT] Library-only mode and optional authentication


abelanger5:
**Is your feature request related to a problem? Please describe.**

We'd like to improve support for lightweight environments by offering a library-only mode version of Hatchet which spawns a sidecar process which runs the Hatchet engine. Most of this initiative is reducing the footprint of the Hatchet engine to make it lightweight from a memory, CPU, and database utilization perspective.

Some other dependent work is:
- Removing the requirement for authentication for local setup
- Removing the polling intervals for `UpdateDAGStatuses` and `UpdateTaskStatuses` as they are the largest contributors to connection utilization and database CPU. This is being worked on in #3554

gregfurman:
> - Removing the requirement for authentication for local setup

Seems relevant to #2821

jamietdavidson:
Any update on this?

abelanger5:
Hi @jamietdavidson , we've made progress by shipping #3554 and we're tackling making local authentication optional in the two weeks or so. These are the two pre-requisites to supporting a full local mode, so we should have a more substantial update after those are shipped.

PR #4325, opened 2026-07-01
feat: local auth disabled mode and dev images

mnafees:
# Description

Introduces the local auth disabled mode. Partially fulfills https://github.com/hatchet-dev/hatchet/issues/3680.

## Type of change

## What's Changed

- [x] Add auth disabled build tags to build `-dev` prefixed images
- [x] Add warnings about dev local usage only
- [x] Add documentation
- [x] Add new `--disable-auth` flag to `hatchet server start`   

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

</details>


api/v1/server/authn/middleware.go
```diff
@@ -138,6 +142,33 @@ func (a *AuthN) handleNoAuth(c echo.Context) error {
...
+	if err != nil {
+		a.l.Error().Ctx(ctx).Err(err).Msg("no-auth mode: could not resolve default user")
+
+		return forbidden
+	}
+
+	c.Set("user", user)
+	c.Set("auth_strategy", "noauth")
+
+	ctx = context.WithValue(ctx, analytics.UserIDKey, user.ID)
```
mnafees: Do we even need the analytics here?

api-contracts/openapi/components/schemas/metadata.yaml
```diff
@@ -33,6 +33,10 @@ APIMeta:
       type: boolean
       description: whether or not a Prometheus federation server is configured (SERVER_PROMETHEUS_SERVER_URL) on this instance
       example: false
+    noAuthEnabled:
```
abelanger5: nit: feel free to disagree, but this field (and the config vars) are confusing to me. I think `authDisabled` would be a much more immediately comprehensible.

api/v1/server/authn/middleware.go
abelanger5: probably not?

abelanger5: not sure if we case on any of this downstream

api/v1/server/authz/middleware.go
```diff
@@ -51,6 +51,8 @@ func (a *AuthZ) authorize(c echo.Context, r *middleware.RouteInfo) error {
 	var err error
 
 	switch c.Get("auth_strategy").(string) {
+	case "noauth":
+		err = a.validateUserTenantPermissions(c, r)
```
abelanger5: this is interesting...so even in the case of no auth, if the "no-auth" user tries to access a tenant which they're not a member of, it'll fail? that seems like a good default actually, scopes the no-auth to a single tenant and user (not that anyone will/should run this in production, but still)

cmd/hatchet-admin/cli/keyset.go
```diff
@@ -118,11 +146,53 @@ func runCreateLocalKeysets() error {
...
+	if err != nil {
+		return err
+	}
+
+	if err := os.WriteFile(encryptionKeyDir+"/noauth_private_ec256.key", privateEc256, 0600); err != nil {
+		return err
+	}
+
+	return os.WriteFile(encryptionKeyDir+"/noauth_public_ec256.key", publicEc256, 0600)
+}
```
abelanger5: should the "no-auth" keyset just be statically defined? I'm not sure I understand the benefit of being able to generate it, if you're generating keysets I'd imagine that we could just use the existing JWT keysets...

cmd/hatchet-admin/cli/token.go
```diff
@@ -65,18 +67,28 @@ func init() {
 		"Expiration duration for the API token",
 	)
 
+	tokenCreateAPICmd.PersistentFlags().BoolVar(
+		&tokenNoAuth,
+		"no-auth",
+		false,
+		"sign the token with the no-auth keyset for the default tenant (local no-auth mode)",
+	)
```
abelanger5: again, similar to above, I'm not sure it makes sense to be able to generate a token for the no-auth user via `hatchet-admin`, as opposed to just supporting a statically defined token which encodes a constant tenant id and unlimited expiration time.

pkg/auth/token/token.go
```diff
@@ -119,6 +144,10 @@ func (j *jwtManagerImpl) ValidateTenantToken(ctx context.Context, token string)
 
 	verifiedJwt, err := j.verifier.VerifyAndDecode(token, validator)
 
+	if err != nil && j.noAuthVerifier != nil {
+		verifiedJwt, err = j.noAuthVerifier.VerifyAndDecode(token, validator)
+	}
```
abelanger5: this makes me nervous -- is there any way we can avoid touching the `auth/token` package here and just support switching on the path for decoding and verifying the jwt from the middleware which requires it? one bad initialization (for example, initializing a zero-valued noAuthVerifier) could lead to auth bypass

api/v1/server/handlers/metadata/get.go
abelanger5, line 1: If auth is disabled, we should probably disable a number of other endpoints and handlers as well, such as: creating tenants, creating users, etc.

.github/workflows/pre-release.yaml
```diff
@@ -26,9 +26,19 @@ jobs:
             --build-arg VERSION=${{steps.tag_name.outputs.tag}} \
             --platform linux/amd64 \
             .
+      - name: Build dev image
```
abelanger5: can we make each of these separate jobs instead? that way they build in parallel and we can cleanly isolate them in their own part of the CI file

api/v1/server/authn/middleware.go
abelanger5, line 41: We could use build tags to re-export `Middleware` so that it only returns `return a.handleAuthDisabledBypass(c)`? This gets us around dealing with the case `if authmode.Disabled`

api/v1/server/authz/middleware.go
```diff
@@ -48,9 +49,15 @@ func (a *AuthZ) authorize(c echo.Context, r *middleware.RouteInfo) error {
 		return nil
 	}
 
+	if authmode.Disabled && rbac.OperationIn(r.OperationID, authDisabledDeniedOperations) {
+		return echo.NewHTTPError(http.StatusForbidden, "This operation is disabled while authentication is disabled")
+	}
```
abelanger5: same with this, could be abstracted into an `authPreflight` method

frontend/app/src/components/layout/auth-disabled-banner.tsx
abelanger5, line 1: I think this banner should render in a warning color (yellow) dismissable somehow?

frontend/app/src/pages/main/v1/side-nav-items.tsx
```diff
@@ -279,16 +280,22 @@ export function sideNavItems(opts: {
...
-          key: 'settings-api-tokens',
-          name: 'API Tokens',
-          to: appRoutes.tenantSettingsApiTokensRoute.to,
-          icon: ({ collapsed }: { collapsed: boolean }) => (
-            <RiKey2Line
-              className={collapsed ? 'size-5' : 'mr-2 size-4 shrink-0'}
-            />
-          ),
-        },
+        // API tokens are unavailable in authdisabled instances (the only token is the
```
abelanger5: should we consider showing the API tokens section, but just showing the user the static auth token instead? I'd generally like to keep parity between top-level UI pages as much as possible, so that users are familiar with the UI when toggling between self-hosted / no-auth / cloud.

api/v1/server/authz/middleware.go
```diff
@@ -48,9 +49,15 @@ func (a *AuthZ) authorize(c echo.Context, r *middleware.RouteInfo) error {
...
+
 	var err error
 
 	switch c.Get("auth_strategy").(string) {
+	case "authdisabled":
+		err = a.validateUserTenantPermissions(c, r)
```
abelanger5: same here, I'd prefer a separately built file to handle this logic. the idea is that we should remove any potential footguns by modifying this auth logic in the future.

.github/workflows/pre-release.yaml
abelanger5, line 1: should we consider adding a test somehow which runs the non-dev engines / API containers and absolutely ensures that auth is enabled? it's a little annoying, but might save us if we accidentally set the wrong build args in the future.

.github/workflows/test.yml
```diff
@@ -738,3 +738,32 @@ jobs:
           TESTING_MATRIX_RABBITMQ_ENABLED: ${{ matrix.rabbitmq-enabled }}
           TESTING_MATRIX_PG_VERSION: ${{ matrix.pg-version }}
           RAMP_UP_DURATION_TIMEOUT: 30s
+
+  auth-enabled:
```
abelanger5: why is this in `test.yaml`? I'd imagine we'd want to run `assert-auth-enabled.sh` as part of `pre-release`, after building the images but before publishing them?

api/v1/server/authn/authdisabled.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/labstack/echo/v4"
+)
+
+func (a *AuthN) authPreflight(c echo.Context) (handled bool, err error) {
+	forbidden := echo.NewHTTPError(http.StatusForbidden, "Please provide valid credentials")
+
+	ctx := c.Request().Context()
+
+	user, err := a.config.V1.User().GetUserByEmail(ctx, a.config.Seed.AdminEmail)
```
grutt, line 14: open question: should this be the seeded user or should we force an email env var or something?

api/v1/server/authz/authdisabled.go
```diff
@@ -0,0 +1,32 @@
...
+
+func (a *AuthZ) authPreflight(c echo.Context, r *middleware.RouteInfo) (handled bool, err error) {
+	if rbac.OperationIn(r.OperationID, authDisabledDeniedOperations) {
+		return true, echo.NewHTTPError(http.StatusForbidden, "This operation is disabled while authentication is disabled")
+	}
+
+	return true, a.validateUserTenantPermissions(c, r)
+}
+
+var authDisabledDeniedOperations = []string{
```
grutt: what's our plan to maintain this list as we continue to add more endpoints?

cmd/hatchet-admin/cli/seed/seed.go
```diff
@@ -100,5 +102,41 @@ func SeedDatabase(dc *database.Layer) error {
...
+	if authmode.Disabled {
+		if err := seedAuthDisabledToken(dc); err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
+
+func seedAuthDisabledToken(dc *database.Layer) error {
```
grutt, line 114: so seed is required in this mode?

pkg/authmode/token.go
```diff
@@ -0,0 +1,12 @@
+package authmode
+
+// Fixed claims of the embedded auth-disabled token. The loader pins the JWT manager to these so the
+// one committed token validates on every authdisabled instance.
+const (
+	EmbeddedTokenIssuer      = "http://localhost:8888" //nolint:gosec // URL, not a credential
```
grutt, line 6: wait so a user cant configure the port?

grutt (commented):
some Qs

api/v1/server/authn/authdisabled.go
mnafees: I chose to simply reuse the seeded tenant and user to avoid adding in more config options

.github/workflows/test.yml
mnafees: seemed natural to me to keep this as part of the CI tests; can change it to pre-release though

cmd/hatchet-admin/cli/seed/seed.go
mnafees: yup, the idea would be similar to how we seed in other environments with non-dev images

pkg/authmode/token.go
mnafees: hmm, good point 🤔 this is intentional for the global single-shared-noauth-token behavior; server-side the loader pins the validator opts to these constants so validation passes regardless of the actual listen port

api/v1/server/authz/authdisabled.go
mnafees: good call out; I have added a `NOAUTH` role (inherits `MEMBER`) to the existing `rbac.yaml` and the noauth'd user will default to this role going forward which should basically get rid of the previous deny list pattern I had implemented

api/v1/server/authn/middleware.go
```diff
@@ -60,6 +60,10 @@ func (a *AuthN) authenticate(c echo.Context, r *middleware.RouteInfo) error {
 		return a.handleNoAuth(c)
 	}
 
+	if handled, err := a.authPreflight(c); handled {
+		return err
+	}
```
mrkaye97, line 68: I'd leave a comment here maybe about what this does, since if someone were to e.g. change this to `!handled` it seems like it'd disable auth globally?

api/v1/server/authz/middleware.go
```diff
@@ -48,6 +48,10 @@ func (a *AuthZ) authorize(c echo.Context, r *middleware.RouteInfo) error {
 		return nil
 	}
 
+	if handled, err := a.authPreflight(c, r); handled {
+		return err
+	}
```
mrkaye97, line 56: same here maybe?

api/v1/server/authz/middleware.go
```diff
@@ -192,14 +196,14 @@ func (a *AuthZ) validateUserTenantPermissions(c echo.Context, r *middleware.Rout
 	return nil
 }
 
-func (a *AuthZ) authorizeTenantOperations(tenantMemberRole sqlcv1.TenantMemberRole, r *middleware.RouteInfo) error {
+func (a *AuthZ) authorizeTenantOperations(roleName string, r *middleware.RouteInfo) error {
```
mrkaye97, line 202: what's the reason for this change?

api/v1/server/handlers/metadata/get.go
```diff
@@ -36,6 +37,8 @@ func (u *MetadataService) MetadataGet(ctx echo.Context, request gen.MetadataGetR
 
 	prometheusServerEnabled := u.config.Prometheus.PrometheusServerURL != ""
 
+	authDisabled := authmode.Disabled
```
mrkaye97: this confused me - I initially read this as hard-coding `authDisabled` to the equivalent of `true` (`disabled`), but this is actually a constant. Can we rename `Disabled` to `IsDisabled` or `IsAuthDisabled` to make this more clear? We need to do a better job of this everywhere 😓

cmd/hatchet-admin/cli/token.go
```diff
@@ -64,14 +65,19 @@ func init() {
...
+	if authmode.Disabled {
+		fmt.Println(authmode.EmbeddedToken())
+		return nil
+	}
+
 	// read in the local config
 	configLoader := loader.NewConfigLoader(configDirectory)
 
-	cleanup, server, err := configLoader.CreateServerFromConfig("", func(scf *server.ServerConfigFile) {
+	cleanup, srv, err := configLoader.CreateServerFromConfig("", func(scf *server.ServerConfigFile) {
```
mrkaye97, line 80: nit, but do we need this naming change? imo, `server` is more clear than `srv`

frontend/app/src/components/layout/auth-disabled-banner.tsx
```diff
@@ -0,0 +1,45 @@
...
+        >
+          <Link
+            to={appRoutes.tenantSettingsApiTokensRoute.to}
+            params={{ tenant: tenantId }}
+          >
+            View worker token
+          </Link>
+        </Button>
+      )}
+      <button
```
mrkaye97: let's use a shadcn button for this

mrkaye97 (approved):
I left some small comments, but generally this looks good.

One thing I'm a bit nervous about is polluting the frontend with more conditional rendering based on this state, but I'm not sure of other ways around it, unfortunately

api/v1/server/authz/middleware.go
mnafees: we now have an internal-only `NOAUTH` role which does not conform to a `sqlcv1.TenantMemberRole` anymore

cmd/hatchet-admin/cli/token.go
mnafees: so I chose to do this since the package name is also `server` and it was confusing me while I was revisiting this code 😅

abelanger5 (approved):

abelanger5 (approved):

PR #4328, opened 2026-07-01
fix: metering bugs (deadlock, phantom context canceled), deadlocks, and index on StepExpression

abelanger5:
# Description

Fixes several engine-side bugs:
1. On the workflow trigger path, we risk deadlocking when we call `Meter` without a cached result, since we acquire a new connection from the pool if we miss the cache.
2. On `ReplayTasks`, `ListStepMatchConditions` was not using the existing tx
3. On the OLAP side, `ListTasks` was incorrectly running two goroutines, one which acquired a new connection from the `readPool` and the other which used the existing tx, and then waiting for the goroutines inside of the existing tx, which is a deadlock risk. This changes the behavior to match `ListWorkflowRuns`, and also changes the behavior of both methods so that failing to count results in an error.
4. Adds an index to `StepExpression` so that we can directly look these up by `stepId`, which was causing unexpectedly high triggering times.
5. Constructs a separate context for the postcommit hook after metering to avoid phantom `context canceled` statements appearing in the logs.  

## Type of change

## What's Changed

See above. 

## Checklist

Changes have been:

- [X] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Claude for debugging and light implementation (particularly on the postcommit hooks)

</details>

### DIFF
--- a/pkg/repository/tenant_limit.go
+++ b/pkg/repository/tenant_limit.go
@@ -39,7 +39,7 @@ type TenantLimitRepository interface {
 
 	DefaultLimits() []Limit
 
-	Meter(ctx context.Context, resource sqlcv1.LimitResource, tenantId uuid.UUID, numberOfResources int32) (precommit func() error, postcommit func())
+	Meter(ctx context.Context, dbtx sqlcv1.DBTX, resource sqlcv1.LimitResource, tenantId uuid.UUID, numberOfResources int32) (precommit func() error, postcommit func())
 
 	Stop()
 }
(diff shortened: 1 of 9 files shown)


pkg/repository/tenant_limit.go
```diff
@@ -39,7 +39,7 @@ type TenantLimitRepository interface {
 
 	DefaultLimits() []Limit
 
-	Meter(ctx context.Context, resource sqlcv1.LimitResource, tenantId uuid.UUID, numberOfResources int32) (precommit func() error, postcommit func())
+	Meter(ctx context.Context, dbtx sqlcv1.DBTX, resource sqlcv1.LimitResource, tenantId uuid.UUID, numberOfResources int32) (precommit func() error, postcommit func())
```
mrkaye97, line 42: should this be a pointer? I sort of dislike this implicitly nullable thing (same with the `pgtype.Foobar` types


```suggestion
	Meter(ctx context.Context, dbtx *sqlcv1.DBTX, resource sqlcv1.LimitResource, tenantId uuid.UUID, numberOfResources int32) (precommit func() error, postcommit func())
```

mrkaye97 (approved):

pkg/repository/tenant_limit.go
abelanger5: `DBTX` is an interface, so it should not be a pointer

mrkaye97: ooo! gotcha, thanks

PR #4355, opened 2026-07-05
feat: library / embedded mode

### LINKED ISSUE #3680: [FEAT] Library-only mode and optional authentication

mnafees:
# Description

Adds the ability to spawn an embedded Hatchet engine from a Go SDK client simply by providing a URL to a PostgreSQL DB.

```go
import (
    hatchet "github.com/hatchet-dev/hatchet/sdks/go"

    _ "github.com/hatchet-dev/hatchet/embed"
)

client, err := hatchet.NewClient(hatchet.WithEmbeddedPostgres(databaseURL))
if err != nil {
    return err
}
defer func() { _ = client.Close(context.Background()) }()
```

This runs a full Hatchet engine (no UI) in-process with the Go binary it is embedded in. 

Partially resolves https://github.com/hatchet-dev/hatchet/issues/3680

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

</details>


embed/keyset.go
mnafees, line 1: The idea here is to have a shared `hatchet_embed.keyset` DB table which keeps the autogenerated keyset information for such embedded instances and the user does not need to generate a keyset themselves.

When `WithKeysets` is provided we always use that instead of this path.

Can convert to required `sqlc` constructs once this approach is approved.

embed/version.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"fmt"
+	"runtime/debug"
+	"strings"
+)
+
+const hatchetModulePath = "github.com/hatchet-dev/hatchet"
+
+func resolveVersion() (string, error) {
```
mnafees, line 11: Just resolving the library tagged version from the respective `go.mod` file

embed/embed.go
```diff
@@ -0,0 +1,245 @@
...
+	if cfg.adminEmail != "" {
+		_ = os.Setenv("ADMIN_EMAIL", cfg.adminEmail)
+	}
+	if cfg.adminPassword != "" {
+		_ = os.Setenv("ADMIN_PASSWORD", cfg.adminPassword)
+	}
+
+	if cfg.runMigrations {
+		migrate.RunMigrations(ctx)
+	}
```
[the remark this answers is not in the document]
mnafees: I think I like this and returning an error for this use-case might actually make sense.

embed/embed.go
```diff
@@ -0,0 +1,245 @@
...
+
+	if err := os.Setenv("DATABASE_URL", cfg.postgresURL); err != nil {
+		return nil, fmt.Errorf("could not set DATABASE_URL: %w", err)
+	}
+	if cfg.adminEmail != "" {
+		_ = os.Setenv("ADMIN_EMAIL", cfg.adminEmail)
+	}
+	if cfg.adminPassword != "" {
+		_ = os.Setenv("ADMIN_PASSWORD", cfg.adminPassword)
+	}
```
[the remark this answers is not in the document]
mnafees: I don't like this since I think it makes sense to simply override these env vars from within the embed engine since we have config level options. Open to what you think @abelanger5

cmd/hatchet-engine/main.go
```diff
@@ -46,7 +46,6 @@ var rootCmd = &cobra.Command{
 }
 
 // Version will be linked by an ldflag during build
-// FIXME: automate this version update on tag, we use it to version the engine for sdks
 var Version = "v0.83.4"
```
abelanger5: Let's bump this version? Avoids the security advisory on startup in dev mode

embed/example/trigger/main.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+type GreetOutput struct {
+	Worker string `json:"worker"`
+}
+
+func main() {
+	ctx := context.Background()
+
+	info := readEngine(env("ENGINE_FILE"))
```
abelanger5: could you explain a bit more what this is supposed to be doing? reading through this it feels like bad developer experience. 

For me the ideal devex is just to instantiate a client from the API using `hatchet.NewClient(hatchet.WithEmbeddedPostgres(databaseUrl))`. 

I should be able to do this from both the API and the workers: the postgres database is the coordination layer and only shared source of truth. 

Reading through this, it seems like we're treating a single engine instance running on the host as a coordination layer, which is bad for a number of reasons:
1. It implies some concept of "leadership" which introduces a bunch of edge cases like - what happens when the worker running that engine is interrupted?
2. It means that if I have a fleet of workers on different machines, they need some way to network together, which most configurations won't have.

pkg/repository/task.go
```diff
@@ -3059,9 +3058,7 @@ func makeEventTypeArr(status sqlcv1.V1TaskEventType, n int) []sqlcv1.V1TaskEvent
 }
 
 func hash(s string) int64 {
```
abelanger5: let's just replace calls to `hash` with this?

embed/example/trigger/main.go
```diff
@@ -23,29 +24,35 @@ type GreetOutput struct {
...
 
-	_ = os.Setenv("HATCHET_CLIENT_TOKEN", info["token"])
-	_ = os.Setenv("HATCHET_CLIENT_HOST_PORT", info["grpcAddress"])
-	_ = os.Setenv("HATCHET_CLIENT_SERVER_URL", info["apiURL"])
-	_ = os.Setenv("HATCHET_CLIENT_TENANT_ID", info["tenantID"])
-	_ = os.Setenv("HATCHET_CLIENT_TLS_STRATEGY", "none")
+	embedOpts, err := portOpts()
+	if err != nil {
+		return err
+	}
```
abelanger5: This is something that we should be automatically figuring out from the embedded engine. We should find an available port and use that as the config.

embed/keyset.go
abelanger5: Just flagging that there's pretty much no benefit to encryption in this case, so we should likely print a warning to the console (at the very least) when running in this mode.

abelanger5: I also wonder if we want to move the embedded keyset management into its own package, and design it similarly to how we did with the pgoutbox schema: https://github.com/hatchet-dev/pgoutbox/blob/main/migrate.go

sdks/go/examples/embedded/go.mod
abelanger5, line 1: can we move this example into the SDK examples section?

examples/go/embedded/trigger/main.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+}
+
+func run() error {
+	ctx := context.Background()
+
+	databaseURL := env("DATABASE_URL")
+	runs := atoiDefault(os.Getenv("RUNS"), 30)
+
+	client, err := hatchet.NewClient(hatchet.WithEmbeddedPostgres(databaseURL))
```
abelanger5, line 38: should we consider looking for a url like `HATCHET_CLIENT_EMBEDDED_DATABASE_URL` or similar? And start in embedded mode if this is set to a non-empty value?

abelanger5 (approved):

PR #4356, opened 2026-07-06
feat: add event time bounds to aggregated analytics flushes

BloggerBust:
# Description

Aggregated Count events are stamped with the flush time, so the real event times are lost. Track the first and last event time in each bucket and include them on the flushed event as first_event_at and last_event_at. The bounds can disagree with the count by one event at an interval boundary. Exact consistency would need a lock in Count, which runs for every event.

## Type of change

## What's Changed

- Each bucket records the first and last event time seen since its last flush. The flush resets them, so the values bracket one flush interval.
- FlushFunc gains firstEventAt and lastEventAt parameters.
- Flushed count events carry the times as first_event_at and last_event_at in UTC, omitted when unknown. Caller properties with those names are dropped, so the values always come from the aggregator.
- Immediate Enqueue events are unchanged.

## Checklist

Changes have been:

- [x] Linted and formatted

## Testing

`go test -count=1 -race ./pkg/analytics/...` passes. The new aggregator tests use an injected clock and direct flush calls, so the assertions are deterministic

---

<details id="ai-disclosure">

- **Details**: Claude Code (Opus 4.8) wrote the tests.

</details>

### DIFF
--- a/pkg/analytics/aggregating.go
+++ b/pkg/analytics/aggregating.go
@@ -31,9 +31,7 @@ type counterEntry struct {
 
 // The first marker is set once and the last marker only advances
 func (e *counterEntry) observe(nanos int64) {
-	if e.firstEventAtNanos.Load() == 0 {
-		e.firstEventAtNanos.CompareAndSwap(0, nanos)
-	}
+	e.firstEventAtNanos.CompareAndSwap(0, nanos)
 
 	for {
 		last := e.lastEventAtNanos.Load()
@@ -198,7 +196,7 @@ func (a *Aggregator) flush() {
 			// marker whose count was already flushed. A bucket's first
 			// event is never misplaced, because Count fills the markers
 			// before publishing a new bucket, so the earliest
-			// first_event_at in a bucket's lifetime is exact. Preventing
+			// firstEventAt in a bucket's lifetime is exact. Preventing
 			// the misplacement would require Count to take a lock.
 			firstNanos := e.firstEventAtNanos.Swap(0)
 			lastNanos := e.lastEventAtNanos.Swap(0)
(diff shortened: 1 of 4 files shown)


pkg/analytics/posthog/posthog.go
```diff
@@ -154,11 +154,22 @@ func (p *PosthogAnalytics) Count(ctx context.Context, resource analytics.Resourc
...
+	// The event-time bounds always come from the aggregator, never from
+	// caller properties.
+	delete(merged, "first_event_at")
+	delete(merged, "last_event_at")
+	if !firstEventAt.IsZero() {
+		merged["first_event_at"] = firstEventAt
+	}
+	if !lastEventAt.IsZero() {
+		merged["last_event_at"] = lastEventAt
+	}
```
mrkaye97: it's a small thing, but I wonder if we should prefix these so we don't accidentally delete an intentional `first_event_at` and cause confusion - e.g. `__aggregator__first_event_at` or something? idk, it's kind of ugly, but would make us less likely to shoot ourselves in the foot. wdyt? (no strong feelings really)

pkg/analytics/aggregating.go
```diff
@@ -24,9 +24,33 @@ type counterEntry struct {
...
+	firstEventAtNanos atomic.Int64
+	lastEventAtNanos  atomic.Int64
 }
 
-type FlushFunc func(resource Resource, action Action, tenantID uuid.UUID, tokenID *uuid.UUID, count int64, properties Properties)
+// The first marker is set once and the last marker only advances
+func (e *counterEntry) observe(nanos int64) {
+	if e.firstEventAtNanos.Load() == 0 {
+		e.firstEventAtNanos.CompareAndSwap(0, nanos)
+	}
```
mrkaye97: small nit: I'm not familiar with `CompareAndSwap` but I looked at the docs - seems like maybe this will only do anything if `e.firstEventAtNanos.Load()` is zero, so we might not need the `if` check? let me know if I'm wrong there, not too sure!

mrkaye97 (approved):

PR #4368, opened 2026-07-07
fix(python): durable evict / restore bug causing sdk to hang

mrkaye97:
# Description

### Fixed

- Fixes an issue in the durable event listener where we could hang indefinitely waiting for an ack. Now, we'll time out and allow the task to retry, so it should be able to recover independently.
- Fixes an issue in the durable execution logic where collisions in id + retry count would cause unexpected behavior - fixed by adding the invocation count to the key, when it's provided.
- Fixes an issue in the durable execution logic where a listener reconnect could cause messages to be stuck in the old request queue, causing the listener to hang indefinitely. Now, we will shovel messages from the old queue to the new queue on reconnect, so that they can be processed normally.

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Used Fable significantly to help diagnose the issues and propose solutions, and take an initial pass at implementation (later edited)

</details>


grutt (approved):
yikes. do we also need this in the other sdks?

mrkaye97:
> yikes. do we also need this in the other sdks?

I think it's pretty likely, but I want to see how this goes and then audit

PR #4369, opened 2026-07-07
fix(cloud): drop tasks past payload store retention

grutt:
# Description

Resolves the FIXME in the dispatcher's populateTaskData: when bulk input retrieval fails, the dispatcher previously requeued every task in the batch regardless of whether the error was transient (e.g. an S3 rate limit) or permanent (the payload no longer exists in the external store). Tasks whose offloaded payloads are permanently gone  would loop forever through assign/requeue, never reaching a terminal state. Because these zombie tasks also hold GROUP_ROUND_ROBIN concurrency slots, they can block all other work behind the same concurrency keys and build up a large backlog for the tenant.

This PR introduces a sentinel error, repository.ErrPayloadNotFound. When the dispatcher sees this error from a bulk retrieval, it retries retrieval per task and splits the batch: tasks whose payloads are permanently missing are failed non-retryably with a descriptive error message.

A companion change in the cloud S3 external store wraps NoSuchKey errors on index file fetches (and external IDs missing from an index file) with the new sentinel; it depends on this PR being released first.

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted
- [] Added to CHANGELOG (where applicable) -- see [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)

---

<details id="ai-disclosure">

- **Details**: Claude (Cursor agent) was used to investigate the production incident that motivated this change

</details>


internal/services/dispatcher/dispatcher.go
```diff
@@ -707,10 +721,35 @@ func (d *DispatcherImpl) populateTaskData(
...
+
+				remaining = append(remaining, task)
+			case errors.Is(taskErr, v1.ErrPayloadNotFound):
+				d.l.Error().Ctx(ctx).Err(taskErr).Int64("task_id", task.ID).Msg("task input payload no longer exists in external store, failing task")
+				fail(task)
+			default:
+				requeue(task)
+			}
+		}
+
```
mrkaye97: this feels pretty risky to me? N+1 query on a critical path?

grutt: yeah fair, let me see if i can figure out how to get the needed context out of cloud

grutt: we can also split the group in half until we have only 1 task failure batch at which point we can test this case....

grutt: ok was able to get some context out of cloud here..

internal/services/dispatcher/dispatcher.go
```diff
@@ -956,6 +992,55 @@ func (d *DispatcherImpl) handleRetries(
...
+	failCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	failGroup := errgroup.Group{}
+
+	for _, _task := range toFail {
+		task := _task
+
+		failGroup.Go(func() error {
```
mrkaye97, line 1011: do we do this kind of thing concurrently in other places? just checking to make sure we're being consistent, not sure off the top of my head

pkg/repository/payloadstore.go
```diff
@@ -431,6 +470,20 @@ func (p *payloadStoreRepositoryImpl) retrieve(ctx context.Context, tx sqlcv1.DBT
...
+
+			for _, externalOpt := range notFoundErr.Missing {
+				if opt, exists := retrieveFromExternalOptsToOpts[externalOpt]; exists {
+					missing = append(missing, opt)
+				}
+			}
+
+			// return the payloads which were found alongside the error so callers can proceed
+			// with the rest of the batch
+			return optsToPayload, &PayloadNotFoundError{Missing: missing}
```
mrkaye97: this sketches me out a little - feels odd to be returning both a result _and_ an error

PR #4372, opened 2026-07-07
fix(engine): raise non-determinism error on incorrect skip entry ordering

mrkaye97:
# Description

Fixes a bug where if we somehow get out-of-order skip entries (from incorrect / jumbled child keys or child indices set on the sdks), we'd potentially return the wrong payload to the caller since we're keyed on the node id. Instead, we check the idempotency key associated with the incoming trigger opts and raise non-determinism, since this is at least a clearer error message.

the underlying bug (sending the wrong index) should have been fixed already in #4368

also added a test that reproduces the bug on main (which shows the same error as the one in the original report):

```python
pydantic_core._pydantic_core.ValidationError: 1 validation error for OutputA
which_child
Input should be 'a' [type=literal_error, input_value='b', input_type=str]
For further information visit https://errors.pydantic.dev/2.13/v/literal_error
```

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

<details id="ai-disclosure">

- **Details**: Fable helped diagnose the issue

</details>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Install Protoc
         uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b # v3.0.0
         with:
-          version: "29.3"
+          version: "29.6"
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Install Task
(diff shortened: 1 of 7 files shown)


mnafees (commented):
LGTM 🚀 nice find

mnafees (approved):

Issue #11, opened 2023-12-19
Feat: introduce a scheduling state


steebchen:
This can be either a new scheduling state or a specific timeout of type ScheduleTimeout

abelanger5:
Scheduling timeout added via #24, will make it fully configurable once the use-case comes up.

PR #4375, opened 2026-07-08
feat: name runs with a CEL display-name expression on the workflow/task definition (#4259)

### LINKED ISSUE #4259: [FEAT] dynamic names for hatchet tasks

When my parent task triggers many child tasks, it is very hard to distinguish between them. For example:

Image

Ideally, I should be able to identify the task I want using something like the company name, similar to what is possible in Inngest. In Hatchet, this is currently very complicated because I have to check each task input one by one.

In the SDKs, we should be able to pass either a name or a prefix when triggering child tasks, so that later in the UI we can recognize them more easily.

---

<details id="ai-disclosure">

- **Details**: [e.g. scanning documentation, solution brain-storming]

</details>

Mdev303:
Name runs with a human-readable label by declaring a [CEL](https://github.com/google/cel-spec) expression on the workflow and/or task **definition**. The engine evaluates it against each run's input at trigger time and stores the result as the run/step display name. Because the expression lives in the definition, **every** trigger source — manual, `run_many`/bulk, child spawn, event, and cron — produces meaningful names automatically, and DAG steps can be named independently.

Fixes #4259

## How it works

The whole string is evaluated as one CEL expression (consistent with concurrency and rate-limit keys — there is no `{{ }}` templating):
- Reference run input with `input.<key>`, e.g. `input.customerName`.
- Quote string literals as CEL strings, e.g. `'Acme Corp'`.
- Guard optional keys with `has()`, e.g. `has(input.name) ? input.name : 'run'`.

**Levels & precedence**
- A **workflow-level** expression names the run — the DAG run for a multi-step workflow, or the task for a single-task workflow (the task *is* the run).
- A **per-task** expression names an individual DAG step.
- On a **single-task** workflow the precedence is: task expression → workflow expression → generated fallback.

**Failures never fail the run.** A malformed expression is rejected at registration (via the `celworkflowrunstr` validator). At run time, any evaluation error — missing key, non-string result, or empty result — silently falls back to the generated `<readableId|name>-<timestamp>` label. A result longer than 255 characters is stored truncated (rune-safe), never rejected.

## Usage

Set the expression on the workflow and/or task definition — it applies to every run regardless of trigger source.

**Python**
```python
customer_workflow = hatchet.workflow(
    name="CustomerWorkflow",
    display_name="input.customerName",   # names the run
)

@customer_workflow.task(display_name="'enrich-' + input.customerName")  # names this DAG step
def enrich(input: WorkflowInput, ctx: Context) -> None: ...
```

**TypeScript**
```ts
const customerWorkflow = hatchet.workflow({
  name: "customer-workflow",
  displayName: "input.customerName",
});

customerWorkflow.task({ name: "enrich", displayName: "'enrich-' + input.customerName", fn });
```

**Go**
```go
wf := client.NewWorkflow("customer-workflow", hatchet.WithWorkflowDisplayName("input.customerName"))
wf.NewTask("enrich", enrichFn, hatchet.WithDisplayName("'enrich-' + input.customerName"))
```

**Ruby**
```ruby
wf = hatchet.workflow(name: "CustomerWorkflow", display_name: "input.customerName")
wf.task(name: "enrich", display_name: "'enrich-' + input.customerName") { |input, ctx| ... }
```

## Design decisions & trade-offs

- **CEL on the definition, not a literal at trigger time.** One declaration names every run of the workflow across all trigger sources (manual/bulk/child/event/cron), and enables independent per-step names — neither of which the trigger-time literal could do.
- **Pure CEL (whole string).** Consistent with concurrency and rate-limit keys; no bespoke `{{ }}` templating to learn or maintain.
- **Input-only environment** (`input`, `additional_metadata`, `workflow_run_id`) — no parent-step outputs, keeping evaluation available at trigger time.
- **Validated at registration, silent at runtime.** A bad expression is caught when you register the workflow (`celworkflowrunstr`); a runtime failure (missing key / non-string / empty) never fails the run — it falls back to the generated name. `>255` runes → truncated (rune-safe), never rejected.
- **Versioned.** The expressions are folded into the workflow-version checksum, so changing an expression produces a new version.
- **Backward compatible.** Both `display_name` fields are optional; omitting them reproduces today's exact generated-name behavior.

## What's changed

- **Contracts:** removed the trigger-time `display_name` (`TriggerWorkflowRequest` #11, `TriggerWorkflowRunRequest` #6) and marked both field numbers/names `reserved`; added `CreateWorkflowVersionRequest.display_name` (#15) and `CreateTaskOpts.display_name` (#16). OpenAPI trigger body field removed. All generated code (Go, REST, frontend, and all 4 SDK proto stubs) regenerated.
- **Storage:** nullable `displayName` columns on `WorkflowVersion` and `Step` (migration `20260721120000_v1_0_126`) + sqlc regen; opts tagged `celworkflowrunstr` and folded into the version checksum.
- **Engine eval:** pure `resolveDisplayName` helper; workflow-level expr evaluated in `createDAGs` → `v1_dag.display_name`, step/workflow expr evaluated in `insertTasks` → `v1_task.display_name`, with silent fallback + `NormalizeDisplayName` (trim + 255-rune truncate).
- **admin service:** `req.DisplayName` / step `DisplayName` threaded through `getCreateWorkflowOpts` / `getCreateTaskOpts` so definition expressions reach the repository via normal gRPC `PutWorkflow`.
- **SDKs (Go, TypeScript, Python, Ruby):** definition-level `display_name` on the workflow and task builders (mirroring `concurrency`), threaded into `CreateWorkflowVersionRequest` / `CreateTaskOpts`; all trigger-time run-call surfaces removed.
- **Docs / changelog:** `run-names.mdx` rewritten for the CEL mechanism; root + per-SDK changelogs updated.

## Type of change

## Checklist

- [x] **Tested** — `resolveDisplayName` unit tests, checksum cases, admin-service conversion regression tests (`getCreateWorkflowOpts`/`getCreateTaskOpts`), per-SDK definition-threading tests (Go 4, TS 9, Python 5, Ruby 7), and a **13-case CEL integration suite** run against live Postgres + RabbitMQ (single-task/DAG, step-vs-workflow precedence, event-triggered runs, fan-out distinct names, fallbacks, 255-rune truncation, invalid-expr rejection).
- [x] Linted / type-checked (Go vet, `tsc` + eslint, `mypy` + ruff, Ruby specs).
- [x] Documented (Run Names guide rewritten).
- [x] Added to CHANGELOG (root + per-SDK).

<details id="ai-disclosure">

- **Details**: Implemented with Claude Code reviewed by me

</details>


Mdev303:
I will fix the conflicts once I get a review , it's mostly changelog stuff or the proto regen

abelanger5:
Hi @Mdev303, thanks for the PR! I think it would be cleaner to use CEL-based templating instead of trigger-based arguments, for the following reasons:
1. It would allow us to set display names on specific steps more easily
2. It would allow for display names when triggering from events or dynamically registered crons
3. It's consistent with how other configuration is written on a per-step basis; priority is one of the exceptions, but even that has a per-step default.

I'm not strictly opposed to trigger-based overrides, but I think I prefer CEL-based for this initial scope. Trigger-based is always possible by just setting a `displayName` on the input.

Mdev303:
@abelanger5 Thanks for the review ! I made the change it now uses CEL based templating ( I also updated the PR title and description ) 

I also tested with python and TS sdk manually 

image


let me know if there is anything else

PR #4379, opened 2026-07-08
fix: invert shouldRefill() condition in v1 scheduler rate limiter

### LINKED ISSUE #4346: [BUG] v1 scheduler rate limiter: shouldRefill() condition is inverted
**Describe the issue**

`shouldRefill()` in the v1 scheduler rate limiter has its condition inverted:

```go
// pkg/scheduling/v1/rate_limit.go:174
func (r *rateLimiter) shouldRefill() bool {
	...
	return r.nextRefillAt.After(time.Now().UTC())
}
```

This returns true while the refill time is still in the future — the opposite of when a refill is due. Compare with the guard inside `flushToDatabase()` (line 292), which no-ops while `nextRefillAt.After(now)` and only does real work once the boundary has passed. The two are meant to agree; `shouldRefill()` got the negation wrong.

Both failure modes land in `use()` at line 118:

- After a window rolls over, `shouldRefill()` reports false, so `use()` never triggers the on-demand refresh. Tasks are evaluated against the depleted counts from the previous window and rejected, with a stale `nextRefillAt` in the result, until the 1s `loopFlush` ticker fires. Worst case ~1s of spurious rate-limit rejections per window rollover, per tenant.
- Before the window rolls over — i.e. almost always — `shouldRefill()` reports true, so every `use()` call runs `flushToDatabase()`, which takes `unflushedMu`, `dbRateLimitsMu`, and `nextRefillAtMu` as write locks and then returns without doing anything. `use()` sits on the per-task scheduling path, so this is constant write-lock churn for zero work.

**Environment**

- `main` @ 0b51d955. The condition is unchanged since the file was introduced, so all released v1 engines have it.

**Expected behavior**

`use()` refreshes from the database once `now >= nextRefillAt`, and skips `flushToDatabase()` entirely while the window is open.

**Code to Reproduce**

Two unit tests in the style of the existing `rate_limit_test.go` fail on `main`:

1. `shouldRefill()` with `nextRefillAt` 1s in the past returns false (want true); 1s in the future returns true (want false).
2. A `rateLimiter` with a depleted in-memory limit (`val: 0`), `nextRefillAt` in the past, and a mock repo returning refreshed rows: `use()` rejects instead of refreshing.

**Additional context**

The fix is a one-line negation to bring `shouldRefill()` in line with the `flushToDatabase()` guard. I have it ready on a branch with the regression tests above and will open a PR if this is accepted. Full diff, per the AI policy's share-code-first guidance: https://github.com/hatchet-dev/hatchet/compare/main...abhay-codes07:hatchet:fix/scheduler-rate-limit-refill

---

- **Details**: Claude Code surfaced this while auditing the v1 scheduler and drafted the fix, tests, and this report under my direction; I reviewed it before submitting.

</details>

CoderSufiyan:
## Description

`shouldRefill()` in the v1 scheduler rate limiter had its condition inverted. It returned `true` when `nextRefillAt` was still in the future (window still open, no refill needed) — the opposite of when a refill is actually due.

## The bug

`pkg/scheduling/v1/rate_limit.go:182`:
```go
return r.nextRefillAt.After(time.Now().UTC())
```

This returns `true` while the refill time is still in the future. Compare with the guard inside `flushToDatabase()` (line 292), which no-ops while `nextRefillAt.After(now)` and only does real work once the boundary has passed. The two are meant to agree; `shouldRefill()` got the negation wrong.

## Impact

Both failure modes land in `use()` at line 118:

1. **After a window rolls over**: `shouldRefill()` reports `false`, so `use()` never triggers the on-demand refresh. Tasks are evaluated against depleted counts from the previous window and rejected until the 1s `loopFlush` ticker fires. Up to ~1s of spurious rate-limit rejections per window rollover, per tenant.

2. **Before the window rolls over** (the common case): `shouldRefill()` reports `true`, so every `use()` call runs `flushToDatabase()`, which takes three write locks (`unflushedMu`, `dbRateLimitsMu`, `nextRefillAtMu`) and immediately returns doing nothing. `use()` sits on the per-task scheduling path — constant write-lock churn for zero work.

## Fix

```go
return !r.nextRefillAt.After(time.Now().UTC())
```

This aligns `shouldRefill()` with the `flushToDatabase()` guard: returns `true` when `now >= nextRefillAt` (refill is due), `false` while the window is still open.

Fixes #4346

### DIFF
--- a/pkg/scheduling/v1/rate_limit.go
+++ b/pkg/scheduling/v1/rate_limit.go
@@ -179,7 +179,7 @@ func (r *rateLimiter) shouldRefill() bool {
 		return false
 	}
 
-	return r.nextRefillAt.After(time.Now().UTC())
+	return !r.nextRefillAt.After(time.Now().UTC())
 }
 
 func (r *rateLimiter) copyDbRateLimits() rateLimitSet {
--- a/pkg/scheduling/v1/rate_limit_test.go
+++ b/pkg/scheduling/v1/rate_limit_test.go
@@ -218,6 +218,72 @@ func TestRateLimiter_FlushToDatabase(t *testing.T) {
 	assert.Empty(t, rateLimiter.unflushed)
 }
 
+func TestRateLimiter_ShouldRefill(t *testing.T) {
+	l := zerolog.Nop()
+
+	t.Run("nil nextRefillAt returns false", func(t *testing.T) {
+		r := &rateLimiter{
+			dbRateLimits: make(rateLimitSet),
+			unacked:      make(map[int64]rateLimitSet),
+			unflushed:    make(rateLimitSet),
+			l:            &l,
+		}
+		assert.False(t, r.shouldRefill())
+	})
+
+	t.Run("future nextRefillAt returns false", func(t *testing.T) {
+		future := time.Now().Add(time.Hour)
+		r := &rateLimiter{
+			nextRefillAt: &future,
+			dbRateLimits: make(rateLimitSet),
+			unacked:      make(map[int64]rateLimitSet),
+			unflushed:    make(rateLimitSet),
+			l:            &l,
+		}
+		assert.False(t, r.shouldRefill())
+	})
+
+	t.Run("past nextRefillAt returns true", func(t *testing.T) {
+		past := time.Now().Add(-time.Hour)
+		r := &rateLimiter{
+			nextRefillAt: &past,
+			dbRateLimits: make(rateLimitSet),
+			unacked:      make(map[int64]rateLimitSet),
+			unflushed:    make(rateLimitSet),
+			l:            &l,
+		}
+		assert.True(t, r.shouldRefill())
+	})
+}
+
+func TestRateLimiter_UseRefreshesFromDbOnExpiredWindow(t *testing.T) {
+	l := zerolog.Nop()
+
+	mockRateLimitRepo := &mockRateLimitRepo{}
+	mockRows := []*sqlcv1.ListRateLimitsForTenantWithMutateRow{
+		{Key: "key1", Value: 10},
+	}
+	past := time.Now().Add(-time.Second)
+	mockRateLimitRepo.On("UpdateRateLimits", context.Background(), mock.Anything, mock.Anything).Return(mockRows, &past, nil)
+
+	rateLimiter := &rateLimiter{
+		dbRateLimits: rateLimitSet{
+			"key1": {key: "key1", val: 0, nextRefillAt: &past},
+		},
+		nextRefillAt:   &past,
+		unacked:        make(map[int64]rateLimitSet),
+		unflushed:      make(rateLimitSet),
+		l:              &l,
+		rateLimitRepo:  mockRateLimitRepo,
+	}
+
+	// use() detects the expired window via shouldRefill(), triggers flushToDatabase,
+	// and the refreshed limits (val: 10) should allow the request
+	res := rateLimiter.use(context.Background(), 1, map[string]int32{"key1": 5})
+	assert.True(t, res.succeeded)
+	mockRateLimitRepo.AssertExpectations(t)
+}
+
 func BenchmarkRateLimiter(b *testing.B) {
 	l := zerolog.Nop()
 


STiFLeR7 (commented):
Verified the fix against the sibling guard in `flushToDatabase` (line ~293: `if r.nextRefillAt.After(time.Now().UTC()) { return nil }`) — the two now agree: `use()` calls `flushToDatabase` exactly when `nextRefillAt` has actually passed, instead of the inverted original which called it while a refill was *not yet* due and stopped calling it once one *became* due. No race between the two separate `time.Now()` calls either, since time only moves forward between them and `nextRefillAt` is fixed — so the two checks can't disagree.

One gap: this ships with no test. Given `shouldRefill`/`nextRefillAt` is guarded by `nextRefillAtMu` and sits on a concurrency-sensitive path, a focused unit test asserting `shouldRefill()` is false while `nextRefillAt` is in the future and true once it's passed (matching `flushToDatabase`'s polarity) would make this much faster to merge with confidence and guard against the exact inversion class of bug regressing later. The original issue reporter (#4346) even offered to include one — might be worth pulling that in before merge.

CoderSufiyan:
Good call on the tests. I've added:

1. TestRateLimiter_ShouldRefill - three subtests covering nil, future, and past nextRefillAt
2. TestRateLimiter_UseRefreshesFromDbOnExpiredWindow - integration-style test verifying use() triggers flushToDatabase when the window has expired, refreshing the depleted rate limit from the DB

All 13 rate limit tests pass.

ApelKristian:
The inverted predicate is real—I hit it on a live self-hosted v0.88.0 engine before finding this PR—and this negation is a necessary part of the fix. Applied on its own, though, it doesn't restore correct behavior: I ran exactly this change against a clean v0.88.0 build, and most of the failure modes persisted.

Acceptance matrix (isolated Compose harness, timestamp-logging tasks; counts are completed starts / expected):

| Scenario | Predicate fix only | With two further fixes |
|---|---:|---:|
| Fresh `1/minute` across five boundaries | `1/6` | `6/6` |
| `5/minute → 1/minute` upsert | `5/10` | `10/10` |
| `1/minute → 5/minute` upsert | `1/12` | `12/12` |
| `minute → second` window change | `5/12` | `12/12` |
| Fresh `1/second`, long horizon | `151/180`, median `2.000025s` | `180/180`, median `0.999963s` |

Tracing the remaining failures led to two additional defects:

1. **Parked work is never rediscovered** — `pkg/scheduling/v1/queuer.go`, `(*Queuer).loopQueue`: `RequeueRateLimitedItems` runs only when queue-local `hasRateLimits` is true, and that flag is learned from `GetTaskRateLimits` on active queue items. Once all remaining work is durably parked, a newly leased or restarted queuer has no active item from which to relearn it, so parked rows are never re-checked after refill. This is why the minute-window scenarios stall even with the predicate corrected.
2. **Prior-window usage charged after refill** — `pkg/scheduling/v1/rate_limit.go`, `(*rateLimiter).loopFlush` / `flushToDatabase`: acknowledged usage accumulates in `unflushed`, but the one-second flush ticker can first persist it after a one-second window has already become refill-eligible. `BulkUpdateRateLimits` then computes the refill and subtracts that prior-window usage in the same update, emptying the fresh bucket—hence the stable ~2s cadence on a `1/second` key. Fixed by flushing more frequently and allowing pending usage to persist before the boundary.

With all three fixes in place, the matrix also passes an engine-only restart variant (same workers and registrations surviving the restart): `6/6`, `10/10`, `12/12`, `12/12`, `180/180`.

I have the complete patch (applies cleanly to current `main`), the disposable reproduction harness, the JSON result artifacts, and can add focused regression coverage. Happy to open a follow-up PR that builds on this one rather than competing with it, if maintainers want that.

<details id="ai-disclosure">

- [x] I acknowledge that LLMs were used in this contribution, in accordance with Hatchet's [AI_POLICY.md](https://github.com/hatchet-dev/hatchet/blob/main/AI_POLICY.md).
- **Tools and extent:** OpenAI GPT-5.6 (as the Pi coding agent) performed source inspection, root-cause tracing, test-harness development, and drafting; Anthropic Claude Fable 5 provided architectural direction and review. I encountered the behavior on my self-hosted deployment, directed the investigation, executed the isolated live-engine matrices, and reviewed and edited this text before posting.

</details>

PR #4393, opened 2026-07-09
feat: task slot cost

BloggerBust:
# Description

A customer asked to set how many worker slots a task consumes, so a task that needs more memory or CPU takes up more of a worker's capacity than a light one. The scheduler has supported multi-slot reservations since #2927, but no SDK exposed a way to request them. This adds a slot cost option to the TypeScript, Python, and Go task definitions. Durable and custom slot types are not exposed. The new docs page in this diff, `frontend/docs/pages/v1/advanced-assignment/slot-cost.mdx`, explains the option and its limits.

The part that needs the closest review is an engine fix. The engine was discarding the slot units a task registered and keeping a one-slot default in their place. Every SDK sent one slot, which matched the default, so no one noticed until the first larger value. The fix is in `pkg/repository/sqlcv1/workflows.sql`.

In the current behavior, a task whose cost exceeds every worker's slot count is cancelled, which will be confusing and hard to diagnose. I recommend a follow-up to improve diagnosability.

## Type of change

## What's Changed

- [x] A slot cost option in each SDK's task definition
- [x] An engine fix so registered slot units persist (`pkg/repository/sqlcv1/workflows.sql`)
- [x] A docs page under advanced assignment, with example snippets and cross-links
- [x] Tests that the scheduler reserves the requested slots, that registration stores them (red before the fix), and that the SDKs send them
- [x] TypeScript SDK 1.25.0 and Python SDK 1.34.0, with changelog entries

## Checklist

Changes have been:

- [x] Linted and formatted

## Testing

The repository test registers a workflow and checks the stored slot units. It stored one slot before the fix and stores the requested units after. I also tested with a local engine and a TypeScript worker, which confirmed the SDK sends the requested units, and is how the engine bug was found.

---

<details id="ai-disclosure">

- **Details**: Claude Code (Claude Fable 5) was used for source investigation & planning as well as tests and validation against a local engine

</details>


sdks/typescript/src/v1/task.ts
```diff
@@ -159,6 +159,17 @@ export type CreateBaseTaskOpts<
...
+  /**
+   * (optional) the number of default worker slots this task consumes.
+   *
+   * A worker has a fixed number of slots (default 100), and a normal task consumes one. Set slotCost
+   * higher for a task that needs more memory or CPU, so a worker runs fewer of them at once. A single
+   * worker must have that many free slots to run it. Durable tasks ignore this option.
+   *
+   * default: 1
+   */
+  slotCost?: number;
```
grutt, line 171: i think we inherit this base config in durable but we're not setting the cost on durable slots. we should either remove it from the durable impl or set it for durable as well. (not sure if this applies to the other sdks)

BloggerBust: I agree and I removed it.

sdks/python/hatchet_sdk/runnables/workflow.py
```diff
@@ -1351,6 +1351,7 @@ def task(
         wait_for: list[Condition | OrGroup] | None = None,
         skip_if: list[Condition | OrGroup] | None = None,
         cancel_if: list[Condition | OrGroup] | None = None,
+        slot_cost: int | None = None,
```
mrkaye97, line 1354: small nit: I wonder if we should make this `slot_cost: int = 1` instead if we end up defaulting to 1 internally anyway?

mrkaye97: I guess maybe this would require a bit of refactoring of how we create `slot_requests` on the `Task` internally 🤷

BloggerBust: Thanks for the feedback @mrkaye97. Looked at this and the behavior is identical either way. `Task.__init__` already sends `{"default": 1}` when the cost is unset, and the server strips a lone `{default: 1}` before version hashing, so `int = 1` has no effect. Also, defaulting to `None` preserves the difference between unset and an explicit `1`.

sdks/python/tests/test_slot_cost.py
```diff
@@ -0,0 +1,88 @@
...
+
+    header = segment({"alg": "none", "typ": "JWT"})
+    payload = segment(
+        {
+            "sub": "tenant-offline",
+            "server_url": "https://localhost",
+            "grpc_broadcast_address": "localhost:7070",
+        }
+    )
+    return f"{header}.{payload}.signature"
```
mrkaye97: I think we maybe don't need these - we can probably use the `hatchet` fixture that a bunch of other tests use, and we shouldn't need a no-op worker fixture I don't think. WDY?T

sdks/python/tests/test_slot_cost.py
```diff
@@ -0,0 +1,88 @@
...
+    standalone = hatchet.task(name="standalone-heavy", slot_cost=5)(dummy)
+
+    assert standalone._task.to_proto("svc").slot_requests == {"default": 5}
+
+
+def test_durable_task_is_unchanged(hatchet: Hatchet) -> None:
+    wf = hatchet.workflow(name="slot-cost-wf")
+    t = wf.durable_task()(dummy_durable)
+
+    assert t.to_proto("svc").slot_requests == {"durable": 1}
```
mrkaye97, line 44: it'd be good to write a genuine E2E test for this. we have a lot of examples throughout the various `test_*` files in the `examples` subdir, but the basic idea would be to write a task that consumes N slots, then use the `on_demand_worker` pytest fixture (there are some other examples of this too) to spin up a worker with that task registered that has something like `2*N - 1` slots, and trigger two of the task and ensure that the first one completes before the second one begins. Happy to help here if you need!

BloggerBust: Thanks, I added the e2e test. Since the `--slots` flag lives on `tests/worker.py`, I registered the task there and kept the tests in `tests/slot_cost`, same as the concurrency strategies test. Let me know if you would rather it live under examples and I will move it.

sdks/python/tests/slot_cost/test_slot_cost_e2e.py
```diff
@@ -0,0 +1,64 @@
...
+    test_run_id = str(uuid4())
+
+    refs = await slot_cost_workflow.aio_run_many(
+        [
+            slot_cost_workflow.create_bulk_run_item(
+                EmptyModel(),
+                additional_metadata={"test_run_id": test_run_id},
+            )
+            for _ in range(2)
+        ],
```
mrkaye97: small nit: I think we can remove this!


```suggestion
        [
            slot_cost_workflow.create_bulk_run_item(
                additional_metadata={"test_run_id": test_run_id},
            )
            for _ in range(2)
        ],
```

sdks/python/tests/slot_cost/test_slot_cost_e2e.py
```diff
@@ -0,0 +1,64 @@
...
+        wait_for_result=False,
+    )
+
+    await asyncio.gather(*[r.aio_result() for r in refs])
```
mrkaye97: nit: think we can simplify a bit:

```suggestion
    )
```

sdks/python/tests/slot_cost/test_slot_cost_e2e.py
```diff
@@ -0,0 +1,64 @@
...
+
+    assert workflows
+
+    workflow = next(w for w in workflows if w.name == slot_cost_workflow.name)
+
+    runs = await hatchet.runs.aio_list(
+        workflow_ids=[workflow.metadata.id],
+        additional_metadata={"test_run_id": test_run_id},
+        limit=100,
+    )
```
mrkaye97: small trick here! we have some sdk helpers for this (note the return type is slightly different):


```suggestion
    runs = await slot_cost_workflow.aio_list_runs(additional_metadata={"test_run_id": test_run_id})
```

sdks/python/tests/slot_cost/workflow.py
```diff
@@ -0,0 +1,19 @@
...
+# Long enough that two runs overlap if the worker can hold both at once, which is the failure the
+# e2e test rules out.
+SLEEP_TIME = 2
+
+slot_cost_workflow = hatchet.workflow(name="slot-cost-e2e")
+
+
+@slot_cost_workflow.task(slot_cost=SLOT_COST)
+async def heavy_task(input: EmptyModel, ctx: Context) -> None:
+    await asyncio.sleep(SLEEP_TIME)
```
mrkaye97: small nit: can simplify a bit (would need to change imports):

```suggestion
@hatchet.task(slot_cost=SLOT_COST)
async def slot_cost_test_heavy_task(input: EmptyModel, ctx: Context) -> None:
    await asyncio.sleep(SLEEP_TIME)
```

sdks/python/tests/test_slot_cost.py
mrkaye97: Thanks! I think putting it there is fine - at some point we should go through and consolidate / standardize what lives where, but not a big deal for now

mrkaye97 (approved):
left a couple more small comments to help simplify the python example / tests, but this looks great to me, nice work! will let Gabe comment on the non-python parts

BloggerBust:
@mrkaye97 I applied your suggestions to help simplify.

mrkaye97 (commented):
a bit late, but the simplification looks great - thanks!

PR #4395, opened 2026-07-10
feat: `hatchet embed-ui` command for embedded instances

mnafees:
# Description

Add new `hatchet embed-ui` subcommand to serve a full Hatchet frontend for embedded Hatchet instances.

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

</details>


cmd/hatchet-cli/cli/embedded_ui.go
abelanger5, line 1: How does the UI "know" which tenant / tenant memberships to render? It seems like this would only work in no-auth mode, otherwise I can't see how this would work without modifying frontend code.

cmd/hatchet-cli/cli/embedded_ui.go
```diff
@@ -0,0 +1,298 @@
...
+		}
+	}
+
+	spa, err := newSPAHandler()
+	if err != nil {
+		return nil, err
+	}
+
+	mux := http.NewServeMux()
+	mux.Handle("/api/", proxy)
```
abelanger5, line 279: I worry about this, users running `hatchet ui` will be opening a port with full access to their target tenant. There are lots of network  misconfigurations where this would be particularly bad, for example Docker Desktop for a while had an issue where ports would be discoverable if you were bound to a network. Same with VPNs like Tailscale, lots of (mis)configurations would allow for remote access to this port. 

I wonder if we should require some form of auth to the `/api/` proxy which we control from the CLI process. Again, this would require some frontend changes to pass some kind of token through (perhaps can hook into the exchange token mechanism).

.github/workflows/cli-release.yaml
```diff
@@ -48,6 +48,16 @@ jobs:
...
+      - name: Set up Node.js
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: "24"
+
+      - name: Set up pnpm
+        uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320 # v5.0.0
+        with:
+          version: 10.16.1
+
```
mnafees, line 60: We need these now since we are generating the UI bundle for `hatchet embed-ui`

cmd/hatchet-cli/cli/internal/ui/assets/.gitkeep
mnafees, line 1: This is used by git to make sure the `assets` empty directory is still kept in the version control history. This directory is used to generate the frontend assets before embedding them in the CLI.

PR #4401, opened 2026-07-10
perf: use cached worker action-hash for replenish

grutt:
# Description

`ListActionsForWorkers` expanded the full worker action join every scheduler replenish cycle (~1s). In production this could produce ~100k+ rows and hundreds of mliliseconds of query latency on large tenants while holding the replenish mutex. Since a worker's action set is immutable after registration (fingerprinted by `Worker.actionHash`), this replaces the per-cycle join with a cheap liveness+hash query (~0.4ms) and an in-process LRU cache of `(tenantId, actionHash) → action names`, only hitting the join tables for legacy worker registrations.

## Type of change

## What's Changed

- [x] Add `ListLiveWorkerActionHashes` (liveness + hash, no joins) as the per-cycle query, and `ListWorkerActionSets` to resolve action sets for one representative worker per uncached hash (materialized CTEs force a hash join over per-pair index lookups)
- [x] Cache resolved action sets in a 10k-entry, 30-min-TTL LRU on the assignment repository, keyed by tenant + action hash; empty results are never cached so a representative worker deleted mid-cycle self-heals next cycle
- [x] Rename the original query to `ListActionsForWorkersLegacyFallback`, used only for workers registered before `actionHash` existed; alias the row type in `types.go` so the `AssignmentRepository` interface and all callers are unchanged
- [x] Add span attributes for live worker count, hash cache misses, and no-hash workers

## Checklist

Changes have been:

- [x] Linted and formatted

## Testing

- `EXPLAIN ANALYZE` on the affected production shard: per-cycle cost drops from ~228ms / 334k buffer hits to ~0.4ms / 349 buffer hits (cache warm); worst-case cold-cache resolution is ~25ms
- `go test ./pkg/scheduling/v1/ ./pkg/repository

### DIFF
--- a/pkg/repository/scheduler_assignment.go
+++ b/pkg/repository/scheduler_assignment.go
@@ -135,8 +135,6 @@ func (d *assignmentRepository) ListActionsForWorkers(ctx context.Context, tenant
 		actions := hashToActions[key]
 
 		if len(actions) == 0 {
-			// preserve the LEFT JOIN behavior of the original query: live
-			// workers with no actions still produce a row
 			rows = append(rows, &sqlcv1.ListActionsForWorkersRow{
 				WorkerId: w.ID,
 				ActionId: pgtype.Text{},
--- a/pkg/repository/scheduler_assignment.go
+++ b/pkg/repository/scheduler_assignment.go
@@ -2,20 +2,36 @@ package repository
 
 import (
 	"context"
+	"time"
 
 	"github.com/google/uuid"
+	"github.com/hashicorp/golang-lru/v2/expirable"
+	"github.com/jackc/pgx/v5/pgtype"
 
 	"github.com/hatchet-dev/hatchet/pkg/repository/sqlcv1"
 	"github.com/hatchet-dev/hatchet/pkg/telemetry"
 )
 
+// A worker's action set is immutable after registration (the actionHash is a
+// sha256 of the registered action names), so cache entries never go stale.
+// The TTL only bounds memory for tenants whose workers have churned away.
+const workerActionCacheTTL = 30 * time.Minute
+
+type workerActionCacheKey struct {
+	actionHash string
+	tenantId   uuid.UUID
+}
+
 type assignmentRepository struct {
 	*sharedRepository
+
+	workerActionCache *expirable.LRU[workerActionCacheKey, []string]
 }
 
 func newAssignmentRepository(shared *sharedRepository) *assignmentRepository {
 	return &assignmentRepository{
-		sharedRepository: shared,
+		sharedRepository:  shared,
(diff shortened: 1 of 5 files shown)


pkg/repository/scheduler_assignment.go
```diff
@@ -2,31 +2,178 @@ package repository
...
+	}
+
+	if len(missedHashRepresentatives) > 0 {
+		representativeIds := make([]uuid.UUID, 0, len(missedHashRepresentatives))
+
+		for _, workerId := range missedHashRepresentatives {
+			representativeIds = append(representativeIds, workerId)
+		}
+
+		records, err := d.queries.ListWorkerActionSets(ctx, d.pool, sqlcv1.ListWorkerActionSetsParams{
```
abelanger5, line 87: do we want to be in a `tx` or do we want to be using `d.pool` here?

pkg/repository/scheduler_assignment.go
```diff
@@ -2,31 +2,178 @@ package repository
...
+
+	telemetry.WithAttributes(span,
+		telemetry.AttributeKV{Key: "list_actions.live_workers", Value: len(liveWorkers)},
+		telemetry.AttributeKV{Key: "list_actions.hash_cache_misses", Value: len(missedHashRepresentatives)},
+		telemetry.AttributeKV{Key: "list_actions.workers_without_hash", Value: len(workersWithoutHash)},
+	)
+
+	// fallback for workers registered before actionHash existed
+	if len(workersWithoutHash) > 0 {
+		fallbackRows, err := d.queries.ListActionsForWorkersLegacyFallback(ctx, d.pool, sqlcv1.ListActionsForWorkersLegacyFallbackParams{
```
abelanger5, line 162: same question about tx here

pkg/repository/scheduler_assignment.go
grutt: no, any cross-query inconsistency resolves as a missing-slots row and self-heals on the next replenish tick... i dont think having these in a tx gives us any real benefits

pkg/repository/scheduler_assignment.go
```diff
@@ -2,31 +2,178 @@ package repository
...
+		if len(w.ActionHash) == 0 {
+			continue
+		}
+
+		key := workerActionCacheKey{tenantId: tenantId, actionHash: string(w.ActionHash)}
+		actions := hashToActions[key]
+
+		if len(actions) == 0 {
+			// preserve the LEFT JOIN behavior of the original query: live
+			// workers with no actions still produce a row
```
abelanger5: nit: comments like this aren't helpful to future readers of the code

abelanger5 (approved):

PR #4403, opened 2026-07-10
fix: force gin for olap lists

grutt:
# Description

Adds GIN and updates queries to favor GINS if tenant entitlement flag is enabled.
Introduces a new AND filter which is more performant and preferred. Sets UI to use this by default.

## Type of change

## What's Changed

- Adds a tenant entitlement for strict additional metadata filters.
- Routes workflow-run and task-list metadata filters through jsonb containment when the entitlement is enabled.
- Adds OLAP GIN indexes and a migration for existing run/task partitions.
- Updates sqlc models and entitlement upsert/read paths for the new flag.

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Cursor was used to draft the PR description and fix a Go lint shadowing issue before opening the PR.

</details>


grutt:
image

image

api/v1/server/handlers/v1/workflow-runs/list.go
```diff
@@ -358,13 +362,35 @@ func (t *V1WorkflowRunsService) V1WorkflowRunList(ctx echo.Context, request gen.
 	spanContext, span := telemetry.NewSpan(ctx.Request().Context(), "v1-workflow-runs-list")
 	defer span.End()
 
+	strictAdditionalMetadata := false
```
mrkaye97: not sure I understand the naming here - what's the difference between strict metadata and using the "and" operator here?

cmd/hatchet-migrate/migrate/migrations/20260713150000_v1_0_126.sql
```diff
@@ -0,0 +1,15 @@
+-- +goose Up
+-- +goose StatementBegin
+
+ALTER TABLE tenant_entitlement
+ADD COLUMN IF NOT EXISTS strict_additional_metadata_filters BOOLEAN NOT NULL DEFAULT FALSE;
```
mrkaye97, line 5: let's rm `IF NOT EXISTS` (and `IF EXISTS` in the down)

cmd/hatchet-migrate/migrate/migrations/20260713150001_v1_0_127.sql
```diff
@@ -0,0 +1,36 @@
...
+-- PARTITION it.
+--
+-- jsonb_path_ops only supports @> (not ? / ?| / ?&), which is all the queries use,
+-- and is smaller and faster to search than the default jsonb_ops.
+
+CREATE INDEX IF NOT EXISTS ix_v1_tasks_olap_additional_metadata_gin
+    ON ONLY v1_tasks_olap USING gin (additional_metadata jsonb_path_ops);
+
+CREATE INDEX IF NOT EXISTS ix_v1_runs_olap_additional_metadata_gin
+    ON ONLY v1_runs_olap USING gin (additional_metadata jsonb_path_ops);
```
mrkaye97, line 26: rm `if not exists` for these too? I'd prefer if we deployed by running your script, and once it completed, we create the parent index

pkg/repository/olap.go
```diff
@@ -1217,11 +1277,24 @@ func (r *OLAPRepositoryImpl) ListWorkflowRuns(ctx context.Context, tenantId uuid
 		countParams.Until = sqlchelpers.TimestamptzFromTime(*until)
 	}
 
-	for key, value := range opts.AdditionalMetadata {
-		params.Keys = append(params.Keys, key)
-		params.Values = append(params.Values, value.(string))
-		countParams.Keys = append(countParams.Keys, key)
-		countParams.Values = append(countParams.Values, value.(string))
+	if opts.StrictAdditionalMetadata && len(opts.AdditionalMetadata) > 0 {
```
mrkaye97: should this check the operator and not this strict flag?

pkg/repository/sqlcv1/olap-overwrite.go
```diff
@@ -38,6 +56,116 @@ WITH filtered AS (
...
+    LIMIT 20000
+)
+
+SELECT COUNT(*)
+FROM filtered
+`
+
+const countTasksMetadataContainsAll = `-- name: CountTasksMetadataContainsAll :one
+WITH filtered AS (
+    SELECT
```
mrkaye97, line 86: the split queries here kind of suck - is there no other way that doesn't require all this duplication?

grutt: not that ensures we use the right indexes afict

cmd/hatchet-migrate/migrate/migrations/20260713150000_v1_0_126.sql
grutt: already ran via script... i think this is fine for now

cmd/hatchet-migrate/migrate/migrations/20260713150001_v1_0_127.sql
grutt: same

api/v1/server/handlers/v1/workflow-runs/list.go
grutt: updated name... but i want to drop this later this week anyway before we lose context on this whole thing

cmd/hatchet-migrate/migrate/migrations/20260713150000_v1_0_126.sql
mrkaye97: fwiw, this is about the local dev experience, not really about any risk in prod - having `IF NOT EXISTS` on things like this just masks bugs with no real value: either we don't need it (the schema is up to date with the migrations), or if we do need it, the schema isn't, which is a bug

mrkaye97 (approved):

PR #4412, opened 2026-07-14
fix: non-interactive manual trigger now searches by workflow name

### LINKED ISSUE #4429: [BUG] CLI "hatchet trigger manual -w" can only trigger a limited list of workflow
The below command though the workflow is clearly defined.
Image

After studying the source code, the list workflow function has a bug that this CLI can only trigger a limited list.
I have submitted a PR

dominicleung209:
# Description

`hatchet-trigger manual -w workflow1 -j '/tmp/hatchet12345.json'` is returning an error `FATA could not read JSON file workflow1`.
The reason is there is a list workflow inside it, and it is only able to trigger workflows that are listed, so limiting to the first page of workflows being listed. It will fail for example if my workflow numbers > 50.

Fixes # (issue)
fixes #4429

## Type of change

## What's Changed

- And workflowName to WorkflowListWithResponse

## Checklist

Changes have been:

- [X] Linted and formatted

<details id="ai-disclosure">

Used in evaluating the impact of the change.
</details>


cmd/hatchet-cli/cli/trigger.go
```diff
@@ -410,8 +410,10 @@ func runManualNonInteractive(profile *profileconfig.Profile, hatchetClient clien
 		cli.Logger.Fatalf("invalid tenant ID: %v", err)
 	}
 
-	// Fetch workflows
-	response, err := hatchetClient.API().WorkflowListWithResponse(ctx, tenantUUID, &rest.WorkflowListParams{})
+	// Fetch workflows — use Name search to find the workflow directly
+	response, err := hatchetClient.API().WorkflowListWithResponse(ctx, tenantUUID, &rest.WorkflowListParams{
+		Name: &workflowName,
```
gregfurman, line 415: If a workflow does not exist, the CLI will now incorrectly return the error in line 422

```go
	if response.JSON200 == nil || response.JSON200.Rows == nil || len(*response.JSON200.Rows) == 0 {
		cli.Logger.Fatal("no workflows available. Deploy a workflow first.")
	}
```

```sh
%  hatchet trigger -w process-message-dne -j bar.json --profile local
2026/07/15 16:35:42 FATA no workflows available. Deploy a workflow first.
exit status 1
```

whereas we'd expect it to return the following if the workflow dne:
```sh
%  go run cmd/hatchet-cli/main.go trigger -w process-message-dne -j bar.json --profile local
2026/07/15 16:35:51 FATA workflow 'process-message-dne' not found
exit status 1
```

So either this approach needs to account for the case of a valid request just not returning any rows OR perhaps the rows should be paginated over instead of just allowing this value to default to 50.

WDYT?

gregfurman (changes requested):
Thanks for the PR! Have recreated this locally and agree its a bug. However, I'm unsure that the below approach correctly solves it. Have left a comment, lmk your thoughts!

cmd/hatchet-cli/cli/trigger.go
dominicleung209: I see. I think the purpose is to give clearer error logs between "workflow not found" and "no workflow available".

I am thinking how about we put an extra cheap API call with Limit 1 and see if workflow is there or not. It will only happen on cases that response is null.

The pros is we can list fewer workflow, and skip the for loop later. I simplied the version logic below too. Any comments?

Tested that it will resolve on the below.
```sh
%  go run ./cmd/hatchet-cli trigger manual workflow -w judge-facts --json /tmp/empty-input.json --profile dom
ℹ Triggering workflow: deep-research-judge-facts

% go run ./cmd/hatchet-cli trigger manual workflow -w judge-facts-dne --
json /tmp/empty-input.json --profile dom
2026/07/16 09:31:38 FATA workflow 'judge-facts-dne' not found
```

PR #4413, opened 2026-07-14
feat(sdks/python): add cron_input to workflow and task declarations

### LINKED ISSUE #3820: [FEAT] Add support for `cron_input` in register_workflow
I would like to supply an input for cron-scheduled task. There is already a TODO for that but I want to open an issue so it get more attention: https://github.com/hatchet-dev/hatchet/blob/9768bf53ffdb0c674d13ea9c8a20db4c67a06375/sdks/python/hatchet_sdk/runnables/workflow.py#L256-L257

Adding a `cron_input` variable inside `BaseWorkflow`

**Additional context**
Thank you for hatchet :D I can send a PR if you are good with the solution above.

adem-loghmari:
## Description

Adds a `cron_input` option to the Python SDK's `workflow()`, `task()`, and `durable_task()` declarations. This lets you declare an input that is passed to runs triggered by a workflow's `on_crons` schedules.

The gRPC contract already carries this field — `CreateWorkflowVersionRequest.cron_input` exists in the generated stubs — but the high-level SDK gave no way to set it. `BaseWorkflow.to_proto` previously hardcoded `cron_input=None` with a `## TODO: Fix this` comment. This PR wires the value through the declaration APIs and serializes it onto the proto using the workflow's `input_validator`.

Closes #3820

## Example

```python
class CronInput(BaseModel):
    name: str
    count: int

workflow = hatchet.workflow(
    name="cron-input-workflow",
    input_validator=CronInput,
    on_crons=["* * * * *"],
    cron_input=CronInput(name="alice", count=3),
)
```

See `sdks/python/examples/cron/cron_input.py`.

## Changes

- Add `cron_input` param to all `workflow` / `task` / `durable_task` overloads on `Hatchet` (with docstrings).
- Add `cron_input` field to `WorkflowConfig`.
- Serialize `cron_input` onto `CreateWorkflowVersionRequest` in `BaseWorkflow.to_proto` (replaces the hardcoded `None`).
- Add an example (`examples/cron/cron_input.py`) and unit tests (`tests/test_cron_input.py`) asserting the value is serialized when provided and left unset otherwise.
- Register the example on the shared e2e-test-worker and add an integration test (`examples/cron/test_cron_input.py`) that runs the workflow end-to-end.


gregfurman:
Thanks for opening! Should solve https://github.com/hatchet-dev/hatchet/issues/3820

sdks/python/examples/cron/cron_input.py
gregfurman, line 1: nit: This should be snake-case i.e. `cron_input.py`

sdks/python/examples/cron/cron_input.py
```diff
@@ -0,0 +1,40 @@
...
+
+class CronInput(BaseModel):
+    name: str
+
+
+# > Cron Trigger with Input
+# You can pass a `cron_input` to a workflow so that runs triggered by its
+# `on_crons` schedules receive a predefined input.
+
+cron_input_workflow = hatchet.workflow(
```
gregfurman: Can we add a test here for the workflow? i.e

https://github.com/hatchet-dev/hatchet/blob/785bb67cd0bc188bb671b93dab8b718b07c9c50d/sdks/python/examples/simple/test_simple_workflow.py#L8-L12

gregfurman (changes requested):
Looks great! Think we should add a workflow test for this, just to be safe 🙂

adem-loghmari:
Thanks for the review! Addressed both:
- Renamed the example to `cron_input.py` (snake_case)
- Added `examples/cron/test_cron_input.py`, which registers `cron_input_workflow` on the shared e2e-test-worker and actually runs it end-to-end (sync + async), asserting on the task output and that `cron_input` is set on the proto — on top of the existing unit tests in `tests/test_cron_input.py`

gregfurman (commented):
Can you include an entry for this in `sdks/python/CHANGELOG.md`? Once done, think it looks good to ship

adem-loghmari:
Added a CHANGELOG entry (`## [1.33.19]`) and bumped the SDK version in `pyproject.toml` to match. Thanks!

frontend/docs/pages/reference/changelog/python.mdx
```diff
@@ -1,5 +1,11 @@
 {/* AUTOGENERATED — do not edit. Run `task sync-changelog` to regenerate from sdks/python/CHANGELOG.md */}
 
+## v1.33.19 - 2026-07-14
```
gregfurman: ~Thanks for adding the entry! FYI this should actually be `[Unreleased]`:~

~The corresponding changelog docs change can then be reverted since this PR won't actually trigger a release on pypi 🙂~

edit: nvm! I see you correctly updated the pyproject.toml

gregfurman (dismissed):

sdks/python/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to Hatchet's Python SDK will be documented in this changelog
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.33.19] - 2026-07-14
```
gregfurman: Should this not be a minor release? Adding support for `cron_input` seems like a behaviour change cc @mrkaye97

gregfurman: nit: also the date is wrong -- this'll require a regen of the docs as well 😅 

```suggestion
## [1.33.19] - 2026-07-15
```

gregfurman (commented):
LGTM! Just need to fix the date and get some clarification on whether this should instead be a minor update before approving.

Otherwise, thanks for the neat PR!

adem-loghmari:
Fixed the date (2026-07-15) and regenerated the docs changelog to match.

On the version bump: happy to make this a minor (`1.34.0`) if that's the call — just flagging that there's precedent for shipping additive `### Added` changes as a patch bump (e.g. `1.33.15` added `oldest_excluding_retries` to the task stats response as a patch), so I matched that convention here. No strong opinion either way — let me know what you and @mrkaye97 land on and I'll update.

sdks/python/examples/cron/cron_input.py
mrkaye97: small nit - let's use `@hatchet.task` for this? it'll improve the typing a bit to simplify the test, and also remove a couple LoC

sdks/python/examples/cron/cron_input.py
```diff
@@ -0,0 +1,40 @@
...
+def greet(input: CronInput, ctx: Context) -> dict[str, str]:
+    return {"message": f"Hello, {input.name}!"}
+
+
+# !!
+
+
+def main() -> None:
+    worker = hatchet.worker(
+        "cron-input-worker", slots=1, workflows=[cron_input_workflow]
```
mrkaye97: this type of thing has caused bugs in the past 😅 

```suggestion
        "cron-input-worker", workflows=[cron_input_workflow]
```

mrkaye97 (approved):
patch version is fine here IMO - I'm not following the "rules" super well, but I've been using minor versions for more significant new features, vs. this feels like a smaller fix 🤷 

in any case, thanks for the contribution!

adem-loghmari:
Rebased onto main to pick up #4393 (task slot cost), which bumped the SDK to `1.34.0` and conflicted with my version. Renumbered this as `1.34.1` (patch, per @mrkaye97's call above) — changelog and docs mirror updated accordingly.

mrkaye97:
Thanks for the contribution @adem-loghmari !

PR #4422, opened 2026-07-14
fix: avoid logging potentially sensitive request data

igor-kupczynski:
Avoid logging request query strings and malformed payload contents.

See in-line comments for details

### DIFF
--- a/api/v1/server/handlers/v1/webhooks/receive.go
+++ b/api/v1/server/handlers/v1/webhooks/receive.go
@@ -144,12 +144,8 @@ func (w *V1WebhooksService) V1WebhookReceive(ctx echo.Context, request gen.V1Web
 					/* url.ParseQuery automatically URL-decodes the payload parameter value */
 					err := json.Unmarshal([]byte(payloadValue), &payloadMap)
 					if err != nil {
-						payloadPreview := payloadValue
-						if len(payloadPreview) > 200 {
-							payloadPreview = payloadPreview[:200] + "..."
-						}
 						errorMsg := "Failed to unmarshal payload parameter as JSON"
-						w.config.Logger.Info().Err(err).Str("webhook", webhookName).Str("tenant", tenantId.String()).Int("payload_length", len(payloadValue)).Str("payload_preview", payloadPreview).Msg(errorMsg)
+						w.config.Logger.Info().Err(err).Str("webhook", webhookName).Str("tenant", tenantId.String()).Int("payload_length", len(payloadValue)).Msg(errorMsg)
 						return gen.V1WebhookReceive400JSONResponse{
 							Errors: []gen.APIError{
 								{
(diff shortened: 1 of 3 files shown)


api/v1/server/handlers/v1/webhooks/receive.go
```diff
@@ -186,12 +182,8 @@ func (w *V1WebhooksService) V1WebhookReceive(ctx echo.Context, request gen.V1Web
 		} else {
 			err := json.Unmarshal(rawBody, &payloadMap)
 			if err != nil {
-				bodyPreview := string(rawBody)
-				if len(bodyPreview) > 200 {
-					bodyPreview = bodyPreview[:200] + "..."
-				}
 				errorMsg := "Failed to unmarshal request body as JSON"
-				w.config.Logger.Info().Err(err).Str("webhook", webhookName).Str("tenant", tenantId.String()).Str("content_type", contentType).Int("body_length", len(rawBody)).Str("body_preview", bodyPreview).Msg(errorMsg)
+				w.config.Logger.Info().Err(err).Str("webhook", webhookName).Str("tenant", tenantId.String()).Str("content_type", contentType).Int("body_length", len(rawBody)).Msg(errorMsg)
```
igor-kupczynski, line 186: Currently, when a webhook form-encoded payload contains invalid JSON, we log its first 200 characters.

I remove it here, in case there are any sensitive data in the malformed JSON. What is still logged is:
* Parse error
* Webhook and tenant
* Payload length

api/v1/server/run/run.go
```diff
@@ -742,7 +741,7 @@ func (t *APIServer) registerSpec(g *echo.Group, spec *openapi3.T) (*populator.Po
 				Dur("latency", v.Latency).
 				Int("status", statusCode).
 				Str("method", v.Method).
-				Str("uri", v.URI).
+				Str("uri", v.URIPath).
```
igor-kupczynski, line 744: It changes `URI` to `URIPath` which in practice drops the query parameters from the request log:
```
Before: /.../foo?code=abc123&state=xyz
After: /.../foo
```

The idea being that e.g. oauth callbacks may contain semi-sensitive values

pkg/repository/match.go
```diff
@@ -1025,7 +1025,7 @@ func (m *sharedRepository) processCELExpressions(ctx context.Context, events []C
 				err := json.Unmarshal(event.Data, &inputData)
 
 				if err != nil {
-					m.l.Warn().Ctx(ctx).Err(err).Msgf("failed to unmarshal user event data %s", string(event.Data))
+					m.l.Warn().Ctx(ctx).Err(err).Msgf("failed to unmarshal user event data. id: %s, key: %s", event.ID, event.Key)
```
igor-kupczynski, line 1028: Similar idea: don't log malformed payload bodies

grutt (approved):

PR #4426, opened 2026-07-14
fix: update trigger for runs status and update schema for consistency

abelanger5:
# Description

Updates the trigger on `v1_runs_status` so that it updates the corresponding `v1_statuses_olap` table using an `INSERT INTO...ON CONFLICT` instead of an `UPDATE`. 

Note the following p99 reduction in latency (milliseconds) on `UpdateTaskStatusesFromMQ` after this fix was applied:

image

This particular database runs TimescaleDB and `v1_statuses_olap` is a hypertable. When running query plans simulating this trigger, we saw that every chunk of the hypertable was being index scanned when calling the `UPDATE`. This seems internal to Timescale's partitioning implementation; we don't see the same thing on updates of our partitioned tables.  

This was ultimately the cause of extremely heavy `LWLock:BufferContent` (called `lwlock:buffer_content` in some systems) due to the index scan touching every partition across 30 days of retention. 

## Type of change

## What's Changed

- [X] Updates the `v1_runs_olap_status_update_function` to use `INSERT INTO...ON CONFLICT` instead of `UPDATE` 

## Checklist

Changes have been:

- [X] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Claude Fable for implementation

</details>

### DIFF
--- a/sql/schema/v1-olap.sql
+++ b/sql/schema/v1-olap.sql
@@ -865,7 +865,11 @@ BEGIN
         kind,
         readable_status
     )
-    SELECT
+    -- DISTINCT ON: nothing constraint-enforces (external_id, inserted_at)
+    -- uniqueness across new_rows, and ON CONFLICT DO UPDATE errors if a single
+    -- statement affects the same row twice. On duplicates, keep the
+    -- highest-priority status.
+    SELECT DISTINCT ON (external_id, inserted_at)
         external_id,
         inserted_at,
         tenant_id,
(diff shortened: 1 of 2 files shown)


sql/schema/v1-olap.sql
```diff
@@ -268,16 +263,12 @@ CREATE TABLE v1_runs_olap (
     additional_metadata JSONB,
     parent_task_external_id UUID,
 
-    PRIMARY KEY (inserted_at, id, readable_status, kind)
+    PRIMARY KEY (inserted_at, id)
```
abelanger5, line 266: drive-by: cleanup

grutt (approved):

PR #4432, opened 2026-07-15
fix: memory leak in Python SDK worker logging

Amiya-w:
# Description

### Root Cause

During worker startup, a `LogForwardingHandler` is added to the logging system.  
Initially, this handler extended `logging.StreamHandler`, which requires a stream object. However, it was initialized with `StringIO`, but this `StringIO` do nothing. StringIO was only present as a placeholder to instantiate StreamHandler and was never actually used.

The `StringIO` buffer is only closed when the worker shuts down. Under normal conditions, the worker runs indefinitely, while `StringIO.emit()` continuously appends data. This results in unbounded memory growth, effectively causing a memory leak.

### Fix

This PR replaces `logging.StreamHandler` with a direct subclass of `logging.Handler`.  
Since the handler only uses `format()` and never performs actual stream output, the `StringIO` dependency is entirely unnecessary and has been removed.

## Type of change

## What's Changed

## Checklist

Changes have been:

- [ ] Linted and formatted

---

<details id="ai-disclosure">

</details>

### DIFF
--- a/sdks/python/hatchet_sdk/worker/runner/utils/capture_logs.py
+++ b/sdks/python/hatchet_sdk/worker/runner/utils/capture_logs.py
@@ -4,7 +4,6 @@ import queue
 import threading
 from collections.abc import Awaitable, Callable
 from dataclasses import dataclass
-from io import StringIO
 from typing import Any, Literal, ParamSpec, TypeVar
 
 from pydantic import BaseModel, Field
@@ -161,15 +160,13 @@ class AsyncLogSender:
         self._thread = None
 
 
-class LogForwardingHandler(logging.StreamHandler):  # type: ignore[type-arg]
-    def __init__(self, log_sender: AsyncLogSender, stream: StringIO):
-        super().__init__(stream)
+class LogForwardingHandler(logging.Handler): # type: ignore[type-arg]
+    def __init__(self, log_sender: AsyncLogSender):
+        super().__init__()
 
         self.log_sender = log_sender
 
     def emit(self, record: logging.LogRecord) -> None:
-        super().emit(record)
-
         log_entry = self.format(record)
         step_run_id = ctx_step_run_id.get()
         task_retry_count = ctx_task_retry_count.get()
@@ -192,8 +189,7 @@ def capture_logs(
 ) -> Callable[P, Awaitable[T]]:
     @functools.wraps(func)
     async def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
-        log_stream = StringIO()
-        log_forwarder = LogForwardingHandler(log_sender, log_stream)
+        log_forwarder = LogForwardingHandler(log_sender)
         log_forwarder.setLevel(logger.level)
 
         if logger.handlers:
@@ -214,7 +210,6 @@ def capture_logs(
         finally:
             log_forwarder.flush()
             logger.removeHandler(log_forwarder)
-            log_stream.close()
 
         return result
 
--- a/sdks/python/tests/unit/test_capture_logs.py
+++ b/sdks/python/tests/unit/test_capture_logs.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import asyncio
 import logging
-from io import StringIO
 from types import SimpleNamespace
 from typing import cast
 
@@ -29,7 +28,7 @@ async def test_log_forwarding_handler_enqueues_correct_record() -> None:
     previous_level = target_logger.level
     target_logger.setLevel(logging.INFO)
 
-    handler = LogForwardingHandler(log_sender, StringIO())
+    handler = LogForwardingHandler(log_sender)
     target_logger.addHandler(handler)
 
     step_token = ctx_step_run_id.set("step-run-id")


sdks/python/hatchet_sdk/worker/runner/utils/capture_logs.py
```diff
@@ -161,15 +160,13 @@ def stop(self, timeout: float = 5.0) -> None:
         self._thread = None
 
 
-class LogForwardingHandler(logging.StreamHandler):  # type: ignore[type-arg]
-    def __init__(self, log_sender: AsyncLogSender, stream: StringIO):
-        super().__init__(stream)
+class LogForwardingHandler(logging.Handler): # type: ignore[type-arg]
```
gregfurman, line 163: Can you provide any steps to reproduce this issue? Since each `emit` call of the `StreamHandler` does a `flush` under-the-hood, I'm unsure whether there is actually unbounded growth in log stream -- else perhaps you can clarify this?

https://github.com/hatchet-dev/hatchet/blob/aad903d3888e7ac61a30c8897d3f61528b7522ee/sdks/python/hatchet_sdk/worker/runner/utils/capture_logs.py#L170-L171

sdks/python/tests/unit/test_capture_logs.py
```diff
@@ -29,7 +28,7 @@ async def test_log_forwarding_handler_enqueues_correct_record() -> None:
     previous_level = target_logger.level
     target_logger.setLevel(logging.INFO)
 
-    handler = LogForwardingHandler(log_sender, StringIO())
+    handler = LogForwardingHandler(log_sender)
```
gregfurman, line 31: suggestion: Perhaps a test case illustrating how we no longer have a memory leak could suffice as a reproducer?

gregfurman (commented):
Have left some comments. Otherwise, the CI is failing as these changes need to be linted. 

Confirming that you've read our [CONTRIBUTING.md](https://github.com/hatchet-dev/hatchet/blob/aad903d3888e7ac61a30c8897d3f61528b7522ee/CONTRIBUTING.md)?

sdks/python/hatchet_sdk/worker/runner/utils/capture_logs.py
Amiya-w: Below is a minimal reproduction demonstrating the issue. In practice, `StringIO.flush()` is a no-op and does **not** clear the underlying buffer.

I constructed a test with 100,000 simple log messages under three configurations:

1. Using `StringIO` as the stream for `StreamHandler`
2. Using `/dev/null` as the stream for `StreamHandler`
3. Using a plain `logging.Handler` **without any stream**

The results confirm the root cause:

- `StringIO` accumulates data indefinitely because `StringIO.flush()` is a no-op.
- `/dev/null` correctly discards output, but still inherits unnecessary `StreamHandler` overhead.

The results clearly show that when `StringIO` is used, memory usage grows continuously even after repeated calls to `flush()`. In contrast, `/dev/null` correctly discards all output.
Notably, the plain `logging.Handler` implementation shows **zero additional memory usage** and **no memory leak** during the same test.

While `/dev/null` achieves similar memory stability, its cross-platform behavior (e.g. `nul` on Windows) introduces unnecessary uncertainty. More importantly, this code never actually performs stream I/O: it only calls `format()`.

Removing the stream entirely and using a bare `logging.Handler` is the cleanest fix: it is simpler, safer across platforms, and semantically correct.

I also isolated the behavior of `StringIO.write()` and `StringIO.flush()`:
- `write()` appends data to the buffer as expected
- `flush()` performs **no operation whatsoever**

Because `StreamHandler.emit()` calls `flush()` internally, and `flush()` does nothing for `StringIO`, the buffer grows unboundedly throughout the worker's lifetime. This is the root cause of the memory leak.
```
import gc
import logging
import os
import tracemalloc
from io import StringIO

N_LOGS = 100_000
MSG = "processing item with some contextual payload for the log line"


def buffer_leaks():
    # replace dev/null
    # log_stream = open(os.devnull, "w")
    log_stream = StringIO()
    logger = logging.getLogger("leak")
    logger.setLevel(logging.INFO)
    logger.handlers.clear()
    logger.addHandler(logging.StreamHandler(log_stream))

    for _ in range(N_LOGS):
        logger.info(MSG)

    # gc.collect() cannot reclaim it: the text is still strongly referenced
    # via logger -> handler -> stream, so it is live, not garbage.
    gc.collect()


class NoBufferHandler(logging.Handler):

    def emit(self, record: logging.LogRecord) -> None:
        self.format(record)


def buffer_fixed():
    logger = logging.getLogger("fixed")
    logger.setLevel(logging.INFO)
    logger.handlers.clear()
    logger.addHandler(NoBufferHandler())

    for _ in range(N_LOGS):
        logger.info(MSG)

    gc.collect()


def flush_vs_close() -> None:
    # Prove with tracemalloc that flush() does NOT free the buffer; only close() does.
    tracemalloc.start()

    stream = StringIO()
    stream.write(MSG * N_LOGS)  # ~6 MB of text written to the in-memory stream
    mem_written = tracemalloc.get_traced_memory()[0]

    # flush() is a no-op for StringIO: the destination IS memory, nothing to push out.
    stream.flush()
    print(f"after flush stream:{len(stream.getvalue())}")
    gc.collect()
    mem_after_flush = tracemalloc.get_traced_memory()[0]

    # close() releases the underlying buffer; the memory is actually reclaimed.
    stream.close()
    gc.collect()
    mem_after_close = tracemalloc.get_traced_memory()[0]

    tracemalloc.stop()

    print(f"StringIO used")
    print(f"mem_written:      {mem_written / 1024 / 1024:>6.1f} MB")
    print(f"mem_after_flush:  {mem_after_flush / 1024 / 1024:>6.1f} MB")
    print(f"mem_after_close:  {mem_after_close / 1024 / 1024:>6.1f} MB")

def measure(fn) -> tuple[int, int]:
    tracemalloc.start()
    fn()
    _, peak = tracemalloc.get_traced_memory()
    tracemalloc.stop()
    return peak


if __name__ == "__main__":
    leak_peak = measure(buffer_leaks)
    fixed_peak = measure(buffer_fixed)

    # Old impl grows with log volume; fixed impl stays flat regardless of volume.
    print(f"StreamHandler(StringIO): {leak_peak / 1024 / 1024:.1f} MB")
    print(f"logging.Handler(no stream): {fixed_peak / 1024 / 1024:.1f} MB")
    print()
    flush_vs_close()


```
image

mrkaye97: Hi @Amiya-w - this isn't a repro in the sense we mean. We're looking for a repro that boots a Hatchet worker, logs, tracks the memory usage, and shows it going up over time.

mrkaye97:
Hi @Amiya-w - I'm closing this for a number of reasons:

1. The repro provided is self-contained, doesn't necessarily indicate a memory leak, and also shows this problem independent of the SDK. We'd love to see a repro that shows the SDK actually leaking memory as you log more and more, but I have a suspicion this is not a problem, since we very likely would've heard about it by now if it were.
2. It doesn't follow our contribution guidelines, which explicitly mention the need to flag any AI use in opening PRs, issues, and so on.
3. It also doesn't follow the contribution guidelines mentioning that we'd like there to be an open, accepted issue before a fix is submitted, so we can triage and agree on whether or not the issue is worth fixing.

Feel free to reopen if you address each of these, and thanks again for the submission!

PR #4442, opened 2026-07-16
fix: cancel in progress has incorrect ordering with new in-memory index

abelanger5:
# Description

When `SERVER_CONCURRENCY_IN_MEMORY_INDEX_ENABLED=true` is set (enabling the new experimental in-memory index), the cancel-in-progress concurrency strategy incorrectly acts like cancel-newest. This is due to the comparator method for the internal heap using the default `priorityCompare` instead of ranking newer slots as greater values in the heap, so they take precedence when popping out of the minHeap. 

Also adds a set of Python tests which use task-level concurrency for cancel-in-progress and cancel-newest, which correctly failed when setting `SERVER_CONCURRENCY_IN_MEMORY_INDEX_ENABLED`. 

## Type of change

## What's Changed

- [X] New unit, integration and e2e tests to ensure we catch this path
- [X] Adding a new comparator for `CancelInProgress`

## Checklist

Changes have been:

- [X] Linted and formatted

---

<details id="ai-disclosure">

</details>

### DIFF
--- a/sdks/python/examples/concurrency_cancel_in_progress_task_level/test_concurrency_cancel_in_progress_task_level.py
+++ b/sdks/python/examples/concurrency_cancel_in_progress_task_level/test_concurrency_cancel_in_progress_task_level.py
@@ -39,7 +39,9 @@ async def test_run(hatchet: Hatchet) -> None:
 
     runs = sorted(
         (
-            await hatchet.runs.aio_list(additional_metadata={"test_run_id": test_run_id})
+            await hatchet.runs.aio_list(
+                additional_metadata={"test_run_id": test_run_id}
+            )
         ).rows,
         key=lambda r: int((r.additional_metadata or {}).get("i", "0")),
     )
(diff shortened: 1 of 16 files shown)


sdks/python/examples/concurrency_cancel_newest_task_level/test_concurrency_cancel_newest_task_level.py
```diff
@@ -0,0 +1,64 @@
...
+    for ref in to_cancel:
+        try:
+            await ref.aio_result()
+        except Exception:
+            pass
+
+    ## wait for the olap repo to catch up
+    await asyncio.sleep(5)
+
+    successful_run = hatchet.runs.get(to_run.workflow_run_id)
```
mrkaye97: it'll be less flaky to use `await hatchet.runs.aio_get_details()` instead - have been trying to do that everywhere (then don't need the `sleep` above this either)

sdks/python/examples/concurrency_cancel_in_progress_task_level/test_concurrency_cancel_in_progress_task_level.py
```diff
@@ -0,0 +1,49 @@
...
+        try:
+            await ref.aio_result()
+        except Exception:
+            continue
+
+    ## wait for the olap repo to catch up
+    await asyncio.sleep(5)
+
+    runs = sorted(
+        hatchet.runs.list(additional_metadata={"test_run_id": test_run_id}).rows,
```
mrkaye97:
```suggestion
        await hatchet.runs.aio_list(additional_metadata={"test_run_id": test_run_id}).rows,
```

mrkaye97 (approved):

PR #4444, opened 2026-07-16
fix(sdks/python): handle empty CREATE payload in durable waits

### LINKED ISSUE #4349: [BUG] Python SDK: durable aio_wait_for_event / aio_sleep_for crash on empty payloads (RuntimeError from PEP 479)
**Describe the issue**

In the Python SDK, the durable helpers `aio_wait_for_event` and `aio_sleep_for` crash with `RuntimeError: coroutine raised StopIteration` when the underlying wait completes with an empty payload.

Both call sites assume the match dict is non-empty:

```python
# hatchet_sdk/context/context.py:1001 (aio_wait_for_event), same shape at :915 (aio_sleep_for)
matches: dict[str, list[dict[str, Any]]] = result.get("CREATE", {})
_, raw_matches = next(iter(matches.items()))   # StopIteration on {}
raw_payload = raw_matches[0]                   # IndexError on []
```

An empty payload is a legitimate engine response: `aio_wait_for` returns `result.payload or {}` (context.py:885), and `DurableTaskEventLogEntryResult.from_proto` maps an empty proto payload to `None` (`clients/listeners/durable_event_listener.py:123-125`) — e.g. when a durable task resumes after eviction. The bare `next()` on an empty dict raises `StopIteration` inside a coroutine, which PEP 479 surfaces as `RuntimeError: coroutine raised StopIteration`, failing the durable task.

The TypeScript SDK guards exactly this case, and has since the Durable Execution Revamp (#2954) — the same PR that introduced the unguarded Python code:

- `sleepFor` (`context.ts:952`): empty match → returns the requested duration
- `waitForEvent` (`context.ts:1080`): empty match → returns `{}` / `payloadSchema.parse({})`

So identical engine responses that TS handles gracefully crash Python durable tasks.

**Environment**

- `main` @ 0b51d955; the code is unchanged since #2954 (2026-03-16), so released 1.x SDKs carry it.

**Expected behavior**

Match the TypeScript semantics: `aio_sleep_for` falls back to the requested duration, `aio_wait_for_event` returns an empty payload (validated through `payload_validator` when one is given) instead of raising.

**Code to Reproduce**

Unit test (pytest, no engine needed) — stub `aio_wait_for` to return `{}` on a `DurableContext` and call either helper:

```python
ctx = object.__new__(DurableContext)
ctx._wait_index = 0
ctx.aio_wait_for = AsyncMock(return_value={})

await ctx.aio_wait_for_event("my-event")   # RuntimeError: coroutine raised StopIteration
await ctx.aio_sleep_for(timedelta(seconds=5))  # same
```

The empty-match-list variant (`{"CREATE": {"event:my-event-0": []}}`) fails the same way via `IndexError`.

**Additional context**

Fix mirroring the TS behavior, with regression tests for both helpers (empty payload, empty match list, `payload_validator` on empty, and the non-empty happy path), is ready here: https://github.com/hatchet-dev/hatchet/compare/main...abhay-codes07:hatchet:fix/python-durable-empty-payload — happy to open a PR if accepted. Note I also have #4346 open; per the one-open-PR rule for new contributors I'd sequence the PRs.

---

- **Details**: Claude Code surfaced this while auditing the SDKs, verified the failure with pytest, and drafted the fix, tests, and this report under my direction; I reviewed it before submitting.

</details>

STiFLeR7:
## Summary

Fixes #4349 — `aio_sleep_for` and `aio_wait_for_event` on `DurableContext` crash with `RuntimeError: coroutine raised StopIteration` (PEP 479) when a durable wait completes with an empty `CREATE` payload, instead of falling back gracefully.

## Root cause

Both methods extracted the first match via:

```python
matches: dict[str, list[dict[str, Any]]] = res.get("CREATE", {})
_, raw_matches = next(iter(matches.items()))
sleep = raw_matches[0]
```

`next(iter(matches.items()))` raises `StopIteration` when `matches` is empty (e.g. `CREATE` missing, `{}`, or a key mapping to an empty list) — and since this executes inside a coroutine, Python 3.7+'s PEP 479 turns that `StopIteration` into a `RuntimeError` instead of letting it propagate as a normal exception. `aio_wait_for` already treats an empty proto payload as valid and falls back to `{}`, so a durable wait genuinely can complete with no payload — this wasn't an edge case, it was a reachable path that just crashed instead of degrading gracefully.

## Fix

Added a shared helper, `_first_wait_match_or_none`, used by both `aio_sleep_for` and `aio_wait_for_event`, that returns `None` for any empty-payload shape (missing `CREATE`, empty dict, or empty match list) instead of raising. Both call sites now do `_first_wait_match_or_none(res) or {}`.

## Testing

- Added `tests/unit/test_durable_wait_empty_payload.py`: parametrized tests for `_first_wait_match_or_none` covering empty/missing payload shapes and normal match extraction, plus async tests constructing `DurableContext` with a mocked `aio_wait_for` to verify `aio_sleep_for`/`aio_wait_for_event` no longer crash and still correctly extract a payload when one is present.
- `poetry run python -m pytest tests/unit/test_durable_wait_empty_payload.py` — 7 passed.
- `poetry run ruff check` / `poetry run ruff format --check` — clean.
- `poetry run mypy` — no issues.
- Rebased onto current `upstream/main`; confirmed no upstream commits touched `context.py` since this branch was created, so no merge risk.

Pre-existing, unrelated: 6 tests in `test_shutdown_ordering.py` fail locally on Windows (`loop.add_signal_handler` raises `NotImplementedError` — `SIGINT` handler registration isn't supported by Windows' asyncio event loop). Verified via `git stash` that this reproduces identically on unmodified `upstream/main`, so it's a pre-existing platform limitation, not a regression from this change.

## Disclosure

Prepared with AI assistance (Claude Code). I read the surrounding `Context`/`DurableContext` code and the PEP 479 mechanics before making the change, wrote the fix and its regression tests, and verified ruff/mypy/pytest all pass locally, including confirming the pre-existing Windows-only test failures are unrelated and reproduce on a clean `upstream/main` checkout.

### DIFF
--- a/sdks/python/hatchet_sdk/context/context.py
+++ b/sdks/python/hatchet_sdk/context/context.py
@@ -98,6 +98,27 @@ def _compute_memo_key(task_run_external_id: str, *args: Any, **kwargs: Any) -> b
     return h.digest()
 
 
+def _first_wait_match_or_none(result: dict[str, Any]) -> dict[str, Any] | None:
+    """
+    Extract the first match from a durable wait result's `CREATE` dict, if present.
+
+    The engine returns an object shaped like `{"CREATE": {"signal_key_1": [{"id": ...}]}}`.
+    A durable wait can legitimately complete with an empty payload (e.g. `aio_wait_for`
+    already falls back to `{}` when the proto payload is empty), in which case `CREATE`
+    may be missing, `{}`, or map to an empty match list. Treat all of those as "no match"
+    instead of letting `next(iter(...))`/list-indexing raise on an empty collection.
+    """
+    matches: dict[str, list[dict[str, Any]]] = result.get("CREATE", {})
+    if not matches:
+        return None
+
+    _, raw_matches = next(iter(matches.items()))
+    if not raw_matches:
+        return None
+
+    return raw_matches[0]
+
+
 class Context:
     def __init__(
         self,
@@ -911,10 +932,10 @@ class DurableContext(Context):
         ## the engine returns an object like this:
         ## {"CREATE": {"signal_key_1": [{"id": ...}]}}
         ## since we have a single match we're looking for, we know that
-        ## the list of matches will only have one item, so we can extract and parse it
-        matches: dict[str, list[dict[str, Any]]] = res.get("CREATE", {})
-        _, raw_matches = next(iter(matches.items()))
-        sleep = raw_matches[0]
+        ## the list of matches will only have one item, so we can extract and parse it.
+        ## An empty payload is a legitimate response (see `_first_wait_match_or_none`),
+        ## so fall back to an empty dict rather than raising.
+        sleep = _first_wait_match_or_none(res) or {}
 
         return SleepResult(
             duration=expr_to_timedelta(
@@ -997,10 +1018,10 @@ class DurableContext(Context):
         ## the engine returns an object like this:
         ## {"CREATE": {"signal_key_1": [{"id": ...}]}}
         ## since we have a single match we're looking for, we know that
-        ## the list of matches will only have one item, so we can extract and parse it
-        matches: dict[str, list[dict[str, Any]]] = result.get("CREATE", {})
-        _, raw_matches = next(iter(matches.items()))
-        raw_payload = raw_matches[0]
+        ## the list of matches will only have one item, so we can extract and parse it.
+        ## An empty payload is a legitimate response (see `_first_wait_match_or_none`),
+        ## so fall back to an empty dict rather than raising.
+        raw_payload = _first_wait_match_or_none(result) or {}
 
         if payload_validator is not None:
             adapter = TypeAdapter(payload_validator)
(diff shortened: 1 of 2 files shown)


sdks/python/hatchet_sdk/context/context.py
```diff
@@ -911,10 +932,10 @@ async def aio_sleep_for(
...
         ## {"CREATE": {"signal_key_1": [{"id": ...}]}}
         ## since we have a single match we're looking for, we know that
-        ## the list of matches will only have one item, so we can extract and parse it
-        matches: dict[str, list[dict[str, Any]]] = res.get("CREATE", {})
-        _, raw_matches = next(iter(matches.items()))
-        sleep = raw_matches[0]
+        ## the list of matches will only have one item, so we can extract and parse it.
+        ## An empty payload is a legitimate response (see `_first_wait_match_or_none`),
+        ## so fall back to an empty dict rather than raising.
+        sleep = _first_wait_match_or_none(res) or {}
```
mrkaye97: needing to fall back here with `or {}` feels a bit odd - can we have `_first_wait_match_or_none` return only `dict[str, Any]` instead, and return `{}` in the `None` cases?

STiFLeR7: Good catch, done in d30672f6c. Renamed it to `_first_wait_match` too since it no longer has an `_or_none` case to speak of.

PR #4451, opened 2026-07-16
perf: hot schedulers

grutt:
# Description

Reworks the v1 scheduler so slot capacity lives in per-worker pools instead of being copied into each action, cutting assignment/replenish overhead on large worker sets. Also fixes unused-slot accounting so expiry cannot make selection skip valid capacity.

## Type of change

## What's Changed

- Own slots in `slotPool` keyed by worker + slot type; actions keep a thin worker-id index
- Assign from pools with `active()` selection (do not hard-gate on `unusedCount`, which can drift on expiry)
- Align pool refresh time with slot expiry on replenish; count only active slots into `unused`
- Add shape / worker-scale scheduler benchmarks and cover unused-counter drift in unit tests

## Checklist

Changes have been:

- [x] Linted and formatted

## Testing

- `go test ./pkg/scheduling/v1/ -count=1`

---

<details id="ai-disclosure">

- **Details**: Used cursor throughout investigation and writing the pr

</details>


pkg/scheduling/v1/scheduler_shape_bench_test.go
```diff
@@ -0,0 +1,445 @@
...
+
+func (f *inventoryFixture) measureInventory() {
+	refs := 0
+	for _, pool := range f.scheduler.pools {
+		refs += len(pool.slots)
+	}
+	f.uniqueSlots = refs
+	f.actionSlotRefs = refs
+	if f.uniqueSlots > 0 {
+		f.fanout = float64(f.actionSlotRefs) / float64(f.uniqueSlots)
```
mrkaye97: is this not trivially going to be `1.0` since we're dividing `refs` by itself?

pkg/scheduling/v1/scheduler_shape_bench_test.go
mrkaye97, line 1: do these tests fail on main?

grutt: these are net new

grutt: i wrote them before making any changes but they require changes on this branch which is bad

pkg/scheduling/v1/scheduler_shape_bench_test.go
grutt: lol yes, artifact of the test written on main's code which would not have been 1.

pkg/scheduling/v1/scheduler_shape_bench_test.go
mrkaye97: I'm not sure I follow? I just ran these tests on main and they all passed, so I'm not convinced they're either 1) not flaky or 2) actually testing anything - I imagine these are supposed to be a regression test?

grutt: They're benchmark tests and just report timing

PR #4453, opened 2026-07-17
fix: non tty profile selection

BloggerBust:
# Description

CLI commands that resolve a profile opened an interactive select form whenever no default profile was set, even with only one profile configured. In a session without a TTY the form cannot open `/dev/tty`, and the CLI dies with an error that does not explain the issue. This issue aplies to any non-interactive environments, such as CI, containers, and coding agents.

Profile resolution now works without a terminal, if a selection still needs a prompt the error explains how to proceed.

## Type of change

## What's Changed

- [x] Commands that take `--profile` use the configured default or only profile without prompting
- [x] `hatchet server start` sets its new profile as the default when none is configured
- [x] A failed selection form reports the `--profile` and `set-default` options to resolve the issue
- [x] CLI help text, the profiles reference page, and the changelog have been updated

## Checklist

Changes have been:

- [x] Linted and formatted

## Testing

Unit tests cover profile resolution, the failure message, and the set-default-if-unset behavior. Manually reproduced the crash and verified the fix by running the built CLI under `setsid` with `TERM=xterm` and no `/dev/tty`, across single-profile, multi-profile, and default-set states.

---

<details id="ai-disclosure">

- **Details**: Claude Code (Claude Fable 5) performed the investigation and implemented the tests.

</details>

### DIFF
--- a/cmd/hatchet-cli/cli/profile.go
+++ b/cmd/hatchet-cli/cli/profile.go
@@ -590,13 +590,8 @@ func selectProfileForm(useDefault bool) string {
 	}
 	sort.Strings(names)
 
-	// If useDefault is true, try to return the default profile without showing the form
-	if useDefault {
-		defaultProfile := cli.GetDefaultProfile()
-		// Verify the default profile is still in the list
-		if defaultProfile != "" && slices.Contains(names, defaultProfile) {
-			return defaultProfile
-		}
+	if name, ok := resolveProfileWithoutForm(names, cli.GetDefaultProfile(), useDefault); ok {
+		return name
 	}
 
 	// Create options in sorted order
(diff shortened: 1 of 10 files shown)


cmd/hatchet-cli/cli/profile.go
```diff
@@ -621,12 +616,44 @@ func selectProfileForm(useDefault bool) string {
...
+		cli.Logger.Fatalf("%s", profileSelectionFailureMessage(err, names))
 	}
 
 	return selectedName
 }
 
+// resolveProfileWithoutForm picks the profile when the form can be skipped: the still-valid
+// default, or the only profile when useDefault is set. Management commands keep the form even
+// with a single profile because they act immediately on the returned name.
+func resolveProfileWithoutForm(names []string, defaultProfile string, useDefault bool) (string, bool) {
```
mrkaye97, line 628: small nit: pointer and return nil for this instead of returning empty strings?

mrkaye97 (approved):

cmd/hatchet-cli/cli/profile.go
BloggerBust: I prefer the (string, bool) pair since it matches `os.LookupEnv`. Also, if a caller forgets the bool check, the worst case is an empty string; with a pointer, a forgotten nil check panics with a runtime error. I am going to merge this, but if you strongly disagree let me know and I will make the change.

PR #4457, opened 2026-07-17
fix: Fix all old golintci-lint errors, add check for un-generated files, regened all proto/rest files for SDKs

juliusgeo:
# Description

Our pre-commit in CI wasn't running on all files, so a lot of lint issues have gone unfixed. This becomes an issue for people trying to do merge commits as they must then fix those issues in their branch before committing, or doing `--no-verify` which is not recommended. 

Fixes # (issue)

## Type of change

## What's Changed

## Checklist

Changes have been:

- [ ] Linted and formatted

---

<details id="ai-disclosure">

</details>


.github/workflows/sdk-python.yml
```diff
@@ -55,6 +55,19 @@ jobs:
       - name: Run Pydoclint
         run: poetry run pydoclint .
 
+      - name: Generate protobuf bindings
+        if: matrix.python-version == '3.14'
+        run: ./generate.sh
```
mrkaye97: any idea how long this takes? it's pretty slow on local because of the dependency resolution step, so makes me a bit nervous

api/v1/server/handlers/users/service.go
```diff
@@ -35,7 +35,7 @@ func (u *UserService) checkUserRestrictionsForEmail(conf *server.ServerConfig, e
 
 var ErrNotInRestrictedDomain = errors.New("email is not in the restricted domain group")
 
-const ErrInvalidCredentials = "invalid credentials"
+const ErrInvalidCredentials = "invalid credentials" // #nosec G101 -- this is an error message, not a credential
```
mrkaye97, line 38: 🤦

internal/services/scheduler/v1/scheduler.go
```diff
@@ -744,11 +744,11 @@ func (s *Scheduler) notifyAfterConcurrency(ctx context.Context, tenantId uuid.UU
 
 		msg, err := tasktypes.CancelledTaskMessage(
 			tenantId,
-			cancelled.TaskIdInsertedAtRetryCount.Id,
```
mrkaye97, line 747: maybe we can completely remove this `TaskIdInsertedAtRetryCount` on `cancelled`? not sure if it's still used with this removal

pkg/repository/olap.go
```diff
@@ -2535,9 +2527,7 @@ func (r *OLAPRepositoryImpl) ReadDAG(ctx context.Context, dagExternalId uuid.UUI
 func (r *OLAPRepositoryImpl) ListTasksByExternalIds(ctx context.Context, tenantId uuid.UUID, externalIds []uuid.UUID) ([]*sqlcv1.FlattenTasksByExternalIdsRow, error) {
 	externalUUIDs := make([]uuid.UUID, 0)
 
-	for _, id := range externalIds {
-		externalUUIDs = append(externalUUIDs, id)
-	}
+	externalUUIDs = append(externalUUIDs, externalIds...)
```
mrkaye97: lol, this is silly 😂 

we should remove the whole block instead and use `externalIds` directly, this just creates a new variable for no reason

mrkaye97 (approved):
thanks! left a couple small comments

.github/workflows/sdk-python.yml
juliusgeo: looks like on Python it takes like 1 minute, but I'm pretty sure the actual slow part is the linting we do as part of it--regardless, not too bad. I can also make it so that it doesn't block merge, though AFAIK none of the steps should block merge.

api/v1/server/handlers/users/service.go
juliusgeo: lol yeah gosec is really annoying, but the 1% of time it's useful it's good

internal/services/scheduler/v1/scheduler.go
juliusgeo: we use it for the retry count below this, IIUC

.github/workflows/sdk-python.yml
mrkaye97: nah if it's a minute it's fine with me

pkg/repository/olap.go
juliusgeo: lol yeah oops, fixed!

internal/services/scheduler/v1/scheduler.go
mrkaye97: why? 😅 if we can swap out the retry count here, can't we there too? maybe I'm missing it, idk

juliusgeo: The `TaskIdInsertedAtRetryCount` is embedded inside `TaskWithCancelledReason` struct--no idea why previously we were calling `cancelled.TaskIdInsertedAtRetryCount` for these fields

PR #4460, opened 2026-07-19
feat(engine): terminal status-based idempotency key release

mrkaye97:
# Description

Adds a second mechanism for releasing idempotency keys, which is to release them once the task that claimed them reaches a terminal status, as requested by a few people. The implementation is pretty simple - in `releaseTasks` we just run a new query that does the cleanup, similar to how the other `release*` queries run

Need to make sure this is okay performance-wise still and push the SDKs and docs

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

<details id="ai-disclosure">

Used Claude Code to replicate Python SDK changes to the other SDKs

</details>


cmd/hatchet-migrate/migrate/migrations/20260719000944_v1_0_132.sql
```diff
@@ -0,0 +1,14 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TYPE idempotency_method AS ENUM ('TTL', 'STATUS');
+
+ALTER TABLE "WorkflowVersion"
+ADD COLUMN "idempotencyMethod" idempotency_method;
```
abelanger5, line 6: shouldn't we add a `DEFAULT` to make this easier on ourselves, so we don't need to deal with the null values?

abelanger5 (approved):

cmd/hatchet-migrate/migrate/migrations/20260719000944_v1_0_132.sql
mrkaye97: I don't think so, since then we'd have most rows with an idempotency method but no expression or ttl - it's probably "correct" to make the idempotency config a separate table instead and then have these all be non-null, but I don't really want to deal with adding joins all over. How strongly do you feel...?

pkg/repository/sqlcv1/idempotency-keys.sql
```diff
@@ -62,3 +62,36 @@ SELECT
...
 ;
+
+-- name: ReleaseIdempotencyKeys :exec
+-- !! IMPORTANT: this only gets called when a task reaches a terminal state (exhausted all retries, completed, etc.)
+-- which means we want to evict any idempotency keys that still are live and tied to the task at this point
+WITH input AS (
+    SELECT
+        UNNEST(@taskIds::BIGINT[]) AS task_id,
+        UNNEST(@taskInsertedAts::TIMESTAMPTZ[]) AS task_inserted_at,
+        UNNEST(@taskRetryCounts::INTEGER[]) AS retry_count
```
abelanger5, line 73: I think this might now lead to some raciness on the behavior of `retry_count`, though I'm not sure. I'm wondering what happens if the terminal event comes in much later, like let's say that the task retries and then a few milliseconds later the completed event is processed. Does the idempotency key get released or not? And is there ever a case where this could lead to an orphaned idempotency key?

abelanger5 (approved):

PR #4468, opened 2026-07-20
perf: bulk write meter tenant resources

grutt:
# Description

Metering tenant resource limits currently issues one `MeterTenantResource` UPDATE per task/event, which caused row-lock contention on `TenantResourceLimit` under high-throughput tenants and contributed to `task_processing_queue_v2` backlog. Meter deltas are now buffered in memory per `(tenant, resource)` and flushed to the database in a single bulk UPDATE once per second, following the same pattern as the scheduler's rate limiter.

## Type of change

## What's Changed

- Add `BulkMeterTenantResources` query that coalesces meter deltas and updates all `TenantResourceLimit` rows in one statement, with deterministic lock ordering to avoid deadlocks between engine replicas
- Buffer meter deltas in `tenantLimitRepository` instead of writing per-meter; a background loop flushes every 1s (map is swapped out before the DB call so hot paths never block on a flush; deltas are merged back on flush error)
- `canCreate` includes buffered-but-unflushed deltas so control stays accurate between flushes
- `Stop()` cancels the flush loop and performs a final flush so counts aren't dropped on graceful shutdown
- Reuse the shared `TenantLimitRepository` instance in `NewRepository` so only one buffer/flusher exists per process

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Implementation and PR description written with Cursor agents (Grok and Claude), guided and reviewed by a human.

</details>

### DIFF
--- a/pkg/repository/tenant_limit.go
+++ b/pkg/repository/tenant_limit.go
@@ -50,6 +50,12 @@ type meterKey struct {
 	tenantId uuid.UUID
 }
 
+// cacheKey is the string form used for the canCreate result cache, which is
+// backed by a string-keyed TTL cache.
+func (k meterKey) cacheKey() string {
+	return fmt.Sprintf("%s:%s", k.resource, k.tenantId)
+}
+
 type meterSet map[meterKey]int32
 
 type tenantLimitRepository struct {
@@ -320,8 +326,8 @@ func (t *tenantLimitRepository) flushToDatabase(ctx context.Context) error {
 	for _, limit := range limits {
 		percent := calcPercent(limit.Value, limit.LimitValue)
 		if percent <= 50 || percent >= 100 {
-			key := fmt.Sprintf("%s:%s", limit.Resource, limit.TenantId)
-			t.c.Set(key, limit.Value < limit.LimitValue)
+			key := meterKey{tenantId: limit.TenantId, resource: limit.Resource}
+			t.c.Set(key.cacheKey(), limit.Value < limit.LimitValue)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/repository/tenant_limit.go
```diff
@@ -209,25 +235,97 @@ func calcPercent(value int32, limit int32) int {
...
 
 	if err != nil {
-		return nil, err
+		// merge the deltas back so the next flush retries them
+		t.unflushedMu.Lock()
+		for key, n := range updates {
+			t.unflushed[key] += n
+		}
+		t.unflushedMu.Unlock()
+
```
mrkaye97, line 322: this feels kind of gnarly to me but I'm not sure of a better way off the top of my head - this type of not-quite-atomic rollbacking thing makes me nervous though (i.e. what happens if two tickers somehow overlap and one of them succeeds in updating the metered values and the other fails or something?)

pkg/repository/tenant_limit.go
```diff
@@ -209,25 +235,97 @@ func calcPercent(value int32, limit int32) int {
...
+		return err
+	}
+
+	for _, limit := range limits {
+		percent := calcPercent(limit.Value, limit.LimitValue)
+		if percent <= 50 || percent >= 100 {
+			key := fmt.Sprintf("%s:%s", limit.Resource, limit.TenantId)
```
mrkaye97: can we make this key a struct instead of a magic string?

mrkaye97 (approved):
left a couple small comments but agreed I think this will help a lot

pkg/repository/tenant_limit.go
grutt: yeah agree, i made one other change there's only one ticker per process since we share shared.m

pkg/repository/tenant_limit.go
grutt: good idea

grutt: the Cacheable util takes a string, but pulled out the magic string into a fn on meterKey

PR #4480, opened 2026-07-22
refactor: separate pub/sub interface from durable task queue

abelanger5:
# Description

The `msgqueue` interface currently has two responsibilities:
1. Non-durable, at-most-once delivery to consumers for pub/sub. The primary examples of this are tenant subscription streams, tenant streaming features, and scheduler listeners for hot-path polling.
2. Durable, at-least-once delivery for critical messages, like the task and OLAP controllers. 

These use-cases are long-term divergent and we'd like to experiment with other pub/sub technology, so this PR splits the pub/sub implementation from the durable task queue. 

## Type of change

## What's Changed

- [X] Modifies the msgqueue to add a `PubSub` implementation
- [X] Modifies callers to use pub/sub where appropriate

## Checklist

Changes have been:

- [X] Linted and formatted

---

<details id="ai-disclosure">

</details>


pkg/repository/multiplexer.go
```diff
@@ -62,7 +62,10 @@ func (m *multiplexedListener) startListening() {
 			if err != nil {
 				return nil, err
 			}
-			return poolConn.Conn(), nil
+			// Hijack removes the connection from the pool's accounting: pgxlisten
+			// owns it and closes it when the listener exits, so the pool can be
+			// closed cleanly and slots aren't leaked across listener reconnects.
+			return poolConn.Hijack(), nil
 		},
```
igor-kupczynski: @abelanger5 this deserves to be in its own PR imho; this is a bug in the current OSS version (bug as in, this can lead a to a very slow connection leak if I understand this correctly).

internal/msgqueue/pubsub.go
```diff
@@ -0,0 +1,102 @@
...
+	"context"
+	"fmt"
+
+	"github.com/google/uuid"
+)
+
+// Topic identifies a best-effort pub/sub destination.
+type Topic struct {
+	name         string
+	tenantStream bool
```
igor-kupczynski: nit, but can you rename to `isTenantStream`? in my mind is conveys the _boolean-ness_ better

internal/msgqueue/pubsub.go
```diff
@@ -0,0 +1,102 @@
...
+type PubSub interface {
+	// Pub publishes a message to the topic. Delivery is best-effort: if no
+	// subscriber is listening, the message is dropped.
+	Pub(ctx context.Context, topic Topic, msg *Message) error
+
+	// Sub subscribes to a topic. Delivery is at-most-once with no ack
+	// semantics: handler errors are logged, never redelivered. It returns a
+	// cleanup function that should be called when the subscription is no
+	// longer needed.
+	Sub(topic Topic, handler AckHook) (func() error, error)
```
igor-kupczynski: I've scratched my head why do we need ACK hook for a best-effort pubsub, but this is just a re-use of an existing class right? Maybe define an alias or refine calling it `Handler`?

internal/msgqueue/pubsub.go
igor-kupczynski: also it's maybe a smell as it's closed for extension

an enum would be preferred imho

internal/services/controllers/task/controller.go
igor-kupczynski: Possibly missing a tenant-stream publish for task-cancelled on the API cancellation path?

Before this PR, this `SendMessage` was implicitly mirrored to the tenant fanout exchange (as the old pubMessage mirrored every durable message with a non-nil TenantID). Now that mirroring is explicit, this path only publishes to the durable queue.

igor-kupczynski (commented):
:100: to have the `PubSub` interface and related primitives; i've raised some comments.

Can we split this PR? At the very least into the `Hijack` fix, but otherwise when ready you can prompt cursor/claude to split into 2-3 stacked PRs for an easier review.

internal/msgqueue/pubsub.go
abelanger5: Yeah, this is because it's reusing `AckHook` from the old implementation. We can rename `AckHook` to `MsgHandler`, it'll be more clear that way.

internal/services/controllers/task/controller.go
abelanger5: I had Fable enumerate the messages consumed from the tenant exchange:

```
┌───────────────────────────────┬───────────────────────┬──────────────────────────┐
│            Msg ID             │        Handler        │         Becomes          │
├───────────────────────────────┼───────────────────────┼──────────────────────────┤
│ created-task                  │ msgsToWorkflowEvent   │ STEP_RUN / STARTED       │
├───────────────────────────────┼───────────────────────┼──────────────────────────┤
│ task-completed                │ msgsToWorkflowEvent   │ STEP_RUN / COMPLETED     │
├───────────────────────────────┼───────────────────────┼──────────────────────────┤
│ task-failed                   │ msgsToWorkflowEvent   │ STEP_RUN / FAILED        │
├───────────────────────────────┼───────────────────────┼──────────────────────────┤
│ task-cancelled                │ msgsToWorkflowEvent   │ STEP_RUN / CANCELLED     │
├───────────────────────────────┼───────────────────────┼──────────────────────────┤
│ task-stream-event             │ msgsToWorkflowEvent   │ STEP_RUN / STREAM        │
├───────────────────────────────┼───────────────────────┼──────────────────────────┤
│ workflow-run-finished         │ both handlers         │ WORKFLOW_RUN / COMPLETED │
│                               │                       │ |FAILED|CANCELLED        │
├───────────────────────────────┼───────────────────────┼──────────────────────────┤
│ workflow-run-finished-candida │ isMatchingWorkflowRun │ wakes the run-finished   │
│ te                            │ V1 only               │ poll                     │
└───────────────────────────────┴───────────────────────┴──────────────────────────┘
```

So yes, `task-cancelled` is consumed - nice catch! Will audit these as well.

pkg/config/loader/loader.go
```diff
@@ -529,8 +533,91 @@ func createControllerLayer(dc *database.Layer, cf *server.ServerConfigFile, vers
...
+
+			if err != nil {
+				return nil, nil, fmt.Errorf("could not init rabbitmq pubsub: %w", err)
+			}
+
+			pubsubv1 = ps
+			cleanupPubSub = cleanupRmq
+		default:
+			return nil, nil, fmt.Errorf("invalid pubsub kind %q, must be 'rabbitmq' or 'postgres'", pubsubKind)
+		}
```
abelanger5: nit: should be refactored into a helper

internal/services/controllers/task/controller.go
abelanger5: Instead of doing each `Pub` manually, we can probably just pull this enum into a helper and automatically pub these messages - seems like it'd be cleaner than the manual pubs everywhere.

internal/msgqueue/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,473 @@
...
+	return func() error {
+		cancel()
+		pubChannelPool.Close()
+		subChannelPool.Close()
+		return nil
+	}, p, nil
+}
+
+func (p *PubSub) IsReady() bool {
+	return p.pubChannels.hasActiveConnection() && p.subChannels.hasActiveConnection()
```
igor-kupczynski, line 159: @abelanger5 let's discuss this, it feeds into the readiness check and liveness check, so if rabbit is disconnected, the pod will be marked as not ready;

readiness check --> can pod serve traffic (arguably with pubsub, it can even if the pubsub is disconnected)

liveness check --> is pod broken (if yes, kubelet will restart it); so this means that we will restart the pod on rabbit disconnect; is it what we want?

internal/msgqueue/postgres/pubsub.go
```diff
@@ -0,0 +1,257 @@
...
+		if ackId != nil {
+			if err := p.repo.AckMessage(subscribeCtx, *ackId); err != nil {
+				p.l.Error().Err(err).Msg("error acking message")
+			}
+		}
+
+		if err := handler(task); err != nil {
+			p.l.Error().Err(err).Msgf("error handling pubsub message %s", task.ID)
+		}
+	}
```
igor-kupczynski, line 205: * Handlers can receive gzipped payloads here when the durable queue is RabbitMQ with `compressionEnabled: true` as `pubMessage` compresses in place (`msg.Payloads = compressed; msg.Compressed = true`) and `PubTenantMessage` calls `SendMessage` before handing the same pointer to `Pub` --> the Postgres pub/sub publishes the mutated message.

* The RabbitMQ pub/sub decompresses on its Sub side; this backend never does, so `handler(task)` gets gzip bytes with `task.Compressed == true` which breaks subscribers.

Check for `.Compressed` also in the postgres handler and decompress as required.

igor-kupczynski (changes requested):

internal/msgqueue/rabbitmq/pubsub.go
abelanger5: I think we should likely continue to expose `IsReady` but not call it from the probes; I agree it doesn't make sense to reject all API work if our pub/sub is down

internal/msgqueue/postgres/pubsub.go
abelanger5, line 1: note: I'd like to consolidate implementations by using pgoutbox instead of the `"MessageQueueItem"` approach. We'll need to migrate message queue items and write them into the outbox tables. But this would give us a single source of truth for outbox-based processing which is ideal.

internal/msgqueue/postgres/pubsub.go
```diff
@@ -0,0 +1,271 @@
...
+
+// PubSub implements msgqueue.PubSub over Postgres LISTEN/NOTIFY. Topic names
+// match the legacy non-durable queue / NOTIFY names, and all traffic
+// multiplexes over the single "hatchet_listener" channel, so mixed-version
+// fleets interoperate.
+//
+// INVARIANT: the repo passed to NewPubSub must be built on a dedicated pool
+// from the direct (non-pgbouncer) database URL — never the shared repository
+// pool. Pub can be called from within durable-write paths, and LISTEN does not
+// survive transaction pooling.
```
abelanger5: nit: don't like this comment, perhaps:

```
PubSub implements msgqueue.PubSub in Postgres, using LISTEN/NOTIFY for messages less than 8kb and the `"MessageQueueItem"` table for messages greater than 8kb. 
```

internal/msgqueue/postgres/pubsub.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+func WithPubSubLogger(l *zerolog.Logger) PubSubOpt {
+	return func(opts *PubSubOpts) {
+		opts.l = l
+	}
+}
+
+// NewPubSub creates a new Postgres-backed PubSub over the given message queue
+// repository.
```
abelanger5, line 67: nit: this is where we should document the `direct (non-pgbouncer) database URL` behavior

internal/msgqueue/postgres/pubsub.go
```diff
@@ -0,0 +1,271 @@
...
+
+	return func() error {
+		c.Stop()
+		return nil
+	}, p, nil
+}
+
+func (p *PubSub) IsReady() bool {
+	return true
+}
```
abelanger5, line 107: nit: this matches the prior implementation, but is this really what we want? we probably want `IsReady` to reflect actual readiness, and let the caller decide what to do with it.

internal/msgqueue/postgres/pubsub.go
abelanger5: should be resolved by https://github.com/hatchet-dev/hatchet/pull/4480/changes/a825a7b664eeb202b0e22cd21bef1cf2ce34429c#diff-0a18c7ff5e3579de9defe982a69865977cb787a893184ce4e1ea2369e56f00ad

internal/msgqueue/rabbitmq/gzip.go
```diff
@@ -4,11 +4,18 @@ import (
...
 	"sync"
 )
 
+// compressor holds gzip compression settings shared by the durable
+// MessageQueueImpl and the best-effort PubSub (compression must agree on both
+// sides for wire compatibility).
+type compressor struct {
+	compressionEnabled   bool
+	compressionThreshold int
+}
```
abelanger5: would it make more sense to move the compressor to the `msgqueue` package instead? it feels like we should support compression from all msgqueue implementations. I wouldn't be opposed to adding this as an env var option which lives at the msgqueue config layer instead, and ensure it's backwards compatible with the old var.

internal/msgqueue/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,473 @@
...
+
+// WithPubSubGzip enables gzip compression of payloads. Compression settings
+// must match the durable MessageQueue's settings for wire compatibility, since
+// both sides publish to the same tenant exchanges in mixed-version fleets.
+func WithPubSubGzip(enabled bool, threshold int) PubSubOpt {
+	return func(opts *PubSubOpts) {
+		opts.compressionEnabled = enabled
+
+		if threshold <= 0 {
+			threshold = 5 * 1024 // default to 5KB
```
abelanger5: nit: pull magic value into top-level constant

internal/msgqueue/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,473 @@
...
+		ctx:         ctx,
+		identity:    identity(),
+		l:           opts.l,
+		pubChannels: pubChannelPool,
+		subChannels: subChannelPool,
+		compressor: compressor{
+			compressionEnabled:   opts.compressionEnabled,
+			compressionThreshold: opts.compressionThreshold,
+		},
+		maxPayloadSize: 16 * 1024 * 1024, // 16 MB
```
abelanger5: nit: magic constant

internal/msgqueue/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,473 @@
...
+		if len(msg.Payloads) == 1 {
+			return fmt.Errorf("message size %d bytes exceeds maximum allowed size of %d bytes", len(body), p.maxPayloadSize)
+		}
+
+		// split the payloads in half and publish recursively until each chunk is
+		// under the max size (same strategy as the durable pubMessage)
+		payloadsPerChunk := max(len(msg.Payloads)/2, 1)
+
+		for chunk := range slices.Chunk(msg.Payloads, payloadsPerChunk) {
+			err := p.Pub(ctx, topic, &msgqueue.Message{
```
abelanger5: nit: much cleaner if `msgqueue.Message` has a `Clone` method

internal/msgqueue/rabbitmq/pubsub.go
```diff
@@ -0,0 +1,473 @@
...
+	return cleanup, nil
+}
+
+// declareSubQueue declares the consumer queue for a topic and returns its
+// name. Tenant topics get a random-suffix exclusive queue bound to the tenant
+// fanout exchange; scheduler topics consume the well-known partition queue on
+// the default exchange.
+func (p *PubSub) declareSubQueue(ch *amqp.Channel, topic msgqueue.Topic) (string, error) {
+	args := make(amqp.Table)
+	args["x-consumer-timeout"] = 300000 // 5 minutes
```
abelanger5: same nit on magic constant (and should probably apply changes to the `rabbitmq.go` file)

internal/services/controllers/olap/process_dag_status_updates.go
```diff
@@ -103,12 +103,10 @@ func (o *OLAPControllerImpl) notifyDAGsUpdated(ctx context.Context, rows []v1.Up
...
 			}
 
-			q := msgqueue.TenantEventConsumerQueue(tenantId)
-
-			err = o.mq.SendMessage(ctx, q, msg)
-
-			if err != nil {
-				return err
+			// fanout-only: the dispatcher's workflow run subscriptions consume
+			// workflow-run-finished off the tenant stream
```
abelanger5: nit: remove this comment, doesn't make sense in this context

internal/services/controllers/olap/process_task_status_updates.go
```diff
@@ -103,12 +103,10 @@ func (o *OLAPControllerImpl) notifyTasksUpdated(ctx context.Context, rows []v1.U
...
 			}
 
-			q := msgqueue.TenantEventConsumerQueue(tenantId)
-
-			err = o.mq.SendMessage(ctx, q, msg)
-
-			if err != nil {
-				return err
+			// fanout-only: the dispatcher's workflow run subscriptions consume
+			// workflow-run-finished off the tenant stream
```
abelanger5: same here

internal/services/health/health.go
```diff
@@ -24,6 +24,10 @@ type Health struct {
 	l          *zerolog.Logger
 }
 
+// New creates a health service over the durable message queue and repository.
+// The best-effort PubSub is deliberately NOT part of the probes: a disconnected
+// pub/sub degrades stream notifications but the pod can still serve traffic,
+// and restarting it on a broker disconnect would make things worse.
```
abelanger5: nit: this is irrelevant to future readers, let's remove

igor-kupczynski (approved):
> [!NOTE]
> Igor: This was posted by AI a bit too eagerly; keeping as is though and will comment below

Did a full pass over the final state of the branch (checked out locally, read the changed files in full, audited producer/consumer call sites, ran `go build ./...` and the msgqueue/loader unit tests).

**What I verified:**

- **Producer parity with the old implicit mirror**: every producer of the seven stream-consumed message IDs (`created-task`, `task-completed`, `task-failed`, `task-cancelled`, `task-stream-event`, `workflow-run-finished`, `workflow-run-finished-candidate`) now publishes explicitly via `PubTenantMessage` or `ps.Pub` — dispatcher completed/failed, scheduler cancel/internal-retry/dead-letter, task controller cancel + run-finished-candidate, OLAP signaler created-task + republish + run-finished, ingestor stream events. The one implicit mirror that was dropped without replacement (`sendTaskCancellationsToDispatcher`, which sends `task-cancelled` with `SignalTaskCancelledPayload` to dispatcher queues) previously decoded to zero-valued `CancelledTaskPayload` junk on the stream, so dropping it is an improvement.
- **Wire compatibility**: tenant topic (`<uuid>_v1`) and scheduler partition (`<p>_scheduler_v1`) names, exchange/queue declare args, and postgres NOTIFY channel names all match legacy, so mixed-version fleets interoperate.
- **Cross-backend compression**: durable `pubMessage` now compresses a clone (no caller mutation), and the postgres subscriber decompresses `Compressed` messages — the earlier review finding is properly fixed at both ends.
- **`SharedBufferedTenantReader`** correctly passes `WithKind(PostAck)`, so the single-handler pubsub Sub maps onto the buffer's active hook.
- **Pool isolation**: postgres pubsub gets its own small pool on the direct (non-pgbouncer) URL; rabbitmq pubsub owns its connections. Health probes no longer gate on pubsub readiness, per the review discussion.
- **Fallback-row hygiene**: >8KB postgres fallback rows are acked by subscribers and reaped by the retention controller's existing `CleanupMessageQueueItems` loop regardless of msgqueue kind.
- Config inheritance is backwards compatible (incl. legacy `SERVER_TASKQUEUE_*` aliases) and covered by `loader_pubsub_test.go`; docs updated; CI fully green on the latest commit.

**Non-blocking follow-ups** (none of these should hold up the merge):

1. `rabbitmq/pubsub.go` `Sub` leaves `msg.Compressed = true` after decompressing (the postgres pubsub resets it to `false`; the durable rabbitmq subscriber has the same pre-existing quirk). Harmless today since no handler re-publishes the same `*Message`, but it's a footgun if one ever does — worth normalizing.
2. `tenantStreamMsgIDs` in `pubsub.go` is a hand-maintained mirror of the dispatcher's stream consumers (`msgsToWorkflowEvent` / `isMatchingWorkflowRunV1`). A small drift test asserting the two stay in sync would prevent a silently dead stream event type when a new msg ID is added.
3. Rabbitmq pubsub `Pub` sets `Expiration: "0"` on all publishes, whereas the old tenant-exchange mirror published without a per-message TTL — messages now drop when a subscriber queue momentarily has no ready consumer. Consistent with the at-most-once contract, just noting the subtle semantic change.
4. Postgres >8KB fallback rows deliver to only one subscriber per message (`ReadMessages`+`Ack`), unlike the rabbitmq fanout. This is pre-existing postgres behavior, but it's now a per-implementation divergence of the `PubSub` contract — worth documenting on the interface.

igor-kupczynski:
Only 1 and 4 make sense to me; I'll push a commit. I'm considering 2 as well, but only if that's small.

igor-kupczynski (approved):
Amazing stuff Alexander, thank you!

igor-kupczynski:
I'm going to merge to unblock some of my work; given the 7 commit history, I'm going to squash.

PR #4498, opened 2026-07-23
feat: add NATS pub/sub backend

igor-kupczynski:
# Description

Adds NATS as a pub/sub backend, selectable per deployment via `SERVER_MSGQUEUE_PUBSUB_KIND=nats` + `SERVER_MSGQUEUE_PUBSUB_NATS_URL`. NATS implements only the best-effort pub/sub side of the split from #4480; durable queues stay on RabbitMQ/Postgres.

Core NATS, no JetStream, so delivery is at-most-once — which matches what the pub/sub path already promises.

**Stacked on #4548** (pub/sub Prometheus instrumentation), so that the metrics used to evaluate this backend are already deployed and have a baseline before NATS is enabled anywhere. Review #4548 first; this PR's diff is NATS-only.

## Type of change

## What's Changed

`internal/msgqueue/nats` plus config, docs, and a compose service.

- **Startup validation.** Connect is synchronous and checks the server's advertised `max_payload` against the 16MiB the durable RabbitMQ path allows. Task stream events routinely exceed the NATS default of 1MiB, and a refused startup is easier to diagnose than an oversized publish failing later under load.
- **Fail fast while disconnected.** `ReconnectBufSize(-1)` means publishes error out rather than silently buffering messages that a best-effort subscriber will never wait for.
- **Chunking parity with RabbitMQ.** Oversized multi-payload messages are split recursively until each chunk fits under `max_payload`.
- **Subject prefix.** `SERVER_MSGQUEUE_PUBSUB_NATS_SUBJECT_PREFIX` (default `hatchet.pubsub`). The other backends namespace installations at the connection level (a RabbitMQ vhost, a Postgres database); NATS namespaces by subject, so this is the parity knob for installations sharing a NATS server.
- **Auth as connect options.** Username and password are passed to the client rather than embedded in the URL, so reconnects to cluster peers discovered through gossip authenticate too.

## Checklist

Changes have been:

- [x] Linted and formatted

## Testing

```bash
docker compose up -d nats
go test -tags integration ./internal/msgqueue/nats/...
go test ./internal/msgqueue/...
```

The integration suite covers tenant fanout to multiple subscribers, a scheduler-topic roundtrip, an 8MiB payload, oversized-message chunking, a compressed roundtrip, and a custom subject prefix (asserted against a raw NATS subscription on the expected subject).

The NATS server needs `max_payload: 16777216`; see `hack/dev/nats-server.conf`, which the compose service mounts.

## Notes for reviewers

Rebased onto `main` after #4480 merged. #4480's review absorbed several refactors this branch originally carried, so they are no longer in the diff:

- The shared gzip decompress extraction — `main` moved the whole `Compressor` into `msgqueue`, going further than this branch did
- `Message.CopyWithPayloads` — `main` landed the same shallow-copy fix as `Clone()`, which the NATS chunking path now uses
- The `rabbitmq` and `postgres` pub/sub changes — after rebasing, those files were identical to `main`

Two things the rebase surfaced, both fixed here:

- A comment in the NATS `Sub` path claimed the durable RabbitMQ path compresses payloads in place. #4480 fixed exactly that, so it is now a short note explaining that the decompress branch is defensive parity with `rabbitmq/pubsub.go`
- The NATS `Sub` path was not resetting `Compressed = false` after decompressing, unlike `rabbitmq/pubsub.go`

One thing worth a reviewer's opinion: `hack/dev/nats-server.conf` sets `max_payload: 16777216` for RabbitMQ parity, and NATS itself warns at startup that payloads over 8MiB "are generally discouraged and could lead to poor performance". Parity is the conservative choice for a drop-in backend, but if we don't actually need >8MiB pub/sub messages it would be better to find that out before rollout.

## Related

- Builds on #4480 (merged) — separated the pub/sub interface from the durable task queue
- Stacked on #4548 — pub/sub Prometheus instrumentation

---

<details id="ai-disclosure">

- **Details**: Cursor agent implemented the NATS pub/sub backend, config/docs/compose wiring, and tests from a reviewed plan, and performed the rebase onto `main` after #4480 merged.

</details>


igor-kupczynski:
GitHub marked this PR as merged in error — the NATS backend has **not** landed on `main`.

While reordering the stack, this PR's base was pointed at `pubsub-metrics`, and that branch was then updated to include this branch's commit. GitHub saw the head commit become reachable from the base and auto-closed the PR as merged, then deleted the head branch. The `mergeCommit` it recorded (`a4707f7cf`) was never on `main`.

Merged PRs can't be reopened, so the work continues in **#4550** — same branch, same change, based on `main`. The pub/sub instrumentation that was originally bundled in here is now a separate PR, **#4548**, stacked on top of #4550.

Nothing was lost; only the PR number changed. The review discussion here is still the earlier context.

PR #4521, opened 2026-07-26
fix(olap): prevent infinite poison message loops

khushhhhh:
## 💡 Executive Summary
This PR fixes a critical infinite poison-message loop in the OLAP task event pipeline. By making the OLAP database inserts strictly idempotent and enabling RabbitMQ message rejection by default, we prevent the event ingester from crashing or dead-looping when processing duplicate `task_id` events under high concurrency.

## 🔧 Technical Architecture & Approach
*   **Root Cause / Context:** The `CreateTaskEventsOLAP` query previously utilized `pgx.CopyFrom`, which natively lacks support for `ON CONFLICT` clauses. When poison messages repeatedly dead-lettered, RabbitMQ would infinitely retry them because `enableMessageRejection` was defaulted to `false`. This caused `pgx.CopyFrom` to throw fatal primary key collision errors, dropping entire event batches and causing systemic ingestion lag.
*   **Solution Design:** 
    1. **Idempotent Bulk Inserts:** Replaced `pgx.CopyFrom` with a batched `INSERT ... SELECT ... FROM UNNEST` syntax in `pkg/repository/sqlcv1/olap.sql`. This unlocks the use of `ON CONFLICT (task_id, task_inserted_at, id) DO NOTHING`.
    2. **Code Generation:** Regenerated type-safe Go structs and adapted the internal OLAP controllers to map event batches into arrays for the new SQL query.
    3. **Poison Message Handling:** Changed `enableMessageRejection` to `true` by default in `internal/msgqueue/rabbitmq/rabbitmq.go` to cleanly discard and acknowledge dead messages once their `maxDeathCount` is exceeded.
*   **Trade-offs Considered:** We evaluated keeping `pgx.CopyFrom` and aggressively deduplicating events in-memory before database insertion. However, that approach fails to protect against cross-batch race conditions across multiple worker nodes. Pushing the idempotency constraint down to the Postgres layer via `UNNEST` arrays guarantees ACID-compliant safety with virtually identical performance to `CopyFrom`.

## 🧪 Verification & Test Proof
*   [x] Updated `CreateTaskEventsOLAP` and successfully regenerated the `sqlc` interface.
*   [x] Executed full local compilation (`go build ./...`) and code generation (`task generate`) — all checks passing without syntax/type errors.
*   [x] Verified that RabbitMQ configurations correctly drop dead-letter events instead of looping.
*   [x] Passed all repository linter and code-formatting checks.


gregfurman:
I think for such a wide-reaching change, especially to the core engine, we need a lot more evidence that this is actually a bug. Normally, we'd expect a corresponding issue with clear reproduction steps that has been recreated + accepted by maintainers. Also, there is no AI declaration which makes me hesitant to review this. 

See https://github.com/hatchet-dev/hatchet/blob/main/CONTRIBUTING.md

Once an issue has been opened with the appropriate reproduction steps, and accepted by maintainers, I'd be happy to revisit the contribution. 

In the meantime, I'm going to close this.

PR #3975, opened 2026-05-20
feat:  Batch flush

juliusgeo:
# Description

https://github.com/hatchet-dev/hatchet/pull/2617
Adds concept of "dynamic batches" that allow the workflow to specify either a max batch size, or a batch timeout time, or both, that will flush the batch to the worker once either of those limits are reached. Batch scheduling is done by moving rows from v1_queue_item to v1_batch_queue_item where they will be picked up by the batch scheduler. The dispatcher will send a "START_BATCH" action to workers, containing all the payloads in the batch. If the payloads in a batch are about to exceed the 4mb gRPC limit, then the batch will automatically be flushed, resulting in lower batch sizes. 

Assorted architectural notes:
 - BatchSchedulers are guarded by tenant leases, scoped per step. Batch keys are handled inside the BatchScheduler
 - Messages sending the batches to and from the engine are always a singular message, (so START / FAILURE / SUCCESS / CANCELLED states) to make sure that batches are always executed atomically. A failure in the batch will propagate back to all of the call sites.
 - If broadcast_output is set to true, then the output from the batch step function will be returned to all call sites as-is. If it is false, then the output (a dictionary mapping task ids to outputs), will send the outputs under each key to the respective call site it originated from.
 - Batch group keys can be used to partition different task runs in the same workflow into separate batches, with the same batch settings. So if you want to make sure all of Tenant A's tasks get executed as part of one function, and same for Tenant B, you set a batch key like `input.tenant_id`. 
 - Batches will be flushed when one of the following conditions is true:
   1. The batches max size is reached
   2. The batches max interval is reached
   3. The payloads of the tasks queued for a batch are about to breach the 4mb GRPC limit.
 

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation change (pure documentation change)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Refactor (non-breaking changes to code which doesn't change any behaviour)
- [ ] CI (any automation pipeline changes)
- [ ] Chore (changes which are not directly related to any business logic)
- [x] Test changes (add, refactor, improve or change a test)
- [x] This change requires a documentation update

## What's Changed

- [x] Engine/scheduler/dispatcher support
- [ ] Go SDK support
- [ ] Typescript SDK support
- [x] Python SDK support
- [ ] Ruby SDK support 

## Checklist

Changes have been:

[...]

PR #4529, opened 2026-07-27
fix: reject TS workflow result failures as Error

### LINKED ISSUE #4487: TS SDK: workflow-run-ref rejects with mixed string|undefined[] when any step fails
## Bug

In `sdks/typescript/src/util/workflow-run-ref.ts` lines 116–121, when a workflow finishes with at least one failed step, the SDK rejects the `result()` promise with an incorrect array.

## Root cause

```ts
if (event.results.some((r) => r.error !== undefined)) {
  const errors = event.results.map((r) =>
    r.error !== '' ? r.error : 'task was cancelled'  // ← bug here
  );
  reject(errors);
}
```

`r.error` is typed `string | undefined`. For steps that succeeded, `r.error` is `undefined`. The guard `r.error !== ''` evaluates to `undefined !== ''` → `true`, so the ternary returns `r.error` (i.e., `undefined`) for every successful step.

The result is `errors` is an `Array<string | undefined>` that includes `undefined` for every step that did NOT fail — and `reject(errors)` rejects with this raw array (not an `Error` object), so callers doing:

```ts
try {
  await run.result();
} catch (e) {
  console.error(e.message); // undefined — e is an array, not an Error
}
```

get no useful error message.

## Suggested fix

```ts
if (event.results.some((r) => r.error !== undefined)) {
  const errors = event.results
    .filter((r) => r.error != null && r.error !== '')
    .map((r) => r.error ?? 'task was cancelled');
  reject(new Error(errors.join('; ')));
  return;
}
```

This ensures only failed steps contribute to the message, and the rejection is a proper `Error` object that callers can `catch` normally.

## Steps to reproduce

Run any multi-step workflow where at least one step succeeds and at least one step fails. Catch the `result()` promise — the caught value is an array, not an Error, and contains `undefined` entries for the passing steps.

jatinder14:
## Summary
- reject failed TypeScript SDK workflow results with a standard `Error`
- retain only failed-step messages in the error text
- add regression coverage for mixed successful and failed steps

Closes #4487

## Test plan
- [x] `pnpm test:unit -- src/util/workflow-run-ref.test.ts`
- [x] `pnpm exec prettier --check src/util/workflow-run-ref.ts src/util/workflow-run-ref.test.ts`
- [x] `pnpm exec tsc --noEmit`

### DIFF
--- a/sdks/typescript/src/util/workflow-run-ref.test.ts
+++ b/sdks/typescript/src/util/workflow-run-ref.test.ts
@@ -1,0 +1,35 @@
+import {
+  RunListenerClient,
+  StepRunEvent,
+} from '@hatchet/clients/listeners/run-listener/child-listener-client';
+import { WorkflowRunEventType } from '../protoc/dispatcher';
+import WorkflowRunRef from './workflow-run-ref';
+
+describe('WorkflowRunRef', () => {
+  it('rejects failed runs with an Error containing only failed step messages', async () => {
+    const finishedEvent = {
+      eventType: WorkflowRunEventType.WORKFLOW_RUN_EVENT_TYPE_FINISHED,
+      results: [
+        { taskName: 'successful-step', error: undefined, output: '{}' },
+        { taskName: 'failed-step', error: 'step failed', output: '{}' },
+      ],
+    } as unknown as StepRunEvent;
+    const client = {
+      get: jest.fn().mockResolvedValue({
+        stream: async function* () {
+          yield finishedEvent;
+        },
+      }),
+    } as unknown as RunListenerClient;
+    const workflowRun = new WorkflowRunRef('workflow-run-id', client);
+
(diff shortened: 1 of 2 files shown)


jatinder-mahajan-certifyos:
@grutt if you have a moment — you've done most of the work in `workflow-run-ref.ts`.

At line 124 the finished-run path does `reject(errors)` where `errors` is a `string[]`, so callers get a raw array instead of an `Error`. `error instanceof Error` is false and `error.message` is undefined in any normal try/catch around `await ref.result()`. This wraps it in `new Error(errors.join('; '))` and adds a test.

Two files, no behaviour change beyond the rejection type. Happy to adjust the join separator if you'd prefer a different shape.

grutt:
I think this was addressed in #3975 by filtering empty errors, going to close for now but can reopen if this is not the case.

PR #4543, opened 2026-07-27
refactor: use outbox for olap message flush

abelanger5:
# Description

Moves the boundary between tasks and OLAP into the `outbox.messages` table, which then flushes to rabbitmq. This improves the guarantees from `best-effort` to `at-least-once`, which we should be tolerant of on the OLAP side with our conflict resolution. 

While OLAP has always been for monitoring -- so we've treated it as a typical o11y destination, like an opentelemetry collector -- there are a sufficient number of users who treat the OLAP database as the canonical source of truth (either via REST API or UI) that we want to improve guarantees here significantly. 

## Type of change

## What's Changed

Enqueue OLAP messages as part of each task controller tx. Each task controller handler is now (mostly) free of stateful side effects after tx commit, which should reduce the number of potential edge cases significantly. The side effects which run after tx commit should be best-effort, such as notifying downstream schedulers, but should not impact system correctness.

## Checklist

Changes have been:

- [X] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Claude Fable for parts of the implementation and testing. 

</details>


internal/msgqueue/outbox/outbox.go
```diff
@@ -0,0 +1,188 @@
...
+	return "mq." + q.Name()
+}
+
+// Relay drains outbox-staged messages and republishes them to their message queue.
+// Producers stage messages under a queue's topic (see Topic) inside their database
+// transactions; after commit, the relay delivers them through the delegate message
+// queue, so consumers are unaffected by the outbox. Every replica runs a relay —
+// concurrent relays drain disjoint batches.
+type Relay struct {
+	delegate msgqueue.MessageQueue
```
abelanger5: nit: let's rename this, not sure what `delegate` means in this context

internal/services/controllers/olap/controller.go
```diff
@@ -525,32 +525,11 @@ func (tc *OLAPControllerImpl) handleBufferedMsgs(tenantId uuid.UUID, msgId strin
 		return tc.handleFailedWebhookValidation(ctx, tenantId, payloads)
 	case "cel-evaluation-failure":
 		return tc.handleCelEvaluationFailure(ctx, tenantId, payloads)
-	case "offload-payload":
```
abelanger5, line 528: note: deprecated message id, we don't use this anywhere and haven't for months

internal/services/controllers/task/controller.go
```diff
@@ -526,7 +515,7 @@ func (tc *TasksControllerImpl) handleTaskCompleted(ctx context.Context, tenantId
 
 	tc.notifyQueuesOnCompletion(ctx, tenantId, res.ReleasedTasks)
 
-	return tc.signaler.SendInternalEvents(ctx, tenantId, res.InternalEvents)
+	return tc.repov1.Signaler().SendInternalEvents(ctx, tenantId, res.InternalEvents)
```
abelanger5: should internal events not be written from (and signaled via) the transactional enqueue?

internal/services/controllers/task/controller.go
```diff
@@ -556,35 +546,21 @@ func (tc *TasksControllerImpl) handleTaskFailed(ctx context.Context, tenantId uu
...
-			span.RecordError(err)
-			span.SetStatus(codes.Error, "could not publish monitoring event message")
-			continue
-		}
+	// send failed tasks to the olap repository
+	if err := tc.repov1.OLAPOutbox().MonitoringEvents(ctx, tenantId, failedPayloads...); err != nil {
+		tc.l.Error().Ctx(ctx).Err(err).Msg("could not publish monitoring event message")
+		err = fmt.Errorf("could not publish monitoring event message: %w", err)
+		span.RecordError(err)
+		span.SetStatus(codes.Error, "could not publish monitoring event message")
```
abelanger5: should this not be written in the tx?

internal/services/controllers/task/controller.go
```diff
@@ -644,7 +620,7 @@ func (tc *TasksControllerImpl) processFailTasksResponse(ctx context.Context, ten
 	tc.notifyQueuesOnCompletion(ctx, tenantId, res.ReleasedTasks)
 
 	// TODO: MOVE THIS TO THE DATA LAYER?
-	err := tc.signaler.SendInternalEvents(ctx, tenantId, internalEventsWithoutRetries)
+	err := tc.repov1.Signaler().SendInternalEvents(ctx, tenantId, internalEventsWithoutRetries)
```
abelanger5: same comment here

internal/services/controllers/task/controller.go
```diff
@@ -727,7 +703,7 @@ func (tc *TasksControllerImpl) handleTaskCancelled(ctx context.Context, tenantId
 	tc.notifyQueuesOnCompletion(ctx, tenantId, res.ReleasedTasks)
 
 	// TODO: MOVE THIS TO THE DATA LAYER?
-	err = tc.signaler.SendInternalEvents(ctx, tenantId, res.InternalEvents)
+	err = tc.repov1.Signaler().SendInternalEvents(ctx, tenantId, res.InternalEvents)
```
abelanger5: same comment here

internal/services/controllers/task/controller.go
```diff
@@ -736,45 +712,23 @@ func (tc *TasksControllerImpl) handleTaskCancelled(ctx context.Context, tenantId
...
+	}
 
-		if err != nil {
-			tc.l.Error().Ctx(ctx).Err(err).Msg("could not publish monitoring event message")
-			err = fmt.Errorf("could not publish monitoring event message: %w", err)
-			span.RecordError(err)
-			span.SetStatus(codes.Error, "could not publish monitoring event message")
-			outerErr = multierror.Append(outerErr, err)
-		}
+	if pubErr := tc.repov1.OLAPOutbox().MonitoringEvents(ctx, tenantId, cancelledPayloads...); pubErr != nil {
```
abelanger5: same comment here

internal/services/controllers/task/controller.go
```diff
@@ -1215,57 +1113,27 @@ func (tc *TasksControllerImpl) pubRetryEvent(ctx context.Context, tenantId uuid.
...
+			RetryCount:     task.RetryCount,
+			EventType:      sqlcv1.V1EventTypeOlapQUEUED,
+			EventTimestamp: time.Now(),
+		})
+	}
 
-		if err != nil {
-			return fmt.Errorf("could not publish monitoring event message: %w", err)
-		}
+	if err := tc.repov1.OLAPOutbox().MonitoringEvents(ctx, tenantId, retryPayloads...); err != nil {
```
abelanger5: same comment here

internal/services/controllers/task/controller.go
abelanger5, line 1: the general theme of comments in this file is that each task controller handler should do exactly one thing, which is run a single database query which ends with a `commit`. everything after that is a side effect which can fail without impacting data or state machine durability.

internal/services/controllers/task/trigger/trigger.go
```diff
@@ -127,7 +91,7 @@ func (tw *TriggerWriter) TriggerFromWorkflowNames(ctx context.Context, tenantId
 		}
 	}
 
-	tasks, dags, idempotencyKeyCollisions, celEvaluationFailures, err := tw.repo.Triggers().TriggerFromWorkflowNames(ctx, tenantId, opts)
+	_, _, idempotencyKeyCollisions, _, err := tw.repo.Triggers().TriggerFromWorkflowNames(ctx, tenantId, opts)
```
abelanger5: we should audit these fields and see if we depend on them anywhere else; if not, we should simply remove them.

internal/services/controllers/task/trigger/trigger.go
```diff
@@ -72,7 +61,7 @@ func (tw *TriggerWriter) TriggerFromEvents(ctx context.Context, tenantId uuid.UU
 		opts = append(opts, opt)
 	}
 
-	result, err := tw.repo.Triggers().TriggerFromEvents(ctx, tenantId, opts)
+	_, err := tw.repo.Triggers().TriggerFromEvents(ctx, tenantId, opts)
```
abelanger5: we can likely change the signature of TriggerFromEvents as well to only return an error

internal/services/controllers/task/controller.go
```diff
@@ -643,33 +592,7 @@ func (tc *TasksControllerImpl) processFailTasksResponse(ctx context.Context, ten
...
-				err = fmt.Errorf("could not publish retry event: %w", err)
-				span.RecordError(err)
-				span.SetStatus(codes.Error, "could not publish retry event")
-				outerErr = multierror.Append(outerErr, err)
-			}
-		}
-	}
-
-	return outerErr
+	return nil
```
abelanger5: nit: should just remove the return value since we appear to always be returning `nil`

internal/services/controllers/task/controller.go
```diff
@@ -889,53 +765,17 @@ func (tc *TasksControllerImpl) handleReplayTasks(ctx context.Context, tenantId u
 		}
 	}
 
-	eg := &errgroup.Group{}
-
+	// the replay repository stages all OLAP messages on the replay transaction and
+	// runs the non-transactional signaling side effects post-commit
 	for _, tasks := range workflowRunIdToTasks {
-		replayRes, err := tc.repov1.Tasks().ReplayTasks(ctx, tenantId, tasks)
+		_, err := tc.repov1.Tasks().ReplayTasks(ctx, tenantId, tasks)
```
abelanger5: perhaps we should change the signature of ReplayTasks as well, if we don't need the signaling

internal/services/controllers/task/controller.go
```diff
@@ -1129,20 +969,14 @@ func (tc *TasksControllerImpl) processUserEventMatches(ctx context.Context, tena
...
-		err = tc.signaler.SignalTasksCreated(ctx, tenantId, matchResult.CreatedTasks)
-
-		if err != nil {
-			return fmt.Errorf("could not signal created tasks: %w", err)
-		}
-	}
-
 	if len(matchResult.SatisfiedDurableEventLogEntries) > 0 {
 		if err := tc.processSatisfiedEventLogEntry(ctx, tenantId, matchResult.SatisfiedDurableEventLogEntries); err != nil {
 			tc.l.Error().Err(err).Msg("could not process satisfied entries")
```
abelanger5, line 974: add a `FIXME` statement here, we need to make this transactionally safe as well.

internal/services/controllers/task/controller.go
```diff
@@ -1167,28 +1001,14 @@ func (tc *TasksControllerImpl) processInternalEvents(ctx context.Context, tenant
...
-		err = tc.signaler.SignalTasksReplayedFromMatch(ctx, tenantId, matchResult.ReplayedTasks)
-
-		if err != nil {
-			return fmt.Errorf("could not signal replayed tasks: %w", err)
-		}
-	}
-
 	if len(matchResult.SatisfiedDurableEventLogEntries) > 0 {
 		if err := tc.processSatisfiedEventLogEntry(ctx, tenantId, matchResult.SatisfiedDurableEventLogEntries); err != nil {
 			tc.l.Error().Err(err).Msg("could not process satisfied entries")
```
abelanger5, line 1009: same fixme here, this is a side effect which is stateful and can't be dropped, but will need to wait until we add the outbox path for the tasks controller.

pkg/repository/olap_signaler.go
```diff
@@ -0,0 +1,591 @@
...
+}
+
+// notifyScheduler wakes the tenant's scheduler partition so queued tasks are
+// scheduled promptly.
+func (s *OLAPSignaler) notifyScheduler(ctx context.Context, tenantId uuid.UUID, tasks []*V1TaskWithPayload) {
+	if s.pubsub == nil || s.tenant == nil {
+		return
+	}
+
+	tenant, err := s.tenant.GetTenantByID(ctx, tenantId)
```
abelanger5, line 446: should audit deadlock risk here; is there any chance this gets called inside of an OptimisticTx?

internal/msgqueue/outbox/outbox.go
```diff
@@ -0,0 +1,188 @@
...
+
+	for _, m := range msgs {
+		eg.Go(func() error {
+			msg := &msgqueue.Message{}
+
+			if err := json.Unmarshal(m.Payload, msg); err != nil {
+				// drop unparseable messages so they don't wedge the topic; this matches
+				// the rabbitmq consumer, which rejects unparseable bodies without requeue
+				f.l.Error().Err(err).Int64("outbox_message_id", m.ID).Msgf("dropping unparseable outbox message for queue %s", f.queue.Name())
+				return nil
```
mrkaye97, line 174: just confirming - returning nil here because returning an error would cause a retry?

pkg/repository/olap_outbox.go
```diff
@@ -0,0 +1,305 @@
...
+	opts := make([]pgoutbox.MessageOpts, 0, len(msgs))
+
+	for _, msg := range msgs {
+		body, err := json.Marshal(msg)
+
+		if err != nil {
+			return fmt.Errorf("could not marshal message %s: %w", msg.ID, err)
+		}
+
+		opts = append(opts, pgoutbox.MessageOpts{Payload: body})
```
mrkaye97, line 328: there's no risk of overflowing whatever the max size of the `payload` column is right? iirc it's 255mb so that'd be fine assuming that's right

mrkaye97 (commented):
The code changes here look fine to me (great, really - it's amazing how much code we can delete and how much return lifting we can remove 🥳), but a couple of questions at a higher level:

1. What happens if we start falling behind reading messages out of the outbox? I guess it's not that different than falling behind on Rabbit now, but it's a second point of failure potentially? Seems like it's probably not a huge deal but was just wondering
2. This is only sort of "at least once" because there's a risk that we pub to rabbit, delete the message, and then rabbit topples over? (or maybe also something else since we have confirms shut off iirc). Also probably low risk and should at least improve the situation a lot, but isn't a 100% guarantee?

internal/services/controllers/task/trigger/trigger.go
mrkaye97: I'd be in favor of this (and same with the other ones) - no reason to keep returning stuff we don't need IMO

internal/msgqueue/outbox/outbox.go
```diff
@@ -0,0 +1,188 @@
...
+				return f.mq.SendMessage(ctx, f.queue, msg)
+			}
+
+			// PubTenantMessage sends durably and mirrors stream-relevant messages
+			// (e.g. created-task) to the tenant topic for the dispatcher's streams
+			return msgqueue.PubTenantMessage(ctx, f.l, f.mq, f.pubsub, f.queue, msg)
+		})
+	}
+
+	return eg.Wait()
```
igor-kupczynski, line 187: IIUC, the callers retry on any error, but some of the messages might have been sent successfully (e.g., in a batch of 10, one error is enough to retry the whole batch)

Technically this is still at-least-once, but it bothers me. Do you think its rare enough that we can accept it?

hack/benchmark/trigger-bench.sh
```diff
@@ -0,0 +1,427 @@
...
+#
+# trigger-bench.sh — compare task-trigger write latency between two git refs.
+#
+# For each ref, this script:
+#   1. checks the ref out into a temporary git worktree,
+#   2. starts a dedicated postgres container on a random localhost port,
+#   3. runs an embedded-mode benchmark driver (built against that worktree's code)
+#      which measures per-trigger latency (RunNoWait) under concurrency,
+#   4. tears the container down,
+# then prints a side-by-side comparison.
```
igor-kupczynski, line 11: this is really neat

igor-kupczynski (commented):
I start to understand what happens here on the high level and I really like it; need one more more _clean code_ pass.

Helpful visual (though the value was probably in creating it):
```mermaid
sequenceDiagram
    box rgb(255, 243, 205) Ephemeral request state
        participant Owner as Transaction owner
        participant Batch as outboxBatch
    end

    box rgb(225, 238, 255) Long lived engine components
        participant Signaler as OLAPSignaler
        participant Outbox as OLAPOutbox and pgoutbox
        participant Relay as Relay
        participant Flusher as relayFlusher
    end

    box rgb(225, 245, 230) External systems
        participant DB as PostgreSQL
        participant Wakeup as RabbitMQ wakeup exchange
        participant Queue as RabbitMQ durable queue
        participant Consumer as Task consumer
    end

    Note over Relay,Flusher: Engine startup
    Relay->>Outbox: Register relayFlusher for topic
    Relay->>Outbox: Subscribe to topic
    Outbox->>Wakeup: Subscribe to wakeups

    Note over Owner,DB: CompleteTasks transaction
    Owner->>DB: Begin transaction and update task state
    Owner->>Batch: Create batch

    Owner->>Signaler: internalEvents with batch and events
    Signaler->>Signaler: Create internal event message
    Signaler->>Batch: Add message under task processing topic

    Owner->>Outbox: Flush batch using owner transaction
    Outbox->>Batch: Read messages grouped by topic
    Outbox->>DB: Insert outbox rows using same transaction
    Outbox->>Batch: Add post commit wakeup hook

    Owner->>DB: Commit task state and outbox rows
    Note over Owner,DB: State and delivery intent are durable

    Owner->>Batch: Notify
    Batch->>Outbox: Execute topic wakeup hooks
    Outbox->>Wakeup: Publish topic wakeup
    Note over Owner,Batch: Batch lifetime ends

    Wakeup-->>Relay: Wake topic subscription
    Relay->>Outbox: Process available messages
    Outbox->>DB: Acquire outbox rows
    DB-->>Outbox: Return locked messages

    Outbox->>Flusher: Flush acquired batch
    activate Flusher
    Flusher->>Queue: Publish messages
    Queue-->>Consumer: Deliver internal events
    Flusher-->>Outbox: Success
    deactivate Flusher

    Outbox->>DB: Delete delivered outbox rows

```

igor-kupczynski (approved):

PR #4544, opened 2026-07-27
fix: evaluate rate limit CEL expressions with parent outputs

### LINKED ISSUE #4380: [BUG] rate limit CEL expressions referencing `parents` pass registration but fail on every run
**Describe the issue**
Dynamic rate limit CEL expressions that reference `parents` pass workflow registration but then fail on **every** run at trigger time.

Registration-time validation compiles `key_expr` / `units_expr` / `limit_expr` with the `celsteprunstr` validator, which uses the step-run CEL env where `parents` is declared (`internal/cel/cel.go`, `stepRunEnv`). At runtime, however, the same expressions are evaluated through `CELParser.ParseAndEvalStepRun`, which compiles via `ParseWorkflowString` — i.e. against the workflow env, which does **not** declare `parents` (`internal/cel/cel.go:221`; presumably it should call `p.ParseStepRun` instead). The compile error inside `createTasks` (`pkg/repository/task.go:2187`) puts the task straight into `FAILED`.

There is a second layer to this: even with the env fixed, the value bound to `parents` at eval time is `task.Input.TriggerData` — a `*repository.MatchData` with only unexported fields whose values are `TaskOutputEvent` wrappers (`pkg/repository/task.go:2119/2191/2574`), not the `map[step_name]output` shape users see in the step-run context. cel-go's default type adapter can't traverse it, so the variable is unusable as passed. The correct shape is built nearby in `ToV1StepRunData` (`pkg/repository/input.go`) but never reaches the eval.

**Environment**
- SDK: any (validation and evaluation are server-side; snippet below uses Python v1.33.x)
- Engine: Self-hosted, main @ `14c6724` (also applies to v0.90.13)

**Expected behavior**
Registration-time validation and runtime evaluation should agree on the variable set: either both accept `parents` (with a usable parents map bound at eval), or registration rejects the expression up front instead of deferring a guaranteed failure to every run.

```python
from hatchet_sdk import Hatchet, RateLimit, RateLimitDuration, EmptyModel, Context

hatchet = Hatchet()
wf = hatchet.workflow(name="parents-in-rate-limit")

@wf.task()
def step1(input: EmptyModel, ctx: Context) -> dict:
    return {"group": "tenant-a"}

@wf.task(
    parents=[step1],
    rate_limits=[RateLimit(
        dynamic_key="parents.step1.group",  # accepted at registration
        units=1, limit=10, duration=RateLimitDuration.MINUTE,
    )],
)
def step2(input: EmptyModel, ctx: Context) -> None: ...
```
Registration succeeds; on trigger, `step2` immediately fails with an initial-state error of the form:
`failed to parse step expression (parents.step1.group): ERROR: <input>:1:1: undeclared reference to 'parents' (in container '') ...`

**Additional context**
- Root cause pointers: `internal/cel/cel.go:221` (`ParseAndEvalStepRun` compiles via `ParseWorkflowString` instead of `ParseStepRun`); validator registration in `pkg/validator/validator.go` (`celsteprunstr` vs `celworkflowrunstr`); runtime eval sites `pkg/repository/task.go:2115/2187/2570`.
- `internal/cel/cel_test.go` has no coverage for `parents` in any expression kind.
- Related feature request that depends on this same env/value fix: allowing `parents` in task-level concurrency expressions (filed separately).
- Happy to submit a PR for the fix if accepted.

---

<details id="ai-disclosure">

- **Details**: Claude Fable 5 was used to explore the codebase (tracing CEL env definitions, validation tags, and runtime eval call sites) and to draft the initial diagnosis and this write-up; findings and text were reviewed and edited by me.

</details>

jatinder14:
## Summary
- compile step expressions with the step-run CEL environment
- bind parent task outputs as the documented parents map during task creation and replay
- cover parent output access in step-run CEL evaluation

## Test plan
- [x] `go test ./internal/cel`

Fixes #4380

### DIFF
--- a/internal/cel/cel.go
+++ b/internal/cel/cel.go
@@ -218,7 +218,7 @@ func (p *CELParser) ParseStepRun(stepRunExpr string) (cel.Program, error) {
 }
 
 func (p *CELParser) ParseAndEvalStepRun(stepRunExpr string, in Input) (*StepRunOut, error) {
-	prg, err := p.ParseWorkflowString(stepRunExpr)
+	prg, err := p.ParseStepRun(stepRunExpr)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/cel/cel_test.go
+++ b/internal/cel/cel_test.go
@@ -185,7 +185,7 @@ func TestCELParserStepRunExpressionWithParents(t *testing.T) {
 		`parents.step1.group`,
 		cel.NewInput(
 			cel.WithParents(map[string]map[string]interface{}{
-				"step1": {"group": "tenant-a"},
+				"step1": {"group": "parent-group-a"},
 			}),
 		),
 	)
@@ -193,5 +193,5 @@ func TestCELParserStepRunExpressionWithParents(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, cel.StepRunOutTypeString, result.Type)
 	assert.NotNil(t, result.String)
-	assert.Equal(t, "tenant-a", *result.String)
+	assert.Equal(t, "parent-group-a", *result.String)
 }
--- a/internal/cel/cel_test.go
+++ b/internal/cel/cel_test.go
@@ -177,3 +177,21 @@ func TestCELParserEventExpression(t *testing.T) {
 		})
 	}
 }
+
+func TestCELParserStepRunExpressionWithParents(t *testing.T) {
+	parser := cel.NewCELParser()
+
+	result, err := parser.ParseAndEvalStepRun(
+		`parents.step1.group`,
+		cel.NewInput(
+			cel.WithParents(map[string]map[string]interface{}{
+				"step1": {"group": "tenant-a"},
+			}),
+		),
+	)
+
+	assert.NoError(t, err)
+	assert.Equal(t, cel.StepRunOutTypeString, result.Type)
+	assert.NotNil(t, result.String)
+	assert.Equal(t, "tenant-a", *result.String)
+}
--- a/pkg/repository/task.go
+++ b/pkg/repository/task.go
@@ -2292,7 +2292,7 @@ func (r *sharedRepository) insertTasks(
 						cel.WithInput(task.Input.Input),
 						cel.WithAdditionalMetadata(additionalMeta),
 						cel.WithWorkflowRunID(task.ExternalId),
-						cel.WithParents(task.Input.TriggerData),
+						cel.WithParents(r.ToV1StepRunData(task.Input).Parents),
 					))
 
 					if err != nil {
@@ -2364,7 +2364,7 @@ func (r *sharedRepository) insertTasks(
 						cel.WithInput(task.Input.Input),
 						cel.WithAdditionalMetadata(additionalMeta),
 						cel.WithWorkflowRunID(task.ExternalId),
-						cel.WithParents(task.Input.TriggerData),
+						cel.WithParents(r.ToV1StepRunData(task.Input).Parents),
 					))
 
 					if err != nil {
@@ -2769,7 +2769,7 @@ func (r *sharedRepository) replayTasks(
 						cel.WithInput(task.Input.Input),
 						cel.WithAdditionalMetadata(additionalMeta),
 						cel.WithWorkflowRunID(task.ExternalId),
-						cel.WithParents(task.Input.TriggerData),
+						cel.WithParents(r.ToV1StepRunData(task.Input).Parents),
 					))
 
 					if err != nil {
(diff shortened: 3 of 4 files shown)


pkg/repository/task.go
```diff
@@ -2292,7 +2292,7 @@ func (r *sharedRepository) insertTasks(
 						cel.WithInput(task.Input.Input),
 						cel.WithAdditionalMetadata(additionalMeta),
 						cel.WithWorkflowRunID(task.ExternalId),
-						cel.WithParents(task.Input.TriggerData),
+						cel.WithParents(r.ToV1StepRunData(task.Input).Parents),
```
gregfurman: While I think this would work, I'm unsure we should be relying on converting task input to that V1 type.

Wdyt about a standalone function that returns `Parents` data as a `map[string]map[string]any` e.g

```go
func toParentsData(t *TaskInput) map[string]map[string]any {
	parents := make(map[string]map[string]interface{})

	if t == nil || t.TriggerData == nil {
		return parents
	}

	for _, stepReadableId := range t.TriggerData.DataKeys() {
		data := t.TriggerData.DataValueAsTaskOutputEvent(stepReadableId)

		if data == nil || !data.IsCompleted() {
			continue
		}

		dataMap := make(map[string]interface{})

	   
       if err := json.Unmarshal(data.Output, &dataMap); err != nil {
            // probably just ignore errors, else we can log them if this is a method
			continue
		}

		parents[stepReadableId] = dataMap
	}

	return parents
}
```

gregfurman (commented):
Think this would work! Just a suggestion on creating a standalone helper function here for this conversion.

internal/cel/cel.go
```diff
@@ -218,7 +218,7 @@ func (p *CELParser) ParseStepRun(stepRunExpr string) (cel.Program, error) {
 }
 
 func (p *CELParser) ParseAndEvalStepRun(stepRunExpr string, in Input) (*StepRunOut, error) {
-	prg, err := p.ParseWorkflowString(stepRunExpr)
+	prg, err := p.ParseStepRun(stepRunExpr)
```
gregfurman, line 221: Why the change here?

pkg/repository/task.go
jatinder14: Done in `c9d2f7a` — agreed, routing through the V1 type was the wrong dependency.

Added `toParentsData` on `sharedRepository`, close to your sketch, and swapped all three CEL call sites (`insertTasks` x2, `replayTasks`) over to it. Keeping it a method so the unmarshal failure can go through `s.l.Warn()` instead of being dropped silently.

Two side benefits: it skips building the `triggers` map that these call sites never read, and it's nil-safe — `ToV1StepRunData(nil)` returns nil, so the old `.Parents` would have panicked on a nil `TaskInput`.

Left `ToV1StepRunData` untouched. It needs `data.IsFailed()` from the same loop for `StepRunErrors`, and `DataValueAsTaskOutputEvent` marshals and unmarshals per key, so having it call the helper would mean paying for a second pass over `DataKeys()`.

internal/cel/cel.go
jatinder14: It's load-bearing for this fix, not a drive-by.

`parents` is only declared on `stepRunEnv`:

```go
stepRunEnv, _ := cel.NewEnv(
	...
	decls.NewVar("parents", decls.NewMapType(decls.String, decls.NewMapType(decls.String, decls.Dyn))),
)
```

`workflowStrEnv` has no such declaration, so `ParseAndEvalStepRun` compiling through `ParseWorkflowString` meant any expression referencing `parents.*` failed at compile time with an undeclared-reference error — no amount of populating the input would have helped. That's the actual bug behind rate limit keys that read from a parent step.

Every caller of `ParseAndEvalStepRun` is a rate limit or concurrency expression (`task.go` 2022, 2291, 2363, 2768), all of which are step-run scoped, so `stepRunEnv` is the environment they should always have been compiled against. `ParseAndEvalWorkflowString` still uses `ParseWorkflowString` and is unaffected.

`TestCELParserStepRunExpressionWithParents` in `cel_test.go` covers it — it fails on the old line and passes on the new one.

internal/cel/cel_test.go
```diff
@@ -177,3 +177,21 @@ func TestCELParserEventExpression(t *testing.T) {
...
 }
+
+func TestCELParserStepRunExpressionWithParents(t *testing.T) {
+	parser := cel.NewCELParser()
+
+	result, err := parser.ParseAndEvalStepRun(
+		`parents.step1.group`,
+		cel.NewInput(
+			cel.WithParents(map[string]map[string]interface{}{
+				"step1": {"group": "tenant-a"},
```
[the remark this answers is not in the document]
jatinder14: Fixed in `7cfd9fa` — `tenant-a` → `parent-group-a` in both the fixture and the assertion. `AGENTS.md` is explicit about it and I missed it:

> Do not include customer names, tenant names, or production namespace names in test names, fixtures, comments, or logs. Describe the scenario by structural properties instead

`parent-group-a` describes what the value actually is here: the `group` field of a parent step's output, which is what the rate limit expression reads.

PR #4550, opened 2026-07-28
feat: add NATS pub/sub backend

igor-kupczynski:
# Description

> [!NOTE]
>This is an experimental prototype. We are not launching this as a feature yet; we need to gain more experience running this internally first.

Adds NATS as an opt-in pub/sub backend, set:
- `SERVER_MSGQUEUE_PUBSUB_KIND=nats`
- `SERVER_MSGQUEUE_PUBSUB_NATS_URL=...`

It covers only the best-effort pub/sub side of the #4480 split: core NATS, at-most-once. Durable queues stay on RabbitMQ/Postgres.

The defaults are unchanged, NATS remain optional, until the config is set.

## Type of change

## What's Changed

`internal/msgqueue/nats` plus config, docs, and a compose service.

Technicalities:
- Connect is synchronous and refuses to start if the server's `max_payload` is under the 16MiB. 16MiB is what RabbitMQ allows, so as a drop in replacement, we want to keep this.
- `SERVER_MSGQUEUE_PUBSUB_NATS_SUBJECT_PREFIX` (default `hatchet.pubsub`) for isolating hatchet installations sharing a NATS server (not unlike like a RabbitMQ vhost or Postgres database).

## Checklist

Changes have been:

- [x] Linted and formatted

## Testing

```bash
docker compose up -d nats
go test -tags integration ./internal/msgqueue/nats/...
go test ./internal/msgqueue/...
```

---

<details id="ai-disclosure">

- **Details**: Cursor agent implemented the backend, config/docs/compose wiring, and tests from a reviewed plan.

</details>


internal/msgqueue/nats/pubsub.go
```diff
@@ -0,0 +1,262 @@
...
+	natsgo "github.com/nats-io/nats.go"
+	"github.com/rs/zerolog"
+
+	"github.com/hatchet-dev/hatchet/internal/msgqueue"
+	"github.com/hatchet-dev/hatchet/pkg/logger"
+)
+
+// requiredMaxPayload is the minimum NATS server max_payload we accept.
+// The durable RabbitMQ path allows 16MiB, so the NATS server must match.
+const requiredMaxPayload = 16 * 1024 * 1024
```
mrkaye97: this makes me a bit nervous since I think this can e.g. change across rabbit versions (pretty sure we ran into this)

internal/msgqueue/nats/pubsub.go
```diff
@@ -0,0 +1,262 @@
...
+
+	if opts.url == "" {
+		return nil, nil, fmt.Errorf("nats pubsub requires a URL to be set")
+	}
+
+	l := opts.l
+
+	connectOpts := []natsgo.Option{
+		natsgo.MaxReconnects(-1),    // reconnect indefinitely
+		natsgo.ReconnectBufSize(-1), // do not buffer pubs during disconnect
```
mrkaye97: should these be configurable?

internal/msgqueue/nats/pubsub.go
```diff
@@ -0,0 +1,262 @@
...
+			if sub != nil {
+				subject = sub.Subject
+			}
+			l.Error().Err(err).Str("subject", subject).Msg("nats pubsub async error")
+		}),
+	}
+
+	if opts.username != "" || opts.password != "" {
+		connectOpts = append(connectOpts, natsgo.UserInfo(opts.username, opts.password))
+	}
```
mrkaye97: do these need to both be non-empty? or just one? I guess maybe it doesn't matter since the connect after this will fail if they're wrong

mrkaye97 (approved):

internal/msgqueue/nats/pubsub.go
igor-kupczynski: Good catch. Our rabbitmq backend hardcodes its own client-side `maxPayloadSize = 16MiB` and never reads the broker's limit. Fixed in 830c4d6 by adding a single shared `msgqueue.MaxMessageSize` used by rabbit and NATS. If the contract ever changes, at least this will happen in one place.

internal/msgqueue/nats/pubsub.go
igor-kupczynski: I don't think either makes sense as a tunable config:

- MaxReconnects: any finite value permanently closes the connection once exhausted, leaving the engine without pub/sub until a process restart (so on a NATS outage, if the broker recovers after the MaxReconnects we'd never know). The rabbitmq backend also retries forever.
- ReconnectBufSize(-1): with a buffer, publishes during a disconnect get queued and flushed on reconnect. But these signals are latency optimizations with polling paths as fallbacks; by the time a buffered publish flushed, polling would have likely covered it.

Documented the rationale inline in 830c4d6 as comments so it doesn't surprise the readers.

internal/msgqueue/nats/pubsub.go
igor-kupczynski: Thanks for raising. NATS marshals this with `omitempty`, so UserInfo("", "") is the same as not setting it at all. Empty credentials never reach the broker.

In 830c4d6 I deleted the conditional and pass UserInfo unconditionally. Whether a lone username (or password) is acceptable is the server's call, and it fails loudly at the synchronous Connect on startup, as you said.

Issue #4204, opened 2026-06-15
CI Health Dashboard


igor-kupczynski:
# CI Health Dashboard

_Window: last 14 days (trend + pass rate) · tables: last 24h · updated 2026-09-04T07:12:04Z · auto-generated, do not edit by hand._

**Gating-CI pass rate** — PR: 85% (2309/2730) · main: 83% (138/166)

## Gating-CI pass-rate trend

```mermaid
xychart-beta
  title "Gating-CI pass rate (%) per day"
  x-axis [21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31, 1, 2, 3, 4]
  y-axis "pass rate %" 0 --> 100
  line "CI" [82, 73, 93, 88, 93, 88, 92, 87, 100, 71, 86, 78, 88, 76, 68]
  line "main" [76, 100, 100, 81, 100, 88, 100, 62, 62, 62, 90, 91, 71, 76, 100]
```

_X-axis = day of month (Aug 21 → Sep 04). Two lines: **CI** (PR gating-CI runs, generally the upper line) and **main** (post-merge main runs, lower). Y-axis = % of that day's gating-CI runs that passed._

## Top 10 failing jobs (last 24h)

| # | job | workflow | fails | recovered | runs | fail rate | flaky? | scope | cause |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 1 | `build` | frontend / docs | 9 | 0 | 24 | 38% | flaky | PR | **infra/CI** — frontend/docs prettier drift on migration-guide-python-v2.mdx |
| 2 | `lint` | ruby | 8 | 0 | 28 | 29% | flaky | PR | **infra/CI** — Ruby protobuf/REST generated bindings out of date |
| 3 | `unit` | test | 7 | 0 | 38 | 18% | flaky | main + PR | **product bug** — duplicate dagoperator methods block test/unit compile |
| 4 | `test-templates` | cli-e2e-tests | 6 | 0 | 14 | 43% | flaky | main + PR | **flaky test** — CLI TestQuickstartTemplates parent E2E template timing |
| 5 | `integration` | test | 6 | 0 | 38 | 16% | flaky | main + PR | **product bug** — duplicate dagoperator methods block test/integration generate |
| 6 | `generate` | test | 6 | 0 | 38 | 16% | flaky | PR | **product bug** — test/generate fails golangci-lint on duplicate dagoperator methods |
[...]

PR #3846, opened 2026-05-07
Add Dockerfile for all SDK e2e tests, fix flaky e2e tests

juliusgeo:
# Description

Adds a built Dockerfile for the go e2e tests.

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation change (pure documentation change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking changes to code which doesn't change any behaviour)
- [ ] CI (any automation pipeline changes)
- [ ] Chore (changes which are not directly related to any business logic)
- [ ] Test changes (add, refactor, improve or change a test)
- [ ] This change requires a documentation update

## What's Changed

- [ ] Add a list of tasks or features here...

PR #4559, opened 2026-07-29
test(python): anchor test_waits on start completing

igor-kupczynski:
# Description

`examples/conditions/test_conditions.py::test_waits` is the flakiest test in the repo — **10 fails / 32 runs (31%)** per the CI Health Dashboard (#4204). It fails as:

```
assert result["skip_on_event"] == {"skipped": True}
AssertionError: assert {'random_number': 17} == {'skipped': True}
```

**Why it fails.** `skip_on_event` has `parents=[start]`, `wait_for=[SleepCondition(30s)]`, `skip_if=[UserEventCondition("skip_on_event:skip")]`. Because it has wait_for/skip_if conditions, its parent-complete condition is registered with action `CREATE_MATCH` rather than `QUEUE` (`getParentInDAGGroupMatch`), so at trigger time neither the sleep nor the user-event match exists. Both are created only when `start` completes, in one transaction, with `sleep_until = CURRENT_TIMESTAMP + 30s`. So the skip event only counts during the 30s after `start` completes — pushed earlier it matches nothing and is dropped, pushed later the sleep has fired and the task ran.

The test measured its 15s pre-push delay from `wait_for_running_status`, which waits for the *run* to report RUNNING. But a run only reads RUNNING while some task holds a `v1_task_runtime` row, and every task here other than `start` is parked on a wait condition — so RUNNING is observable for a few milliseconds at a time (`start` at the very beginning, then `wait_for_sleep`/a branch near the 10s mark; in between the run reads QUEUED). That wait commonly returned ~10s in, the test then slept its full 15s, and the push landed ~25s into a 30s window: **~5s of margin**, which CI load eats. Because sustained load rather than chance drives it, all four retry attempts fail together with a different random number each time — exactly the observed signature.

**Regression history.** This was introduced by #3846 (May 2026), which inserted `wait_for_running_status` *in front of* a pre-existing `sleep(15)` that had been counted from run creation. The added wait can only delay the push, so it strictly reduced the late-side margin. Worth noting the test was already racy before, in the opposite direction: a fixed 15s from run creation risked arriving *before* `start` completed, which drops the event and fails identically. #3846 appears to have been fixing that early-side race and overcorrected. Hence the structural fix here rather than a retuned constant — any single fixed delay is exposed to one side or the other.

## Type of change

## What's Changed

- Anchor on the window's own opening edge: poll `runs.aio_get` until the `start` task reads `COMPLETED`, then push. Unlike the gRPC `aio_get_details`, that REST endpoint exposes live per-task status mid-run (prior art: `examples/on_failure/test_on_failure.py`). Margin goes from ~5s to ~28s and can't drift.
- Push twice, 2s apart: the read model can surface `start` as `COMPLETED` marginally before the engine commits the match registration, and a too-early push fails identically to a too-late one.
- Drop `wait_for_running_status` and both fixed sleeps from this test. All seven original assertions are unchanged.

Not touched on purpose: `examples/conditions/worker.py` (published docs snippets), `test_utils.py`, and the engine.

**Follow-up, not in this PR:** `wait_for_running_status` in `examples/test_utils.py` silently returns when its timeout expires instead of raising — its `for _ in range(max_iters)` loop just falls off the end. That turns "the run never started" into a confusing assertion failure much later. It also counts iterations rather than tracking a deadline, so a nominal 60s becomes longer under load. Three other tests still use it; left alone here to keep this PR's blast radius at zero.

## Checklist

Changes have been:

- [x] Tested (black, ruff, mypy, pydoclint all clean locally; the e2e itself runs in CI)
- [x] Linted and formatted
- [ ] Documented (n/a)
- [ ] Added to CHANGELOG (n/a — test-only)

## Testing

The e2e needs docker compose plus a live engine/API/worker, so it runs in CI here. A single local pass couldn't demonstrate a load-dependent flake is gone anyway; the argument for this change is the margin going from ~5s to ~28s and the anchor no longer being a proxy that sits ~10s downstream of the thing it's supposed to track.

Worth watching in review: `t.display_name.startswith("start-")` relies on a DAG task's display name being `<step_readable_id>-<unix>` (`pkg/repository/task.go`, `fmt.Sprintf("%s-%d", stepConfig.ReadableId.String, unix)`) — no namespace or workflow prefix.

---

<details id="ai-disclosure">

- **Details**: Cursor agent diagnosed the flake against the engine source, then implemented and reviewed the test change from a plan I approved.

</details>

### DIFF
--- a/sdks/python/examples/conditions/test_conditions.py
+++ b/sdks/python/examples/conditions/test_conditions.py
@@ -1,22 +1,49 @@
 import asyncio
+import time
 
 import pytest
 
 from examples.conditions.worker import task_condition_workflow
-from examples.test_utils import wait_for_running_status
-from hatchet_sdk import Hatchet
+from hatchet_sdk import Hatchet, V1TaskStatus
+
+
+async def _wait_for_start_to_complete(
+    hatchet: Hatchet, workflow_run_id: str, timeout: float = 60.0
+) -> None:
+    interval = 0.5
+    deadline = time.monotonic() + timeout
+
+    while time.monotonic() < deadline:
+        details = await hatchet.runs.aio_get(workflow_run_id)
+
+        # a task's display name is "<step readable id>-<unix timestamp>"
+        if any(
+            t.status == V1TaskStatus.COMPLETED and t.display_name.startswith("start-")
+            for t in details.tasks
+        ):
+            return
+
+        await asyncio.sleep(interval)
+
+    raise TimeoutError(
+        f"start did not complete within {timeout}s for run {workflow_run_id}"
+    )
 
 
 @pytest.mark.asyncio(loop_scope="session")
 async def test_waits(hatchet: Hatchet) -> None:
     ref = task_condition_workflow.run(wait_for_result=False)
 
-    await wait_for_running_status(hatchet, ref.workflow_run_id)
-    await asyncio.sleep(15)
+    # skip_on_event's skip match and its 30s sleep are both registered when start
+    # completes, so the skip event only counts during the 30s after that.
+    await _wait_for_start_to_complete(hatchet, ref.workflow_run_id)
 
-    hatchet.event.push("skip_on_event:skip", {})
-    hatchet.event.push("wait_for_event:start", {})
-    await asyncio.sleep(5)
+    # Push twice: the read model can show start COMPLETED just before the engine
+    # registers the match, and an unmatched event is dropped silently.
+    for _ in range(2):
+        await hatchet.event.aio_push("skip_on_event:skip", {})
+        await hatchet.event.aio_push("wait_for_event:start", {})
+        await asyncio.sleep(2)
 
     result = await ref.aio_result()
 


sdks/python/examples/conditions/test_conditions.py
```diff
@@ -1,22 +1,49 @@
...
+
+
+async def _wait_for_start_to_complete(
+    hatchet: Hatchet, workflow_run_id: str, timeout: float = 60.0
+) -> None:
+    interval = 0.5
+    deadline = time.monotonic() + timeout
+
+    while time.monotonic() < deadline:
+        details = await hatchet.runs.aio_get(workflow_run_id)
```
mrkaye97, line 17: I think we'll want to use `hatchet.runs.aio_get_details` here (in v2 we'll name these better b/c it's confusing, but that one goes over gRPC to the core db so it's not prone to race conditions caused by replicating data to the OLAP db)

mrkaye97 (approved):

sdks/python/examples/conditions/test_conditions.py
igor-kupczynski: Hey, I think aio_get_details won't work here (this is `GetWorkflowRunResultDetails`) and it only surfaces QUEUED/RUNNING/EVICTED mid-run. COMPLETED is applied only after `ListFinalizedWorkflowRuns`, which requires the whole DAG to be finalized.

If I'm misreading the code I'll raise a followup, hope that's OK.

PR #4564, opened 2026-07-29
chore: remove managed worker compute cost UI

grutt:
## Summary

- Removes the Managed Compute monthly usage card and `getComputeCost` query.
- Updates the generated cloud API client after the cloud compute-cost endpoint / managed worker meters OLAP path were removed.
- Billing-required page no longer fetches credits from that endpoint.

Depends on the cloud PR that drops `/api/v1/cloud/tenants/{tenant}/managed-worker/cost`.

## Test plan

- [ ] Managed Compute list page loads without the Monthly Usage card
- [ ] Billing-required gate still renders pricing + billing CTA
- [ ] No remaining references to `computeCostGet` / `MonthlyComputeCost`


mrkaye97 (approved):
can we just remove the feature entirely.....

grutt:
> can we just remove the feature entirely.....

soon

PR #4565, opened 2026-07-29
fix: buffer and coalesce RefreshTimeout updates

grutt:
## Summary
- Hot-looping `refreshTimeout` callers can stampede `RefreshTimeoutBy` against a single `v1_task_runtime` row, creating large `Lock:transactionid` / `Lock:tuple` load on shared Postgres.
- Buffer `IncrementTimeoutBy` per task in-memory on each engine, sum pending increments, and flush at most once every 500ms with a single UPDATE.
- Concurrent callers coalesce via `singleflight`; a trailing flush applies increments that arrive during the debounce window so extensions are preserved rather than dropped.
- Per-engine buffering is enough — a few writes/sec across engines is still much better than hundreds of contended row updates.

## Test plan
- [x] `go test ./internal/services/dispatcher/ -run TestRefreshTimeoutBuffer`
- [x] `go build ./internal/services/dispatcher/`
- [ ] Spam `RefreshTimeout` for one task and confirm ~2 DB updates/sec/engine with `timeout_at` extended by the summed increments
- [ ] Confirm a normal single refresh still succeeds and updates `timeout_at`
- [ ] After deploy, check Performance Insights / `pg_stat_activity` for `RefreshTimeoutBy` lock waits under previously noisy tenants

### DIFF
--- a/internal/services/dispatcher/server.go
+++ b/internal/services/dispatcher/server.go
@@ -1848,17 +1848,70 @@ func (d *DispatcherImpl) refreshTimeoutV1(ctx context.Context, tenant *sqlcv1.Te
 		return nil, status.Errorf(codes.InvalidArgument, "Invalid request: %s", apiErrors.String())
 	}
 
+	increment, err := time.ParseDuration(request.IncrementTimeoutBy)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid increment timeout by %s: %v", request.IncrementTimeoutBy, err)
+	}
+
 	cacheKey := fmt.Sprintf("refresh-timeout:%s:%s", tenantId, taskExternalId)
+	d.refreshTimeoutBuf.add(cacheKey, tenantId, taskExternalId, increment)
 
 	timeoutAt, err, _ := d.refreshTimeoutGroup.Do(cacheKey, func() (interface{}, error) {
-		if cached, ok := d.cache.Get(cacheKey); ok {
-			return cached.(time.Time), nil
+		return d.flushRefreshTimeout(ctx, cacheKey)
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return &contracts.RefreshTimeoutResponse{
+		TimeoutAt: timestamppb.New(timeoutAt.(time.Time)),
+	}, nil
+}
+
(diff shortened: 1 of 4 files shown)


internal/services/dispatcher/server.go
```diff
@@ -1848,40 +1848,94 @@ func (d *DispatcherImpl) refreshTimeoutV1(ctx context.Context, tenant *sqlcv1.Te
...
-		tasktypes.CreateMonitoringEventPayload{
-			TaskId:         taskRuntime.TaskID,
-			RetryCount:     taskRuntime.RetryCount,
-			WorkerId:       workerId,
-			EventTimestamp: time.Now(),
-			EventType:      sqlcv1.V1EventTypeOlapTIMEOUTREFRESHED,
-			EventMessage:   fmt.Sprintf("Timeout refreshed by %s", request.IncrementTimeoutBy),
-		},
-	)
+	cacheKey := fmt.Sprintf("refresh-timeout:%s:%s", tenantId, taskExternalId)
```
mrkaye97: struct for this instead of magic string?

internal/services/dispatcher/server.go
```diff
@@ -1848,40 +1848,94 @@ func (d *DispatcherImpl) refreshTimeoutV1(ctx context.Context, tenant *sqlcv1.Te
...
-	if err != nil {
-		return nil, err
+func (d *DispatcherImpl) flushRefreshTimeout(ctx context.Context, cacheKey string) (time.Time, error) {
+	if !d.refreshTimeoutBuf.shouldFlush(cacheKey) {
+		if timeoutAt, ok := d.refreshTimeoutBuf.lastTimeout(cacheKey); ok {
+			d.refreshTimeoutBuf.scheduleTrailingFlush(cacheKey, func(key string) {
+				flushCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+				defer cancel()
+
+				_, err, _ := d.refreshTimeoutGroup.Do(key, func() (interface{}, error) {
```
mrkaye97: looks like we call this from inside here and also from line 1859, which then calls this method - is that a problem?

mrkaye97 (approved):

internal/services/dispatcher/server.go
grutt: this is ok

PR #4571, opened 2026-07-30
fix(go): preserve result subscriptions across reconnect

igor-kupczynski:
# Description

Context: On the client, during reconnect, we have basically two stream: the old (now broken), the new (the one being established). Let's call this process a "handoff" from the old stream to the new stream. Concretely (in `connectOnce`) the following steps happen during the handoff:

1. Replay. Re-send every subscription currently in the registry onto the new stream, so the server knows about them on the new connection .
2. Publish. Swap the `client` field to the new stream and bump generation, so every future `snapshot()` returns the new stream.  [until this completes, the `snapshot()` returns the old stream].
3. Retire. `CloseSend` the old stream.

The bug is that 1 and 2 are not on the same critical section: if a new registration happened after 1, but before 2, it would be registered on the old broken stream (so therefore wouldn't happen).

This is a minimal fix that brings all 3 steps under a single `sendMu` mutex. (The side effect is that the registrations are stalled during the handoff, but it's the right tradeoff here).

## Type of change

## What's Changed

- [x] Serialize replay and stream installation with workflow-result registration sends
- [x] Add a deterministic regression for registration during reconnect handoff

## Testing

I've iterated a bit on this, and settled on `pkg/client/listener_reconnect_invariant_test.go`. It tests the invariant not the mechanism so even if it +300 it's worth adding. The invariant under test: **every `AddWorkflowRun` that returns no error has its run ID sent on the valid stream**.

Two tests:

1. `TestAddWorkflowRunDuringReconnectLandsOnReplacement` targets the bug.
2. `TestWorkflowRunsListenerReconnectChaosPreservesRegistrations` chaos testing.

Both tests were run against the parent commit (red) and fix (green).

## Checklist

Changes have been:

- [x] Linted and formatted
- [ ] Documented (not applicable; no public API change)
- [ ] Added to CHANGELOG (not applicable)

---

<details id="ai-disclosure">

- **Details**: Used Cursor agents to investigate the concurrency interleaving, implement the focused fix and regression test, and review the resulting diff. All listed tests and linters were run locally.

</details>


pkg/client/reconnecting_stream.go
```diff
@@ -104,6 +105,14 @@ func (s *reconnectingStream[C]) setInitialClient(client C) {
 }
 
 func (s *reconnectingStream[C]) installClient(client C) error {
+	s.sendMu.Lock()
```
juliusgeo: should this function be removed?

igor-kupczynski: GC no production callers; its used in one of the existing tests though -- let me see what that would look like if we drop it

igor-kupczynski: That looks OK: 8a4f0e56d

juliusgeo (approved):

PR #4576, opened 2026-07-30
feat: `tenant_queued_to_assigned` broken down by workflow name, queue depth metric

abelanger5:
# Description

Adds several new Prometheus metrics:

| Metric Name                                      | Type      | Description                                                                          |
| ------------------------------------------------ | --------- | ------------------------------------------------------------------------------------ |
| `hatchet_tenant_queued_to_assigned_by_workflow`  | Counter   | The total number of unique tasks that were queued and later got assigned to a worker, by workflow name |
| `hatchet_tenant_queued_to_assigned_time_seconds_by_workflow` | Histogram | Buckets of time in seconds spent in the queue before being assigned to a worker, by workflow name |
| `hatchet_tenant_queue_size`                      | Gauge     | The current number of queued items, by queue and workflow name                       |
| `hatchet_tenant_additional_metadata_queue_size`  | Gauge     | The current number of queued items, by additional metadata key-value pair            |

## Type of change

## What's Changed

New metrics, see above. 

## Checklist

Changes have been:

- [X] Linted and formatted

---

<details id="ai-disclosure">

</details>


frontend/docs/pages/self-hosting/prometheus-metrics.mdx
```diff
@@ -229,6 +233,19 @@ The `slot_type` label separates the worker's slot pools (e.g. `default` and `dur
   querying.
 </Callout>
 
+The `hatchet_tenant_queued_to_assigned_by_workflow` counter and the `hatchet_tenant_queued_to_assigned_time_seconds_by_workflow` histogram are per-workflow variants of the queued-to-assigned metrics, carrying a `workflow_name` label so queue latency can be attributed to individual workflows.
+
+The queue size gauges are polled from the database every 15 seconds, so they reflect the backlog with up to that much delay. `hatchet_tenant_queue_size` reports the number of queued items per `queue` and `workflow_name`, and is safe to sum. `hatchet_tenant_additional_metadata_queue_size` reports the number of queued items carrying each additional metadata `(key, value)` pair, using the `key` and `value` Prometheus labels; only scalar metadata values (strings, numbers, booleans) are reported. Like the worker label slot metrics, an item counts towards every metadata key it carries, so do not sum this gauge across different `key` values. Items queued behind a concurrency strategy are not counted until they enter the queue.
```
abelanger5: should be condensed and added to notes in the description of the relevant metrics, instead of dumped at the bottom

frontend/docs/pages/self-hosting/prometheus-metrics.mdx
```diff
@@ -229,6 +233,19 @@ The `slot_type` label separates the worker's slot pools (e.g. `default` and `dur
...
+
+<Callout type="warning">
+  Every distinct additional metadata `(key, value)` pair on a queued task
+  creates its own Prometheus series. Avoid high-cardinality metadata values
+  (e.g. per-run identifiers) on workflows you run at volume, or your metrics
+  storage will pay for it.
+</Callout>
+
+When a queue drains, its series report `0` on the next poll and are removed shortly after, so autoscalers observe the drop to zero. Guard queries against the series being absent (e.g. `sum(...) or vector(0)`) if you scale to zero.
```
abelanger5: just drop this

frontend/docs/pages/v1/autoscaling-workers.mdx
```diff
@@ -204,3 +204,13 @@ sum by (tenant_id) (avg_over_time(hatchet_tenant_used_worker_slots{tenant_id="<t
 /
 sum by (tenant_id) (avg_over_time(hatchet_tenant_worker_slots{tenant_id="<tenant-id>"}[5m]))
 ```
+
+### Scaling Based on Queue Backlog by Metadata Tag
+
+To scale on queue depth via Prometheus instead of the Task Stats API, Hatchet exposes queue size gauges, including one grouped by the additional metadata `(key, value)` pairs on queued tasks. If each of your products tags its workflow runs with additional metadata (e.g. `product: search`), each product's backlog can drive its own scaler:
+
+```promql
+sum(hatchet_tenant_additional_metadata_queue_size{tenant_id="<tenant-id>", key="product", value="search"}) or vector(0)
```
abelanger5: should use a consistent example, as above - i.e. gpus

pkg/repository/sqlcv1/queue.sql
```diff
@@ -417,6 +417,57 @@ WHERE
...
+)
+SELECT
+    s.tenant_id,
+    s.key::text AS key,
+    s.value::text AS value,
+    s.count::bigint AS count
+FROM
+    sizes s
+WHERE
+    s.tenant_id = ANY(@tenantIds::uuid[]);
```
abelanger5: note: these are known seq scan risks; there's no other way to do it, and we effectively do the same thing on the `task-stats` endpoint in the worst case. it would be nice to materialize this into a view, which will add a small write overhead but probably worth the tradeoff. not an immediate priority though.

internal/services/scheduler/v1/queue_metrics.go
```diff
@@ -0,0 +1,214 @@
...
+	for _, row := range sizes {
+		sizesByTenant[row.TenantID] = append(sizesByTenant[row.TenantID], row)
+	}
+
+	metadataByTenant := make(map[uuid.UUID][]*sqlcv1.GetQueueSizesByMetadataRow)
+	for _, row := range metadata {
+		metadataByTenant[row.TenantID] = append(metadataByTenant[row.TenantID], row)
+	}
+
+	active := make(map[uuid.UUID]struct{}, len(tenantIds))
```
mrkaye97: naming nit: `activeTenants`?

pkg/repository/sqlcv1/queue.sql
```diff
@@ -417,6 +417,57 @@ WHERE
...
+    SELECT
+        qi.tenant_id,
+        qi.queue,
+        qi.workflow_id,
+        COUNT(*) AS count
+    FROM
+        v1_queue_item qi
+    GROUP BY
+        qi.tenant_id, qi.queue, qi.workflow_id
+)
```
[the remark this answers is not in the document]
mrkaye97: I imagine you're thinking we need to seq scan either way? but for small sets of tenant ids, I agree with this? putting the tenant here will pick up that index at least which should help limit how much stuff we need to scan

pkg/repository/sqlcv1/queue.sql
```diff
@@ -417,6 +417,57 @@ WHERE
...
+SELECT
+    s.tenant_id,
+    s.queue,
+    COALESCE(w."name", '')::text AS workflow_name,
+    SUM(s.count)::bigint AS count
+FROM
+    sizes s
+LEFT JOIN "Workflow" w ON w."id" = s.workflow_id
```
mrkaye97: inner join here? `workflow_id` is non-null

pkg/repository/sqlcv1/queue.sql
```diff
@@ -417,6 +417,57 @@ WHERE
...
+        COUNT(*) AS count
+    FROM
+        v1_queue_item qi
+    GROUP BY
+        qi.tenant_id, qi.queue, qi.workflow_id
+)
+SELECT
+    s.tenant_id,
+    s.queue,
+    COALESCE(w."name", '')::text AS workflow_name,
```
mrkaye97: don't think we need to coalesce if we inner join

pkg/repository/sqlcv1/queue.sql
```diff
@@ -417,6 +417,57 @@ WHERE
...
+WITH sizes AS MATERIALIZED (
+    SELECT
+        qi.tenant_id,
+        kv.key AS key,
+        (kv.value #>> '{}') AS value,
+        COUNT(*) AS count
+    FROM
+        v1_queue_item qi
+    JOIN v1_task t ON t.id = qi.task_id AND t.inserted_at = qi.task_inserted_at
+    CROSS JOIN LATERAL jsonb_each(t.additional_metadata) AS kv(key, value)
```
mrkaye97: this seems like a bad idea....

mrkaye97: Ran it against Cloud: https://explain.dalibo.com/plan/gc3h7afb638fd2d3

pkg/scheduling/v1/workflow_names_test.go
```diff
@@ -0,0 +1,85 @@
+package v1
```
mrkaye97: I'm not convinced we need these tests

pkg/repository/sqlcv1/queue.sql
```diff
@@ -417,6 +417,57 @@ WHERE
...
+    WHERE
+        jsonb_typeof(kv.value) IN ('string', 'number', 'boolean')
```
mrkaye97: I think filtering by the tenant would help here in the average case at least

pkg/repository/scheduler.go
```diff
@@ -75,14 +79,31 @@ type schedulerRepository struct {
...
+		lease:            newLeaseRepository(shared),
+		queueFactory:     newQueueFactoryRepository(shared),
+		rateLimit:        newRateLimitRepository(shared),
+		batchQueue:       newBatchQueueFactoryRepository(shared),
+		assignment:       newAssignmentRepository(shared),
+		optimistic:       newOptimisticSchedulingRepository(shared),
+	}
+}
+
+func (d *schedulerRepository) ListWorkflowNamesByIds(ctx context.Context, workflowIds []uuid.UUID) (map[uuid.UUID]string, error) {
```
abelanger5: don't we have a cache which we can use for this? I think there's a cache in the sharedRepository

mrkaye97:
bumping the original comments I left, I think most of them are still relevant

pkg/repository/sqlcv1/queue.sql
abelanger5: I refactored this query pretty significantly to filter by a specific tenant

pkg/repository/sqlcv1/queue.sql
abelanger5: also refactored this significantly to remove the `cross join lateral` and create a smaller working set in the original CTE

pkg/repository/sqlcv1/queue.sql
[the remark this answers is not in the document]
abelanger5: yep, modified this

internal/services/scheduler/v1/queue_metrics.go
abelanger5: done 👍

pkg/scheduling/v1/workflow_names_test.go
abelanger5: moved the cache to the repo level and dropped the tests

mrkaye97 (approved):

PR #4578, opened 2026-07-30
refactor: actor pattern for improving scheduling performance

abelanger5:
# Description

Refactors the scheduler with three primary differences:
- Use a single event loop to mutate any objects in the scheduler to get rid of the pessimistic locking path where we locked `actionsMu` followed by each action individually. All the mutating calls are moved into `.do`
- Removes the database reads from the scheduler mutation path so we read in worker slots from the DB and coalesce them with the unackedSlots in memory. This is pessimistic in nature: if the unackedSlots won't flush it means we undercount non-utilized slots (until the next read), but it makes the hot-path scheduler performance much, much faster, so likely an ok tradeoff.
- Rewrites the worker/action indexing based off of the original #4451 

To get a sense of the changes against the existing scheduler, you can view this diff: https://github.com/hatchet-dev/hatchet/compare/main...736f400f764d5e027e31c105af6607bb9c5510e3

For rolling this out, I've basically forked the scheduler into a `v1alpha` package which is enabled via an env var so we can easily turn this on and off. 

## Type of change

## What's Changed

See above. 

## Checklist

Changes have been:

- [X] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Claude w/ Fable to assist with testing and parts of the implementation

</details>


pkg/scheduling/v1alpha/slot_pool.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// slotPool is the single owner of scheduling capacity for one worker and slot
+// type. Actions index workers; they do not copy these slot slices.
+//
+// Expiry is pool-level: every slot in a pool is built from the same replenish
+// read, so the whole pool goes stale together. The freelist is exact — only the
+// run loop mutates it, and expiry invalidates the pool as a whole rather than
+// draining individual slots — so free counts never drift.
+type slotPool struct {
```
abelanger5, line 30: note: only RW from the `run` method, so we don't need mutexes here -- `slotPool` isn't otherwise concurrency safe. worth a comment here.

pkg/scheduling/v1/scheduler.go
```diff
@@ -683,62 +625,176 @@ func (s *Scheduler) loopReplenish(ctx context.Context) {
 			}
 			cancel()
 		}
-
 	}
 }
 
 func (s *Scheduler) loopSnapshot(ctx context.Context) {
 	ticker := randomticker.NewRandomTicker(10*time.Millisecond, 90*time.Millisecond)
```
abelanger5, line 691: we should likely reduce the ticker interval here, since we're snapshotting every time. once per second is probably a good QOS.

pkg/scheduling/v1/scheduler.go
```diff
@@ -62,92 +73,188 @@ func newScheduler(cf *sharedConfig, tenantId uuid.UUID, rl *rateLimiter, exts *E
...
-func (s *Scheduler) setWorkers(workers []*v1.ListActiveWorkersResult) {
-	s.workersMu.Lock()
-	defer s.workersMu.Unlock()
+// ack marks assignments as flushed to the database. The slots stay used until a
+// later replenish observes the flush in its availability read.
+func (s *Scheduler) ack(ids []int) {
+	var callbacks []func()
 
-	newWorkers := make(map[uuid.UUID]*worker, len(workers))
+	s.do(context.Background(), func() {
```
abelanger5: lots of usage of `context.Background()` which I wonder if we should be using here

pkg/scheduling/v1/scheduler.go
```diff
@@ -683,62 +625,176 @@ func (s *Scheduler) loopReplenish(ctx context.Context) {
 			}
 			cancel()
 		}
-
 	}
 }
 
 func (s *Scheduler) loopSnapshot(ctx context.Context) {
```
abelanger5, line 690: nit: let's move the snapshotting into its own file so it doesn't pollute the core logic in this file

.github/workflows/scheduler-sync.yml
abelanger5, line 1: note: hacky script to make sure we're propagating changes between the two files while we roll this out.

frontend/docs/pages/self-hosting/configuration-options.mdx
```diff
@@ -157,6 +157,7 @@ Variables marked with ⚠️ are conditionally required when specific features a
 | `SERVER_GRPC_TRIGGER_WRITE_SLOTS`              | Number of slots for gRPC writes                                                               | `5`                     |
 | `SERVER_OPTIMISTIC_SCHEDULING_ENABLED`         | Enable optimistic scheduling                                                                  | `true`                  |
 | `SERVER_OPTIMISTIC_SCHEDULING_SLOTS`           | Slots to allocate for optimistic scheduling                                                   | `5`                     |
+| `SERVER_V1ALPHA_SCHEDULER_ENABLED`             | Run the v1alpha (event loop) scheduler on this shard instead of the v1 scheduler              | `false`                 |
```
abelanger5: let's not document this here: not something that we'd like to expose, and we'll remove this after rollout

pkg/scheduling/v1alpha/action.go
abelanger5, line 1: note: lots of changes in the diff, but most of these are just lifted from the existing scheduler; this is just the easiest way to support an "alpha" version of the scheduler. the core changes are:
- action.go
- scheduler.go
- slot_pool.go

pkg/scheduling/v1alpha/rate_limit.go
```diff
@@ -0,0 +1,340 @@
...
+func (r *rateLimiter) shouldRefill() bool {
+	r.nextRefillAtMu.Lock()
+	defer r.nextRefillAtMu.Unlock()
+
+	if r.nextRefillAt == nil {
+		return false
+	}
+
+	return r.nextRefillAt.After(time.Now().UTC())
+}
```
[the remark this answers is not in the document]
abelanger5: implemented this fix as a drive-by, seems correct

pkg/scheduling/v1alpha/concurrency.go
```diff
@@ -0,0 +1,270 @@
...
+
+		c.isCleanedUp = true
+		cancel()
+	}
+
+	go c.loopConcurrency(ctx)
+	go c.loopCheckActive(ctx)
+
+	// run once on startup instead of waiting for the first tick
+	c.notify(context.Background())
```
grutt, line 108: should this be the shared context?

grutt (approved):

PR #4579, opened 2026-07-31
fix(dashboard): bundle Monaco editor with inline workers to fix CORS

koushik717:
Re-applies #4247 with a fix for the production CORS failure that caused the revert in #4478.

## Root cause of the revert

The original PR used Vite's `?worker` suffix, which emits each worker as a **separate JS chunk** fetched by URL at runtime. In production environments where assets are served with strict CORS headers, those cross-origin worker URL fetches are blocked by the browser, breaking JSON schema autocomplete.

## What changed

Switched `?worker` to `?worker&inline` in `monaco-environment.ts`:

```ts
// before
import editorWorker from 'monaco-editor/esm/vs/editor/editor.worker?worker';
import jsonWorker from 'monaco-editor/esm/vs/language/json/json.worker?worker';

// after
import editorWorker from 'monaco-editor/esm/vs/editor/editor.worker?worker&inline';
import jsonWorker from 'monaco-editor/esm/vs/language/json/json.worker?worker&inline';
```

`?worker&inline` bundles each worker as a base64 blob URL inside the main chunk. No cross-origin fetch occurs at runtime, so CORS headers on the asset server are irrelevant.

Everything else is unchanged from the original PR.

## Testing

- `pnpm run typecheck` - same pre-existing codegen errors as main, no new errors
- Verified no separate worker chunk files are emitted (workers are inlined)
- The trade-off: slightly larger main bundle vs. no CORS dependency

---
AI disclosure (per AI_POLICY.md): I used an AI coding assistant for parts of this change and reviewed/tested everything above myself before opening this PR.

### DIFF
--- a/frontend/app/package.json
+++ b/frontend/app/package.json
@@ -20,7 +20,7 @@
     "typecheck": "tsc -p tsconfig.json",
     "typecheck:build": "tsc -p tsconfig.build.json",
     "typecheck:cypress": "tsc -p tsconfig.cypress.json",
-    "docs:gen": "cd ../snippets && python3 -X utf8 generate.py"
+    "docs:gen": "cd ../snippets && python3 generate.py"
   },
   "dependencies": {
     "@heroicons/react": "^2.2.0",
@@ -69,6 +69,7 @@
     "js-confetti": "^0.13.1",
     "lodash": "^4.18.1",
     "lucide-react": "^1.8.0",
+    "monaco-editor": "^0.55.1",
     "monaco-themes": "^0.4.8",
     "ms": "^2.1.3",
     "posthog-js": "^1.369.2",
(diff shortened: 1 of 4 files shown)


frontend/app/package.json
```diff
@@ -20,7 +20,7 @@
     "typecheck": "tsc -p tsconfig.json",
     "typecheck:build": "tsc -p tsconfig.build.json",
     "typecheck:cypress": "tsc -p tsconfig.cypress.json",
-    "docs:gen": "cd ../snippets && python3 -X utf8 generate.py"
+    "docs:gen": "cd ../snippets && python3 generate.py"
```
gregfurman, line 23: Why the change here?

frontend/app/package.json
```diff
@@ -112,41 +112,32 @@
...
+    "jsdom": "^29.1.1"
   },
   "pnpm": {
     "overrides": {
       "preact": "10.28.2",
       "prismjs": "^1.30.0",
       "qs": ">=6.14.2",
-      "js-yaml": "^4.2.0",
-      "seroval": "^1.5.3",
+      "js-yaml": "^4.1.1",
```
gregfurman, line 132: Think these deps need a rebase on `main` 😬

gregfurman (changes requested):

PR #4580, opened 2026-07-31
fix(frontend): normalize Windows path separators in generate.py

Chelizzzzz:
## Summary
`generate.py` builds the docs/snippet trees consumed by the Hatchet dashboard. It strips `pages_dir` and splits glob results on `/`, but on Windows `glob`/`os.path.join` return backslash paths, so the prefix is never stripped. `docsPages` then nests under garbage keys like `{"..":{"..":{"frontend\docs\pages\":{...}}}}`, and the dashboard reads `docsPages.v1.quickstart` -> `undefined` -> TypeError, making every sidebar link throw "Something went wrong".

Normalizing path separators to `/` before computing the keys fixes Windows while staying a no-op on Linux/macOS.

## Test plan
- [ ] Windows: `python frontend/snippets/generate.py` -> `docsPages` top-level contains `v1`, `cookbooks`, `self-hosting`, ...
- [ ] Linux/macOS: generated `docs/index.ts` is unchanged (no-op)

### DIFF
--- a/frontend/snippets/generate.py
+++ b/frontend/snippets/generate.py
@@ -123,7 +123,7 @@ def parse_snippets(ctx: SDKParsingContext, filename: str) -> list[Snippet]:
     with open(filename) as f:
         content = f.read()
 
-    code_path = f"examples/{ctx.name.lower()}{filename.replace(base_path, '')}"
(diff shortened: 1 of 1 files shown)


mrkaye97:
Hi @Chelizzzzz, I'm going to close this. I'd recommend using WSL instead, as we also don't officially support running workers on Windows (at least Python workers)

PR #4588, opened 2026-08-02
fix(python): logging memory leak

### LINKED ISSUE #4587: [BUG] Python SDK: worker log capture retains every log record for the lifetime of the worker process
**Describe the issue**

`capture_logs` retains every log record the worker emits for the lifetime of the worker process.

`WorkerActionRunLoopManager.aio_start` wraps the *entire* action run loop in `capture_logs`. `capture_logs` creates one `StringIO`, wraps it in a `LogForwardingHandler` (a `logging.StreamHandler`), and closes the stream only in its `finally` — i.e. when the worker stops. `LogForwardingHandler.emit` calls `super().emit(record)`, which appends the formatted record to that `StringIO`.

Nothing ever reads it. Those five lines are every reference to the stream in the SDK:

```
worker/runner/utils/capture_logs.py:183:  class LogForwardingHandler(logging.StreamHandler):
worker/runner/utils/capture_logs.py:214:      log_stream = StringIO()
worker/runner/utils/capture_logs.py:215:      log_forwarder = LogForwardingHandler(log_sender, log_stream)
worker/runner/utils/capture_logs.py:228:      if not any(h for h in logger.handlers if isinstance(h, LogForwardingHandler)):
worker/runner/utils/capture_logs.py:236:      log_stream.close()
```

No `read`, no `getvalue`, no `seek`, no `truncate`. The stream exists only to satisfy `StreamHandler.__init__` so the handler can reuse `Formatter.format()`. Every byte written to it is retained until the worker exits.

Two details that widen the blast radius:

- `super().emit(record)` runs *before* the `if not step_run_id: return` guard, so records emitted outside a task run — which can never be forwarded anywhere — are retained too.
- `ClientConfig.logger` defaults to `getLogger()`, the **root** logger. Out of the box the SDK retains every stdlib log record the process emits, from any library.

The outbound half of `emit()` is correctly bounded (`log_queue_size`, drops on overflow). The `StringIO` is the one unbounded thing in the design, which is why this reads as an oversight rather than a deliberate trade-off.

**Environment**
- SDK: Python SDK 1.33.12 (verified unchanged on `main` as of today)
- Engine: Self-hosted — though the engine is not involved; no token or network needed to reproduce.

**Expected behavior**

Log capture forwards records to the log sender. It should not accumulate them in memory. A worker's RSS should not be a function of how much it has logged since it started.

Self-contained, no engine/token/network. The only logging handler involved is the one the SDK installs on itself: [`hatchet_leak_repro.py`](https://gist.github.com/blakewatters/fdbb910edd8fffc29c4af0354fa04ff9)

```
hatchet-sdk 1.33.12 | python 3.13.3

  records |   log text |     SDK buffer |   RSS kept |  control
---------------------------------------------------------------
        0 |      0 MiB |              0 |    0.0 MiB |  0.0 MiB
   10,000 |     23 MiB |     24,010,000 |   24.5 MiB |  0.0 MiB
   20,000 |     46 MiB |     48,020,000 |   45.8 MiB |  0.0 MiB
   40,000 |     92 MiB |     96,040,000 |   91.6 MiB |  0.0 MiB
   80,000 |    183 MiB |    192,080,000 |  183.2 MiB |  0.0 MiB

ClientConfig.logger default -> <RootLogger root (WARNING)> (is root: True)
```

`control` is the identical workload with `capture_logs` skipped — what `disable_log_capture=True` gives you. Retained RSS is linear in log volume with capture on, and flat with it off. `flush()` reclaims nothing; the memory comes back only when the worker stops.

**Production impact**

This took down a 7-replica worker fleet of ours. All 7 pods OOMKilled inside a 20-minute window, ~12.5 hours after a deploy, growing ~100 MiB/hour from a ~800 MiB baseline into a 2 GiB limit. That worker logs ~6 lines/second averaging ~2.3 KB — nothing exotic. We mitigated in production with `disable_log_capture=True`, which stopped the growth immediately and cost us the dashboard logs. We have since written a local subclass that overrides `emit` to skip the stream write; measured against the same workload it retains 0 bytes where the stock handler retains 92.8 MiB, and it is rolling out now.

**Additional context**

This was already reported in #4432 by @Amiya-w, with a correct diagnosis and essentially the fix below. It was closed with three objections, so let me address each rather than re-litigate:

> The repro provided is self-contained, doesn't necessarily indicate a memory leak, and also shows this problem independent of the SDK.

The repro above defines no handler and no stream. It calls `hatchet_sdk.capture_logs` exactly the way `aio_start` does, and measures process RSS *inside* the wrapped call, where a live worker spends its whole life. The buffer read back is the SDK's own `StringIO`, reached through `handler.stream` on the handler the SDK attached. The control arm isolates `capture_logs` as the sole variable. Independently of any measurement: a buffer that is written on every record and never read is a leak by construction.

> We'd love to see a repro that shows the SDK actually leaking memory as you log more and more.

That is the table — 10k → 80k records, RSS 24.5 → 183.2 MiB, linear, unreclaimed.

> I have a suspicion this is not a problem, since we very likely would've heard about it by now.

Respectfully, absence of reports is weak evidence here, for reasons specific to this bug:

1. The symptom is an OOMKill attributed to the *application*, not to Hatchet. Nothing points at the SDK. We only found it by reading `capture_logs.py` after ruling out our own code.
2. It needs sustained volume *through the configured logger*. Workers that use `ctx.log()` (which bypasses this handler entirely) or that log lightly never accumulate enough to notice.
3. It needs uptime. Anything that redeploys frequently or scales to zero resets the buffer before it matters.
4. The retained bytes live in a private attribute of a handler attached to a logger. No heap tool or metric names Hatchet; it looks like generic heap growth.

We hit all four conditions at once and it still took a production incident to find. I'd expect others to be paying this quietly.

**Proposed fix**

Same as #4432: make `LogForwardingHandler` subclass `logging.Handler` rather than `logging.StreamHandler`, and drop the `StringIO`. The handler only ever needs `format()`; the stream has no consumer. It is a small, behavior-preserving change — the forwarded records, their levels, and the `step_run_id`/retry metadata are all unchanged.

I have this written and tested against 1.33.12 and am happy to open the PR, with tests, as soon as this is triaged and labeled `accepted` — per CONTRIBUTING.md and AI_POLICY.md I'm not opening a drive-by PR ahead of that. Just say the word.

---

<details id="ai-disclosure">

- **Details**: Claude Code (Opus) traced the code path, wrote and ran the reproduction script, and drafted this issue text. The production incident and all measurements are from our own systems.

</details>

mrkaye97:
# Description

Fixes #4587

Memory leak via `StringIO` in the log sender - turns out we don't need `StringIO` at all, and this is actually a very old relic of the V0 SDK

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

No AI

</details>


sdks/python/hatchet_sdk/worker/runner/utils/capture_logs.py
```diff
@@ -161,15 +160,13 @@ def stop(self, timeout: float = 5.0) -> None:
         self._thread = None
 
 
-class LogForwardingHandler(logging.StreamHandler):  # type: ignore[type-arg]
-    def __init__(self, log_sender: AsyncLogSender, stream: StringIO):
```
gregfurman, line 165: Is this not a breaking change since the init no longer requires this extra param? Should we perhaps extent this signature to be:

```python
def __init__(self, log_sender: AsyncLogSender, *args):
```

Or maybe overload it like:

```python
@typing.overload
def __init__(self, log_sender: AsyncLogSender, stream: StringIO): ...

@typing.overload
def __init__(self, log_sender: AsyncLogSender): ...

def __init__(self, log_sender: AsyncLogSender, stream: StringIO | None = None): ...
```

mrkaye97: I can't see how it'd be breaking since this is internal-only, unless someone is using this directly which they shouldn't be? Maybe I'm missing it

gregfurman: Ah well if we don't expect this to be used anywhere else but internally then I suppose it's fine.

gregfurman (approved):
LGTM!

PR #4593, opened 2026-08-02
refactor(dashboard): remove uses of `v1_statuses_olap` 

mrkaye97:
# Description

Part 1 of a few bits of cleanup that'll help us consolidate the cloud repo and this one. I can't see why we need this table, although maybe I'm missing it. We only rely on it for hypertables in cloud and for the status metrics in the OSS, and we could also just use the runs table for those AFAICT, and get rid of all of the churn on this table.

Let me know if I'm missing something obvious here 🤷  I'm thinking we can ship this and then just drop that table after, or at least shut off the triggers into it...

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

No AI use here

</details>

### DIFF
--- a/pkg/repository/sqlcv1/models.go
+++ b/pkg/repository/sqlcv1/models.go
@@ -1798,48 +1798,6 @@ func (ns NullV1RunKind) Value() (driver.Value, error) {
 	return string(ns.V1RunKind), nil
 }
 
-type V1StatusKind string
-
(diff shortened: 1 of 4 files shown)


mrkaye97:
going to close this for now, but probably will bring it back in a week or two

PR #4619, opened 2026-08-05
fix: strip unsupported unicode escapes before writing payloads to jsonb

### LINKED ISSUE #4611: [BUG] task-failed pre-ack fails writing payloads: unsupported Unicode escape sequence (SQLSTATE 22P05)
## Describe the issue

Failing a task via the RabbitMQ `task-failed` consumer can hard-fail in pre-ack when storing task event payloads into `v1_payload.inline_content` (JSONB). Postgres rejects the write with:

```
ERROR: unsupported Unicode escape sequence (SQLSTATE 22P05)
```

Observed error:

```
2026-08-04T13:22:52.387Z ERR error in pre-ack on msg task-failed: could not fail tasks: failed to fail tasks: failed to store task event payloads: failed to write payloads: ERROR: unsupported Unicode escape sequence (SQLSTATE 22P05) service=rabbitmq
```

This blocks task failure acknowledgment (`FailTasks` → store `TASKEVENTDATA` payloads → `WritePayloads`), so the failure path cannot complete cleanly.

## Environment

- Engine: Cloud (seen on cloud-shard-2 / RabbitMQ consumer)
- Path: `task-failed` msgqueue pre-ack → `Tasks().FailTasks` → `payloadStore.Store` → `WritePayloads` (`inline_content` JSONB)

## Suspected source

Likely bad Unicode (commonly `\u0000` / NUL) in a payload that ends up on the failure-event write path. It's probably coming from e.g. an LLM-generated error message or some such on a worker, which causes (probably) the error message we write to the db to include the invalid unicode without being stripped.

Possible areas to inspect:
- Task failure event payload construction (`createTaskEvents` / `FailTasks`) before `StorePayloadOpts` with `V1PayloadTypeTASKEVENTDATA`
- Whether any internal path bypasses `ValidateJSONB` that API handlers use for cron/trigger/event inputs

## Expected behavior

- Task failure should succeed (or fail with a clear validation error) even if upstream payload data contains unsupported Unicode escapes.
- Ideally: sanitize or reject `\u0000` / invalid Unicode before writing JSONB in the payload store, so a single bad payload cannot wedge `task-failed` pre-ack.

## Code pointers

- Error site: `pkg/repository/payloadstore.go` (`failed to write payloads`)
- Call chain: `pkg/repository/task.go` (`failed to store task event payloads`) → `FailTasks`
- Consumer log: `internal/msgqueue/rabbitmq/rabbitmq.go` (`error in pre-ack on msg task-failed`)
- Existing NUL guard (not applied on payload store writes): `pkg/repository/jsonb.go` (`ValidateJSONB` / `isUnicodeValid`)
- SQL: `pkg/repository/sqlcv1/payload-store.sql` (`WritePayloads` → `inline_content` JSONB)

## Additional context

Seen in production logs on 2026-08-04. Appears tied to the fail-task / task-event payload write rather than the RabbitMQ layer itself (`service=rabbitmq` is the consumer).

<details id="ai-disclosure">

- **Details**: Issue drafted from production error log and codepath investigation in Cursor; human provided error and suspected cron/scheduler/trigger writer origin.

</details>

Nahid-NHB:
# Description

Failing a task can hard-fail in the `task-failed` pre-ack path when the error message contains a NUL character:

```
ERR error in pre-ack on msg task-failed: could not fail tasks: failed to fail tasks: failed to store task event payloads: failed to write payloads: ERROR: unsupported Unicode escape sequence (SQLSTATE 22P05)
```

Postgres can't represent `\u0000` in `jsonb`, so it rejects the insert. `FailTasks` marshals the worker's error message into the FAILED event data and hands it to the payload store, which writes it into `v1_payload.inline_content` (JSONB). Nothing on that path goes through `ValidateJSONB`, which only runs on API/gRPC ingest boundaries. So one task failing with a NUL in its error message (an LLM-generated message, some binary junk in a stack trace) fails the whole batch write, the message never gets acked, and the consumer keeps redelivering it.

The same output also goes to `OLAPRepository.PutPayloads` → `v1_payloads_olap.inline_content`, which is JSONB too, so that path can fail the same way.

Since these are internal write paths reacting to data that's already in the system, rejecting isn't useful, there's nothing the consumer can do with a rejection except get stuck again. Sanitizing keeps the failure path moving and loses only characters Postgres refused to store anyway.

Fixes #4611

## Type of change

## What's Changed

- Added `SanitizeJSONB` in `pkg/repository/jsonb.go`. Strips `\u0000` escapes and raw NUL bytes. The scan tracks string/escape state, so `\\u0000` (an escaped backslash followed by the literal text `u0000`) is left alone, matching what `ValidateJSONB` already considers valid.
- Applied it in `payloadStoreRepositoryImpl.Store` and `OLAPRepositoryImpl.PutPayloads`, with a warn log naming the tenant and payload when something is stripped.
- Unit tests for the sanitizer, including a check that sanitized output passes `ValidateJSONB` for every case the existing rejection test covers.

Clean payloads are returned as-is with no copy, so the cost on the normal path is one `bytes.Contains` plus one `IndexByte`.

## Checklist

Changes have been:

- [x] Tested (unit)
- [x] Linted and formatted

## Testing

```
go test ./pkg/repository/ -run 'TestSanitizeJSONB|TestValidateJSONB' -v
```

All 16 pass. `go build ./pkg/... ./internal/...` and `go vet ./pkg/repository/` are clean.

Not covered by an integration test against a real Postgres, the existing tests in this package that need a DB spin up testcontainers and I didn't add one for this. Happy to if you'd prefer it.

---

<details id="ai-disclosure">

- **Details**: Claude Code (Opus) used for tracing the write path, drafting the sanitizer and its tests, and drafting this description. Reviewed and edited by me.

</details>

### DIFF
--- a/pkg/repository/jsonb.go
+++ b/pkg/repository/jsonb.go
@@ -23,6 +23,75 @@ func ValidateJSONB(jsonb []byte, fieldName string) error {
 	return nil
 }
 
+// nullEscape is the JSON escape sequence Postgres rejects with SQLSTATE 22P05
+// ("unsupported Unicode escape sequence") when casting to jsonb. JSON only
+// allows a lowercase `u` here and `0000` has no hex letters, so a plain byte
+// comparison covers every way it can be spelled.
+var nullEscape = []byte{'\\', 'u', '0', '0', '0', '0'}
+
+// SanitizeJSONB strips the characters Postgres refuses to store in a jsonb
+// column: `\u0000` escapes and raw NUL bytes. Both come in from data we don't
+// control (worker error messages, task outputs), and internal write paths like
+// the payload store can't reject them without wedging the queue consumer that
+// is trying to write them.
+//
+// The input is returned untouched when there is nothing to strip, so the common
+// case is one scan and no allocation.
+func SanitizeJSONB(jsonb []byte) []byte {
+	if !bytes.Contains(jsonb, nullEscape) && bytes.IndexByte(jsonb, 0) < 0 {
+		return jsonb
+	}
+
+	out := make([]byte, 0, len(jsonb))
+	inString := false
+
+	for i := 0; i < len(jsonb); {
+		c := jsonb[i]
+
+		switch {
+		case c == 0:
+			// a raw NUL byte is never valid inside a JSON document
+			i++
+		case !inString:
+			if c == '"' {
+				inString = true
+			}
+
+			out = append(out, c)
+			i++
+		case c == '\\':
+			// consume the escape sequence as a unit, otherwise an escaped
+			// backslash would make us misread the bytes after it as an escape
+			// of their own
+			if i+1 >= len(jsonb) {
+				out = append(out, c)
+				i++
+
+				continue
+			}
+
+			if i+len(nullEscape) <= len(jsonb) && bytes.Equal(jsonb[i:i+len(nullEscape)], nullEscape) {
+				i += len(nullEscape)
+
+				continue
+			}
+
+			out = append(out, c, jsonb[i+1])
+			i += 2
+		case c == '"':
+			inString = false
+
+			out = append(out, c)
+			i++
+		default:
+			out = append(out, c)
+			i++
+		}
+	}
+
+	return out
+}
+
 func isUnicodeValid(jsonb []byte) bool {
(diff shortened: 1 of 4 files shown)


mrkaye97:
Going to close this in favor of #4640

PR #4624, opened 2026-08-05
fix(engine): durable task performance improvements

mrkaye97:
# Description

Makes a couple significant performance improvements to durable tasks:

1. We used to serialize durable task requests at a worker level, which caused a ton of backpressure and slowdown when many durable tasks were running on the same worker. Now we run each in a goroutine so it doesn't block. Importantly, this is safe because we always call `getAndLockLogFile` inside of that method, so Postgres basically handles the serialization for us (similar to an in-memory mutex might)
2. Added a bunch more tracing (including `NewRootSpan` to the durable path to make it easier to visualize each call
3. Simplified the load test for durable a lot to not sleep, since we don't want the timings to be susceptible to polling on the engine

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Claude Code to help investigate issues, but I wrote the code

</details>

### DIFF
--- a/internal/services/dispatcher/server_v1.go
+++ b/internal/services/dispatcher/server_v1.go
@@ -352,7 +352,6 @@ func (d *DispatcherServiceImpl) DurableTask(server contracts.V1Dispatcher_Durabl
 	registeredTasks := make(map[uuid.UUID]struct{})
 
 	var reqWg sync.WaitGroup
-	defer reqWg.Wait()
 
 	defer func() {
 		for taskId := range registeredTasks {
@@ -363,6 +362,8 @@ func (d *DispatcherServiceImpl) DurableTask(server contracts.V1Dispatcher_Durabl
 		}
 		d.workerInvocations.Delete(invocation.workerId)
 	}()
+
+	defer reqWg.Wait()
 
 	registerTask := func(externalIdStr string) {
 		taskExtId, err := uuid.Parse(externalIdStr)
--- a/internal/services/dispatcher/server_v1.go
+++ b/internal/services/dispatcher/server_v1.go
@@ -343,6 +343,11 @@ func (d *DispatcherServiceImpl) DurableTask(server contracts.V1Dispatcher_Durabl
 	deregister := d.streamSessions.Register(cancel)
 	defer deregister()
 
+	rateLimitToken := tenantId.String()
+	if tokenId, ok := ctx.Value(analytics.APITokenIDKey).(uuid.UUID); ok && tokenId != uuid.Nil {
+		rateLimitToken = tokenId.String()
+	}
+
 	invocation := &durableTaskInvocation{
 		server:   server,
 		tenantId: tenantId,
@@ -438,6 +443,15 @@ func (d *DispatcherServiceImpl) DurableTask(server contracts.V1Dispatcher_Durabl
 				registerTask(msg.TriggerRuns.DurableTaskExternalId)
 			case *contracts.DurableTaskRequest_WaitFor:
 				registerTask(msg.WaitFor.DurableTaskExternalId)
+			}
+
+			// blocks until the tenant's dispatcher quota has capacity, so a client
+			// sending requests faster than they can be processed is throttled here
+			// instead of piling up unbounded goroutines
+			if d.rateLimiter != nil {
+				if err := d.rateLimiter.WaitForDispatcherCapacity(ctx, rateLimitToken); err != nil {
+					continue
+				}
 			}
 
 			reqWg.Add(1)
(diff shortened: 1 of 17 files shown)


juliusgeo (approved):

internal/services/dispatcher/server_v1.go
```diff
@@ -423,9 +439,14 @@ func (d *DispatcherServiceImpl) DurableTask(server contracts.V1Dispatcher_Durabl
...
 			}
 
-			if err := d.handleDurableTaskRequest(ctx, invocation, r.req); err != nil {
-				d.l.Error().Err(err).Msg("error handling durable task request")
-			}
+			reqWg.Add(1)
+			go func(req *contracts.DurableTaskRequest) {
+				defer reqWg.Done()
+
+				if err := d.handleDurableTaskRequest(ctx, invocation, req); err != nil {
```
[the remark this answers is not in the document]
abelanger5: @mrkaye97 I do sort of agree with this, because it bypasses the grpc rate limit protection - perhaps we can just hook into our tenant rate limiter here

internal/services/dispatcher/server_v1.go
abelanger5, line 1: not really part of these changes, but related: our OTEL gRPC middleware probably shouldn't be creating traces for our stream endpoints, should they? it tends to pollute durations and isn't helpful because of how large those span trees are. Not sure what best practice is but I think it'd clean up our telemetry.

abelanger5 (approved):

internal/services/dispatcher/server_v1.go
mrkaye97: oh yeah this is why I added this `NewRootSpan` thing - it splits the traces for every individual request on this path

abelanger5: yep, I got that, but there's still the trace duration pollution issue right?

mrkaye97: I think we could just not create non-root spans at all on these paths? idk if that's what you're suggesting

internal/services/dispatcher/server_v1.go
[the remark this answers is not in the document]
mrkaye97: will try that out!

internal/services/dispatcher/server_v1.go
abelanger5: it's the otel middleware which is polluting our data though, so we probably need to modify that somehow. can deal with this in a follow-up PR though

pkg/repository/task.go
```diff
@@ -1825,6 +1825,15 @@ func (r *TaskRepositoryImpl) EvictTask(ctx context.Context, tenantId uuid.UUID,
 
 	defer rollback()
 
+	_, err = r.queries.GetAndLockLogFile(ctx, tx, sqlcv1.GetAndLockLogFileParams{
+		Durabletaskid:         task.Id,
+		Durabletaskinsertedat: task.InsertedAt,
+		Tenantid:              tenantId,
+	})
```
igor-kupczynski, line 1832: This is the query:
```
-- name: GetAndLockLogFile :one
SELECT *
FROM v1_durable_event_log_file
WHERE
    durable_task_id = @durableTaskId::BIGINT
    AND durable_task_inserted_at = @durableTaskInsertedAt::TIMESTAMPTZ
    AND tenant_id = @tenantId::UUID
FOR UPDATE
;
```

Why does it need both `durable_task_id` and `durable_task_inserted_at`? Isn't the ID unique per tenant?

igor-kupczynski (commented):
I like using `GetAndLockLogFile` to serialize.

The PR description doesn't mention rate limiting. How does it fit?

pkg/repository/task.go
mrkaye97: Yeah, we've been doing this sort of as a matter of convention, but it's actually practically helpful here - the convention is because on the OLAP side we need to filter by both since the `id` isn't unique (on the engine this isn't true), but passing the `inserted_at` here will help the query planner prune partitions, which will be a slight performance improvement

mrkaye97:
> I like using `GetAndLockLogFile` to serialize.
> 
> The PR description doesn't mention rate limiting. How does it fit?

Thanks! Added that in response to a PR comment, will test a bit more and then update the description

igor-kupczynski:
oh I see https://github.com/hatchet-dev/hatchet/pull/4624#discussion_r3729852113 GC

mrkaye97:
edit: I'm just gonna revert the rate limit change actually, I think we don't have enough volume for this to matter yet so I'd rather think about it more and see if it becomes an issue

pkg/repository/task.go
igor-kupczynski: I see, thanks for the explanation

PR #4626, opened 2026-08-05
fix: keep side panel open updates from stale history index

grutt:
## Summary
- Side panel `open()` was closing over a stale history index, so clicking another row while the panel was already open could leave the panel open with no content
- Read the current index via a ref, and include `open` / `triggerNow` in the scheduled runs column memo deps

## Test plan
- [ ] Open a scheduled run from the list — detail panel shows the correct run
- [ ] With the panel still open, click a different scheduled run — panel updates to that run
- [ ] Use back/forward in the panel header — history still works
- [ ] Repeat for events / filters side panels if convenient

### DIFF
--- a/frontend/app/src/hooks/use-side-panel.tsx
+++ b/frontend/app/src/hooks/use-side-panel.tsx
@@ -20,6 +20,7 @@ import {
   useContext,
   useEffect,
   useMemo,
+  useRef,
   useState,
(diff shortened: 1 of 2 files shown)


grutt:
Converting to an issue: #4671

grutt:
Closing in favor of #4671

PR #4627, opened 2026-08-06
fix(engine): add unique index to `v1_event`, remove `v1_event_lookup_table` and `v1_event_to_run`

mrkaye97:
# Description

Fixes an issue we've been seeing where we store duplicate payloads coming from user events with the same external id + seen at but different ids. This just makes that impossible by preventing those writes. 

Also dropped `v1_event_lookup_table` and `v1_event_to_run` since they're unused

### To do

- [x] Make sure we don't break `v1_event_to_run_olap` and `v1_event_lookup_table_olap` by deduping here - might need to figure out how to map correctly with the `ON CONFLICT` in the `v1_event` write
- [x] Figure out how to delete existing dupes from `v1_event` before making the index

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

No AI

</details>


sql/schema/v1-core.sql
```diff
@@ -2457,53 +2457,7 @@ CREATE TABLE v1_event (
...
-
-CREATE TABLE v1_event_to_run (
-    run_external_id UUID NOT NULL,
-    event_id BIGINT NOT NULL,
-    event_seen_at TIMESTAMPTZ NOT NULL,
-    filter_id UUID,
-
-    PRIMARY KEY (event_id, event_seen_at, run_external_id)
-) PARTITION BY RANGE(event_seen_at);
+CREATE UNIQUE INDEX v1_event_external_id_seen_at ON v1_event (external_id, seen_at);
```
abelanger5, line 2460: we don't need a corresponding `ON CONFLICT` change on the application side?

abelanger5 (approved):

sql/schema/v1-core.sql
mrkaye97: yep: https://github.com/hatchet-dev/hatchet/pull/4627/changes#diff-7eceb1bc33b6f091989ad285925efaa7a33fcd38bac6e158ff181811bf283876R1020

PR #4628, opened 2026-08-06
fix: `head-prepend` frontend base path config

mnafees:
# Description

Blank pages resulted since the base path was being appended to the generated `index.html` whereas they should be prepended before the generated React JS bundle file.

`head-prepend` is the default as per Vite's [documentation](https://vite.dev/guide/api-plugin#transformindexhtml) so removing those lines works in our case.

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

</details>

### DIFF
--- a/frontend/app/vite.config.ts
+++ b/frontend/app/vite.config.ts
@@ -33,12 +33,12 @@ export default defineConfig({
               {
                 tag: 'base',
                 attrs: { href: '{{ .BasePath }}' },
-                injectTo: 'head',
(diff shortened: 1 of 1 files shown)


igor-kupczynski (approved):
The explanation makes sense to me; dropping `injectTo: head` changes where vite inserts its tags from at the end of head to the start of head `injectTo: head-prepend` is the default.

PR #4630, opened 2026-08-06
fix: stop RefreshTimeout buffer races from returning Internal

grutt:
# Description

Fixes a race in the RefreshTimeout coalesce buffer where empty-buffer singleflight results were returned as gRPC `Internal`, surfacing as nice-grpc `ClientError` on `ctx.refreshTimeout`.

## Type of change

## What's Changed

- Treat empty RefreshTimeout buffer flushes as no-ops instead of `Internal`
- Re-check pending increments after empty `take`, and force-flush once if a shared no-op stranded the caller's add
- Map finished/missing task runtimes (`pgx.ErrNoRows`) to `NotFound` and drop the buffered increment
- Keep OLAP timeout-refreshed publish best-effort so pub failures don't fail the RPC after a successful extension
- Add a unit test covering empty-take then add/`hasPending`

## Checklist

Changes have been:

- [ ] Linted and formatted

## Testing

- `go test ./internal/services/dispatcher/ -run TestRefreshTimeoutBuffer`

---

<details id="ai-disclosure">

- **Details**: Investigation via o11y, implementing the buffer race fix and tests, drafting the PR description

</details>

### DIFF
--- a/internal/services/dispatcher/server.go
+++ b/internal/services/dispatcher/server.go
@@ -1859,29 +1859,59 @@ func (d *DispatcherImpl) refreshTimeoutV1(ctx context.Context, tenant *sqlcv1.Te
 	}
 	d.refreshTimeoutBuf.add(key, increment)
 
-	timeoutAt, err, _ := d.refreshTimeoutGroup.Do(key.String(), func() (interface{}, error) {
-		return d.flushRefreshTimeout(ctx, key)
-	})
+	timeoutAt, err := d.doFlushRefreshTimeout(ctx, key, false)
 	if err != nil {
 		return nil, err
 	}
 
+	// A singleflight leader (often a trailing flush) may have observed an empty
+	// buffer before this call's add() was visible. Force-flush once so our
+	// pending increment is not stranded behind a shared no-op result.
+	if timeoutAt.IsZero() && d.refreshTimeoutBuf.hasPending(key) {
+		timeoutAt, err = d.doFlushRefreshTimeout(ctx, key, true)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if timeoutAt.IsZero() {
+		if estimated, ok := d.refreshTimeoutBuf.lastTimeout(key); ok {
+			timeoutAt = estimated
+		} else {
+			return nil, status.Errorf(codes.FailedPrecondition, "could not refresh timeout for task %s", taskExternalId)
+		}
+	}
+
 	return &contracts.RefreshTimeoutResponse{
-		TimeoutAt: timestamppb.New(timeoutAt.(time.Time)),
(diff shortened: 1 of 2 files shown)


internal/services/dispatcher/server.go
```diff
@@ -1859,29 +1859,59 @@ func (d *DispatcherImpl) refreshTimeoutV1(ctx context.Context, tenant *sqlcv1.Te
...
+
 	return &contracts.RefreshTimeoutResponse{
-		TimeoutAt: timestamppb.New(timeoutAt.(time.Time)),
+		TimeoutAt: timestamppb.New(timeoutAt),
 	}, nil
 }
 
-func (d *DispatcherImpl) flushRefreshTimeout(ctx context.Context, key refreshTimeoutKey) (time.Time, error) {
-	if !d.refreshTimeoutBuf.shouldFlush(key) {
+func (d *DispatcherImpl) doFlushRefreshTimeout(ctx context.Context, key refreshTimeoutKey, force bool) (time.Time, error) {
```
mrkaye97, line 1890: should we have this return a pointer instead of a zero value? seems silly to check for `time.Zero()` instead of `nil`

internal/services/dispatcher/server.go
```diff
@@ -1859,29 +1859,59 @@ func (d *DispatcherImpl) refreshTimeoutV1(ctx context.Context, tenant *sqlcv1.Te
...
 
+	// A singleflight leader (often a trailing flush) may have observed an empty
+	// buffer before this call's add() was visible. Force-flush once so our
+	// pending increment is not stranded behind a shared no-op result.
+	if timeoutAt.IsZero() && d.refreshTimeoutBuf.hasPending(key) {
+		timeoutAt, err = d.doFlushRefreshTimeout(ctx, key, true)
+		if err != nil {
+			return nil, err
+		}
+	}
```
mrkaye97, line 1875: I'm not sure I really understand this 🤔 we call flush multiple times here?

internal/services/dispatcher/server.go
```diff
@@ -1900,14 +1930,23 @@ func (d *DispatcherImpl) flushRefreshTimeout(ctx context.Context, key refreshTim
 			if cached, ok := d.refreshTimeoutBuf.lastTimeout(key); ok {
 				return cached, nil
 			}
-			return time.Time{}, fmt.Errorf("no buffered refresh timeout for %s", key)
+			// A singleflight waiter may have add()'d after take saw empty.
+			if d.refreshTimeoutBuf.hasPending(key) {
+				continue
+			}
```
mrkaye97, line 1936: this too, sort of confusing / surprising that either 1) singleflight doesn't protect against this and 2) we don't use a mutex instead to block until the flush completes / make sure this doesn't happen?

PR #4663, opened 2026-08-10
fix(engine): jitter operation interval startup to avoid query storms

grutt:
## Summary
- Phase-spread per-tenant activity gauges so ~10k tickers are not synchronized every 5s
- Randomize the first operation trigger across the full interval window
- Lazy-load persisted intervals instead of blocking in `NewInterval` during tenant discovery

This reduces the DB connection / latency spike when controllers restart or scale and discover thousands of tenants at once.

## Test plan
- [x] `go test ./internal/operation/ -count=1`
- [ ] Deploy / restart cloud controllers; confirm RDS connection + latency spike on roll is reduced

### DIFF
--- a/internal/operation/interval.go
+++ b/internal/operation/interval.go
@@ -13,9 +13,9 @@ import (
 	v1 "github.com/hatchet-dev/hatchet/pkg/repository"
 )
 
-const (
-	gaugeInterval = time.Second * 5
-)
+// gaugeInterval is the period between activity-gauge polls. It is a var so
+// tests can shrink it; production always uses the 5s default.
+var gaugeInterval = 5 * time.Second
 
 // IntervalGauge is a function that determines whether or not to increase or reset the interval.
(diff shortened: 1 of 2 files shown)


internal/operation/interval.go
```diff
@@ -49,38 +58,62 @@ func NewInterval(
...
 
-	if currInterval < startInterval {
-		currInterval = startInterval
-	}
+	i.intervalMu.Lock()
+	defer i.intervalMu.Unlock()
 
-	if currInterval > maxInterval {
-		currInterval = maxInterval
+	if currInterval < i.startInterval {
```
juliusgeo, line 108: hmm, this doesn't seem right--it's replacing maxInterval?

grutt: no the diff is just borked... max interval is below

juliusgeo: ohh, I see

juliusgeo (approved):

PR #4667, opened 2026-08-11
fix(api): treat client-aborted requests as 499 instead of server errors

igor-kupczynski:
# Description

Failed unauthenticated session saves on requests the client already abandoned were surfacing as 500s and error-level logs. They now return 499 (client closed request), log at info, and no longer mark OTel spans as errors, since a client disconnect is not a server fault.

## Type of change

## What's Changed

- [x] Return 499 and log at info when an unauthenticated session save fails because the request context is already done
- [x] Add a `StatusClientClosedRequest` (499) constant and demote 499s to info in the API access log
- [x] Stop marking OTel spans as errors for 499 responses
- [x] Unit tests for `sessionSaveError` and `accessLogLevel`

## Checklist

Changes have been:

- [x] Tested (unit)
- [x] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Cursor agent assisted with the implementation and tests.

</details>


mnafees:
Any strong reason to use a non-standard HTTP response code?

igor-kupczynski:
> Any strong reason to use a non-standard HTTP response code?

There's no standard http code to mark a client that _abandoned_ a request.

I'd argue that 499 is well adopted at this point; AFAIK nginx invented it (so it's widespread in k8s world), and it's adopted e.g. by cloudflare https://developers.cloudflare.com/support/troubleshooting/http-status-codes/4xx-client-error/error-499/ and referenced in multiple guides online https://betterstack.com/community/questions/possible-reasons-fro-nginx-499-error-codes/ 

I bet it's in the training data of most of the LLMs nowadays, e.g. terra on no reasoning:
Screenshot 2026-08-11 at 15 30 15


**edit** also, given the client has left, this is not send back to client, only logged in the access.log to make that case clear (i bet http doesn't have a standard code, because it looked at it from the protocol, not telemetry perspective)

grutt (approved):

PR #4669, opened 2026-08-11
chore(dispatcher): warn on task assigned bulk send failures

grutt:
## Summary
- Downgrade `failed to handle task assigned bulk message` from error to warning in the dispatcher.
- These failures (especially worker send flow control) are already requeued via the retry path, so error-level logging is noisy.

## Test plan
- [ ] Confirm dispatcher still requeues tasks when send fails / flow control is active
- [ ] Confirm the log appears as `WRN` rather than `ERR` under flow control


igor-kupczynski (approved):
makes sense to me

PR #4687, opened 2026-08-13
fix(engine): merge single-task workflow concurrency onto the task

grutt:
# Description

Workflow-level concurrency on a one-task DAG was registered as a parent DAG strategy. Merge it onto the task at registration so the run stays a standalone task.

## Type of change

## What's Changed

- Merge workflow concurrency onto the sole DEFAULT task (prepended before existing task concurrency) when there is no on-failure handler
- Leave parent workflow concurrency in place for multi-task DAGs and single-task + on-failure
- Add unit tests for merge order, on-failure/multi-task skip, checksum equivalence, and write-path persistence

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Cursor Grok 4.6 implemented the merge in `createWorkflowVersionTxs` and added unit/write-path tests.

</details>


pkg/repository/workflow_concurrency_merge_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSingleTaskWorkflowConcurrencyPersistsAsTaskLevel(t *testing.T) {
```
mrkaye97: I don't think this is a useful test - if the unit tests in the previous file pass, this has to also pass unless the whole thing is broken, which we'd catch in other tests

pkg/repository/workflow_concurrency_merge_test.go
```diff
@@ -0,0 +1,148 @@
...
+	assert.Equal(t, strategy, got[0].strategy)
+	assert.Nil(t, got[0].parentStrategyID, "merged workflow concurrency must not have a parent strategy")
+
+	assert.Equal(t, "input.task_id", got[1].expression)
+	assert.Equal(t, taskMaxRuns, got[1].maxConcurrency)
+	assert.Equal(t, taskStrategy, got[1].strategy)
+	assert.Nil(t, got[1].parentStrategyID)
+}
+
+func TestMultiTaskWorkflowConcurrencyKeepsParentStrategy(t *testing.T) {
```
mrkaye97: this too - feels like we should have another test that catches this?

grutt: can you draft a skill for "test acceptance criteria"

mrkaye97: I would really prefer we use our brains to decide if tests are valuable and not rely on skills

grutt: i dont think that's really fair, i did look at these and generally err on the side of more tests that wont flake are better. i think a skill would codify a happy middle ground.

grutt: and would remove one cognitive load thing when hopping between smaller prs

mrkaye97 (approved):

PR #4688, opened 2026-08-13
refactor(engine): split txs for durable event ingest

mrkaye97:
# Description

Splitting durable event ingestion into two separate transactions, one which writes into the event log under the lock, and the second that actually does the triggering of runs/waits. The hope is this will make us hold connections for much less time and hold that lock for less time too, and maybe also help us do more buffering of the trigger writes now that they're decoupled from the log file stuff

The key implementation detail here is I added a `triggered_at` column to the event log which starts as null and then is stamped when the run/wait is actually processed in the second tx, which basically functions like an outbox where we write the row and then we can retry the trigger as many times as we need in a tx-safe way until it succeeds and stamps the row with the timestamp

One note is we'd need to fix forward if this breaks, but it'd be easy enough (can just revert the code changes, leave the schema as-is, and regenerate)

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Claude Code to help with the copy / paste into helper functions

</details>

### DIFF
--- a/cmd/hatchet-migrate/migrate/migrations/20260813070123_v1_0_140.sql
+++ b/cmd/hatchet-migrate/migrate/migrations/20260813070123_v1_0_140.sql
@@ -1,0 +1,9 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE v1_durable_event_log_entry ADD COLUMN triggered_at TIMESTAMPTZ;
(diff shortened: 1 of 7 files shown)


mrkaye97:
closing in favor of #4698

PR #4697, opened 2026-08-14
fix(engine): round-robin workers with equal label affinity

grutt:
## Summary
- Label ranking was resetting the assignment ring to offset 0, so workers with the same `desired_worker_labels` score packed onto the first UUID until its slots filled.
- Keep higher affinity first, but round-robin within the highest-rank tied group (and skip the sort when every worker is tied).
- Regression test assigns two equal-affinity workers with spare slots and asserts they are not packed onto the same worker.

## Test plan
- [x] `go test ./pkg/scheduling/v1/`
- [x] `TestScheduler_AssignSingleton_EqualLabelsRoundRobin` fails against main packing behavior and passes on this branch
- [ ] Confirm sticky HARD/SOFT still prefer the desired worker
- [ ] Spot-check a labeled fleet: consecutive tasks should cycle workers instead of filling worker 1..N

### DIFF
--- a/pkg/scheduling/v1/scheduler.go
+++ b/pkg/scheduling/v1/scheduler.go
@@ -1020,23 +1020,23 @@ func (s *Scheduler) assignSingleton(
 	candidates := a.workerIds
 	offset := a.ringOffset
 	a.ringOffset++
-	tiedLen := len(candidates)
+	topRankCount := len(candidates)
 
 	if qi.Sticky != sqlcv1.V1StickyStrategyNONE || len(labels) > 0 {
-		candidates, tiedLen = s.rankWorkerIds(qi, labels, a.workerIds)
+		candidates, topRankCount = s.rankWorkerIds(qi, labels, a.workerIds)
 	}
 
-	if len(candidates) == 0 || tiedLen == 0 {
+	if len(candidates) == 0 || topRankCount == 0 {
 		r.noSlots = true
 		return
 	}
 
-	offset %= tiedLen
+	offset %= topRankCount
 
 	var selected []*slot
 
-	for i := 0; i < tiedLen; i++ {
-		workerId := candidates[(offset+i)%tiedLen]
+	for i := 0; i < topRankCount; i++ {
+		workerId := candidates[(offset+i)%topRankCount]
 
 		if sel, ok := selectSlotsFromPools(s.poolsByWorker[workerId], requests, now); ok {
 			selected = sel
@@ -1045,7 +1045,7 @@ func (s *Scheduler) assignSingleton(
 	}
 
 	if selected == nil {
-		for i := tiedLen; i < len(candidates); i++ {
+		for i := topRankCount; i < len(candidates); i++ {
 			workerId := candidates[i]
 
 			if sel, ok := selectSlotsFromPools(s.poolsByWorker[workerId], requests, now); ok {
@@ -1204,17 +1204,17 @@ func (s *Scheduler) rankWorkerIds(
 		return []uuid.UUID{}, 0
 	}
 
-	tiedLen := len(ranked)
+	topRankCount := len(ranked)
 	for i := 1; i < len(ranked); i++ {
 		if ranked[i].rank != ranked[0].rank {
 			slices.SortStableFunc(ranked, func(left, right rankedWorker) int {
 				return right.rank - left.rank
 			})
 
-			tiedLen = 1
+			topRankCount = 1
 			topRank := ranked[0].rank
-			for tiedLen < len(ranked) && ranked[tiedLen].rank == topRank {
-				tiedLen++
+			for topRankCount < len(ranked) && ranked[topRankCount].rank == topRank {
+				topRankCount++
 			}
 			break
 		}
@@ -1224,7 +1224,7 @@ func (s *Scheduler) rankWorkerIds(
 	for index := range ranked {
 		result[index] = ranked[index].id
 	}
-	return result, tiedLen
+	return result, topRankCount
 }
 
 type assignedQueueItem struct {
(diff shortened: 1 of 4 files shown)


pkg/scheduling/v1/scheduler_test.go
```diff
@@ -490,6 +490,48 @@ func TestScheduler_AssignSingleton_StickyHardForcesRanking(t *testing.T) {
 	require.Equal(t, desiredWorkerId, res.workerId)
 }
 
+func TestScheduler_AssignSingleton_EqualLabelsRoundRobin(t *testing.T) {
+	tenantId := uuid.New()
+	workerIds := []uuid.UUID{uuid.New(), uuid.New()}
+
+	regionLabel := &sqlcv1.ListManyWorkerLabelsRow{
+		Key:      "region",
+		StrValue: pgtype.Text{String: "us-east-1", Valid: true},
```
mrkaye97: small nit, but I've been trying to use `sqlchelpers.TextFromStr` instead (when I have time to do stuff, I want to keep removing more uses of `pgtype.Foobar` from everywhere

pkg/scheduling/v1/scheduler_test.go
```diff
@@ -490,6 +490,48 @@ func TestScheduler_AssignSingleton_StickyHardForcesRanking(t *testing.T) {
...
+	s := newTestScheduler(t, tenantId, &mockAssignmentRepo{})
+	s.setWorkers(workers)
+	a := seedActionPools(t, s, "A", slots...)
+
+	first := assignOne(t, s, a, testQI(tenantId, "A", 1), desired, defaultRequest(), nil, nil)
+	require.True(t, first.succeeded)
+
+	second := assignOne(t, s, a, testQI(tenantId, "A", 2), desired, defaultRequest(), nil, nil)
+	require.True(t, second.succeeded)
+	require.NotEqual(t, first.workerId, second.workerId, "equal-affinity workers with free slots must not pack onto the same worker")
```
mrkaye97, line 533: I struggle with these types of tests a bit because there are actually more cases here. There's the (alleged) old case, which was us always scheduling on the first worker (although unclear to me if "first" is coming from db ordering, in-memory ordering, etc.). Then there's the new one which is where we try to round robin. But there's also a possibility where we would assign semi-randomly, in which case this test would sometimes pass if it just happens that we get round robin behavior because e.g. the ordering changes for some reason.

This is the same issue we have in e.g. the tests for dynamic label assign, and so I basically wrote those tests to be "probabilistic," meaning that if we run N tasks with label X provided on two workers with labels X and Y, the probability of all N running on a single worker by random chance (the second case) is (1/2)^N, so then we just make N sufficiently big to be confident the feature is working.

I think we should do that here too

pkg/scheduling/v1/scheduler.go
```diff
@@ -1020,30 +1020,41 @@ func (s *Scheduler) assignSingleton(
 	candidates := a.workerIds
 	offset := a.ringOffset
 	a.ringOffset++
+	tiedLen := len(candidates)
```
mrkaye97: nit: it took me like 10 mins of reading the code to understand that this was the number of workers that are tied for "first place" in the rank order - I feel like naming it something to do with that would be good?

pkg/scheduling/v1/scheduler.go
```diff
@@ -1020,30 +1020,41 @@ func (s *Scheduler) assignSingleton(
...
 
-	offset %= len(candidates)
+	offset %= tiedLen
 
 	var selected []*slot
 
-	for i := 0; i < len(candidates); i++ {
-		workerId := candidates[(offset+i)%len(candidates)]
+	for i := 0; i < tiedLen; i++ {
+		workerId := candidates[(offset+i)%tiedLen]
```
mrkaye97: I'm a little confused by the modulus math here - to me, this read like it'll still select the first candidate? maybe I'm misunderstanding why this works

pkg/scheduling/v1/scheduler_test.go
grutt: this path is not probabilistic and this covers the case on main (and fails). this wont cover reliably if we decide to make this path probabilistic in the future, but it should be fine for today

mrkaye97: Sorry to be clear, the other path I was describing also isn't probabilistic. But the question is 1) if it becomes probabilistic, this test will be flaky instead of reliably failing and 2) how confident are you that it's not probabilistic?

pkg/scheduling/v1/scheduler_test.go
grutt: pre-existing condition on the other tests. this is a good candidate for a linter rule 😅 because afik no one else is reviewing for this.

pkg/scheduling/v1/scheduler_test.go
grutt: 1. yes, but do we expect to make this path probabilistic? 
2. 100%

pkg/scheduling/v1/scheduler.go
grutt: added some clarifying comments here

pkg/scheduling/v1/scheduler_test.go
mrkaye97: maybe I'm misunderstanding 😅 these methods: 1) loop over things from the database and 2) loop over maps. At the very least, map iteration in Go is intentionally non-deterministic, so are we really 100% sure?

mrkaye97: Ah, okay I see the thing I was missing - there's a byte sort of the worker ids on line 6something that gets rid of the risk from the maps here, so I think this is fine 👍

grutt: just double checked to be extra certain. 

the replenish step is where the nondeterminism is possible, but we sort there. 
we're not looping over a map on this path

grutt: github race, yes

mrkaye97 (approved):

PR #4703, opened 2026-08-15
feat(engine): workflow pause

mrkaye97:
# Description

Adds the ability to pause and unpause workflows (with a little bit of config), which basically results in shoveling runs triggered under those workflows into a separate `v1_paused_workflow_queue_item` table and back instead of triggering. When you unpause, we shovel everything back into the queue.

There's a bit of config you can specify:

1. The TTL for runs to be queued (before they get cleaned up)
2. The behavior of cron and scheduled runs (which you can choose to either have drop or queue)

~Still need to write some E2E tests for this, but I'm landing soon so I'll do it later~

Pause modal:

Screenshot 2026-08-15 at 2 42 59 PM

Paused state:

Screenshot 2026-08-15 at 2 28 19 PM

Run queued state:

Screenshot 2026-08-15 at 2 28 04 PM

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

<details id="ai-disclosure">

- **Details**: Used Claude Code a bit on the frontend

</details>


api/v1/server/handlers/workflows/update.go
```diff
@@ -1,11 +1,61 @@
...
 	"github.com/hatchet-dev/hatchet/api/v1/server/oas/gen"
+	"github.com/hatchet-dev/hatchet/api/v1/server/oas/transformers"
+	"github.com/hatchet-dev/hatchet/internal/msgqueue"
+	tasktypes "github.com/hatchet-dev/hatchet/internal/services/shared/tasktypes/v1"
+	"github.com/hatchet-dev/hatchet/pkg/repository"
+	"github.com/hatchet-dev/hatchet/pkg/repository/sqlcv1"
 )
 
-func (t *WorkflowService) WorkflowUpdate(ctx echo.Context, request gen.WorkflowUpdateRequestObject) (gen.WorkflowUpdateResponseObject, error) {
-	panic("deprecated")
```
mrkaye97, line 10: I think un-deprecating is fine? we could also add a `/stable/workflow` thing

pkg/repository/task.go
```diff
@@ -1557,6 +1559,56 @@ func (r *TaskRepositoryImpl) ProcessTaskReassignments(ctx context.Context, tenan
...
+
+	for _, item := range expiredItems {
+		tasks = append(tasks, TaskIdInsertedAtRetryCount{
+			Id:         item.TaskID,
+			InsertedAt: item.TaskInsertedAt,
+			RetryCount: item.RetryCount,
+		})
+	}
+
+	res, err := r.cancelTasks(ctx, tx, tenantId, tasks)
```
mrkaye97, line 1606: needing to do this is a pretty big risk IMO, since we could have queue items piling up faster than we can cancel them. I'm not sure what to do about this though...

api-contracts/openapi/components/schemas/workflow.yaml
```diff
@@ -34,12 +46,51 @@ Workflow:
...
+
+PauseWorkflowRequest:
+  oneOf:
+    - $ref: "#/PauseWorkflowRequestPause"
+    - $ref: "#/PauseWorkflowRequestUnpause"
+
+WorkflowUpdateRequest:
+  type: object
+  description: >-
+    Deprecated: this request no longer accepts any fields. Use the dedicated
```
juliusgeo: Is the previous behavior of WorkflowUpdate that it would just drop scheduled runs? Or did it not work? Slightly confused by this. Also, would it be better to just update this endpoint rather than deprecating?

sql/schema/v1-core.sql
```diff
@@ -723,6 +723,42 @@ CREATE INDEX v1_batched_queue_item_step_priority_idx ON v1_batched_queue_item (
     id ASC
 );
 
+-- v1_paused_workflow_queue_item stores queue items for workflows that are currently paused.
+CREATE TABLE v1_paused_workflow_queue_item (
+    paused_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
juliusgeo: can we just use task_inserted_at instead of creating a new column?

api-contracts/openapi/components/schemas/workflow.yaml
mrkaye97: `WorkflowUpdate` is a deprecated V0 endpoint and does nothing, not sure if that's what you're asking. We could change that, but I kinda like this toggle behavior w/ the discriminated union to make the params much clearer (when pausing you provide settings, on onpause you only need the workflow id). WDYT?

sql/schema/v1-core.sql
mrkaye97: Yeah good call, I think we probably can - lemme see

internal/services/controllers/task/controller.go
```diff
@@ -1167,6 +1180,31 @@ func (tc *TasksControllerImpl) handleProcessDurableRunTrigger(ctx context.Contex
...
+			newlyUnpausedWorkflows = append(newlyUnpausedWorkflows, msg.WorkflowID)
+		}
+	}
+
+	if err := tc.repov1.Workflows().MovePausedWorkflowQueueItems(ctx, tenantId, newlyPausedWorkflows); err != nil {
+		return fmt.Errorf("failed to move newly paused workflows: %w", err)
+	}
+
+	if err := tc.repov1.Workflows().RequeuePausedWorkflowQueueItems(ctx, tenantId, newlyUnpausedWorkflows); err != nil {
+		return fmt.Errorf("failed to requeued tasks for newly unpaused workflows: %w", err)
```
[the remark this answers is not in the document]
mrkaye97: this isn't going to happen in practice

api-contracts/openapi/components/schemas/workflow.yaml
juliusgeo: Ah ok gotcha, it wasn't clear if WorkflowUpdate was already being used. I do like the toggle behavior as well, but perhaps we could just resurrect `WorkflowUpdate` in case we want to add more options later, so we don't need extra endpoints

mrkaye97: Hah I was thinking I'd add `/v1/stable/workflows/{workflow}` to make it more of a "v1" endpoint, lemme see about that

mrkaye97: fixed! rolled it into the old pause endpoint

juliusgeo (approved):

PR #4712, opened 2026-08-17
fix(go-sdk): poll run status if the result listener stays silent

### LINKED ISSUE #4711: [BUG] Go SDK: child Run can hang after child already COMPLETED (shared result listener)
> Claude Fable on Cursor commenting on behalf of Igor:

**Describe the issue**
Go SDK `Client.Run` / `ref.Result()` blocks until `step_timeout` while the child shows COMPLETED in the dashboard within seconds. The shared `SubscribeToWorkflowRuns` stream stays silent — no error, reconnect, or GOAWAY. Retry spawns a second child → duplicate side effects.

**How to hit it**: Go worker with concurrent `Run` waits in one process. First wait in a burst can succeed; later waits on already-COMPLETED children hang. Cloud, v0.100.2–v0.100.18, reproduced on a worker <2 min old (no deploy/reconnect).

**Why existing fixes don't cover it**: #4668's pending-result log can't fire from `Client.Run` (nil context), and `ResultWithContext` only logs, never fails the wait. #4491/#4571 address reconnect handoff — not in play here.

**Root-cause candidates** (both consistent with `dispatcher/server.go`):
- (A) ack-without-delivery: `server.Send` returns nil → run dropped from acks though bytes never reach the waiter; keepalives stay green.
- (B) 1s poller never classifies the run finalized (`ListFinalizedWorkflowRuns`) despite dashboard COMPLETED.

Discriminator: `subscribe_to_workflow_runs_v1.send_event`/`.iter` spans vs DB step/task/output-event counts. Candidate mitigation: fail the wait if REST says terminal while the listener is silent.

<details id="ai-disclosure">

- [x] _LLM used per [AI_POLICY.md](https://github.com/hatchet-dev/hatchet/blob/main/AI_POLICY.md)._ Claude Fable in Cursor; human-reviewed.

</details>

igor-kupczynski:
# Description

Fixes #4711.

`Client.Run` / `Result()` wait on a shared `SubscribeToWorkflowRuns` stream. If the run is already terminal but the stream never delivers — e.g. it went half-open across a reconnect — the waiter hangs until `step_timeout` or a manual cancel. This PR polls `GetRunDetails` after a short grace period and returns the authoritative result (or terminal error) instead of hanging.

**Production reproduction (2026-08-17, hatchet-ci-prod, Go SDK v0.86.3, run `117f8ed0-034c-4b8d-b561-372fc0e52b54`):** `deploy-high` spawned 13 children on its own worker and waited on their results. All 13 completed by 19:32:41 UTC and the engine recorded SIGNAL_COMPLETED at 19:32:43–44, but the parent never unblocked. The worker's listener had re-established at 19:32:30, and the engine's 13 result sends at 19:32:48 reported success with nothing arriving client-side — consistent with a half-open stream. The run was cancelled by hand at 19:35:06. With this fix the poll would have returned each child result within ~1s of it turning terminal and the DAG would have completed normally. This also widens #4711: v0.86.3 is affected, and a mid-wait listener re-establish can be the trigger; the poll doesn't care which mechanism ate the event.

## Type of change

## What's Changed

- After 5s with no listener event, poll run details every 1s (`RunWorkflow`, `SpawnWorkflow`, `SpawnWorkflows` waiters).
- COMPLETED → return the result. FAILED/CANCELLED → error. Fetch errors keep waiting on the stream.
- Prefer a listener event if one is already queued when a poll fires.
- Warn when a result is returned via the poll path — today these incidents leave zero client or engine logs.

Cost: each in-flight `Result()` adds one unary `GetRunDetails`/s after the grace period, even with a healthy listener (13 concurrent waiters in the incident above ⇒ 13 rps).

## Test plan

- [x] `go test ./pkg/client/ -race -count=1` (silent listener + COMPLETED/FAILED fallback)
- [x] `go test ./pkg/worker/ -count=1`

## Testing

Silent-listener unit test: transient fetch error, then RUNNING, then COMPLETED with an empty task error pointer still returns output. FAILED returns the task error.

---

<details id="ai-disclosure">

- **Details**: Cursor (Claude Fable planned/reviewed and validated against the production occurrence; GPT implemented; Cursor Grok 4.6 folded review findings and opened the PR). Human-in-the-loop.

</details>


igor-kupczynski:
@greptileai review this draft

PR #4717, opened 2026-08-17
refactor(engine): simplify child id gen for the durable path

mrkaye97:
# Description

Moves some unnecessary database i/o off of the durable path for child id generation. Previously, we were locking the task, looking up a row in the lookup table, etc., none of which we really needed to be doing. Should help shed some more load from the db / get rid of some more locking

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

<details id="ai-disclosure">

- **Details**: Claude Opus for code review to make sure I didn't mess anything up

</details>

### DIFF
--- a/internal/services/dispatcher/server.go
+++ b/internal/services/dispatcher/server.go
@@ -1248,7 +1248,10 @@ func (s *DispatcherImpl) sendStepActionEventV1(ctx context.Context, request *con
 		return nil, status.Errorf(codes.InvalidArgument, "invalid task external run id %s: %v", request.TaskRunExternalId, err)
 	}
 
-	task, err := s.repov1.Tasks().GetTaskForActionEvent(ctx, tenant.ID, taskExternalId)
+	// if there's no retry count, we need to read it from the task, so we can't skip the cache
+	skipCache := request.RetryCount == nil
+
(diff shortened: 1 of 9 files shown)


internal/services/dispatcher/server.go
```diff
@@ -1242,15 +1242,13 @@ func (s *DispatcherImpl) taskEventsToWorkflowRunEvent(tenantId uuid.UUID, finali
 func (s *DispatcherImpl) sendStepActionEventV1(ctx context.Context, request *contracts.StepActionEvent) (*contracts.ActionEventResponse, error) {
 	tenant := ctx.Value("tenant").(*sqlcv1.Tenant)
 
-	// if there's no retry count, we need to read it from the task, so we can't skip the cache
-	skipCache := request.RetryCount == nil
```
juliusgeo, line 1246: why is this removed?

juliusgeo (approved):

internal/services/dispatcher/server.go
mrkaye97: Initially I was thinking we don't need it because we're never caching on this path after this change, but now I'm reconsidering that since we might lose out on some performance gains from the cache - going to see about adding it back

mrkaye97: the `skipCache` itself btw is because we'd need to force a db read if we have no retry count available here

PR #4718, opened 2026-08-17
feat: add per-member canViewPayloads for payload visibility

grutt:
# Description

Adds a per-membership `canViewPayloads` flag (default true) so tenant OWNER/ADMIN can hide run and event payloads from specific MEMBERs and VIEWERs. Restricted requests still return 200: required `input`/`output` become `{}`, additional metadata is omitted, and `payloadsRestricted` is set so the UI can distinguish restricted from empty.

## Type of change

## What's Changed

- Persist `canViewPayloads` on `TenantMember` and `TenantInviteLink`; create/update invite and update-member APIs accept the flag
- `CanViewPayloads`: API tokens unrestricted, OWNER/ADMIN always see payloads (flag ignored), MEMBER and VIEWER follow the boolean
- Task/event/workflow-run transformers strip payloads via `WithPayloads` instead of 403
- Dashboard: Payloads column on members and pending invites; invite/edit-member checkbox (greyed out for OWNER/ADMIN); run Output/Input empty state; runs list Metadata shows Hidden
- Org invite, org member role, and user-group forms expose the same flag (cloud UI in this frontend)

## Checklist

Changes have been:

- [ ] Linted and formatted

## Testing

- Unit: `CanViewPayloads` (no tenant-member → allow; MEMBER/VIEWER false/true; ADMIN/OWNER with flag false → allow)
- Manual: invite a MEMBER with payloads unchecked; Metadata cells show Hidden; run detail shows Payloads hidden; same run as OWNER still shows JSON; responses stay 200

---

<details id="ai-disclosure">

- **Details**: Implementation and PR description drafted with Cursor (Grok 4.6).

</details>


grutt:
add-tenant-member-can-view-payloads
edit-member-can-view-payloads
org-invite-can-view-payloads
org-team-payloads-column
owner-run-output-visible
restricted-member-run-payloads-hidden
restricted-member-runs-metadata-hidden
tenant-members-payloads-column
user-group-can-view-payloads

api/v1/server/authz/payloads.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/labstack/echo/v4"
+
+	"github.com/hatchet-dev/hatchet/pkg/repository/sqlcv1"
+)
+
+// CanViewPayloads reports whether the current request may see task and event payloads.
+// API tokens (no tenant-member in context) are unrestricted. OWNER and ADMIN always can.
+func CanViewPayloads(c echo.Context) bool {
+	member, ok := c.Get("tenant-member").(*sqlcv1.PopulateTenantMembersRow)
+	if !ok || member == nil {
```
juliusgeo, line 14: I think this should be false in this branch?

api/v1/server/authz/payloads.go
```diff
@@ -0,0 +1,22 @@
...
+		return true
+	}
+
+	if member.Role == sqlcv1.TenantMemberRoleOWNER || member.Role == sqlcv1.TenantMemberRoleADMIN {
+		return true
+	}
```
mrkaye97, line 20: I wonder if we don't want this - I could imagine some people not wanting _anyone_ to be able to view payloads. WDYT?

mrkaye97 (approved):
this implementation looks good to me 👍 nice work on this one - I wonder if we can get rid of some changes to the API handlers and keep this logic _only_ at the transformer level? I was thinking that a middleware that runs on the after side of the request might work really nicely for this?

another thing - there might be endpoints on the gRPC side we need to check on too, such as the `GetDetails` thing. sort of annoying we'll have split paths

api/v1/server/authz/payloads.go
grutt: yeah i thought of this... the issue becomes who sets this permission. i decided owners and admins can set it. so if they can set it they might as well see it...

mrkaye97: I wonder if adding this to the audit logs could help here...? I see what you mean though - I'm thinking maybe you can audit when someone toggled this for themselves and that prevents the risk of an admin toggling it on and off?

api/v1/server/authz/payloads.go
grutt: confusing, but this is the api token case which needs to default to true. added a comment.

api/v1/server/authz/payloads.go
grutt: good idea on adding this to the audit log, but i think we should keep it like this for now and change it someone asks. 

midterm goal here is i think we should allow customers to create their own roles including this permission....

mrkaye97: works for me 👍

api/v1/server/authz/payloads.go
juliusgeo: ah gotcha

PR #4719, opened 2026-08-17
fix(engine): prevent POWER overflow when computing retry backoff

grutt:
# Description

`v1_task_update_function` computed `retry_after` as `LEAST(retry_max_backoff, POWER(backoff_factor, app_retry_count))`. `POWER` is evaluated first and overflows float8 (`SQLSTATE 22003`) once `backoff_factor ^ app_retry_count` exceeds ~1e308, so `LEAST` never caps the value.

That rolls back `FailTaskAppFailure`, nacks `task-failed` forever, and also breaks timeout processing for the same tenant. Cap the exponent in log space before `POWER` and fall back to `retry_max_backoff`.

## Type of change

## What's Changed

- [x] Compute retry backoff in log space in `v1_task_update_function` so `POWER` cannot overflow float8
- [x] Add migration `v1_0_144` to replace the trigger function

## Checklist

Changes have been:

- [x] Linted and formatted

## Test plan

- [ ] Apply `v1_0_144` (or `CREATE OR REPLACE` the function) on a DB that currently overflows
- [ ] Confirm `FailTaskAppFailure` succeeds for a task with backoff and a high `app_retry_count` (e.g. factor 2 with count >= 1024, or factor 1000 with count >= 103)
- [ ] Confirm `v1_retry_queue_item.retry_after` is `now + retry_max_backoff` seconds rather than overflowing
- [ ] Confirm `task-failed` pre-ack errors and "could not list step runs to timeout" stop for the affected tenant

---

<details id="ai-disclosure">

- **Details**: Cursor was used to diagnose the production overflow from engine/pg logs and to write the overflow-safe trigger expression and migration.

</details>


sql/schema/v1-core.sql
```diff
@@ -1404,8 +1404,23 @@ BEGIN
...
             nt.retry_count,
             nt.tenant_id,
-            -- Convert the retry_after based on min(retry_backoff_factor ^ retry_count, retry_max_backoff)
-            NOW() + (LEAST(nt.retry_max_backoff, POWER(nt.retry_backoff_factor, nt.app_retry_count)) * interval '1 second') AS retry_after
+            -- NOTE: cap in log space before POWER. POWER(backoff_factor, app_retry_count)
+            -- overflows float8 (SQLSTATE 22003) for large retry counts, and LEAST cannot
+            -- prevent that because POWER is evaluated first.
+            NOW() + (
+                LEAST(
+                    COALESCE(nt.retry_max_backoff, 86400)::double precision,
```
mrkaye97, line 1414: do we need this `COALESCE`? seems net-new

sql/schema/v1-core.sql
```diff
@@ -1404,8 +1404,23 @@ BEGIN
...
+                    CASE
+                        WHEN nt.retry_backoff_factor <= 1 THEN
+                            POWER(nt.retry_backoff_factor, nt.app_retry_count)
+                        WHEN LN(nt.retry_backoff_factor) * nt.app_retry_count
+                            >= LN(GREATEST(COALESCE(nt.retry_max_backoff, 86400), 1)::double precision) THEN
```
mrkaye97, line 1419: same here, maybe don't need to coalesce?

mrkaye97 (approved):
the fix looks fine, but I needed to write out the log math by hand. it'd be helpful to leave a comment about how this works because saying `b^n > cap` =>> `n * ln(b) > ln(cap)`

sql/schema/v1-core.sql
grutt: both are needed afict. retry_max_backoff is nullable, and without a finite cap postgres would either overflow power or insert a null retry_after

PR #4723, opened 2026-08-17
fix(python): rest client ssl/tls creds, memory leak, durable callback ordering, improved shutdown behavior

### LINKED ISSUE #4635: Python worker shutdown hangs when REST pause fails or never returns
## Problem

`Worker.exit_gracefully()` treats the engine REST pause as a prerequisite for the worker's own teardown. If that best-effort call raises or never returns during SIGTERM, local shutdown stops before the listener stream closes, runners drain, queues close, or the loop stops.

This is present in Python SDK 1.33.12 and still present in current 1.37.1:

- `Worker._pause_task_assignment()` awaits `self._client.workers.aio_pause(worker_id)` without a guard.
- `Worker.exit_gracefully()` awaits `_pause_task_assignment()` before every local teardown step.
- `WorkersClient.aio_pause()` is `asyncio.to_thread(self.pause, worker_id)` with no shutdown-scoped timeout.
- The SIGTERM handler schedules `exit_gracefully()` with `create_task()` and never awaits/observes that task, so an exception abandons shutdown.

Source (1.37.1):
- https://github.com/hatchet-dev/hatchet/blob/py/1.37.1/sdks/python/hatchet_sdk/worker/worker.py#L681-L685
- https://github.com/hatchet-dev/hatchet/blob/py/1.37.1/sdks/python/hatchet_sdk/worker/worker.py#L744-L790
- https://github.com/hatchet-dev/hatchet/blob/py/1.37.1/sdks/python/hatchet_sdk/features/workers.py#L92-L112

## Observed impact

During an engine/DNS failure, logs stopped after `pausing task assignment...`; `task assignment paused` never appeared, `worker.start()` never returned, and the worker kept accepting work after it had announced draining. A 503, timeout, engine roll, or other REST error has the same control-flow result.

## Expected behavior

Pausing task assignment should be finite and best-effort during shutdown. If it times out or fails, log/surface the error and continue the worker's local teardown and runner drain. The engine courtesy must not be a precondition for closing the worker itself.

A regression test should inject an exception and a never-returning call from `aio_pause()` and prove `exit_gracefully()` still reaches runner drain, queue/listener teardown, and loop stop.

mrkaye97:
# Description

Pass tls config through to the api client like we do for gRPC. Also wraps task assignment pausing in a timeout and handles possible errors there so we don't hang on shutdown.

Fixes #4635

Also merged in #4736 which fixes a possible memory leak
Also merged in #4733 which has a fix for durable callback ordering on the SDK

## Type of change

## Checklist

Changes have been:

- [x ] Linted and formatted

---

<details id="ai-disclosure">

No AI

</details>


examples/python/cron/cron_input.py
```diff
@@ -16,7 +16,7 @@ class CronInput(BaseModel):
 
 @hatchet.task(
     input_validator=CronInput,
-    on_crons=["* * * * *"],
+    on_crons=["0 0 * 1 0"],
```
juliusgeo: why are these included?

juliusgeo (approved):

PR #4737, opened 2026-08-19
fix(sdks/python): retry aio_push/aio_bulk_push on retryable grpc errors

### LINKED ISSUE #4605: [BUG] Python SDK: tenacity_retry does nothing on aio_push
**Describe the issue**

`tenacity.retry` dispatches on `iscoroutinefunction`. A `grpc.aio` multicallable isn't one — it's a sync callable returning an awaitable — so tenacity returns the un-awaited call and never sees the failure. `aio_push` gets 1 attempt, not `max_attempts`, even for `UNAVAILABLE`.

- `clients/events.py` has both patterns: `aio_stream` wraps `_put_stream_event` (an `async def`) and retries fine;
- `aio_push` wraps `aio_client.Push` directly and doesn't.
- `aio_bulk_push` has the same shape. `admin.py` is unaffected — it wraps sync stubs via `asyncio.to_thread`.

**Environment**
- SDK: Python v1.35.1 (unchanged on `main` as of 2026-08-03)
- Engine: Self-hosted

**Expected behavior**

`aio_push` retries retryable statuses up to `max_attempts`, like sync `push` does.

**Code to Reproduce**

```python
cfg = TenacityConfig()                            # max_attempts = 5
await tenacity_retry(some_async_def, cfg)()        # 5 calls
await tenacity_retry(aio_stub.SomeMethod, cfg)()   # 1 call
```

**Additional context**

Found during an engine rollout: client got `UNAVAILABLE ... GOAWAY ...
graceful_stop`, which `tenacity_should_retry` says is retryable, but it wasn't
retried — it propagated into our handler and 500'd a live request.

Fix: wrap an `async def` like `_put_stream_event` does, or route
awaitable-returning callables through `AsyncRetrying` so it can't recur.

---
<details id="ai-disclosure">

- **Details**: Claude Code (Opus) — diagnosis and initial draft. Reviewed and revised by me over several iterations before submission.
</details>

harsh-sikarwar:
## Summary

`aio_push` and `aio_bulk_push` were only ever making **1 attempt**, never
retrying, even on retryable gRPC errors like `UNAVAILABLE`.

`tenacity_retry` (in `tenacity_utils.py`) picks between the sync and async
retry loop using `inspect.iscoroutinefunction(func)`. `aio_client.Push` and
`aio_client.BulkPush` are `grpc.aio` generated methods — callables that
*return* an awaitable, but aren't themselves coroutine functions. So
`iscoroutinefunction` returned `False`, tenacity used the **sync** `Retrying`
loop, the call returned a `UnaryUnaryCall` object without raising, and
tenacity considered the attempt successful. The real `AioRpcError` only
surfaced later when the caller awaited the result — outside the retry loop
entirely.

This was caught in production during an engine rollout: a transient
`UNAVAILABLE`/`GOAWAY` (which `tenacity_should_retry` correctly flags as
retryable) propagated straight up and 500'd a live request instead of being
retried.

`aio_stream` doesn't have this bug, since it already wraps a real `async def`
helper (`_put_stream_event`) rather than the raw stub method directly.

## Fix

Mirrors the existing `_put_stream_event` pattern instead of changing
`tenacity_retry`'s shared dispatch logic (other call sites depend on it):

- Added `_push_event` / `_bulk_push_event` as real `async def` wrappers
  around `aio_client.Push` / `aio_client.BulkPush`.
- Create their `tenacity_retry(...)`-wrapped versions once, in
  `EventClient.__init__`.
- `aio_push` / `aio_bulk_push` now await those wrapped callables instead of
  wrapping the raw stub method on every call.

Since the wrappers are genuine coroutine functions, `iscoroutinefunction`
now correctly returns `True`, so tenacity uses `AsyncRetrying` and retries
actually happen up to `max_attempts`.

## Testing

Added `tests/unit/test_aio_push_retry.py` with regression tests for both
`aio_push` and `aio_bulk_push`, verifying:

- Retries up to `max_attempts` on `UNAVAILABLE`
- Does **not** retry on `INVALID_ARGUMENT` (non-retryable)
- Final error propagates correctly once attempts are exhausted

Ran `pytest`, `ruff check`, `black --check`, and `mypy` — all clean, no
regressions in the existing suite (57/57 passing).

Closes #4605

## AI Disclosure

I acknowledge that an LLM was used in the creation of this PR, in accordance
with Hatchet's `AI_POLICY.md`.

Details: Claude was used to help diagnose the root cause, implement the fix,
and draft regression tests. I reviewed the diagnosis, the code changes, and
the tests before submission.


mrkaye97 (approved):
thanks for the fix! removed the tests, didn't think we needed them. will merge once CI finishes spinning

PR #4739, opened 2026-08-19
chore: enable filtered Prometheus scrapes

igor-kupczynski:
## Summary

- Upgrade `github.com/prometheus/client_golang` from v1.23.2 to v1.24.1.
- Allow `/metrics` requests with repeated `name[]` query parameters to encode only the requested metric families
- Use v1.24.1 instead of v1.24.0 to include the promhttp nil-URL panic fix.

## Rollout

This PR only upgrades the library. It does not change Alloy or any other scrape configuration. Older engines ignore `name[]` and return the full metrics response, so Alloy must not send the parameter until this version has shipped.

## Test plan

- [x] `go test ./pkg/integrations/metrics/prometheus/...`
- [x] `go test ./cmd/hatchet-engine/...`


go.mod
```diff
@@ -253,7 +253,7 @@ require (
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/pelletier/go-toml/v2 v2.2.4 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
-	github.com/prometheus/client_golang v1.23.2
```
igor-kupczynski, line 256: This is the main cause for the PR

grutt (approved):

PR #4742, opened 2026-08-19
feat: add support for embedded mode to TS and Python SDKs

mnafees:
# Description

Adds support for embedded mode to TS and Python SDKs. The PR also adds the `hatchet embedded-ui` command.

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

</details>


sdks/typescript/src/v1/embedded.ts
```diff
@@ -0,0 +1,271 @@
...
+}
+
+async function ensureSidecarBinary(version?: string, checksum?: string): Promise<string> {
+  const tag = await resolveVersion(version);
+  const asset = sidecarAssetName();
+  const binPath = path.join(os.homedir(), '.hatchet', 'embedded', tag, asset);
+
+  // verified on every start, not just at download; a cached binary that no
+  // longer matches the expected checksum is re-downloaded
+  const expected = checksum ?? (await expectedChecksum(tag, asset));
```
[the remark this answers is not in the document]
mnafees: Intentional for the default path: the checksum guards integrity, not channel trust. A compromised GitHub release is the same trust model as a compromised npm/PyPI package. Users who need an independent anchor can pin version + checksum, or supply binary_path. Once the first hatchet-embedded release is cut, we'll pin default sidecar checksums inside the SDK packages.

sdks/typescript/package.json
```diff
@@ -57,12 +57,20 @@
...
     "@aws-sdk/client-s3": "^3.1042.0",
     "@eslint/js": "^10.0.1",
+    "@grpc/grpc-js": "^1.14.3",
+    "@modelcontextprotocol/sdk": "^1.29.0",
+    "@openai/agents": "0.14.2",
+    "@opentelemetry/api": "^1.9.0",
+    "@opentelemetry/core": "^2.0.0",
+    "@opentelemetry/exporter-trace-otlp-grpc": "^0.221.0",
+    "@opentelemetry/instrumentation": "^0.221.0",
+    "@opentelemetry/sdk-trace-base": "^2.0.0",
```
abelanger5: why are these moved into `devDependencies`?

cmd/hatchet-cli/cli/embedded_ui.go
```diff
@@ -0,0 +1,452 @@
...
+		Embedded bool `json:"embedded"`
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(&meta); err != nil {
+		return fmt.Errorf("could not parse /api/v1/meta response from %s: %w", target, err)
+	}
+
+	if !meta.Embedded {
+		return fmt.Errorf("%s is not an embedded Hatchet instance; 'hatchet embedded-ui' only serves the UI for embedded instances", target)
+	}
```
abelanger5: should we not be using the generated openapi client here?

sdks/python/hatchet_sdk/embedded.py
```diff
@@ -0,0 +1,316 @@
...
+    sidecar = start_embedded_sidecar(options)
+
+    os.environ["HATCHET_CLIENT_TOKEN"] = sidecar.token
+    os.environ["HATCHET_CLIENT_TENANT_ID"] = sidecar.tenant_id
+    os.environ["HATCHET_CLIENT_HOST_PORT"] = sidecar.grpc_address
+    os.environ["HATCHET_CLIENT_TLS_STRATEGY"] = "none"
+    if sidecar.api_url:
+        os.environ["HATCHET_CLIENT_SERVER_URL"] = sidecar.api_url
+
+    server_url = {"server_url": sidecar.api_url} if sidecar.api_url else {}
```
mrkaye97: doing this and then unpacking it is really bizarre 😅

sdks/python/hatchet_sdk/embedded.py
```diff
@@ -0,0 +1,316 @@
...
+
+        time.sleep(0.2)
+
+    process.kill()
+    raise TimeoutError(
+        f"hatchet embedded sidecar did not become ready within {timeout_seconds}s"
+    )
+
+
+def HatchetEmbedded(options: EmbeddedOptions | None = None) -> "Hatchet":  # noqa: N802
```
mrkaye97: this isn't very idiomatic (hence why the linter is complaining), and I think we shouldn't do it. It's an issue for a couple reasons, but the main one is that we use `pydantic-settings` to manage config, and manually setting / updating environment variables like this is absolutely an anti-pattern in that realm.

I'd need to think a bit more about the DX here. The problem with this is that it'll run on import, which may not be desirable (idk) - it's sort of a problem with our client setup in general... I wonder if making this a `classmethod` or something on the `Hatchet` client itself would be marginally better though, so that we could do:

```
hatchet = Hatchet.from_embedded(...)
```
 
or some such - WDYT?

sdks/python/hatchet_sdk/embedded.py
```diff
@@ -0,0 +1,316 @@
...
+    # worker subprocesses re-import the main module; connect them to the
+    # parent's engine (via the env vars exported below) instead of booting
+    # a second one. parent_process() is None while a spawn child is still
+    # importing the main module, so also check the _inheriting flag set
+    # during that phase.
+    in_child = multiprocessing.parent_process() is not None or getattr(
+        multiprocessing.current_process(), "_inheriting", False
+    )
+    if in_child:
+        return Hatchet()
```
mrkaye97: initializingthe client here (and also below) means that the user doesn't get to pass any config in, except what they have set in env vars, which I think is pretty likely to break. For instance:

1. If they have `HATCHET_CLIENT_TOKEN` set as an env var pointing at e.g. Cloud, won't this line pick up that token and use Cloud instead of the embedded engine?
2. If they want to provide other config, like for logging, retries, etc. etc., they can't AFAICT

sdks/python/hatchet_sdk/embedded.py
```diff
@@ -0,0 +1,316 @@
...
+from pydantic import BaseModel
+
+if TYPE_CHECKING:
+    from hatchet_sdk.hatchet import Hatchet
+
+REPO_URL = "https://github.com/hatchet-dev/hatchet-embedded"
+DEFAULT_READY_TIMEOUT_SECONDS = 300.0
+
+
+class EmbeddedOptions(BaseModel):
```
mrkaye97: I don't think we should do this if it's not using `pydantic-settings`. We should either pass these as args directly to the function taking them, or (preferably), we should add a new `embedded` set of pydantic-settings on the `config.py` which will let you load these from an env var, e.g. `HATCHET_CLIENT_EMBEDDED_VERSION`, etc.

sdks/python/hatchet_sdk/embedded.py
```diff
@@ -0,0 +1,316 @@
...
+    return f"hatchet-embedded-sidecar_{system}_{arch}"
+
+
+class _NoRedirectHandler(urllib.request.HTTPRedirectHandler):
+    def redirect_request(self, *args: object, **kwargs: object) -> None:
+        return None
+
+
+def _resolve_version(version: str | None) -> str:
+    requested = version or os.environ.get("HATCHET_EMBEDDED_VERSION") or "latest"
```
mrkaye97: we shouldn't read from env vars directly, this should come from pydantic settings

mrkaye97 (commented):
left some comments on the python side!

sdks/python/hatchet_sdk/embedded.py
```diff
@@ -0,0 +1,340 @@
...
+    checksum_file = bin_path.with_name(asset + ".sha256")
+
+    # fall back to the checksum cached at download time so a pinned, already
+    # verified binary still starts when GitHub is unreachable
+    try:
+        expected = _expected_checksum(tag, asset)
+    except (urllib.error.URLError, OSError):
+        if bin_path.exists() and checksum_file.exists():
+            return checksum_file.read_text().strip()
+        raise
```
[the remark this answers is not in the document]
mnafees: This case is invalid since this PR is going to start the release process.

cmd/hatchet-cli/cli/embedded_ui.go
```diff
@@ -0,0 +1,451 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("could not bind to %s:%d: %w", host, port, err)
+		}
+
+		return ln, nil
+	}
+
+	const base = 8080
+
+	for p := base; p < base+100; p++ {
```
juliusgeo: there shouldn't really be any downside to increasing this, right, because it will early exit? Could we make this `1000` rather than 100 just in case someone is using a ton of ports already, or make it configurable?

juliusgeo (approved):

PR #4744, opened 2026-08-19
fix: return NotFound when RefreshTimeout misses a task runtime

grutt:
## Summary
- `Dispatcher.RefreshTimeout` currently returns raw `pgx.ErrNoRows` when `v1_task_runtime` is already gone (complete/cancel/timeout). The gRPC error interceptor treats that as Unknown, logs ERR, alerts, and rewrites it to Internal.
- Map that miss to `NotFound` so the SDK gets the real error, Go/Python clients stop retrying Internal, and we no longer error-log or alert on this path.

## Test plan
- [ ] Trigger `RefreshTimeout` for a task with no `v1_task_runtime` row and confirm the client receives `NotFound` (`task run not found`)
- [ ] Confirm engine logs no longer emit `error="no rows in result set"` / `grpc.code=Internal` for that RPC
- [ ] Confirm a live running task can still refresh its timeout successfully


cmd/hatchet-migrate/migrate/migrations/20260819154500_v1_0_146.sql
grutt, line 1: stupid ai

igor-kupczynski: what's going on here?

igor-kupczynski (approved):
looks good to me (- accidental commit of the migration?)

cmd/hatchet-migrate/migrate/migrations/20260819154500_v1_0_146.sql
grutt: instead of fixing the broken test it decided to add a file to make the test happy lol

igor-kupczynski: lol

PR #4757, opened 2026-08-20
fix: return NotFound for missing rows instead of Internal

grutt:
# Description

Map `pgx.ErrNoRows` to gRPC `NotFound` (and HTTP 404 for traces) instead of rewriting it to Internal. A missing task, runtime, or trace is an expected client miss; Internal caused SDK retries and error-log noise.

## Type of change

## What's Changed

- gRPC error interceptor maps `ErrNoRows` to `NotFound` without error-logging or alerting
- `PutLog`, `PutStreamEvent`, `RegisterDurableEvent`, `ReleaseSlot`, and `RefreshTimeout` return `NotFound` when the task or runtime is gone
- Traces API returns 404 when the run has no trace
- Interceptor tests for no-rows, cancel, Internal+alert, and pass-through status

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Cursor agent used to implement the interceptor/handler mapping and tests, and to draft this description.

</details>


internal/services/dispatcher/server_v1.go
```diff
@@ -44,6 +43,10 @@ func (d *DispatcherServiceImpl) RegisterDurableEvent(ctx context.Context, req *c
 	task, err := d.repo.Tasks().GetTaskByExternalId(ctx, tenantId, taskId, false)
 
 	if err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			return nil, status.Errorf(codes.NotFound, "task run not found: %s", taskId)
```
grutt, line 49: @mrkaye97 do we want to retry this internally a few times with a small backoff? I'm confused how we'd even hit this case but we are.

internal/services/grpc/middleware/error.go
```diff
@@ -27,6 +25,12 @@ func contextErrorStatus(err error) error {
 		return status.Error(codes.DeadlineExceeded, "request deadline exceeded")
 	}
 
+	if goerrors.Is(err, pgx.ErrNoRows) {
+		// A missing row is a client-visible miss (task already complete, webhook
+		// deleted, etc.), not an engine failure.
+		return status.Error(codes.NotFound, "not found")
+	}
```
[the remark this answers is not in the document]
grutt: I think this is correct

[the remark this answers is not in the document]
grutt: ok traced this, this should not be a factor.

internal/services/dispatcher/server_v1.go
mrkaye97: maybe this could happen after the retention window?

mrkaye97 (approved):

internal/services/dispatcher/server_v1.go
grutt: good guess.. its probably fine...

PR #4762, opened 2026-08-20
fix(engine): durable dag optimizations

mrkaye97:
# Description

Adding a couple simple optimizations that should help some on the dag path:

1. Caching step match conditions and workflow versions so we don't look them up incessantly.
2. Run started event send in a goroutine to not block
3. Get rid of a bunch of update churn on the `Worker`

## Type of change

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: Used the o11y skill to help figure out some of these bottlenecks

</details>

### DIFF
--- a/pkg/operator/dagoperator/operator.go
+++ b/pkg/operator/dagoperator/operator.go
@@ -290,6 +290,12 @@ func (d *DAGOperator) run(action *contracts.AssignedAction) error {
 		attribute.Int("dagoperator.durable_invocation_count", int(action.GetDurableTaskInvocationCount())),
 	)
 
+	externalId, err := uuid.Parse(action.TaskRunExternalId)
+
+	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/operator/dagoperator/operator.go
```diff
@@ -290,15 +290,26 @@ func (d *DAGOperator) run(action *contracts.AssignedAction) error {
...
+	}()
+
+	awaitStarted := func() { <-startedReported }
+
+	defer awaitStarted()
 
 	externalId, err := uuid.Parse(action.TaskRunExternalId)
 
 	if err != nil {
+		awaitStarted()
```
juliusgeo: do we need this if we already deferred?

mrkaye97: yes, but I'll rearrange - it's because it's a race condition but I think I can just move the uuid parsing above the defer

mrkaye97: fixed!

juliusgeo (approved):

PR #4775, opened 2026-08-22
fix(python): Make action listener process unalive itself when parent is dead

### LINKED ISSUE #4731: [BUG] Python SDK: Dead worker, but action listener lives on
**Describe the issue**

In a python worker, I ran a task that died very suddenly due to some memory corruption thing (in my own code).

However the action listener lived on, leading to a bad state where it kept taking jobs that would then timeout.
It logs the error "THE TIME TO START THE RUN IS TOO LONG, THE EVENT LOOP MAY BE BLOCK. See [...]" many times which lead me to believe that the issue was that the worker was blocked and not dead.

Also the health check presumably keeps saying it's healthy, meaning that in a deployed version it would not get restarted

**Environment**
- SDK:  Python 1.34.1
- Engine: Self-hosted v0.90.0

**Expected behavior**
The action listener should also die when the worker process dies

It's a bit difficult to recreate, if needed I can try. I also know that I'm not on the latest hatchet_sdk version, but it looks to me like this should still be a problem. The action listener will wait for a `STOP_LOOP` message that will never come.

**Additional context**

OS: Ubuntu 25.04.3 LTS running in WSL

Maybe the action_listener should get the pid of the worker process and check if it's still alive?
I feel there must be a fairly standard pattern for this

---

<details id="ai-disclosure">

- NO AI

</details>

StefanBRas:
# Description

Fixes #4731 

Action listener will no longer live on after the parent process is dead.

The PR is more of an RFC now, I'll write some questions/comments directly in the file/review tab

## Type of change

- [ ] Performance improvement (non-breaking change which improves performance)

## Checklist

Changes have been:

 ## Testing

Added a test which I think solves it fairly well.

---

<details id="ai-disclosure">

- **Details**: No AI

</details>


sdks/python/hatchet_sdk/worker/action_listener_process.py
```diff
@@ -350,13 +356,23 @@ async def start(self, retry_attempt: int = 0) -> None:
...
     async def start_event_send_loop(self) -> None:
         while True:
-            event = await self._get_event()
+            try:
+                event = await self._get_event(timeout_seconds=1.0)
+            except Empty:
+                if self._parent_is_dead():
+                    logger.error("stopping event send loop, parent is dead...")
+                    break
+                continue
```
StefanBRas, line 375: The `_parent_is_dead` call could run in the background, but since the logic right now assumes that we will always get the STOP_LOOP event we need to occasionally stop to see if the parent is dead here

The timeout duration is arbitrarily chosen, i have no idea about performance implications (but doubt the `self._parent_is_dead` takes a long time).
If it's a problem it could run in the background and then set an event or something

sdks/python/hatchet_sdk/worker/action_listener_process.py
```diff
@@ -606,23 +622,30 @@ async def run() -> None:
...
                 task.cancel()
                 with contextlib.suppress(asyncio.CancelledError):
                     await task
 
+        # TODO: don't actually merge this, see comment in github
+        if process._parent_is_dead():
+            import os
+
+            os.kill(os.getpid(), signal.SIGKILL)
+
```
StefanBRas, line 650: There's something odd here. If I don't do the os.kill  (even if i do sys.exit) the process will not stop.
I will then see two process still, something like:
```
 129865 pts/1    S      0:00 /home/stefan/dev/hatchet3/sdks/python/.venv/bin/python3 -c from multiprocessing.resource_tracker import main;main(7)
 129892 pts/1    Sl     0:01 /home/stefan/dev/hatchet3/sdks/python/.venv/bin/python3 -c from multiprocessing.spawn import spawn_main; spawn_main(tracker_fd=8, pipe_handle=25) --multiprocessing-fork

```

So it looks like the resource_tracker is keeping it alive. Or maybe it is, because it also looks like it never returns from the `asyncio.run(run())` below somehow. If I put a log statement after it, that doesn't get fired

With the os.kill it will still say something like `/usr/lib/python3.10/multiprocessing/resource_tracker.py:224: UserWarning: resource_tracker: There appear to be 14 leaked semaphore objects to clean up at shutdown
`

StefanBRas (commented):
Also generally I'm not sure how you would prefer the shutdown sequence when the parent dies. I tried to make it as little of a change as possible and piggyback of the standard shutdown sequence

sdks/python/hatchet_sdk/worker/action_listener_process.py
```diff
@@ -184,6 +185,11 @@ async def _monitor_event_loop(self) -> None:
             ):
                 self._event_loop_blocked_since = None
 
+    def _parent_is_dead(self) -> bool:
+        if (parent_process := multiprocessing.parent_process()) is None:
+            return False
```
StefanBRas, line 190: I think this should actually return `True`, but I guess now there's a guard against it being killed if it's run in a non-multiprocess context.

Will have to check the implications of `.parent_process()` being null

abelanger5:
Thanks @StefanBRas ! Adding @mrkaye97 here to take a look 👍

mrkaye97:
hey @StefanBRas, thanks for the fix, also not sure why this was auto-closed - I'll look into that

mrkaye97:
hey @StefanBRas, merging this into my branch to get it over the line - thanks for the fix!

PR #4790, opened 2026-08-25
feat: count live crons, scheduled runs, and webhooks per tenant

grutt:
## Summary
- Add per-tenant live counts for enabled latest-version crons, pending scheduled runs, and incoming webhooks (used by shard billing).
- BestEffort-notify after create/delete, and after cron enable/disable, so allocated usage can be recounted without waiting for the hourly sync.
- Count crons from `WorkflowTriggerCronRef` with a newer-version anti-join instead of `DISTINCT ON` every workflow version.

## Test plan
- [ ] `TestCountAllocatedResourcesByTenant` covers disabled API crons, pending vs fired schedules, and webhooks
- [ ] Create/delete a cron, schedule, and webhook and confirm the allocated-resource callback fires
- [ ] Enable/disable a cron and confirm the same callback fires


pkg/repository/workflow_schedules.go
```diff
@@ -159,6 +170,23 @@ func newWorkflowScheduleRepository(shared *sharedRepository) WorkflowScheduleRep
 	}
 }
 
+func (w *workflowScheduleRepository) RegisterAllocatedResourceChangeCallback(callback TenantScopedCallback[struct{}]) {
+	w.allocatedChangeCallbacks = append(w.allocatedChangeCallbacks, callback)
+}
```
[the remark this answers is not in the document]
grutt: Dismissing: this is the library surface. Cloud registers the callbacks in hatchet-api and hatchet-engine (`RegisterAllocatedResourceChangeCallbacks`); self-host leaves the list empty on purpose because `BillingHatchetClient` is nil. See https://github.com/hatchet-dev/hatchet-cloud/pull/474

pkg/repository/workflow_schedules.go
```diff
@@ -570,6 +626,7 @@ func (w *workflowScheduleRepository) CreateCronWorkflow(ctx context.Context, ten
 		return nil, fmt.Errorf("failed to fetch cron workflow")
 	}
 
+	w.notifyAllocatedResourceChange(tenantId)
```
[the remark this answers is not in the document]
grutt: Dismissing: worker-registered DEFAULT crons and ticker `DeleteInvalidCron` are intentionally hourly-only. On-change notify is for API create/delete/enable-disable; DEFAULT churn on workflow-version publish would stampede recounts, and the hourly sync is the backstop.

pkg/repository/sqlcv1/allocated_resources.sql
```diff
@@ -0,0 +1,76 @@
+-- name: CountAllocatedResourcesByTenant :many
+-- One row per tenant with live allocated-resource counts on this shard.
+-- cron_count: enabled, non-deleted cron refs on the latest non-deleted
+-- workflow version (API and DEFAULT).
+-- scheduled_run_count: pending scheduled refs only (not deleted, and no
+-- WorkflowRunTriggeredBy.scheduledId — fired runs are excluded).
+-- webhook_count: row count on v1_incoming_webhook.
+-- NULL tenantIds counts every tenant on the shard; otherwise only those IDs.
```
juliusgeo: this seems like a footgun that could lead to overcounting usage lol, when would we want to count every tenant on a shard rather than passing in every tenant for an org?

grutt: good catch

grutt: sorry this is confusing but it is needed this way for one of the fanouts, it doesnt sum in the query. will update the comment

juliusgeo (approved):

PR #4804, opened 2026-08-26
fix(scheduler): unstick rate-limited tasks holding filled concurrency slots

grutt:
# Description

Tasks that hit a rate limit while holding a filled `GROUP_ROUND_ROBIN` concurrency slot could deadlock their queue permanently: parked queue items were only requeued when the queuer had already observed rate limits in-process, so after a scheduler restart they never came back, and the schedule-timeout pass only deleted unfilled slots, so the filled slots never released. New runs then waited on the full group until they hit their scheduling timeout, indefinitely. Observed on a production tenant.

## Type of change

## What's Changed

- [x] Queuer always calls `RequeueRateLimitedItems` each loop (removed the in-memory `hasRateLimits` gate), so items parked in `v1_rate_limited_queue_items` are requeued even by a fresh queuer after a restart or lease change
- [x] `RunGroupRoundRobin` / `RunChildGroupRoundRobin` schedule-timeout now cancels any slot past `schedule_timeout_at` with no `v1_task_runtime` row (not actually running), instead of only unfilled slots
- [x] Add queuer unit tests and an integration test reproducing the filled-slot deadlock (all fail without the fix)

## Checklist

Changes have been:

- [x] Linted and formatted

## Testing

- `go test ./pkg/scheduling/v1/ -run TestQueuerRequeuesRateLimitedItems`
- `go test -tags integration ./pkg/scheduling/v1/` (full suite, all 22 tests pass; the new test and unit tests fail on `main`)

---

<details id="ai-disclosure">

- **Details**: Root-cause analysis, fix, and tests written with Cursor (Claude), reviewed by a human.

</details>

### DIFF
--- a/pkg/repository/sqlcv1/concurrency.sql
+++ b/pkg/repository/sqlcv1/concurrency.sql
@@ -239,7 +239,16 @@ WITH eligible_slots_per_group AS (
         tenant_id = @tenantId::uuid AND
         strategy_id = @strategyId::bigint AND
         schedule_timeout_at < NOW() AND
-        is_filled = FALSE
+        -- Filled slots are only running if they have a v1_task_runtime row.
+        -- Rate-limited tasks stay is_filled while parked, so they must time out too.
+        NOT EXISTS (
+            SELECT 1
+            FROM v1_task_runtime tr
+            WHERE
+                tr.task_id = v1_concurrency_slot.task_id AND
+                tr.task_inserted_at = v1_concurrency_slot.task_inserted_at AND
+                tr.retry_count = v1_concurrency_slot.task_retry_count
+        )
     ORDER BY
         task_id, task_inserted_at
     FOR UPDATE
--- a/pkg/scheduling/v1/queuer.go
+++ b/pkg/scheduling/v1/queuer.go
@@ -62,8 +62,6 @@ type Queuer struct {
 	unassigned   map[int64]*sqlcv1.V1QueueItem
 	unassignedMu mutex
 
-	hasRateLimits bool
-
 	// consecutiveEmptyPolls counts loop iterations whose refill returned no items. It is only
 	// accessed from the loopQueue goroutine.
 	consecutiveEmptyPolls int
@@ -221,15 +219,8 @@ func (q *Queuer) loopQueue(ctx context.Context) {
 
 		start := time.Now()
 		checkpoint := start
-		var err error
 
-		if q.hasRateLimits {
-			_, err := q.repo.RequeueRateLimitedItems(ctx, q.tenantId, q.queueName)
-
-			if err != nil {
-				q.l.Error().Ctx(ctx).Err(err).Msg("error requeuing rate limited items")
-			}
-		}
+		q.requeueRateLimitedItems(ctx)
 
 		qis, err := q.refillQueue(ctx)
 
@@ -268,10 +259,6 @@ func (q *Queuer) loopQueue(ctx context.Context) {
 
 			q.unackedToUnassigned(qis)
 			continue
-		}
-
-		if len(rls) > 0 {
-			q.hasRateLimits = true
 		}
 
 		rateLimitTime := time.Since(checkpoint)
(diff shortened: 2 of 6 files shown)


pkg/scheduling/v1/queuer_test.go
```diff
@@ -0,0 +1,78 @@
...
+		l:         &l,
+	}
+
+	q.requeueRateLimitedItems(context.Background())
+
+	require.Equal(t, 1, repo.callCount(),
+		"a fresh queuer must requeue parked rate-limited items even when the live queue is empty")
+}
+
+func TestQueuerRequeuesRateLimitedItemsWhenAlreadyObserved(t *testing.T) {
```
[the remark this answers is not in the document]
grutt: Dropped the test instead of adding a godoc. After removing `hasRateLimits`, it was the same path as `TestQueuerRequeuesRateLimitedItemsOnFreshProcess`.

pkg/scheduling/v1/queuer.go
```diff
@@ -221,15 +219,8 @@ func (q *Queuer) loopQueue(ctx context.Context) {
 
 		start := time.Now()
 		checkpoint := start
-		var err error
 
-		if q.hasRateLimits {
```
juliusgeo, line 226: why are we removing this conditional?

pkg/repository/sqlcv1/concurrency.sql
```diff
@@ -239,7 +239,16 @@ WITH eligible_slots_per_group AS (
         tenant_id = @tenantId::uuid AND
         strategy_id = @strategyId::bigint AND
         schedule_timeout_at < NOW() AND
-        is_filled = FALSE
+        -- Filled slots are only running if they have a v1_task_runtime row.
```
juliusgeo, line 242: slightly confused by the second line of this comment--the fix here is making it so that is_filled is predicated on whether a v1_task_runtime row exists, otherwise the slot isn't filled? Isn't this logic an inversion of that, where it will only be set to `is_filled` if there *isnt* a runtime row?

Scalahansolo:
We put together a standalone reproduction for both defects covered by this PR: [hatchet-rate-limit-repro](https://github.com/Scalahansolo/hatchet-rate-limit-repro).

The same tests run against the pinned PR base and this PR’s proposed fix:

| Scenario | PR base | PR fix |
| --- | --- | --- |
| Fresh scheduler recovers an eligible parked task | Bug reproduced 10/10 | Passed 10/10 |
| Expired filled concurrency slot releases the waiting follow-up | Bug reproduced 10/10 | Passed 10/10 |

It uses the real scheduler and PostgreSQL repositories. To make the trigger deterministic, it seeds the parked-task state directly and constructs a fresh scheduling pool; it does not submit through an SDK or kill a full engine process. The recovery test checks that the queuer completed assignment passes, then confirms a direct repository call can recover the stuck row. The timeout test checks cancellation of the expired holder and queuing of the follow-up.

Run `python3 reproduce.py` after cloning; Docker, Git, and Python are the only local prerequisites. The repo includes [recorded test evidence](https://github.com/Scalahansolo/hatchet-rate-limit-repro/tree/main/evidence), and the [comparison also passed on a clean GitHub Actions runner](https://github.com/Scalahansolo/hatchet-rate-limit-repro/actions/runs/34620378605).

This gives us a repeatable check for both halves of the fix. Hope it helps with review and verification.

pkg/scheduling/v1/queuer.go
grutt: hasRateLimits starts false and only flips after the queuer sees rate limits on items already in the live queue. Parked tasks are not in that queue, so after a restart the flag stays false and they are never requeued. They keep their filled slots and the queue stays stuck. Requeuing every tick is what picks those rows back up once requeue_after has passed.

pkg/repository/sqlcv1/concurrency.sql
grutt: This doesnt set is_filled. It only picks which expired slots get deleted as SCHEDULING_TIMED_OUT. Filling still happens in updated_slots, and that still requires is_filled = FALSE.

The old check treated filled as running, so a parked rate-limited task (is_filled, no v1_task_runtime row) never timed out. Now any expired slot that is not actually running times out. A real run has a runtime row, so it is left alone.

juliusgeo (approved):

PR #4807, opened 2026-08-26
perf(engine): add tenant/time index on v1_statuses_olap

igor-kupczynski:
# Description

Queries that filter `v1_statuses_olap` by tenant and time range (tenant status metrics, task point metrics) have no supporting index, so each call reads far more of the table than it should. The existing indexes (`(external_id, inserted_at)` and time-only indexes) do not help tenant-scoped reads, so every dashboard query scans all tenants' rows in the time range. This adds an index on `(tenant_id, inserted_at)`.

Some deployments run `v1_statuses_olap` as a TimescaleDB hypertable, and TimescaleDB does not support `CREATE INDEX CONCURRENTLY` on hypertables (timescale/timescaledb#504), so a plain SQL migration would fail there and block the whole migration run. The migration is therefore a Go migration that branches:

- On plain Postgres it uses `CREATE INDEX CONCURRENTLY IF NOT EXISTS`. It also checks `pg_index.indisvalid` first and drops an INVALID leftover from a previously failed concurrent build, which `IF NOT EXISTS` alone would silently keep.
- On a hypertable it uses `CREATE INDEX ... WITH (timescaledb.transaction_per_chunk)`, which builds the index one chunk at a time instead of locking the whole hypertable. Since the create is idempotent, operators of very large hypertables can pre-create the index out of band at a time of their choosing and the migration becomes a no-op.
- Down mirrors the branching, since `DROP INDEX CONCURRENTLY` is also unsupported on hypertables.

The plain `CREATE INDEX` in `sql/schema/v1-olap.sql` is unchanged and applies to fresh installs.

I considered a covering variant with `INCLUDE (readable_status)` and rejected it. Updates to this table change `readable_status`, so putting it in the index disables HOT updates and roughly triples WAL volume, while the index-only-scan benefit disappears whenever the visibility map is stale. The plain two-column index gets nearly all of the win without the write amplification.

## Type of change

## Checklist

Changes have been:

## Testing

Tests run the migration against real Postgres and TimescaleDB containers and cover both branches, including the invalid-index rebuild path on plain Postgres.

---

<details id="ai-disclosure">

- **Details**: A Cursor agent (Claude) wrote the Go migration, the tests, and this description, and benchmarked the rejected `INCLUDE` variant. I reviewed the changes before submitting.

</details>


cmd/hatchet-migrate/migrate/migrations/20260826195354_v1_0_149.go
```diff
@@ -0,0 +1,138 @@
...
+			"CREATE INDEX IF NOT EXISTS %s ON v1_statuses_olap (tenant_id, inserted_at) WITH (timescaledb.transaction_per_chunk)",
+			statusesOlapTenantInsertedAtIndexName,
+		))
+
+		if err != nil {
+			return fmt.Errorf("failed to create index %s on hypertable: %w", statusesOlapTenantInsertedAtIndexName, err)
+		}
+
+		return nil
+	}
```
igor-kupczynski, line 58: For reviewers: is this the right way to go with timescaleDB?

cmd/hatchet-migrate/migrate/migrations/20260826195354_v1_0_149_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	_ "github.com/jackc/pgx/v5/stdlib"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/modules/postgres"
+	"github.com/testcontainers/testcontainers-go/wait"
+)
+
+func setupContainerWithStatusesOlap(t *testing.T, image string) *sql.DB {
```
igor-kupczynski, line 19: are these tests useful at all or slop?

igor-kupczynski:
Closing for now, the consensus is that we want a larger rework of OLAP instead of optimizing it.

PR #4809, opened 2026-08-26
perf(engine): rewrite durable event queries to use existing indexes

igor-kupczynski:
# Description

Completing one child of a durable task got more expensive as the parent grew. Two queries on that path re-read every event or log entry of the parent on each call, because their shape stopped the planner from using indexes that already exist.

`LockSignalCreatedEvents` joined the unnested input on only `(task_id, task_inserted_at)` and filtered `event_key` late, so a parent with ~6k signal events read all of them to lock a handful of rows. The query now takes flat predicates for one parent with `event_key = ANY(...)` and the planner uses the unique index on `(tenant_id, task_id, task_inserted_at, event_type, event_key)`. On that same ~6k-event parent this cut buffer reads per call by ~840x. A new Go helper groups a batch by parent and runs one lookup per parent in sorted order, so batches spanning multiple parents behave as before.

`UpdateDurableEventLogEntriesSatisfied` filtered with a 4-column `IN (SELECT ... FROM inputs)`, which scanned all of the parent's log entries through a 2-column index prefix before joining. It now joins the de-duplicated inputs on all four primary key columns, so each input row probes the primary key directly.

## Type of change

## Checklist

Changes have been:

## Testing

New tests run both queries against a real Postgres. They cover batches spanning multiple parents, keys requested under one parent not matching events of another, and satisfied-order assignment, including that re-satisfying an entry keeps its original order.

---

<details id="ai-disclosure">

- **Details**: A Cursor agent (Claude) wrote the query rewrites, the tests, and this description, and ran the buffer-read measurements. I reviewed the changes before submitting.

</details>

### DIFF
--- a/pkg/repository/durable_event_log_satisfied_test.go
+++ b/pkg/repository/durable_event_log_satisfied_test.go
@@ -1,0 +1,142 @@
+//go:build !e2e && !load && !rampup && !integration
+
+package repository
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/hatchet-dev/hatchet/pkg/repository/sqlchelpers"
+	"github.com/hatchet-dev/hatchet/pkg/repository/sqlcv1"
+)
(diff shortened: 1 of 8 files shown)


igor-kupczynski:
after internal discussion: this doesn't make much sense for now; closing

PR #4816, opened 2026-08-27
test(python): add repro for durable non-determinism errors after eviction when children use thread/process offloading

grutt:
# Description

Adds a reproduction for reported non-determinism errors on the evictable ("virtualized") durable engine: a durable parent running concurrent child spawns via `asyncio.gather` fails on replay after eviction+restore with `non-determinism error in task … at node N:1` when the non-durable children (and the code between spawns) use `asyncio.to_thread` / `ProcessPoolExecutor` — even though all child inputs are fully deterministic. Reproduces at roughly a 25% rate per eviction/restore cycle on current `main` with SDK 1.38.1 (which already includes the #4733 callback-ordering fix); the same shape without thread/process offloading passes.

## Type of change

## What's Changed

- Add `examples/bug_tests/durable_child_thread_nondeterminism/worker.py`: durable parent gathering 8 two-stage child pipelines; children offload to `asyncio.to_thread` and a `ProcessPoolExecutor`, with a thread hop between the two durable spawns; eviction policy with a 5s TTL
- Add `driver.py`: triggers runs, restores on every eviction (and optionally replays completed runs), and tallies non-determinism failures across `threads` / `nothreads` / `evict` / `evict-nothreads` modes

## Checklist

Changes have been:

- [x] Tested (manually: `poetry run python -m examples.bug_tests.durable_child_thread_nondeterminism.driver 10 evict` against a local engine; 2/10 and 2/6 iterations fail with non-determinism errors, 0/6 without thread offloading)
- [x] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: repro worker/driver and PR description written with LLM assistance (Cursor); failure rates verified by running the driver locally

</details>

### DIFF
--- a/sdks/python/examples/bug_tests/durable_child_thread_nondeterminism/driver.py
+++ b/sdks/python/examples/bug_tests/durable_child_thread_nondeterminism/driver.py
@@ -1,0 +1,139 @@
+"""Driver for the durable thread/process-pool nondeterminism repro.
+
+Per iteration:
+1. trigger document_pipeline
+2. while it runs, restore it whenever the engine evicts it (each restore
+   replays the durable function from the top against the recorded event log)
+3. after completion, replay the whole run once more via the runs API
+4. record any failure containing "non-determinism"
+
+Usage:
+    poetry run python -m examples.bug_tests.durable_child_thread_nondeterminism.driver [iterations] [mode]
+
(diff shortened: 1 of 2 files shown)


mrkaye97:
gonna close this so I don't confuse myself

PR #4820, opened 2026-08-27
fix(engine): bind workflow event subscribe before OLAP wait

grutt:
# Description

Bind the tenant stream for `SubscribeToWorkflowEvents` before waiting on OLAP so chunks published during hydration are not dropped. When the run is already terminal, send a hangup event instead of returning an empty OK.

## Type of change

## What's Changed

- [x] Attach the tenant fanout subscribe before the OLAP poll on workflow-run event streams
- [x] Send a hangup `WorkflowEvent` (COMPLETED/FAILED/CANCELLED) instead of silently closing when OLAP already shows a terminal status
- [x] Log run id and status on that path, and add span attributes for run id and status
- [x] Add unit coverage for terminal status mapping and hangup release through the stream buffer

## Checklist

Changes have been:

- [x] Linted and formatted

---

<details id="ai-disclosure">

- **Details**: implemented the bind-first subscribe reorder, hangup-on-terminal path, and unit tests

</details>

### DIFF
--- a/internal/services/dispatcher/server.go
+++ b/internal/services/dispatcher/server.go
@@ -2222,8 +2222,7 @@ func (s *DispatcherImpl) subscribeToWorkflowEventsByWorkflowRunIdV1(workflowRunI
 			msgId,
 			payloads,
 			func(events []*contracts.WorkflowEvent) ([]*contracts.WorkflowEvent, error) {
-				workflowRunIds := make([]uuid.UUID, 0)
-				workflowRunIdsToEvents := make(map[string][]*contracts.WorkflowEvent)
+				res := make([]*contracts.WorkflowEvent, 0, len(events))
 
 				for _, e := range events {
 					wri, err := uuid.Parse(e.WorkflowRunId)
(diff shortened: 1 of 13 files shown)


internal/services/dispatcher/tenant_stream_events.go
```diff
@@ -107,15 +116,62 @@ var workflowEventConverters = map[string]func(payloads [][]byte) []*contracts.Wo
...
 
 	if convert, ok := workflowEventConverters[msgId]; ok {
-		workflowEvents = convert(payloads)
+		var err error
+
+		workflowEvents, err = convert(payloads)
+
+		if err != nil {
+			return nil, err
+		}
```
mrkaye97, line 174: nit: kind of an awkward diff? maybe I'm missing something


```suggestion
		workflowEvents, err := convert(payloads)

		if err != nil {
			return nil, err
		}
```

internal/services/dispatcher/server.go
```diff
@@ -714,9 +715,16 @@ type StreamEventBuffer struct {
 	cancel                    context.CancelFunc
 	timeoutDuration           time.Duration
 	gracePeriod               time.Duration
+	hangupQuietPeriod         time.Duration
```
mrkaye97, line 718: what's this? like how long the stream needs to be quiet before we hang up?

mrkaye97 (approved):

internal/services/dispatcher/tenant_stream_events.go
juliusgeo: yeah it's awkward but necessary, because we're also assigning to the outside scope's `workflowEvents` so we can't do `:=`

internal/services/dispatcher/server.go
juliusgeo: ok yeah so this is lowkey confusing. The quiet period is how long it waits after *any* event before sending the hangup, the hangupMaxWait is how long it will wait with items in the buffer before sending the hangup regardless

PR #4825, opened 2026-08-28
fix: Go SDK's `RunMany` should return results in the same order of the inputs

mnafees:
# Description

Go SDK's `RunMany` randomized the results which led to confusion. This PR fixes that and keeps it in-order in terms of the inputs.

## Type of change

---

<details id="ai-disclosure">

No AI

</details>


sdks/go/workflow.go
```diff
@@ -1001,16 +1001,15 @@ func (w *Workflow) RunMany(ctx context.Context, inputs []RunManyOpt) ([]Workflow
 		return durableRefs, nil
 	}
 
-	var workflowRefs []WorkflowRunRef
+	results := make([]*WorkflowRunRef, len(inputs))
```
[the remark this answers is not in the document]
mnafees: will release a new version once the PR is merged

mnafees:
@greptileai

pkg/client/admin.go
```diff
@@ -168,6 +168,8 @@ type AdminClient interface {
 	RunWorkflow(workflowName string, input interface{}, opts ...RunOptFunc) (*Workflow, error)
 
 	BulkRunWorkflow(workflows []*WorkflowRun) ([]string, error)
+	// BulkRunWorkflows triggers multiple workflow runs in a single request. The returned runs are in the same order as the inputs.
```
igor-kupczynski, line 171: nit, but I wonder if we need the comment here; if yes, should it also go on the BulkRunWorkflow above? What about other methods?

sdks/go/workflow.go
```diff
@@ -1001,56 +1000,106 @@ func (w *Workflow) RunMany(ctx context.Context, inputs []RunManyOpt) ([]Workflow
...
+	return workflowRefs, nil
+}
 
-	var wg sync.WaitGroup
-	var otherErrs []error
-	var collisions []*IdempotencyCollisionError
-	var errsMutex sync.Mutex
-	var workflowRefsMutex sync.Mutex
-	wg.Add(len(inputs))
+// ponytail: fixed chunk size like the TS SDK; chunk by grpc message size like Python if payloads get large
```
igor-kupczynski, line 1013: what's a ponytail?

igor-kupczynski: Would fixed 500-run chunks break on large payloads? `GRPCMaxMsgSize` defaults to 4MB, so 500 * 8K would result in a `ResourceExhausted`. Given that Python chunks by byte size and caps at 1000 (as mentioned in the comment) can we do the same here?

I think this is a regression, as the previous code would handle these paloads fine.

sdks/go/workflow.go
```diff
@@ -1001,56 +1000,106 @@ func (w *Workflow) RunMany(ctx context.Context, inputs []RunManyOpt) ([]Workflow
...
+
+		runs, err := triggerRunManyChunk(ctx, otelCtx, v0, workflowName, chunk)
+		for _, run := range runs {
+			refs = append(refs, WorkflowRunRef{RunId: run.RunId(), v0Workflow: run})
+		}
+
+		if err != nil {
+			var bulkErr *v0Client.BulkIdempotencyViolationErr
+			if !errors.As(err, &bulkErr) {
+				return refs, err
```
igor-kupczynski, line 1030: Can you check my thinking here: previously every input was attempted and all errors joined; now an error in chunk N means chunks N+1 and next are never attempted. It's esp problematic for the collisions.

igor-kupczynski (dismissed):
I understand and agree why this change is needed. However, on the implementation, I have two blockers (the regression on the large payloads, and the error handling of chunks after the error).

Also, given the order is deterministic now maybe a good time to add a test? Otherwise we'd need a manual proof this runs as intended.

WDYT?

sdks/go/workflow.go
mnafees: good catch, fixed this in my latest commit

sdks/go/workflow.go
mnafees: ahh good catch again, fixed it in my latest commit

abelanger5 (approved):

PR #4828, opened 2026-08-28
fix(engine): shovel concurrency slots over to paused queue items table on pause

mrkaye97:
# Description

Fixes a bug where when a workflow had a concurrency config set, workflow pause wouldn't work because we'd leave the concurrency slots and not shovel them over to the paused workflow queue items table. This will fix that by deleting concurrency slots in addition to deleting queue items on pause.

Also fixed one other thing, which was that we weren't pubbing a message to the scheduler on unpause so we'd fall back to polling for triggering the re-queued runs

Aaaaand one more, which is that there's a race condition on unpause where some workflow runs can still be written into the paused queue items table (I think from caching), so we need to also check for any stranded runs in that table while the workflow is unpaused to make sure nothing gets stuck

## Type of change

## Checklist

Changes have been:

---

<details id="ai-disclosure">

No AI

</details>

### DIFF
--- a/examples/python/bug_tests/workflow_pause/worker.py
+++ b/examples/python/bug_tests/workflow_pause/worker.py
@@ -1,0 +1,18 @@
+from hatchet_sdk import (
+    Context,
+    EmptyModel,
+    Hatchet,
+)
+import asyncio
+from datetime import timedelta
+
+hatchet = Hatchet()
+
+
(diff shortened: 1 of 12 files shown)


juliusgeo (approved):

mrkaye97:
@greptileai

PR #4829, opened 2026-08-28
fix(engine): duped rows causing on conflict error on assign

mrkaye97:
# Description

We've been seeing a bunch of this:

```
ERR error marking queue items processed error="ERROR: ON CONFLICT DO UPDATE command cannot affect row a second time (SQLSTATE 21000)"
```

Which I think is coming from the same task but with two different retry counts being inserted at the same time, and failing. Instead, we should only try to insert a runtime for the current retry, and drop the old one on the floor. Just doing a join + threading the retry counts here through to achieve that

## Type of change

## Checklist

Changes have been:

---

<details id="ai-disclosure">

No AI

</details>

### DIFF
--- a/pkg/repository/sqlcv1/tasks.sql
+++ b/pkg/repository/sqlcv1/tasks.sql
@@ -928,7 +928,7 @@ FROM
 
 -- name: PreflightCheckTasksForReplay :many
 -- Checks whether tasks can be replayed by ensuring that they don't have any active runtimes,
--- concurrency slots, or retry queue items. Returns the tasks which cannot be replayed.
+-- concurrency slots, retry queue items, or pending queue items. Returns the tasks which cannot be replayed.
 WITH input AS (
     SELECT
         UNNEST(@taskIds::bigint[]) AS task_id,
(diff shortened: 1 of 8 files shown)


mrkaye97:
@greptileai can you review this :)

pkg/repository/sqlcv1/tasks.sql
```diff
@@ -952,6 +952,14 @@ LEFT JOIN
     v1_concurrency_slot cs ON cs.task_id = t.id AND cs.task_inserted_at = t.inserted_at AND cs.task_retry_count = t.retry_count
 LEFT JOIN
     v1_retry_queue_item rqi ON rqi.task_id = t.id AND rqi.task_inserted_at = t.inserted_at AND rqi.task_retry_count = t.retry_count
+LEFT JOIN
```
grutt, line 955: are we concerned at all about performance here?

grutt (approved):

pkg/repository/sqlcv1/tasks.sql
mrkaye97: don't think so 🤷 if the task set is big we're hosed either way, and if it's small this is fine

PR #4844, opened 2026-08-31
feat: new `ci` param and `embedded` mode security check

mnafees:
# Description

Send embedded and CI information to security endpoint.

## Type of change

---

<details id="ai-disclosure">

No AI

</details>

### DIFF
--- a/pkg/security/security.go
+++ b/pkg/security/security.go
@@ -34,6 +34,7 @@ type DefaultSecurityCheck struct {
 	MQKind         string
 	OAuthProviders []string
 	AuthDisabled   bool
+	Embedded       bool
 
 	startTime time.Time
 }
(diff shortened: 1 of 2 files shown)


pkg/security/security.go
```diff
@@ -48,16 +49,30 @@ func NewSecurityCheck(opts *DefaultSecurityCheck, repo v1.SecurityCheckRepositor
...
 		return "github_actions"
 	case os.Getenv("CI") != "":
 		return "ci"
+	default:
+		return ""
+	}
+}
+
+func detectEnvironment(embedded bool) string {
+	if embedded {
```
grutt, line 69: we wont have embeded-ci?

mnafees: we would and that's why I am keeping this backwards compatible but also adding a new `ci` field for PostHog to case on so that we can construct how many embedded instances are originating from CI.

Here: https://github.com/hatchet-dev/hatchet/pull/4844/changes#diff-545c8e3aa8a9d4fd92409061fee8e80019a1386a4b5e48371b9ef428d97fd225R145

grutt (approved):

PR #4845, opened 2026-08-31
feat(engine): shared concurrency strategies

abelanger5:
# Description

Adds engine support for tenant-scoped concurrency strategies which can be shared across workflows. Example (eventual) usage from the Python SDK:

```py
# Shared across workflows: every task declaring this name uses the same limit.
gpu_limit = SharedConcurrency(
    name="gpu-limit",
    expression="input.model",
    max_runs=2,
    limit_strategy=ConcurrencyLimitStrategy.GROUP_ROUND_ROBIN,
)

wf_a = hatchet.workflow(name="WorkflowA")
wf_b = hatchet.workflow(name="WorkflowB")


@wf_a.task(concurrency=[gpu_limit])
def task_a(input, ctx): ...


@wf_b.task(concurrency=[gpu_limit])
def task_b(input, ctx): ...
```

Implementation-wise, this adds a new table to the database to track tenant-scoped strategies. This was easier than modifying the existing `v1_step_concurrency`, which involved a pkey change which seemed very unappealing. The tricky thing here is that we have a bunch of plumbing which assumes that concurrency slots are referencing a strategy_id on the `v1_step_concurrency` table (things like writing into the outbox for a topic called `concurrency.<tenant_id>.<strategy_id>` which seemed really risky to change). 

So rather than mess with that, **we re-use the sequence on `v1_step_concurrency` for `v1_tenant_concurrency`**, so that strategy ids are unique in the union of the two tables. This means that we can treat the tenant concurrency strategies as a regular strategy id. 

(This is still adding complexity to an already complex path, I'm hoping that once the DAGs-to-durable-tasks work is completed, we can get rid of `v1_workflow_concurrency` and removes the old path, so on net we will still reduce complexity in a few months). 

The risk here is mostly on the database migration path if users do something manual with `pglogical` and need to reset the sequence values after the migration: these will need to be set to the max of both tables. This is documented pretty thoroughly in the schema. 

Fixes # (issue)

## Type of change

## Checklist

Changes have been:

---

<details id="ai-disclosure">

- **Details**: Claude w/ Fable for implementation, testing, etc. Lots of manual testing as well. 

</details>


pkg/scheduling/v1/concurrency.go
```diff
@@ -64,7 +64,10 @@ func newConcurrencyManager(conf *sharedConfig, tenantId uuid.UUID, strategy *sql
 	ctx, cancel := context.WithCancel(context.Background())
 
 	var concurrencyStrategy *concurrency.ConcurrencyStrategy
-	if conf.concurrencyInMemoryIndexEnabled && !strategy.ParentStrategyID.Valid {
+	// Tenant-scoped strategies (zero-uuid workflow columns in their descriptor) are only
+	// supported by the in-memory index, so they always take that path regardless of the
+	// config flag.
+	if (conf.concurrencyInMemoryIndexEnabled || strategy.WorkflowID == uuid.Nil) && !strategy.ParentStrategyID.Valid {
```
abelanger5: this seems incorrect / outdated, we should be checking for the `tenant_strategy_id` column value instead

pkg/scheduling/v1/concurrency.go
```diff
@@ -218,6 +221,13 @@ func (c *ConcurrencyManager) loopConcurrency(ctx context.Context) {
 }
 
 func (c *ConcurrencyManager) loopCheckActive(ctx context.Context) {
+	// Tenant-scoped strategies are not tied to a workflow version: they only deactivate
+	// via the stale sweep and reactivate on slot insert, so there is no
+	// workflow-version-based active check to run for them.
+	if c.strategy.WorkflowID == uuid.Nil {
+		return
+	}
```
abelanger5: same comment as above. also, if we're running `loopCheckActive` on the step-level concurrency, we should likely be running it on the tenant-level concurrency as well (we do update through the trigger, but this other path must exist for a reason).

pkg/scheduling/v1/tenant_manager.go
```diff
@@ -314,8 +314,17 @@ func (t *tenantManager) setConcurrencyStrategies(strategies []*sqlcv1.V1StepConc
 	newArr := make([]*ConcurrencyManager, 0, len(strategies))
 
 	for _, c := range t.concurrencyStrategies {
-		if _, ok := strategiesSet[c.strategy.ID]; ok {
-			newArr = append(newArr, c)
+		if strat, ok := strategiesSet[c.strategy.ID]; ok {
+			// If the definition changed (e.g. a shared strategy was re-registered with a
+			// different max_concurrency or expression), replace the manager so the
+			// in-memory index is rebuilt against the new definition.
+			if concurrencyStrategyChanged(c.strategy, strat) {
```
abelanger5: I think this would be cleaner to be a definition on the strategy, i.e. `strat.Diff()`, so the logic doesn't leak to the tenant manager.

pkg/repository/sqlcv1/concurrency-strategy-rows.go
abelanger5, line 1: let's place this in `repository` instead, if we need this logic at all (which I question). we pretty much only want generated sqlc in here.

pkg/repository/sqlcv1/concurrency.sql
```diff
@@ -5,29 +5,105 @@ FROM v1_workflow_concurrency_slot
...
-    sc.is_active = TRUE;
+    sc.is_active = TRUE AND
+    sc.tenant_strategy_id IS NULL
+UNION ALL
+SELECT
+    tc.id,
+    NULL::bigint AS parent_strategy_id,
+    '00000000-0000-0000-0000-000000000000'::uuid AS workflow_id,
+    '00000000-0000-0000-0000-000000000000'::uuid AS workflow_version_id,
+    '00000000-0000-0000-0000-000000000000'::uuid AS step_id,
```
abelanger5: this feels fragile, I'd rather just run a second query and get proper typed results back when we get a `tenant_strategy_id` back

pkg/repository/sqlcv1/concurrency.sql
```diff
@@ -5,29 +5,105 @@ FROM v1_workflow_concurrency_slot
...
     sc.tenant_id = @tenantId::uuid AND
-    sc.id = @id::bigint;
+    sc.id = @id::bigint
+UNION ALL
+SELECT
+    tc.id,
+    NULL::bigint AS parent_strategy_id,
+    '00000000-0000-0000-0000-000000000000'::uuid AS workflow_id,
+    '00000000-0000-0000-0000-000000000000'::uuid AS workflow_version_id,
+    '00000000-0000-0000-0000-000000000000'::uuid AS step_id,
```
abelanger5: this also feels fragile, let's use proper types instead

sql/schema/v1-core.sql
```diff
@@ -243,9 +254,37 @@ CREATE TABLE v1_step_concurrency (
     expression TEXT NOT NULL,
```
abelanger5, line 254: would it make sense to add an FK constraint which updates the shared columns (expression, max_concurrency, etc) so we can just read from v1_step_concurrency directly? these are low-volume tables so FKs should be fine.

pkg/scheduling/v1/concurrency.go
abelanger5: Done in 9ca807d — the gate now checks `strategy.TenantStrategyID.Valid` on the descriptor (set by `TenantConcurrencyDescriptor` for tenant strategies) instead of the zero-uuid sentinel.

pkg/scheduling/v1/concurrency.go
abelanger5: Done in 9ca807d — same marker change, and `loopCheckActive` now runs for tenant strategies too: `UpdateConcurrencyStrategyIsActive` branches to a tenant check that retires the strategy when no workflow's latest version references it and no slots remain (the trigger handles reactivation; this handles prompt retirement instead of waiting 25h for the stale sweep). Covered by TestConcurrency_SharedStrategy_ActiveCheck.

pkg/scheduling/v1/tenant_manager.go
abelanger5: Done in 9ca807d — moved onto the manager as `(*ConcurrencyManager).strategyDiffers(next)`; the tenant manager just calls it during reconcile.

pkg/repository/sqlcv1/concurrency-strategy-rows.go
abelanger5: Done in 9ca807d — file deleted. With the queries returning proper types the converters mostly disappeared; the one remaining piece (building the scheduler descriptor from a `V1TenantConcurrency` row) lives in `repository.TenantConcurrencyDescriptor`.

pkg/repository/sqlcv1/concurrency.sql
abelanger5: Done in 9ca807d — the UNION is gone. `ListActiveConcurrencyStrategies` is a plain typed read over step strategies (excluding tenant references), and `ListActiveTenantConcurrencyStrategies` returns `V1TenantConcurrency` rows; the lease repository runs both and combines them.

pkg/repository/sqlcv1/concurrency.sql
abelanger5: Done in 9ca807d — `GetConcurrencyStrategyById` is back to a plain typed read; on ErrNoRows the repository falls back to `GetTenantConcurrencyStrategyById` (ids are unique across both tables, so at most one matches).

sql/schema/v1-core.sql
abelanger5: Done in 9ca807d — added the FK plus a statement trigger on v1_tenant_concurrency that propagates strategy/expression/max_concurrency to referencing rows on update, so all per-step reads now come straight from v1_step_concurrency with no join or COALESCE (this is also what let the fragile queries above collapse into plain typed reads). The only resolution left in Go is using tenant_strategy_id as the effective strategy id for task slots. Covered by the updated UpsertInPlace test, which asserts the sync propagates.

pkg/repository/workflow.go
```diff
@@ -85,7 +85,20 @@ type CreateConcurrencyOpts struct {
 	LimitStrategy *string `validate:"omitnil,oneof=CANCEL_IN_PROGRESS GROUP_ROUND_ROBIN CANCEL_NEWEST CANCEL_QUEUED_EXCEPT_NEWEST CANCEL_QUEUED_EXCEPT_OLDEST"`
 
 	// (required) a concurrency expression for evaluating the concurrency key
-	Expression string `validate:"celworkflowrunstr"`
+	Expression string `json:"expression" validate:"celworkflowrunstr"`
```
abelanger5: are we sure that renaming this field to `expression` won't create new workflow versions if this is part of the prior hashes?

pkg/repository/workflow.go
```diff
@@ -318,6 +331,306 @@ func (r *workflowRepository) ListStepMatchConditions(ctx context.Context, tenant
...
+	names := make([]string, 0)
+	seen := make(map[string]bool)
+
+	for _, entry := range entries {
+		if !entry.IsTenantScoped() {
+			continue
+		}
+
+		if seen[entry.Name] {
+			return nil, fmt.Errorf("tenant-scoped concurrency strategy %s is referenced more than once by step %s", entry.Name, stepReadableId)
```
abelanger5: should these errors be known return types? are we passing them to the user or not? if we are, we should probably enum them and audit that we're not reflecting any unwanted data back to the user.

abelanger5: same for all errors out of this package

sdks/python/examples/concurrency_shared/test_concurrency_shared.py
abelanger5, line 1: we should also have a mixed-strategy test which tests multiple tenant-scoped concurrency strategies and step-scoped strategies chained together, using different strategies.

abelanger5:
@greptileai review pls

pkg/repository/workflow.go
abelanger5: Good catch — it would have: the checksum serializes CreateConcurrencyOpts via ToJSONMap, so the json tag renamed the key from "Expression" to "expression" and every existing workflow with concurrency would have re-versioned on redeploy. Reverted in a740e63 (tag removed, with a comment explaining why it must stay untagged). The new Name/TenantScoped fields are omitempty, so existing workflows' hashes are unaffected by them.

pkg/repository/workflow.go
abelanger5: Done in a740e63 — these are user-facing (they surface through PutWorkflow), so they're now a typed TenantConcurrencyError which the admin server maps to InvalidArgument. Audited the messages: they contain only caller-supplied identifiers (strategy names, step readable ids, and workflow names/ids belonging to the caller's own tenant).

sdks/python/examples/concurrency_shared/test_concurrency_shared.py
abelanger5: Done in a740e63 — added test_shared_concurrency_multi_strategy_chain: a chain of [tenant-scoped GROUP_ROUND_ROBIN, workflow-scoped CANCEL_IN_PROGRESS, tenant-scoped GROUP_ROUND_ROBIN], asserting a newer run with distinct tenant keys but a colliding CANCEL_IN_PROGRESS key cancels the older in-flight run and completes.

sdks/python/hatchet_sdk/runnables/workflow.py
```diff
@@ -226,7 +226,7 @@ def to_proto(self) -> CreateWorkflowVersionRequest:
         if isinstance(self._config.concurrency, list):
             _concurrency_arr = [c.to_proto() for c in self._config.concurrency]
             _concurrency = None
-        elif isinstance(self._config.concurrency, ConcurrencyExpression):
+        elif isinstance(self._config.concurrency, (ConcurrencyExpression, SharedConcurrency)):
             _concurrency_arr = []
             _concurrency = self._config.concurrency.to_proto()
```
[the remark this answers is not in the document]
abelanger5: Fixed in a740e63 — scalar SharedConcurrency now rides concurrency_arr instead of the deprecated singular field, and the server's singular-field mapping also carries name/tenant_scoped now for good measure.

pkg/repository/workflow.go
```diff
@@ -352,6 +665,17 @@ func (r *workflowRepository) PutWorkflowVersion(ctx context.Context, tenantId uu
...
 	defer rollback()
 
+	// Upsert tenant-scoped concurrency definitions (entries carrying both a name and an
+	// expression) first, in the same transaction, so the entries created below can
+	// reference them. This runs before the checksum early-return on purpose: definitions
+	// are tenant-level state, so re-registering an unchanged workflow with a changed
+	// definition must still update the strategy in place.
+	for _, entry := range collectTenantConcurrencyDefs(opts) {
+		if _, err := r.upsertTenantConcurrencyStrategy(ctx, tx, tenantId, entry); err != nil {
+			return nil, fmt.Errorf("could not upsert tenant concurrency strategy %s: %w", entry.Name, err)
```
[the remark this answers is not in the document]
abelanger5: Addressed in a740e63 — after a registration commit that upserted tenant strategy definitions, the process invalidates its cached per-step strategy lists for every referencing step, so task creation on that node picks up the new expression immediately. Caveat: the cache is per-process, so other engine replicas remain stale up to the 5-minute TTL; fully closing that would need distributed invalidation (e.g. an mq message), which I left out — happy to add if you want it.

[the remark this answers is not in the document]
abelanger5: Closed properly in 140da97 with the read-through approach: per-step strategy lists are never cached for steps referencing tenant-scoped strategies. The referencing rows are kept in sync in the database by the v1_tenant_concurrency update trigger, so every engine reads the current definition on every insert batch (one cheap indexed query); the bounded-staleness contract is gone entirely, cross-replica included. The local cache invalidation from the previous commit was removed as redundant. Workflow-scoped strategies (immutable per version) keep the cache.

pkg/repository/workflow.go
abelanger5: As of 140da97 every user-facing error introduced by this PR in the repository package goes through the typed TenantConcurrencyError (ordering conflicts, workflow-level-unsupported, duplicate references, missing name, unknown names). Internal errors (wrapped DB failures) still propagate the way the rest of the package's errors do today; a package-wide error taxonomy felt out of scope for this PR, but happy to take it as a follow-up.

pkg/repository/scheduler_concurrency.go
```diff
@@ -193,6 +260,13 @@ func (c *ConcurrencyRepositoryImpl) RunConcurrencyStrategy(
 		}
 	}
 
+	// Strategies with no runnable branch (e.g. NONE, which exists so tasks of a removed
+	// strategy can still enqueue) must return an empty result rather than nil: the
+	// scheduler dereferences the embedded result and would panic on nil.
```
abelanger5: does this edge case still exist? I think this happened locally due to being able to remove the tenant strategies during runtime, but we should be able to do that any more. I think we can still have the `res == nil` check but we should likely log an error message

api-contracts/v1/workflows.proto
```diff
@@ -152,10 +153,19 @@ enum ConcurrencyLimitStrategy {
...
+// A tenant-scoped entry (tenant_scoped = true) defines (or updates in place) a strategy
+// shared across workflows, keyed by name: every task declaring the same name consumes the
+// same concurrency limit. Registrations whose chains order the same tenant-scoped
+// strategies inconsistently are rejected, since inconsistent orders can deadlock.
 message Concurrency {
     string expression = 1; // (required) the expression to use for concurrency
     optional int32 max_runs = 2; // (optional) the maximum number of concurrent workflow runs, default 1
     optional ConcurrencyLimitStrategy limit_strategy = 3; // (optional) the strategy to use when the concurrency limit is reached, default CANCEL_IN_PROGRESS
+    optional string name = 4; // (required when tenant_scoped) the strategy name; unique per tenant for tenant-scoped strategies
+    optional bool tenant_scoped = 5; // (optional) when true, the entry is a tenant-scoped strategy shared across workflows, default false
```
mrkaye97: nit, but I really feel like we should use `is_` as a prefix for bools like this

cmd/hatchet-migrate/migrate/migrations/20260831120000_v1_0_150.sql
```diff
@@ -0,0 +1,276 @@
...
+    -- Unique per tenant: registration upserts by (tenant_id, name).
+    name TEXT NOT NULL,
+    is_active BOOLEAN NOT NULL DEFAULT TRUE,
+    -- last_active_at is refreshed at most once per hour when a new slot is inserted for this strategy.
+    last_active_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    strategy v1_concurrency_strategy NOT NULL,
+    expression TEXT NOT NULL,
+    max_concurrency INTEGER NOT NULL,
+    CONSTRAINT v1_tenant_concurrency_pkey PRIMARY KEY (id),
+    CONSTRAINT v1_tenant_concurrency_tenant_name_ux UNIQUE (tenant_id, name)
```
mrkaye97:
```suggestion
    CONSTRAINT v1_tenant_concurrency_tenant_name_uq UNIQUE (tenant_id, name)
```

`uq` for `unique`? feel like this is the normal shorthand

cmd/hatchet-migrate/migrate/migrations/20260831120000_v1_0_150.sql
```diff
@@ -0,0 +1,276 @@
...
+);
+
+CREATE INDEX v1_step_concurrency_tenant_strategy_id_idx
```
mrkaye97, line 30: concurrently for this? or no risk since the index is empty initially?

cmd/hatchet-migrate/migrate/migrations/20260831120000_v1_0_150.sql
```diff
@@ -0,0 +1,276 @@
...
+    ON v1_step_concurrency (tenant_strategy_id)
+    WHERE tenant_strategy_id IS NOT NULL;
+
+-- These are low-volume tables, so a real FK is fine here (unlike the high-volume v1
+-- tables, which avoid them).
+ALTER TABLE v1_step_concurrency
+    ADD CONSTRAINT v1_step_concurrency_tenant_strategy_id_fkey
+    FOREIGN KEY (tenant_strategy_id) REFERENCES v1_tenant_concurrency (id);
```
mrkaye97, line 38: do we want an FK here? do we have any others in the V1 schema?

internal/services/admin/v1/server.go
```diff
@@ -1024,17 +1032,34 @@ func getCreateWorkflowOpts(req *contracts.CreateWorkflowVersionRequest) (*v1.Cre
...
+		}
+
 		var limitStrategy *string
 
 		if req.Concurrency.LimitStrategy != nil && req.Concurrency.LimitStrategy.String() != "" {
 			s := req.Concurrency.LimitStrategy.String()
 			limitStrategy = &s
 		}
 
+		var name string
```
mrkaye97, line 1051: nit: should this be a pointer instead of defaulting to an empty string in most cases?

mrkaye97: (same below)

pkg/repository/sqlcv1/concurrency.sql
```diff
@@ -1362,3 +1370,176 @@ WHERE task_id = $2
...
+    tenant_id = @tenantId::uuid AND
+    name = ANY(@names::text[]);
+
+-- name: DeactivateStaleTenantConcurrency :exec
+WITH stale_tenant_concurrencies AS (
+    SELECT tc.id
+    FROM v1_tenant_concurrency tc
+    WHERE tc.tenant_id = @tenantId::UUID
+        AND tc.is_active = TRUE
+        -- we use 25 hours because there's a 1-hour cache on updating last_active_at
```
mrkaye97: 1 hour or 1 day?

pkg/repository/sqlcv1/concurrency.sql
```diff
@@ -1362,3 +1370,176 @@ WHERE task_id = $2
...
+        AND tc.last_active_at < NOW() - INTERVAL '25 hours'
+        AND NOT EXISTS (
+            SELECT 1 FROM v1_concurrency_slot cs
+            WHERE
+                cs.strategy_id = tc.id
+                AND cs.tenant_id = @tenantId::UUID -- tenant id filter to force index usage
+        )
+    FOR UPDATE SKIP LOCKED
```
mrkaye97, line 1435: order by here? guess since we have `skip locked` it maybe doesn't matter? idk

pkg/repository/workflow.go
```diff
@@ -352,6 +688,16 @@ func (r *workflowRepository) PutWorkflowVersion(ctx context.Context, tenantId uu
 
 	defer rollback()
 
+	// Upsert tenant-scoped concurrency definitions first, in the same transaction, so the
+	// entries created below can reference them. This runs before the checksum early-return
+	// on purpose: definitions are tenant-level state, so re-registering an unchanged
+	// workflow with a changed definition must still update the strategy in place.
+	for _, entry := range collectTenantConcurrencyDefs(opts) {
+		if _, err := r.upsertTenantConcurrencyStrategy(ctx, tx, tenantId, entry); err != nil {
+			return nil, fmt.Errorf("could not upsert tenant concurrency strategy %s: %w", entry.Name, err)
```
mrkaye97: intentional n+1?

sdks/python/examples/concurrency_shared/worker.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+@concurrency_shared_workflow_a.task(concurrency=[shared_limit])
+def task_a(input: WorkflowInput, ctx: Context) -> dict[str, int]:
+    start = int(time.time() * 1000)
+    time.sleep(1.5)
+    return {"start_ms": start, "end_ms": int(time.time() * 1000)}
+
+
+@concurrency_shared_workflow_b.task(concurrency=[shared_limit])
```
mrkaye97: do these need to be workflow level or can we make them standalone tasks?

sdks/python/examples/concurrency_shared/worker.py
```diff
@@ -0,0 +1,144 @@
...
+)
+
+concurrency_shared_workflow_b = hatchet.workflow(
+    name="ConcurrencySharedB",
+    input_validator=WorkflowInput,
+)
+
+
+@concurrency_shared_workflow_a.task(concurrency=[shared_limit])
+def task_a(input: WorkflowInput, ctx: Context) -> dict[str, int]:
```
mrkaye97: would be great to return a pydantic model or a dataclass here so we can get better type checking in this example / show that as recommended in the docs / rely more on the type checker in the tests

sdks/python/hatchet_sdk/types/concurrency.py
```diff
@@ -44,3 +44,36 @@ def from_int(max_runs: int) -> "ConcurrencyExpression":
             max_runs=max_runs,
             limit_strategy=ConcurrencyLimitStrategy.GROUP_ROUND_ROBIN,
         )
+
+
+class SharedConcurrency(BaseModel):
```
mrkaye97: need to think about this a bit, I don't love that we have two classes for virtually the same thing with the only difference being that one is for tenant-level and the other is task- and workflow-level, feels kind of awkward to me

sdks/python/hatchet_sdk/runnables/workflow.py
```diff
@@ -223,9 +223,13 @@ def to_proto(self) -> CreateWorkflowVersionRequest:
             t.to_proto(service_name) if (t := self._on_failure_task) else None
         )
 
-        if isinstance(self._config.concurrency, list):
+        if isinstance(self._config.concurrency, (list, tuple)):
```
mrkaye97: can this be a tuple? oh is this because we're using `Sequence`?

mrkaye97 (commented):
looks great to me, left a couple small questions, mostly on naming and such

api-contracts/v1/workflows.proto
```diff
@@ -18,6 +18,7 @@ service AdminService {
     rpc BranchDurableTask(BranchDurableTaskRequest) returns (BranchDurableTaskResponse);
 }
 
+
```
abelanger5: nit: whitespace

cmd/hatchet-migrate/migrate/migrations/20260831120000_v1_0_150.sql
abelanger5: yep I think this is low risk

cmd/hatchet-migrate/migrate/migrations/20260831120000_v1_0_150.sql
abelanger5: we do want an FK here, ideally we rely on them on the workflow config side of things. the only reason we don't have FKs everywhere is the performance hit, but I think this is actually useful

pkg/repository/sqlcv1/concurrency.sql
abelanger5: yep added the order by

pkg/repository/scheduler_concurrency.go
```diff
@@ -114,6 +116,10 @@ func (c *ConcurrencyRepositoryImpl) UpdateConcurrencyStrategyIsActive(
 	tenantId uuid.UUID,
 	strategy *sqlcv1.V1StepConcurrency,
 ) error {
+	if strategy.TenantStrategyID.Valid {
+		return c.updateTenantConcurrencyStrategyIsActive(ctx, tenantId, strategy)
+	}
```
abelanger5, line 129: I wonder if this is-active check should be called from the concurrency manager as a separate endpoint specific to the tenant manager. I'm worried about massive contention on the advisory lock if there are very many step concurrencies which are sharing a tenant concurrency.

pkg/scheduling/v1/tenant_manager.go
```diff
@@ -314,8 +314,17 @@ func (t *tenantManager) setConcurrencyStrategies(strategies []*sqlcv1.V1StepConc
...
-			newArr = append(newArr, c)
+		if strat, ok := strategiesSet[c.strategy.ID]; ok {
+			// If the definition changed (e.g. a shared strategy was re-registered with a
+			// different max_concurrency or expression), replace the manager so the
+			// in-memory index is rebuilt against the new definition.
+			if c.strategyDiffers(strat) {
+				go c.cleanup()
+
+				newArr = append(newArr, newConcurrencyManager(t.cf, t.tenantId, strat, t.concurrencyResultsCh, t.concurrencyAdvisoryLock, t.concurrencyParentAdvisoryLock))
+			} else {
```
abelanger5, line 328: let's investigate the overhead of updating these values in-place in the concurrency manager. there's a greater risk of contention with an async cleanup and init.

sdks/python/hatchet_sdk/types/concurrency.py
abelanger5: feel free to put updated thoughts on this PR: https://github.com/hatchet-dev/hatchet/pull/4873

sdks/python/examples/concurrency_shared/worker.py
abelanger5: should be addressed in the new PR

sdks/python/examples/concurrency_shared/worker.py
abelanger5: these are standalone now, nice catch

pkg/repository/workflow.go
abelanger5: fixed!

pkg/repository/sqlcv1/concurrency.sql
abelanger5: 1 day + 1 hour (the comment is correct, we'd like to expire after 24 hours inactive, but we add 1 hour because of caching)

grutt (approved):

PR #4854, opened 2026-09-01
refactor(engine): buffer durable event ingestion

mrkaye97:
# Description

Sets up a buffer for `IngestDurableTaskEvent` so that we can try to batch db operations a little bit instead of having many writes for every call of that function. I think this is the last big lever that's relatively straightforward that we have for improving the performance of this path

## Type of change

## Checklist

Changes have been:

---

<details id="ai-disclosure">

- **Details**: Fable to help wire up the buffer after I wrote the new queries and env vars

</details>

### DIFF
--- a/pkg/config/server/server.go
+++ b/pkg/config/server/server.go
@@ -342,7 +342,7 @@ type ConfigFileRuntime struct {
 
 	// DurableEventBufferMaxSize caps how many durable task events a single flush
 	// batches together, so a burst of traffic still flushes promptly.
-	DurableEventBufferMaxSize int `mapstructure:"durableEventBufferMaxSize" json:"durableEventBufferMaxSize,omitempty" default:"20"`
+	DurableEventBufferMaxSize int `mapstructure:"durableEventBufferMaxSize" json:"durableEventBufferMaxSize,omitempty" default:"100"`
 
 	// DurableEventBufferMaxConcurrentFlushes bounds how many durable event ingestion
 	// transactions run concurrently.
--- a/pkg/config/server/server.go
+++ b/pkg/config/server/server.go
@@ -336,6 +336,18 @@ type ConfigFileRuntime struct {
 	// WorkflowRunBufferSize is the buffer size for workflow run event batching in the dispatcher
 	WorkflowRunBufferSize int `mapstructure:"workflowRunBufferSize" json:"workflowRunBufferSize,omitempty" default:"1000"`
 
+	// DurableEventBufferFlushInterval is how long a durable task event waits in the
+	// ingestion buffer to accumulate a batch before the batch is flushed to the database.
+	DurableEventBufferFlushInterval time.Duration `mapstructure:"durableEventBufferFlushInterval" json:"durableEventBufferFlushInterval,omitempty" default:"10ms"`
+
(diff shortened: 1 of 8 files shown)


mrkaye97:
@greptileai

pkg/repository/durable_events.go
```diff
@@ -164,12 +165,192 @@ type DurableEventsRepository interface {
...
+		if taskErr, ok := taskErrors[taskId]; ok {
+			request.responseCh <- durableEventIngestResponse{err: taskErr}
+			continue
+		}
+
+		request.responseCh <- durableEventIngestResponse{result: results[taskId]}
+	}
+
+	if len(deferredRequests) > 0 {
+		go b.flush(deferredRequests)
```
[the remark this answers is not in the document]
mrkaye97: Not a regression — submission-order FIFO across concurrent same-task requests has never existed in this path. The dispatcher handles each durable-task stream message in its own goroutine (server_v1.go), so ordering is lost before requests reach the buffer, and pre-batching, concurrent requests for one task raced for the FOR UPDATE row lock in arbitrary order.

The guarantees that actually matter both hold: (1) mutual exclusion per task — same-task duplicates never share a batch, and any two flushes containing the same task serialize on the log file row lock; (2) each ingest observes previously committed log file state — a deferred flush starts only after the current batch's ingest returns, post-commit, and a request submitted after another's response always lands in a later flush.

The buffer's orderings are a strict subset of the old code's: for requests A (batched) / B (deferred) / C (next window), the old code allowed all six interleavings including C before A, while the buffer guarantees A before B and only leaves B/C to the lock. So no node/branch ordering or non-determinism error is reachable that wasn't already reachable before batching.

mrkaye97:
@greptileai re-review please

pkg/config/server/server.go
```diff
@@ -336,6 +336,18 @@ type ConfigFileRuntime struct {
 	// WorkflowRunBufferSize is the buffer size for workflow run event batching in the dispatcher
 	WorkflowRunBufferSize int `mapstructure:"workflowRunBufferSize" json:"workflowRunBufferSize,omitempty" default:"1000"`
 
+	// DurableEventBufferFlushInterval is how long a durable task event waits in the
+	// ingestion buffer to accumulate a batch before the batch is flushed to the database.
+	DurableEventBufferFlushInterval time.Duration `mapstructure:"durableEventBufferFlushInterval" json:"durableEventBufferFlushInterval,omitempty" default:"10ms"`
```
grutt, line 341: should we try shorter first?

grutt (commented):
there's quick a bit of for update is it possible we deadlock on the db if we write two concurrent batches?

juliusgeo (approved):
LGTM!

pkg/config/server/server.go
grutt: my thinking here the non-100 batch case is not really our problem; if we're not hitting that path it doesnt really make sense imo to slow things down

mrkaye97: yeah, can try

mrkaye97:
> there's quick a bit of for update is it possible we deadlock on the db if we write two concurrent batches?

Don't think so since we're locking log files ordered by the task id + inserted at

PR #4857, opened 2026-09-01
fix: add new dag shapes to load test to stress test dag operator

juliusgeo:
# Description

Adds new dag workloads to the load test that test a variety of complex DAG shapes.
Used these to fix several issues with the dag operator:
- Rabbit blips causing OLAP messages to not send, causing COMPLETED events to never arrive, parking runs in RUNNING
- Dag operator rows only being inserted into OLAP *after* status updates had arrived, also parking runs in RUNNING
- Dag operators being stuck in evicted state because of mismatched child ids in dag event log
- Engines rolling also causing dag operators to become stuck in evicted state because of durable invocation messages not being retried

Fixes # (issue)

## Type of change

- [ ] Performance improvement (non-breaking change which improves performance)

## Checklist

Changes have been:

---

<details id="ai-disclosure">

</details>


pkg/repository/sqlcv1/olap.sql
```diff
@@ -2525,3 +2525,44 @@ SELECT
...
+        UNNEST(@dagIds::BIGINT[]) AS dag_id,
+        UNNEST(@dagInsertedAts::TIMESTAMPTZ[]) AS dag_inserted_at
+), latest_event AS (
+    SELECT DISTINCT ON (e.task_id, e.task_inserted_at)
+        e.task_id,
+        e.task_inserted_at,
+        e.readable_status,
+        e.retry_count
+    FROM v1_task_events_olap e
+    JOIN inputs i ON (e.task_id, e.task_inserted_at) = (i.dag_id, i.dag_inserted_at)
```
mrkaye97, line 2544: can we leave a comment here that this is specifically looking at the durable task (the operator)'s task events since it's what's using the dag id? was a bit confusing initially

pkg/repository/sqlcv1/tasks.sql
```diff
@@ -1153,6 +1153,35 @@ SELECT
...
+SELECT
+    t.id,
+    t.inserted_at,
+    t.external_id,
+    rt.retry_count
+FROM v1_task_runtime rt
+JOIN v1_task t ON (t.id, t.inserted_at) = (rt.task_id, rt.task_inserted_at)
+WHERE rt.tenant_id = @tenantId::uuid
+    AND rt.evicted_at IS NOT NULL
+    AND rt.evicted_at < NOW() - @gracePeriod::interval
```
mrkaye97: nit: I think having the `IS NOT NULL` here is redundant?

pkg/repository/sqlcv1/tasks.sql
```diff
@@ -1153,6 +1153,35 @@ SELECT
           AND NOT is_satisfied
     ) AS has_unsatisfied_durable_events;
 
+-- name: ListStuckEvictedDurableOrchestrators :many
+-- DAG-orchestrator tasks whose runtime has been evicted past the grace period and whose durable
+-- event log entries are ALL satisfied -- the orchestrator is ready to resume but the
+-- edge-triggered restore (a child callback arriving while evicted -> DurableRestoreTask) never
+-- fired: the callback was lost on an engine roll, or every entry was satisfied before the
+-- eviction so there was no later callback. The caller re-queues these via DurableRestoreTask.
+SELECT
```
mrkaye97, line 1162: have you pulled a query plan for this yet? just making sure the indexes will work correctly

pkg/repository/durable_events.go
```diff
@@ -192,9 +192,9 @@ func newDurableEventsRepository(shared *sharedRepository, opts DurableEventBuffe
 type durableTaskId int64
 
 type appendDurableEventLogResult struct {
-	logEntries                  []*EventLogEntryWithPayloads
-	nodeIdBranchIdToTriggerOpts map[NodeIdBranchIdTuple]*WorkflowNameTriggerOpts
-	childrenToReplay            map[uuid.UUID]bool
+	logEntries                   []*EventLogEntryWithPayloads
+	childExternalIdToTriggerOpts map[uuid.UUID]*WorkflowNameTriggerOpts
```
mrkaye97, line 196: I feel like I might've _just_ changed this to fix a bug. I wonder if @greptileai can check the history / blame to see

mrkaye97: So is this a regression?

pkg/repository/sqlcv1/tasks.sql
juliusgeo: is this because the comparison would just result in NULL which is falsey?

mrkaye97: I think `rt.evicted_at < NOW() - @gracePeriod::interval` implies that it's not null

mrkaye97: (but yeah, similar idea)

pkg/repository/sqlcv1/tasks.sql
juliusgeo: just pulled a query plan, not seeing any seq scans, all index scans

pkg/repository/durable_events.go
mrkaye97: @greptileai is this a regression?

pkg/repository/sqlcv1/tasks.sql
juliusgeo: the column is null in the schema IIUC

mrkaye97: ah yeah but I mean the planner can figure out that when you say `x < @y::interval`, that it should only consider non-null values of `x`, since no null value of `x` will ever satisfy that condition. so having the `is not null` + the `<` will result in the same plan as only having the `<`

pkg/repository/durable_events.go
mrkaye97: @greptileai you need to look at commits of mine (mrkaye97) farther back (the past few weeks). I'm pretty sure I put in a PR in the past few weeks that changed this to use the node id + branch id instead of the external id as the key, because it was a bug.

pkg/repository/sqlcv1/olap.sql
juliusgeo: done!

mrkaye97 (approved):

PR #4863, opened 2026-09-01
feat(engine): dynamic per-group max runs via CEL expression

abelanger5:
# Description

Adds engine-side support for making `max_runs` dynamic on a per-key basis. This is useful if you'd like certain organizations or tiers to have different concurrency limits from other organizations or tiers. 

The unlock here is the in-memory concurrency index, which tracks each group in its own heap, which means we can have per-heap settings. This isn't possible in the old SQL definition without a near-guaranteed seq scan, because of how the `join lateral` needed to be constructed. 

This PR just adds support to the API contract and SQL schema, but doesn't implement the SDK changes yet. It branches off of `belanger/shared-concurrency` to avoid merge conflicts, since both of these branches modify concurrency strategies and contracts heavily. 

## Type of change

## Checklist

Changes have been:

---

<details id="ai-disclosure">

</details>


pkg/repository/scheduler_concurrency.go
```diff
@@ -303,11 +304,10 @@ func (c *ConcurrencyRepositoryImpl) RunConcurrencyStrategy(
...
 	}
 
-	// No strategy branch ran. This should no longer be reachable (every enum value has a
-	// branch above), so log it as an error, but still return an empty result rather than
-	// nil: the scheduler dereferences the embedded result and would panic on nil.
+	// Strategies with no runnable branch (e.g. NONE, which exists so tasks of a removed
+	// strategy can still enqueue) must return an empty result rather than nil: the
+	// scheduler dereferences the embedded result and would panic on nil.
 	if res == nil {
-		c.l.Error().Ctx(ctx).Msgf("no result for concurrency strategy %s (strategy ID: %d); returning empty result", strategy.Strategy, strategy.ID)
```
abelanger5, line 310: this seems incorrect to remove?

pkg/repository/task.go
```diff
@@ -2203,6 +2203,27 @@ func (r *sharedRepository) evalBatchGroupKey(
...
+		return pgtype.Int4{}, fmt.Errorf("failed to parse max runs expression (%s): %w", strat.MaxRunsExpression.String, err)
+	}
+
+	if res.Int == nil {
+		return pgtype.Int4{}, fmt.Errorf("failed to parse max runs expression (%s): expected integer output", strat.MaxRunsExpression.String)
+	}
+
+	if *res.Int < 1 {
+		return pgtype.Int4{}, fmt.Errorf("failed to evaluate max runs expression (%s): result must be a positive integer, got %d", strat.MaxRunsExpression.String, *res.Int)
+	}
```
abelanger5: should a user be able to set concurrency to 0? it might make sense, i.e. for a disabled account so they can just hit scheduling timeouts and not run any workload. not sure if this has downstream implications.

abelanger5:
@greptileai review

pkg/repository/scheduler_concurrency.go
mrkaye97: agreed, seems like we should leave this?

pkg/repository/task.go
```diff
@@ -2205,6 +2205,27 @@ func (r *sharedRepository) evalBatchGroupKey(
 
 // insertTasks inserts new tasks into the database. note that we're using Postgres rules to automatically insert the created
 // tasks into the queue_items table.
+// evalMaxRunsExpression evaluates a strategy's max_runs_expression against one task's
+// input. The returned value rides on the task's concurrency slot; a NULL (invalid)
+// value means the strategy's static max_concurrency applies.
+func (r *sharedRepository) evalMaxRunsExpression(strat *sqlcv1.V1StepConcurrency, in cel.Input) (pgtype.Int4, error) {
```
mrkaye97: nit: should we have this return `*int32` instead? I keep wanting to remove this `pgtype.FooBar` because I hate how they mask nil values, but not a big deal here

mrkaye97 (approved):
two tiny comments, but this lgtm - it's great how simple it is

cmd/hatchet-migrate/migrate/migrations/20260901120000_v1_0_151.sql
abelanger5, line 1: note: need to manually validate triggers here

internal/services/admin/v1/server.go
```diff
@@ -1055,11 +1055,12 @@ func getCreateWorkflowOpts(req *contracts.CreateWorkflowVersionRequest) (*v1.Cre
...
-			LimitStrategy: limitStrategy,
-			Expression:    req.Concurrency.Expression,
-			MaxRuns:       req.Concurrency.MaxRuns,
-			Name:          name,
-			TenantScoped:  tenantScoped,
+			LimitStrategy:     limitStrategy,
+			Expression:        req.Concurrency.Expression,
+			MaxRuns:           req.Concurrency.MaxRuns,
+			Name:              name,
+			TenantScoped:      tenantScoped,
```
abelanger5: was this a bad rebase or did we not update this to `IsTenantScoped` on the repository layer as well?

pkg/repository/sqlcv1/concurrency.sql
```diff
@@ -1380,28 +1381,34 @@ WITH input AS (
         unnest(@names::text[]) AS name,
         unnest(cast(@strategies::text[] as v1_concurrency_strategy[])) AS strategy,
         unnest(@expressions::text[]) AS expression,
-        unnest(@maxConcurrencies::int[]) AS max_concurrency
+        unnest(@maxConcurrencies::int[]) AS max_concurrency,
+        -- empty string means "no expression" (NULLIF below); keeps the parameter a
+        -- plain text[] while still writing NULL for static entries
+        unnest(@maxRunsExpressions::text[]) AS max_runs_expression
```
abelanger5: this isn't how we do it elsewhere, we construct a manual query in an `-overwrite` file. but I wonder if we can leverage sqlc.yaml replacements for this instance, to make this generate as `pgtype.Text` instead of `string` in the query

pkg/repository/sqlcv1/concurrency.sql
```diff
@@ -1380,28 +1381,34 @@ WITH input AS (
...
+    max_runs_expression
 )
 SELECT
     @tenantId::uuid,
     i.name,
     i.strategy,
     i.expression,
-    i.max_concurrency
+    i.max_concurrency,
+    NULLIF(i.max_runs_expression, '')
```
abelanger5: same comment here

pkg/repository/workflow.go
```diff
@@ -371,6 +380,11 @@ func (r *workflowRepository) upsertTenantConcurrencyStrategies(ctx context.Conte
 		params.Strategies[i] = string(strategy)
 		params.Expressions[i] = opts.Expression
 		params.Maxconcurrencies[i] = maxRuns
+
+		// empty string writes NULL (NULLIF in the query)
```
abelanger5: hopefully this cruft can get fixed with the SQL changes

cmd/hatchet-migrate/migrate/migrations/20260901120000_v1_0_151.sql
```diff
@@ -0,0 +1,1451 @@
...
+-- Per-strategy max-runs values evaluated from max_runs_expression at insert time,
+-- parallel to concurrency_strategy_ids. A NULL element means that strategy's static
+-- max_concurrency applies.
+ALTER TABLE v1_task ADD COLUMN concurrency_max_runs INTEGER[];
+
+-- max_runs is this task's insert-time evaluation of the strategy's max_runs_expression
+-- (NULL = static max_concurrency applies); next_max_runs carries the values for the
+-- rest of the chain, peeled forward like next_keys.
+ALTER TABLE v1_concurrency_slot ADD COLUMN max_runs INTEGER;
+ALTER TABLE v1_concurrency_slot ADD COLUMN next_max_runs INTEGER[];
```
abelanger5: these `alter table` commands risk deadlocks when inside the same tx. we should split each into a separate TX (mostly worried about v1_concurrency_slot and v1_task, since we have lots of transactions which touch both)

PR #4867, opened 2026-09-02
fix(sdks/ts): recover from inactive listener stream on heartbeat rejection

### LINKED ISSUE #4823: TypeScript SDK: worker /health always returns HTTP 200, even when UNHEALTHY — breaks Kubernetes liveness probes
<details open>

- [x] _I acknowledge that an LLM was used in the creation of this issue, in accordance with Hatchet's AI_POLICY.md._
- **Details**: Claude Code was used for source analysis and drafting; all findings were verified by a human against SDK source and a live deployment.

</details>

## Describe the bug

The TypeScript SDK's worker health server always responds to `GET /health` with HTTP **200**, regardless of worker status — the status is only reported inside the JSON body.

`sdks/typescript/src/v1/client/worker/health-server.ts` (`handleHealth`, line ~73; identical behavior in published 1.22.1):

```ts
res.writeHead(200, { 'Content-Type': 'application/json' });
await res.end(JSON.stringify(response)); // { status: this.getStatus(), ... }
```

There is no branch that returns a non-200 for an `UNHEALTHY` worker — the only 503 in the file belongs to `/metrics` when `prom-client` is not installed.

This breaks the primary use case for the endpoint: a standard Kubernetes `httpGet` liveness probe evaluates the HTTP status code and cannot inspect the body, so an `UNHEALTHY` TypeScript worker still passes its probe and is never restarted.

The Python SDK does this correctly — `sdks/python/hatchet_sdk/worker/action_listener_process.py`:

```python
return web.json_response(response, status=200 if ok else 503)
```

## Expected behavior

Parity with the Python SDK: `/health` returns 200 for `HEALTHY`/`STARTING`/`INITIALIZED` and **503 for `UNHEALTHY`**, so `httpGet` liveness probes work as documented.

## Environment

- `@hatchet-dev/typescript-sdk` 1.22.1 (published) — behavior unchanged on current `main`
- Node.js 24, self-hosted Hatchet engine, workers on Kubernetes (GKE)

## Contribution

Happy to submit a PR (status-code change + test) if this is accepted.

### LINKED ISSUE #4824: TypeScript SDK: worker becomes a permanent zombie when the engine loses its listener stream — heartbeats rejected forever, no recovery, health stays HEALTHY
<details open>

- [x] _I acknowledge that an LLM was used in the creation of this issue, in accordance with Hatchet's AI_POLICY.md._
- **Details**: Claude Code assisted with incident investigation, SDK/engine source analysis, and the reproduction harness; all findings were verified against production logs and source by a human.

</details>

## Describe the bug

If the engine loses a worker's listener stream while the gRPC transport stays healthy, a TypeScript worker enters a **permanent zombie state**: the engine rejects every heartbeat, never assigns new tasks, the SDK never recovers, and the worker's health status stays `HEALTHY` so orchestrators (Kubernetes) never intervene.

In our case (self-hosted engine behind an envoy gRPC proxy on GKE), the client-side HTTP/2 connection survived whatever severed the stream server-side, producing exactly the condition the engine guards against in `internal/services/dispatcher/server.go` (~line 457):

```go
if worker.LastListenerEstablished.Valid && !worker.IsActive {
    span.SetStatus(telemetry_codes.Error, "worker stream is not active")
    return nil, status.Errorf(codes.FailedPrecondition, "Heartbeat rejected: worker stream is not active: %s", req.WorkerId)
}
```

Why the SDK cannot recover from this today (verified in published 1.22.1 and current `main`):

1. **The listener reconnect loop never fires.** `clients/dispatcher/action-listener.ts` re-establishes the stream when the `for await` throws or completes — a silently-dead stream just blocks forever. gRPC keepalive (configured at 10s pings in `util/grpc-helpers.ts`) doesn't help: pings succeed because the transport is genuinely alive; the rejection is an application-level *response*, not a transport error. (An abort is treated as shutdown — `isAbortError → break` — so there is also no external reconnect handle.)
2. **The heartbeat thread observes the exact failure signal and takes no action.** In `heartbeat/heartbeat-worker.ts`, the non-`UNIMPLEMENTED` failure path logs and posts a message which the controller re-logs — nothing else. (`main` additionally counts `missedHeartbeats`, but only to classify log severity; there is still no reconnect, exit, or status change. `main` also adds a 5s RPC deadline — a good fix for stalled connections, but not for this served-rejection case.)
3. **Worker health status stays `HEALTHY`.** The v1 worker sets `UNHEALTHY` in exactly three places (graceful exit, health-server start failure, action-loop throw) — none occur here, so even the health endpoint reports a healthy worker throughout.

One engine detail that makes this worse: the `Heartbeat` handler deliberately **records the heartbeat before rejecting it** (so a stale stream doesn't cause in-flight task reassignment). Correct engine behavior — but it means the zombie's `lastHeartbeatAt` stays fresh, in-flight tasks are never reassigned, and no engine-side timeout ever clears the state. Recovery is entirely the SDK's responsibility, and the SDK never acts.

## Impact

Production, three occurrences within 24h (2026-08-27/28):

- One worker (fixed single-replica deployment) logged the rejection every ~4s for ~58 minutes (~867 consecutive matched error lines in its pod log, sampled lines all heartbeat rejections) with zero task assignments — every cron task targeting it failed with `Could not send task to worker...` until a manual pod restart.
- Two more workers in a different deployment hit the identical state concurrently (observed via centralized logs; both pods on the same node, consistent with a node-level network event severing streams engine-side).

Kubernetes never intervenes: the pod is Running, 0 restarts, and `/health` reports HEALTHY (and always returns 200 — filed separately as #4823). Fixed-replica deployments are fully exposed; autoscaled ones are partially masked by pod churn.

## Reproduction

Harness that drives the **real** SDK heartbeat pipeline — `new Heartbeat(dispatcherClient, workerId)` from `clients/dispatcher/heartbeat/heartbeat-controller`, real worker thread, real gRPC channel — against an in-process `nice-grpc` server whose `heartbeat` handler does:

```ts
throw new ServerError(Status.FAILED_PRECONDITION, `Heartbeat rejected: worker stream is not active: ${workerId}`);
```

The SDK loops on the rejection indefinitely, matching production byte-for-byte. Happy to contribute the harness as a test.

## Related work — this is the recovery gap, not the prevention gap

#3280 (closed as stale, unfixed) documents the most common *cause* of this state: L7 proxies killing idle `ListenV2` streams, with gRPC PINGs unable to prevent it because proxies answer them locally. #3281 (open) proposes opt-in server-side stream keepalives — **prevention**. This issue is the complementary half: **when a stream dies anyway** (proxy idle timeout before/without #3281, node-level network event, engine restart), the TypeScript SDK cannot detect or recover from the resulting state, even though the engine hands it an unambiguous signal every 4 seconds. Our incidents match #3280's mechanism (self-managed envoy in the path, streams dying after quiet periods), which is likely how many self-hosted deployments will encounter this.

## Suggested fixes (maintainer preference welcome — either or both)

1. **Self-heal**: on N consecutive stream-dead heartbeat rejections, tear down and re-establish the listener stream (analogous to a Kafka consumer rejoining after fencing). Requires distinguishing a reconnect-abort from the current shutdown-abort in the listener loop.
2. **At minimum, surface it**: mark the worker `UNHEALTHY` on the same condition, so the health endpoint (with the 503 fix) lets an orchestrator restart the worker.

## Environment

- `@hatchet-dev/typescript-sdk` 1.22.1 (core behavior verified unchanged on `main`, modulo the heartbeat deadline/severity changes noted above)
- Node.js 24, self-hosted engine, GKE + envoy gRPC proxy

## Contribution

Happy to submit a PR once there's direction on approach 1 vs 2.

RealBhupesh:
## Summary

Fixes #4824 — when the engine loses a worker's ListenV2 stream but unary heartbeats still reach the server, the TypeScript SDK now self-heals instead of entering a permanent zombie state.

- **Detect**: heartbeat worker tracks consecutive `FailedPrecondition` rejections whose message is `worker stream is not active`; after 3 consecutive failures, posts a `stream_inactive` IPC message
- **Reconnect**: `ActionListener` aborts the current ListenV2 with a reconnect-specific reason (distinct from shutdown), stops the heartbeat thread, and the existing `actions()` loop re-establishes via `getListenClient()` / `listenV2`. Capped at 10 reconnects with 3s backoff
- **Surface**: worker status flips to `UNHEALTHY` during reconnect; `/health` returns **503** when unhealthy (also fixes #4823 for K8s `httpGet` probes)

No engine changes.

## Test plan

- [x] `isWorkerStreamInactiveError` unit tests
- [x] shutdown vs reconnect abort distinction tests
- [x] action-listener reconnect does not exit generator; shutdown abort does
- [x] health-server returns 200 when HEALTHY, 503 when UNHEALTHY
- [ ] Manual validation against the nice-grpc repro harness from #4824 (happy to run if maintainers want)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/sdks/typescript/src/clients/dispatcher/action-listener.test.ts
+++ b/sdks/typescript/src/clients/dispatcher/action-listener.test.ts
@@ -1,10 +1,19 @@
 import { ActionType, AssignedAction } from '@hatchet/protoc/dispatcher';
 import sleep from '@util/sleep';
-// import { ServerError, Status } from 'nice-grpc-common';
+import { AbortError } from '@hatchet/util/abort-error';
 import { DEFAULT_LOGGER } from '@clients/hatchet-client/hatchet-logger';
 import { DispatcherClient } from './dispatcher-client';
 import { ActionListener } from './action-listener';
 import { mockChannel, mockFactory } from '../../legacy/legacy-client.test';
+import {
+  LISTENER_RECONNECT_REASON,
+  LISTENER_SHUTDOWN_REASON,
+} from './listener-abort';
+
+jest.mock('@util/sleep', () => ({
+  __esModule: true,
+  default: jest.fn(() => Promise.resolve()),
+}));
 
 let dispatcher: DispatcherClient;
 
@@ -253,5 +262,66 @@ describe('ActionListener', () => {
     //   expect(unsubscribeSpy).toHaveBeenCalled();
     //   expect(res.workerId).toEqual('WORKER_ID');
     // });
+  });
+
+  describe('inactive stream reconnect', () => {
+    it('reconnects on listener reconnect abort without exiting the generator', async () => {
+      const listener = new ActionListener(dispatcher, 'WORKER_ID', 1, 5);
+      const getListenClientSpy = jest.spyOn(listener, 'getListenClient');
+
+      getListenClientSpy
+        .mockResolvedValueOnce(
+          (async function* first() {
+            yield mockAssignedActions[0] as AssignedAction;
+            throw new AbortError(LISTENER_RECONNECT_REASON);
+          })()
+        )
+        .mockResolvedValueOnce(
+          (async function* second() {
+            yield mockAssignedActions[0] as AssignedAction;
+            throw new AbortError(LISTENER_SHUTDOWN_REASON);
+          })()
+        );
+
+      const actions = listener.actions();
+      const first = await actions.next();
+      expect(first.done).toBe(false);
+
+      const second = await actions.next();
+      expect(second.done).toBe(false);
+
+      expect(getListenClientSpy).toHaveBeenCalledTimes(2);
+    });
+
+    it('exits the generator on shutdown abort', async () => {
+      const listener = new ActionListener(dispatcher, 'WORKER_ID', 1, 5);
+      const getListenClientSpy = jest.spyOn(listener, 'getListenClient');
+
+      getListenClientSpy.mockResolvedValueOnce(
+        (async function* shutdown() {
+          yield mockAssignedActions[0] as AssignedAction;
+          throw new AbortError(LISTENER_SHUTDOWN_REASON);
+        })()
+      );
+
+      const actions = listener.actions();
+      const results = [];
+      for await (const action of actions) {
+        results.push(action);
+      }
+
+      expect(results).toHaveLength(1);
+      expect(getListenClientSpy).toHaveBeenCalledTimes(1);
+    });
+
+    it('marks worker unhealthy when stream goes inactive', () => {
+      const listener = new ActionListener(dispatcher, 'WORKER_ID');
+      const statusChanges: string[] = [];
+      listener.setWorkerStatusCallback((status) => statusChanges.push(status));
+
+      (listener as any).handleStreamInactive();
+
+      expect(statusChanges).toEqual(['UNHEALTHY']);
+    });
   });
 });
--- a/sdks/typescript/src/clients/dispatcher/action-listener.ts
+++ b/sdks/typescript/src/clients/dispatcher/action-listener.ts
@@ -11,9 +11,18 @@ import { Logger } from '@hatchet/util/logger';
 import { DispatcherClient } from './dispatcher-client';
 import { Heartbeat } from './heartbeat/heartbeat-controller';
 import { classifyListenerFailure } from './listener-severity';
+import {
+  isListenerReconnectAbort,
+  isListenerShutdownAbort,
+  LISTENER_RECONNECT_REASON,
+  LISTENER_SHUTDOWN_REASON,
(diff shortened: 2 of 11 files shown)


RealBhupesh:
No concerns with the implementation here — this was closed automatically by the `close-unaccepted-prs` workflow because #4824 is not yet labelled `accepted` and assigned (per CONTRIBUTING.md). That's expected for first-time contributors on this repo.

**Status:** implementation is complete on `RealBhupesh:fix/ts-sdk-listener-stream-recovery` (commit `62662bb`). I've asked maintainers on #4824 to add the `accepted` label and assign the issue; I'll reopen this PR as soon as that's in place.

@ubajwa-sketch — your verbal go-ahead is noted; we just need the formal triage step so the bot stops closing it. Thanks for offering to test against your repro once review starts.

PR #4910, opened 2026-09-07
feat(engine): engine-side gRPC operator support for out-of-process operators

abelanger5:
# Description

Extends operators to support gRPC-based operators and creates a new `operators.proto` service definition.  

These are basically just workers with some special and lower-level capabilities. In particular, they can: 
- Decompose durable tasks into async request / response streams, to enable lower-level operations (like mapping request / response streams to a websocket connection if necessary)
- Dynamically register actions, since they are in charge of registering their own workflows

These operators are tenant-scoped, so a separate gRPC operator needs to be enabled per tenant. They use the existing `HATCHET_CLIENT_TOKEN` auth mechanism. 

Can be enabled on the engine by setting `SERVER_GRPC_OPERATORS_ENABLED=true`. 

## Type of change

## Checklist

Changes have been:

---

<details id="ai-disclosure">

- **Details**: Claude w/ Fable 5.1 for planning, implementation, testing

</details>


cmd/hatchet-engine/engine/run.go
```diff
@@ -472,6 +473,27 @@ func runV0Config(ctx context.Context, sc *server.ServerConfig, cleanup *cleanup.
 			grpcOpts = append(grpcOpts, grpc.WithInsecure())
 		}
 
+		// the operator service is only registered behind the kill switch; without it callers
+		// get Unimplemented
```
abelanger5: remove this comment

internal/services/grpc/server_test.go
abelanger5, line 1: this file isn't really testing anything other than default gRPC server behavior - let's remove

api-contracts/v1/operator.proto
```diff
@@ -0,0 +1,195 @@
...
+    // Listen activates a registered worker for the lifetime of the stream, streams assigned
+    // actions to it, and accepts heartbeats and action set deltas from it.
+    //
+    // Protocol:
+    //   1. The first client message MUST be start, naming the worker returned by Register. Any
+    //      other first message is rejected with InvalidArgument, and a second start on the same
+    //      stream is also rejected. The server waits at most 30 seconds for it.
+    //   2. Every server message is an OperatorListenResponse: either an assigned action from the
+    //      dispatcher fan-out or an ack for an actions delta.
+    //   3. The client sends a heartbeat every 4 seconds. A worker whose heartbeat goes stale is
```
abelanger5: Operator heartbeats are kind of like worker heartbeats, but they're not out-of-stream. This is a simpler model than workers, but is justified because we are targeting the Go SDK for operators which is less subject to event loop blocked errors, which are the primary reason heartbeats run in a different thread from the stream.

api-contracts/v1/operator.proto
```diff
@@ -0,0 +1,195 @@
...
+    // Requires the hatchet-operator-id metadata.
+    rpc Listen(stream OperatorListenRequest) returns (stream OperatorListenResponse);
+
+    // PauseWorker stops the scheduler assigning to one of the operator's workers, or lets it be
+    // assigned to again with paused = false. It returns once the change is committed, so an
+    // operator that pauses before draining knows no further work will arrive, and it works while
+    // the Listen stream is being torn down or reconnected. Register clears the pause when it
+    // resumes a worker, so an operator that crashed while paused comes back assignable.
+    // Requires the hatchet-operator-id metadata; the worker_id must belong to the operator.
+    rpc PauseWorker(OperatorPauseWorkerRequest) returns (OperatorPauseWorkerResponse);
```
abelanger5: Should pause be a message on the listen stream? Ideally the Listen stream is somewhat stateful, I wonder if we should be internally requeuing messages which arrive after the pause event.

cmd/hatchet-engine/engine/operators.go
```diff
@@ -0,0 +1,107 @@
...
+func startOperatorClaimer(sc *server.ServerConfig, d *dispatcher.DispatcherImpl, adminv1Svc adminv1.AdminService) (stop func() error, err error) {
+	l := sc.Logger.With().Str("service", "operator-claimer").Logger()
+
+	svc, err := operatorsvc.New(
+		operatorsvc.Deps{
+			Operators:    sc.V1.Operators(),
+			Workers:      sc.V1.Workers(),
+			Dispatcher:   operatorsvc.NewDispatcherBackend(d),
+			DispatcherId: d.DispatcherId(),
+		},
```
abelanger5: while this `Deps` pattern isn't bad per se (I do kind of like it), it's inconsistent with how other things are initialized, where we only use variadic functions as args, even for required args. Let's refactor the constructors to keep them consistent.

cmd/hatchet-engine/engine/operators.go
```diff
@@ -0,0 +1,107 @@
...
+
+	if err != nil {
+		return nil, fmt.Errorf("could not create operator service: %w", err)
+	}
+
+	host, err := hostinproc.New(hostinproc.Deps{
+		Service:    svc,
+		Tenants:    sc.V1.Tenant(),
+		Heartbeats: sc.V1.Workers(),
+		Workflows:  inprocWorkflows{admin: adminv1Svc, workflows: sc.V1.Workflows()},
```
abelanger5: I'm confused why this is even necessary? Don't we just want to provide access to the `adminv1Svc` to the operator? This feels like an unnecessary wrapper.

pkg/repository/worker.go
```diff
@@ -123,6 +130,26 @@ type WorkerRepository interface {
 	// CreateNewWorker creates a new worker for a given tenant.
 	CreateNewWorker(ctx context.Context, tenantId uuid.UUID, opts *CreateWorkerOpts) (*sqlcv1.Worker, error)
 
+	// AddWorkerActions links actionIds to the worker and recomputes its action hash from the
+	// resulting set. Actions the worker already has are skipped. It returns the number of
+	// actions actually linked. The worker must belong to tenantId; otherwise nothing is
+	// mutated and an error wrapping pgx.ErrNoRows is returned.
+	AddWorkerActions(ctx context.Context, tenantId uuid.UUID, workerId uuid.UUID, actionIds []string) (added int, err error)
```
abelanger5: Should add/remove not be transactional if it shares an operation / sequence number?

pkg/repository/worker.go
```diff
@@ -534,35 +561,81 @@ func (w *workerRepository) GetWorkerForEngine(ctx context.Context, tenantId uuid
...
 	for _, units := range slotConfig {
 		slots += units
 	}
 
-	preWorkerSlot, postWorkerSlot := w.m.Meter(ctx, nil, sqlcv1.LimitResourceWORKERSLOT, tenantId, slots)
+	// Operator workers are excluded from the WORKER and WORKER_SLOT limit counts by the
+	// "operatorId" IS NULL filters in workers.sql and tenant_limits.sql, so metering them here
+	// would charge for workers the limit queries never see.
+	postWorker := func() {}
+	postWorkerSlot := func() {}
```
abelanger5: operator workers should count towards the worker limits, though, I think this is incorrect

pkg/operator/operator.go
```diff
@@ -78,37 +82,44 @@ type DAGStepTriggerResult struct {
...
 
+// TaskEventWriter is the engine-internal surface for engine-internal operators (the DAG
+// operator): calls that only exist inside the engine and are never available over gRPC. The
+// dispatcher implements it. Everything an operator needs that both hosts offer (events,
+// durable invocations, the action set) is on Session instead.
 type TaskEventWriter interface {
-	SendStepActionEvent(ctx context.Context, request *contracts.StepActionEvent) (*contracts.ActionEventResponse, error)
-
+	// CancelTaskEvent reports a cancelled task with a custom reason.
 	CancelTaskEvent(ctx context.Context, request *contracts.StepActionEvent) (*contracts.ActionEventResponse, error)
```
abelanger5: should this not be available over gRPC? or perhaps it needs a better name, like `CancelTaskEventCustom` to disambiguate

pkg/operator/operator.go
```diff
@@ -133,24 +163,90 @@ func (s *SharedOperator[T]) Logger() *zerolog.Logger {
...
+	return true, nil
 }
 
 func (s *SharedOperator[T]) TriggerDAGStep(ctx context.Context, req *DAGStepTriggerRequest) (*DAGStepTriggerResult, error) {
 	if s.taskEventWriter == nil {
 		return nil, fmt.Errorf("operator has no task event writer configured")
 	}
 
-	ctx = context.WithValue(ctx, tenantContextKey, &sqlcv1.Tenant{ID: s.tenantId})
+	ctx = context.WithValue(ctx, tenantContextKey, &sqlcv1.Tenant{ID: s.tenantId}) // nolint:staticcheck // key must match the dispatcher's
```
abelanger5: this feels super janky and prone to error

pkg/operator/operator.go
```diff
@@ -133,24 +163,90 @@ func (s *SharedOperator[T]) Logger() *zerolog.Logger {
...
+	add := make([]string, 0)
+
+	for _, id := range actions {
+		if _, ok := desired[id]; ok {
+			continue
+		}
+
+		desired[id] = struct{}{}
+
+		if _, ok := s.lastActions[id]; !ok {
```
abelanger5, line 212: do we need to guard `lastActions` here at all?

pkg/operator/hostgrpc/durable.go
```diff
@@ -0,0 +1,599 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+	"time"
+
+	v1 "github.com/hatchet-dev/hatchet/internal/services/shared/proto/v1"
+	"github.com/hatchet-dev/hatchet/pkg/client" //nolint:staticcheck // OperatorService's client lives in the legacy client package
```
abelanger5: this feels ugly. I imagine this is so we can re-use `reconnecting_stream` or some of the other shared listener logic, but I wonder if we can just move those into a shared set of helper packages and create an `operatorclient` package instead.

pkg/operator/dagoperator/operator.go
```diff
@@ -72,17 +70,14 @@ type DAGOperator struct {
...
-	// touches it.
-	lastActions []string
-
 	// runCancels lets a CANCEL_STEP_RUN action stop one run without cancelling d.ctx.
 	runCancels syncx.Map[string, context.CancelFunc]
 
 	slots int
 }
 
+var _ operator.Operator = (*DAGOperator)(nil)
```
abelanger5: this feels...out of place? what's the purpose of this?

pkg/config/server/server.go
```diff
@@ -874,6 +888,9 @@ func BindAllEnv(v *viper.Viper) {
 	_ = v.BindEnv("runtime.allowedOriginsString", "SERVER_ALLOWED_ORIGINS")
 	_ = v.BindEnv("runtime.operatorInfraBlockedCIDRsString", "SERVER_OPERATOR_INFRA_BLOCKED_CIDRS")
 	_ = v.BindEnv("runtime.dagOperatorDefaultSlots", "SERVER_DAG_OPERATOR_DEFAULT_SLOTS")
+	_ = v.BindEnv("runtime.grpcOperatorsEnabled", "SERVER_GRPC_OPERATORS_ENABLED")
+	_ = v.BindEnv("runtime.grpcOperatorMaxListenStreamsPerOperator", "SERVER_GRPC_OPERATOR_MAX_LISTEN_STREAMS_PER_OPERATOR")
+	_ = v.BindEnv("runtime.grpcOperatorMaxActionsPerOperator", "SERVER_GRPC_OPERATOR_MAX_ACTIONS_PER_OPERATOR")
```
abelanger5, line 893: didn't see all of these documented in `frontend/docs/content/docs/self-hosting/configuration-options.mdx`

sql/schema/v0.sql
```diff
@@ -874,7 +874,15 @@ CREATE TABLE "Worker" (
...
     "durableTaskDispatcherId" UUID,
+    -- The canonical digest of the worker's linked action set (ComputeWorkerActionHash). NULL
+    -- for workers registered before the column existed, and for operator workers between a
+    -- delta that changed their links and the refresh at the end of the delta sequence; the
+    -- scheduler reads such workers' actions through the join instead of the hash cache.
     "actionHash" BYTEA,
+    -- The number of "_ActionToWorker" rows the worker holds. Every path that links or unlinks
+    -- actions maintains it, so the per-operator action budget is the sum over the operator's
+    -- workers rather than a count of their links.
+    "actionCount" INTEGER NOT NULL DEFAULT 0,
```
abelanger5: nit: rename this to `"operatorActionCount"` instead if it's zero for regular workers

api-contracts/v1/operator.proto
```diff
@@ -0,0 +1,205 @@
...
+    // Action ids the worker can now run.
+    repeated string add = 1;
+
+    // Action ids the worker no longer runs.
+    repeated string remove = 2;
+
+    // Sequence number of this delta on the stream: positive and strictly increasing. The
+    // server acknowledges the delta with an OperatorActionsAck carrying the same value once
+    // the change is committed. 0 asks for no ack.
+    uint64 sequence = 3;
```
mrkaye97, line 170: I think this is sort of a confusing name, although I get the intention - I wonder if `index` or something would be better

api-contracts/v1/operator.proto
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// OperatorPause stops the scheduler assigning to the stream's worker, or lets it be assigned to
+// again. The server answers with an OperatorPauseAck once the change is committed; after the ack
+// of a pause nothing is delivered on the stream until the pause is lifted.
+message OperatorPause {
+    // True to stop the scheduler assigning to the worker, false to let it be assigned to again.
+    bool paused = 1;
```
mrkaye97: `is_paused`?

cmd/hatchet-engine/engine/run.go
```diff
@@ -369,7 +370,6 @@ func runV0Config(ctx context.Context, sc *server.ServerConfig, cleanup *cleanup.
 			dispatcher.WithStreamEventBufferTimeout(sc.Runtime.StreamEventBufferTimeout),
 			dispatcher.WithVersion(sc.Version),
 			dispatcher.WithAnalytics(sc.Analytics),
-			dispatcher.WithDAGOperatorDefaultSlots(sc.Runtime.DagOperatorDefaultSlots),
```
mrkaye97, line 372: intentional to remove this? (also below)

cmd/hatchet-migrate/migrate/migrations/20260919090000_v1_0_156.sql
```diff
@@ -0,0 +1,79 @@
+-- +goose NO TRANSACTION
+-- Runs without a wrapping transaction on purpose: goose then sends each statement on its own in
+-- autocommit mode. Postgres refuses to use a value that ALTER TYPE ... ADD VALUE added in the
+-- same transaction, and the backfills below use the values this file adds, so the statements
+-- have to commit one at a time. The cost is that a failure part way through leaves the earlier
+-- statements applied; undo them by hand with the Down section before re-running.
```
mrkaye97: I disagree with this? unless I'm missing it, we don't actually backfill with this value in this tx (could be two separate migrations if we did though to maintain the tx in the migrations)

cmd/hatchet-migrate/migrate/migrations/20260919090000_v1_0_156.sql
```diff
@@ -0,0 +1,79 @@
...
+    AND (first.created_at, first.id) < (dup.created_at, dup.id);
+
+CREATE UNIQUE INDEX v1_operator_tenant_name_kind_key ON v1_operator (tenant_id, name, kind);
+
+-- "operatorActionCount" is the number of "_ActionToWorker" rows the worker holds, kept for the
+-- per-operator action budget: every path that links or unlinks actions maintains it, so the
+-- budget is a sum over the operator's workers rather than a count of their links. Only operator
+-- workers are read, and only they are backfilled; an SDK worker's count is zero until it next
+-- registers and nothing reads it. "actionHash" is untouched: the hash encoding is the one
+-- workers already carry.
```
mrkaye97: can we get rid of comments like this? it's obvious from the name `operatorActionCount` what it's doing, and it makes it tough to read the migration

cmd/hatchet-migrate/migrate/migrations/20260919090000_v1_0_156.sql
```diff
@@ -0,0 +1,79 @@
...
+
+UPDATE "Worker" w
+SET "operatorActionCount" = (SELECT count(*) FROM "_ActionToWorker" aw WHERE aw."B" = w."id")
+WHERE w."operatorId" IS NOT NULL;
+
+-- Whether a worker counts against the tenant's WORKER and WORKER_SLOT limits is decided when
+-- the worker is created, by whoever hosts it: the in-process host exempts every worker it
+-- creates, the wire meters every worker it registers. It is a fact about the worker, not about
+-- the operator row, so the limit queries read it here rather than through a join.
+ALTER TABLE "Worker" ADD COLUMN "exemptFromLimits" BOOLEAN NOT NULL DEFAULT false;
```
mrkaye97: `isExemptFromLimits`?

cmd/hatchet-migrate/migrate/migrations/20260919090000_v1_0_156.sql
```diff
@@ -0,0 +1,79 @@
...
+-- Who keeps an operator alive is a separate axis from what it is. For a DISPATCHER row the
+-- dispatcher claims the row through ClaimOperators and builds the operator from a factory
+-- inside the engine; a SELF row keeps itself alive, through a Listen stream out of process or
+-- its own leaser in process, and is never claimed.
+CREATE TYPE v1_operator_leasing_manager AS ENUM ('SELF', 'DISPATCHER');
+
+ALTER TABLE v1_operator ADD COLUMN leasing_manager v1_operator_leasing_manager NOT NULL DEFAULT 'SELF';
+
+-- Every DAG operator row was claimed by a dispatcher before the column existed.
+UPDATE v1_operator SET leasing_manager = 'DISPATCHER' WHERE kind = 'DAG';
```
mrkaye97, line 19: check constraint for this maybe?

cmd/hatchet-migrate/migrate/migrations/20260919090000_v1_0_156.sql
```diff
@@ -0,0 +1,79 @@
...
+-- creating a tenant's DAG operator) is renamed rather than deleted: the oldest row keeps its
+-- name, a later one gets its id as a suffix.
+UPDATE v1_operator dup
+SET name = dup.name || '-' || dup.id::text
+FROM v1_operator first
+WHERE
+    first.tenant_id = dup.tenant_id
+    AND first.name = dup.name
+    AND first.kind = dup.kind
+    AND (first.created_at, first.id) < (dup.created_at, dup.id);
```
mrkaye97, line 33: do we need this? we only have operators turned on on a couple of dbs, and they only seem to have one row each in this table right now (sometimes one row per tenant), and I can't imagine anyone in the OSS is using this since they'd need to have manually turned on the row in the `tenant_entitlements`

cmd/hatchet-migrate/migrate/migrations/20260919090000_v1_0_156.sql
```diff
@@ -0,0 +1,79 @@
...
+UPDATE "Worker" w
+SET "exemptFromLimits" = true
+FROM v1_operator op
+WHERE op.id = w."operatorId" AND op.kind = 'DAG';
+
+-- +goose Down
+
+ALTER TABLE "Worker" DROP COLUMN "exemptFromLimits";
+
+ALTER TABLE "Worker" DROP COLUMN "operatorActionCount";
```
mrkaye97: single statement for this?


```suggestion
ALTER TABLE "Worker" 
  DROP COLUMN "exemptFromLimits",
  DROP COLUMN "operatorActionCount"
;
```

cmd/hatchet-migrate/migrate/migrations/20260919090000_v1_0_156.sql
```diff
@@ -0,0 +1,79 @@
...
+
+DROP INDEX IF EXISTS v1_operator_tenant_name_kind_key;
```
mrkaye97: don't need `IF EXISTS` once we make this run in a tx

cmd/hatchet-engine/engine/run.go
mrkaye97: ah I see, we do this in `claimer.go` now?

internal/services/dispatcher/server.go
```diff
@@ -1646,10 +1646,10 @@ func (s *DispatcherImpl) handleTaskFailed(inputCtx context.Context, task *sqlcv1
...
 }
 
-func (d *DispatcherImpl) CancelTaskEvent(ctx context.Context, request *contracts.StepActionEvent) (*contracts.ActionEventResponse, error) {
-	tenant := ctx.Value("tenant").(*sqlcv1.Tenant)
-	tenantId := tenant.ID
-
+// CancelTaskEventCustom reports a cancelled task with a custom cancellation reason on behalf of
+// an engine-internal operator (operator.TaskEventWriter). It is not a gRPC handler: the tenant
+// is an argument, not the one the auth middleware puts on a request context.
+func (d *DispatcherImpl) CancelTaskEventCustom(ctx context.Context, tenantId uuid.UUID, request *contracts.StepActionEvent) (*contracts.ActionEventResponse, error) {
```
mrkaye97: kind of odd naming?

internal/services/grpcoperator/register.go
```diff
@@ -0,0 +1,69 @@
...
+		TenantId:   reg.TenantId.String(),
+		OperatorId: reg.OperatorId.String(),
+		WorkerId:   reg.WorkerId.String(),
+		Resumed:    reg.Resumed,
+	}, nil
+}
+
+// parseResumeWorkerId reads the optional worker id a client sends to resume its previous worker.
+// An absent or empty value asks for a new worker.
+func parseResumeWorkerId(raw *string) (*uuid.UUID, error) {
```
mrkaye97, line 58: nit: would be great to make a helper like `uuidFromMaybeString` and use it everywhere, I imagine we do this a lot

pkg/operator/dagoperator/dag.go
```diff
@@ -99,10 +99,6 @@ type dag struct {
 	// (its condition is committed to the DB before the ack is sent); rechecked on each ack.
 	pendingEntryCompletions []pendingEntryCompletion
 
-	// responses received while blocked sending on requestCh (see dag.send); drained by the
-	// main loop before it blocks on responseCh
-	queuedResponses []*v1contracts.DurableTaskResponse
```
mrkaye97, line 104: intentional?

pkg/operator/dagoperator/dag.go
mrkaye97, line 1: we should make sure that all these changes look good relative to main when this one is at the top of the merge stack, since I've been modifying this file a lot lately and it seems pretty likely there would be a merge conflict issue

pkg/operator/dagoperator/operator.go
```diff
@@ -72,11 +70,6 @@ type DAGOperator struct {
 	ctx    context.Context
 	cancel context.CancelFunc
 
-	// lastActions is the most recently registered action set, used to avoid redundant
-	// dispatcher writes when the workflow list is unchanged. Only the polling goroutine
-	// touches it.
-	lastActions []string
```
mrkaye97, line 78: intentional?

pkg/operator/dagoperator/operator.go
```diff
@@ -340,34 +332,19 @@ func (d *DAGOperator) run(action *contracts.AssignedAction) error {
...
+		}
 
-	select {
-	case requestCh <- &v1contracts.DurableTaskRequest{
-		Message: &v1contracts.DurableTaskRequest_RegisterWorker{
-			RegisterWorker: &v1contracts.DurableTaskRequestRegisterWorker{
-				WorkerId: d.WorkerId().String(),
-			},
-		},
-	}:
```
mrkaye97, line 358: intentional?

pkg/operator/hostgrpc/durable.go
```diff
@@ -0,0 +1,645 @@
...
+//
+// Requests reach the listener through the hub's bounded outbound queue and one pump
+// goroutine: the listener's own enqueue has no cancellation, so only the pump ever blocks on
+// it, while every channel's Send selects on its own close signal.
+type durableHub struct {
+	listener *durableTaskListener
+	channels map[string]map[int32]*durableChannel
+	outbound chan outboundRequest
+	stopped  chan struct{}
+	pumpDone chan struct{}
```
mrkaye97: I have a thing in my `CLAUDE.md` to tell it to not use words like "pump" because I have no idea what this means - what is a pump?!?!

pkg/operator/hostgrpc/durable.go
```diff
@@ -0,0 +1,645 @@
...
+
+func (h *durableHub) open(taskId string, invocation int32) (*durableChannel, error) {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+
+	if h.closed {
+		return nil, operator.ErrSessionClosed
+	}
+
+	byInvocation := h.channels[taskId]
```
mrkaye97: naming thing, but _what_ by invocation? task?

pkg/repository/sqlcv1/workers.sql
```diff
@@ -492,6 +500,139 @@ INSERT INTO "_ActionToWorker" (
...
+    gen_random_uuid(),
+    a.action,
+    @tenantId::uuid
+FROM unnest(@actions::text[]) AS a(action)
+ON CONFLICT ("tenantId", "actionId") DO NOTHING
+RETURNING "id", "actionId";
+
+-- name: LinkActionsToWorkerReturning :many
+-- Links the worker to the given action rows and returns the action row ids that were newly
+-- linked, so the caller knows whether the set changed. The Worker and Action rows are joined
```
mrkaye97: naming here is kind of awkward

pkg/repository/sqlcv1/workers.sql
```diff
@@ -492,6 +500,139 @@ INSERT INTO "_ActionToWorker" (
...
+    AND "tenantId" = @tenantId::uuid
+FOR UPDATE;
+
+-- name: InsertMissingActions :many
+-- Creates the Action rows in @actions that do not exist yet and returns the rows it created.
+-- Existing rows are left untouched (no lock, no rewrite), so concurrent transactions that
+-- reference the same actions do not contend on them. @actions must be lower-cased, free of
+-- duplicates and sorted: two transactions creating overlapping sets then take their row locks
+-- in the same order and cannot deadlock. An action another transaction creates concurrently is
+-- absent from the result; the caller resolves it with ListActionsByActionIds afterwards.
```
mrkaye97: can we remove this comment? not helpful imo

also a bit confused about this query honestly, does this not do a pretty similar thing to the existing action insert in `UpsertAction`? (I guess that one is not bulk, but it could be)

pkg/repository/worker.go
```diff
@@ -717,6 +804,349 @@ func (w *workerRepository) CreateNewWorker(ctx context.Context, tenantId uuid.UU
...
+func (w *workerRepository) refreshWorkerActionHash(ctx context.Context, tx pgx.Tx, workerId uuid.UUID) error {
+	hash, err := w.queries.ComputeWorkerActionHash(ctx, tx, workerId)
+
+	if err != nil {
+		return fmt.Errorf("could not compute worker actions hash: %w", err)
+	}
+
+	if err := w.queries.UpdateWorkerActionsHash(ctx, tx, sqlcv1.UpdateWorkerActionsHashParams{
+		Workerid:   workerId,
+		Actionhash: hash,
```
mrkaye97: not sure I follow this - why not consolidate `ComputeWorkerActionHash` and `UpdateWorkerActionsHash` into a single query?

pkg/repository/worker_action_hash_test.go
```diff
@@ -0,0 +1,343 @@
...
+	_, err = addWorkerActions(repo, ctx, tenantB, worker, []string{"svc:run"})
+	require.NoError(t, err)
+
+	_, err = removeWorkerActions(repo, ctx, tenantA, worker, []string{"svc:run"})
+	require.Error(t, err, "a worker of another tenant must not be unlinked")
+	assert.Equal(t, []string{"svc:run"}, linkedActions(t, ctx, pool, worker))
+
+	var cross int
+	require.NoError(t, pool.QueryRow(ctx,
+		`SELECT count(*) FROM "_ActionToWorker" aw JOIN "Action" a ON a.id = aw."A" JOIN "Worker" w ON w.id = aw."B" WHERE aw."B" = $1 AND a."tenantId" <> w."tenantId"`,
```
mrkaye97: claude for some reason really seems to like writing raw SQL in tests, but I feel like it'd be better to use the existing action getter by worker id and then doing the count in memory so we need to maintain less SQL in the tests

pkg/repository/sqlcv1/workers.sql
```diff
@@ -492,6 +500,139 @@ INSERT INTO "_ActionToWorker" (
...
+
+-- name: ListActionsByActionIds :many
+-- Resolves action ids to their rows. @actionIds are compared as stored (lower-cased).
+SELECT "id", "actionId"
+FROM "Action"
+WHERE
+    "tenantId" = @tenantId::uuid
+    AND "actionId" = ANY(@actionIds::text[]);
+
+-- name: UnlinkActionsFromWorkerReturning :many
```
mrkaye97: similar awkward naming here, feels odd to put "Returning" in the query name

pkg/repository/sqlcv1/workers.sql
```diff
@@ -541,6 +682,20 @@ WHERE
...
+-- session owns, so this returns no rows in that case.
+UPDATE "Worker"
+SET
+    "isPaused" = @paused::boolean,
+    "updatedAt" = CURRENT_TIMESTAMP
+WHERE
+    "id" = @id::uuid
+    AND "tenantId" = @tenantId::uuid
+    AND "lastListenerSessionId" = @sessionId::uuid
+RETURNING "id";
```
mrkaye97: can we use the existing set worker paused query for this by adding an optional session id with our normal `AND (@sessionId::UUID IS NULL OR "lastListenerSessionId" = @sessionId::UUID)` thing instead of adding a new query?

sql/schema/v1-core.sql
```diff
@@ -2860,22 +2860,34 @@ CREATE TABLE v1_durable_event_log_branch_point (
...
     kind v1_operator_kind NOT NULL,
+    leasing_manager v1_operator_leasing_manager NOT NULL DEFAULT 'SELF',
     config JSONB NOT NULL,
     worker_id UUID,
     created_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
     updated_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
     CONSTRAINT v1_operator_pkey PRIMARY KEY (id)
 );
 
+-- Operators are upserted by (tenant, name, kind) on registration, so that triple is unique.
```
mrkaye97: rm comment here?

mrkaye97 (commented):
left some comments, mostly on the sql, lots of code to review here 😰

cmd/hatchet-migrate/migrate/migrations/20260919090000_v1_0_156.sql
abelanger5: yeah nice catch, this comment is wrong. I added this because we're doing a bunch of `ALTER TABLE` on high volume tables followed by backfill queries, which is risky and has bitten us in the past. At the very least, I'd prefer to do two `ALTER TABLE` commands at the top of the file, ideally in a single idempotent statement, and then do the backfills in a separate tx or migration. I still think a single migration file per PR is cleanest.

api-contracts/v1/operator.proto
abelanger5: yep, done

api-contracts/v1/operator.proto
abelanger5: I think `sequence` works? at least, it makes sense to me

cmd/hatchet-engine/engine/run.go
abelanger5: yep exactly

cmd/hatchet-migrate/migrate/migrations/20260919090000_v1_0_156.sql
abelanger5: done 👍

cmd/hatchet-migrate/migrate/migrations/20260919090000_v1_0_156.sql
abelanger5: done 👍

cmd/hatchet-migrate/migrate/migrations/20260919090000_v1_0_156.sql
abelanger5: moved this up and consolidated, see the other comment

internal/services/dispatcher/server.go
abelanger5: renamed to `CancelTaskWithReason`

pkg/operator/dagoperator/dag.go
abelanger5: yes, we moved this to the session

pkg/operator/dagoperator/operator.go
abelanger5: yep, also moved to the session layer, we need to reuse this across operators

pkg/operator/dagoperator/operator.go
abelanger5: this is on `Session.OpenDurable` now

pkg/operator/hostgrpc/durable.go
abelanger5: agreed that `pump` is terrible, replaced it with `recvLoop` instead

pkg/operator/hostgrpc/durable.go
abelanger5: it's a set of channels for that task, so renamed to `channelsByInvocation`

pkg/repository/sqlcv1/workers.sql
abelanger5: fixed

pkg/repository/worker.go
abelanger5: yep, good call

pkg/repository/worker_action_hash_test.go
abelanger5: fixed

cmd/hatchet-migrate/migrate/migrations/20260919090000_v1_0_156.sql
abelanger5: added 👍

cmd/hatchet-migrate/migrate/migrations/20260919090000_v1_0_156.sql
abelanger5: we probably don't, but going to leave it just to be safe

cmd/hatchet-migrate/migrate/migrations/20260919090000_v1_0_156.sql
abelanger5: removed for the down migration since it's a single tx

grutt (approved):
Chatted with the diff in cursor, I think this is safe to merge

mrkaye97 (approved):

PR #4911, opened 2026-09-07
feat(cli): `hatchet mcp serve` with a new MCP implementation

abelanger5:
# Description

New Hatchet MCP implementation which hooks into the local Hatchet CLI and profiles to support tool calls: trigger_run (with wait), get_run, list_run_events, list_workers, replay_run, engine_status, and submit_feedback. 

This replaces the old docs MCP implementation, which wasn't useful since agents are getting much better at reading docs markdown files directly (and we've done a lot of work to improve reference links / docs for agents). 

This is implemented as a local stdio MCP server so that we don't need to build a separate OAuth layer, which would need to be enabled on every Hatchet instance. When running `hatchet mcp install`, it prompts the user to select profiles they'd like to make available to the MCP, which can be controlled via `hatchet mcp auth`. The actual profile grants live in `~/.hatchet/mcp-grants.yaml`. Note that **a running embedded instance is implicitly granted**, otherwise the interaction doesn't make sense. 

The auth layer is meant to guard against accidental writes or triggers against production data, but not meant as a security boundary for the agent (since the mcp server runs local, the grants file is local, etc).

## Type of change

## Checklist

Changes have been:

---

<details id="ai-disclosure">

- **Details**: Claude w/ Fable 5.1

</details>


cmd/hatchet-cli/cli/internal/mcp/engine.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/google/uuid"
+
+	"github.com/hatchet-dev/hatchet/pkg/client" //nolint:staticcheck
+	"github.com/hatchet-dev/hatchet/pkg/client/rest"
+	cliconfig "github.com/hatchet-dev/hatchet/pkg/config/cli"
+)
+
+// Engine is the narrow slice of the Hatchet clients that the MCP tools need.
+// Tool handlers depend on this interface so they can be tested with fakes.
+type Engine interface {
```
grutt, line 17: can we call this something different from engine to avoid confusion...

grutt (approved):

mnafees (approved):

PR #4913, opened 2026-09-07
feat: pay-as-you-go billing (usage page + retention UI)

grutt:
## Summary
- Exposes tenant `dataRetentionPeriod` and gates events, logs, and runs history with a retention upgrade modal / expired-run state instead of a silent empty table.
- Adds an Invoices section to the organization Billing & Usage page (upcoming preview dialog, previous invoices via Autumn hosted URLs). Invoice loading is isolated so a failure does not break plan management.
- Combines the improved billing/usage UI with the enforce-retention dashboard work on a single `feat/pay-as-you-go-billing` branch.

Supersedes https://github.com/hatchet-dev/hatchet/pull/4812. Depends on the control-plane companion: https://github.com/hatchet-dev/hatchet-control-plane/pull/134

## Test plan
- [ ] Retention period is present on tenant memberships after the control-plane deploy
- [ ] Time pickers past retention open the upgrade modal (events, logs, runs)
- [ ] Expired run detail shows the expired-run empty state
- [ ] Billing & Usage page shows previous invoices for the seeded `sdks` org
- [ ] Upcoming invoice dialog on an org with a live paid Stripe subscription
- [ ] Previous invoice row opens the hosted invoice in a new tab
- [ ] Invoice query error leaves subscription UI usable


api/v1/server/handlers/v1/retention.go
```diff
@@ -0,0 +1,22 @@
+package v1
+
+import "time"
+
+const defaultRetention = 720 * time.Hour
+
+func retentionBoundary(retentionPeriod string) time.Time {
```
mrkaye97, line 7: can we make this a `time.Duration` and get rid of the parse inside? feels better typed that way 

```suggestion
func retentionBoundary(retentionPeriod time.Duration) time.Time {
```

frontend/app/src/components/sso/SsoDeleteConfirmationDialog.tsx
```diff
@@ -17,9 +17,9 @@ export function SsoDeleteConfirmationDialog() {
     <Dialog open={open} onOpenChange={onOpenChange}>
       <DialogContent>
         <DialogHeader>
-          <DialogTitle>Delete SSO Configuration?</DialogTitle>
+          <DialogTitle>Delete Single Sign On Configuration?</DialogTitle>
```
mrkaye97, line 20: grammatical nit: I think this should be `Single Sign-On` everywhere

frontend/app/src/components/v1/cloud/billing/invoice-formatters.ts
```diff
@@ -0,0 +1,96 @@
...
+export function formatQuantity(quantity: number) {
+  return new Intl.NumberFormat('en-US').format(quantity);
+}
+
+export type InvoiceStatusTone = {
+  label: string;
+  variant: BadgeProps['variant'];
+};
+
+export function invoiceStatusBadge(status?: string): InvoiceStatusTone {
```
mrkaye97, line 74: seems like `status` should be an enum and we should have an exhaustiveness check in the switch

frontend/app/src/components/v1/cloud/billing/subscription-plan-code.ts
```diff
@@ -1,8 +1,28 @@
...
 
 export function resolveSubscriptionPlanCode(
   subscription: OrganizationBillingStateSubscription | undefined,
   fallback: string | null,
 ) {
   return subscription?.planCode ?? fallback;
 }
+
+export function planCodeBase(planCode?: string | null) {
+  return (planCode ?? '').split('_')[0];
```
mrkaye97, line 14: should we make plan code an enum too? maybe it could be an enum in the openapi spec and that would enforce it's the same on both sides

frontend/app/src/components/v1/cloud/billing/upgrade-gate-dialog.tsx
```diff
@@ -0,0 +1,330 @@
...
+    ?.flatMap((group) => group.features)
+    .find((feature) => feature.featureId === featureId);
+}
+
+function formatIncluded(feature?: SubscriptionPlanFeature, fallback = '—') {
+  if (!feature) {
+    return fallback;
+  }
+  if (feature.unlimited) {
+    return '∞';
```
mrkaye97, line 79: does this render right everywhere? seems like the kind of thing we might need to escape or use some custom character for

frontend/app/src/components/v1/cloud/billing/upgrade-gate-dialog.tsx
```diff
@@ -0,0 +1,330 @@
...
+    tenants: {
+      title: 'Take Hatchet to production',
+      context: PAYG_SUBHEAD,
+    },
+    users: {
+      title: 'Bring your whole team',
+      context: PAYG_SUBHEAD,
+    },
+    retention: {
+      title: 'Debug with a week of history',
```
mrkaye97, line 145: just flagging that at some point, this is the sort of thing we'll need templating for (both for i18n purposes and to make it easier to edit without needing to be in the code). nothing to do here for now, but some day

frontend/app/src/components/v1/cloud/billing/upgrade-gate-dialog.tsx
```diff
@@ -0,0 +1,330 @@
...
+  retentionPeriod,
+}: UpgradeGateProps) {
+  const { canBill, isControlPlaneEnabled } = useControlPlane();
+  const { tenant } = useTenantDetails();
+  const { entitlements } = useOrganizationEntitlements(organizationId);
+  const plansQuery = useQuery({
+    ...queries.controlPlane.subscriptionPlans(),
+    enabled: isControlPlaneEnabled && canBill,
+  });
+  const payg = payAsYouGoPlan(plansQuery.data?.plans);
```
mrkaye97, line 215: nit from reading the code outside of an IDE: is this a boolean? if yes, can if be `isPayg`?

frontend/app/src/lib/utils/retention.ts
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * Returns true if the given date falls before the retention boundary.
+ * Uses a 5-minute tolerance so that filter windows matching the retention
+ * period (e.g. "1 day" filter with 24h retention) don't falsely trigger.
+ */
+const RETENTION_TOLERANCE_MS = 5 * 60 * 1000;
+
+export function isBeforeRetention(
+  date: string | Date,
```
mrkaye97, line 39: why not enforce this is a date instead of allowing a string?

mrkaye97 (approved):
It's a lot to review so I looked through the API surface and the helpers, Claude might be able to extrapolate from my comments a bit. Mostly looks good, but one thing to flag is unless I'm misunderstanding, it seems like we're gating in the date filter in the UI, right? If yes, then the SDKs can still get around this?

frontend/app/src/components/v1/cloud/billing/upgrade-gate-dialog.tsx
grutt: agree

frontend/docs/content/docs/v1/single-sign-on.mdx
```diff
@@ -7,7 +7,7 @@ import { Callout, Card, Cards, Steps, Tabs } from "@/components/nextra-compat";
 # Single-Sign-On
 
 <Callout type="info">
-  Single Sign On is only available through our Cloud Scale plan. See
+  Single Sign-On is only available through our Cloud Scale plan. See
   [pricing](https://hatchet.run/pricing) for more details.
 </Callout>
```
abelanger5: probably need to update all references to the scale plan to update to custom

api/v1/server/handlers/v1/events/get.go
```diff
@@ -4,14 +4,24 @@ import (
...
 
 func (t *V1EventsService) V1EventGet(ctx echo.Context, request gen.V1EventGetRequestObject) (gen.V1EventGetResponseObject, error) {
+	tenant := ctx.Get("tenant").(*sqlcv1.Tenant)
 	event := ctx.Get("v1-event").(*v1.EventWithPayload)
 
+	if ts := event.EventSeenAt; ts.Valid && v1handlers.IsBeforeRetention(ts.Time, tenant.DataRetentionPeriod) {
+		t.config.Analytics.Count(ctx.Request().Context(), analytics.Event, analytics.Get, analytics.Properties{
+			"outside_retention": true,
+		})
+	}
```
abelanger5: just to clarify we don't block these queries if they fall outside the retention? wouldn't a proper error be better here?

abelanger5 (approved):
approving to unblock

api/v1/server/handlers/v1/events/get.go
grutt: discussed offline

PR #4916, opened 2026-09-08
fix: guard worker `isActive` using a session id instead of just a timestamp

abelanger5:
# Description

There's a very rare edge case where a worker which is reconnecting via `ListenV2` is marked as inactive when it's still active, because the successful reconnect `ListenV2()` happens before the previous `ListenV2` stream has fully shut down. There are really two separate issues here:
1. `lastListenerEstablished` is a `TIMESTAMP(3)` (thanks Prisma) which means it truncates to milliseconds, so if this happened in the same millisecond it's super racy
2. The fact that a new `ListenV2` could be superseded by an old `ListenV2` which no longer had worker stream ownership in the first place 

There's some improved / more explicit timestamp handling for the first issue, and a new column `lastListenerSessionId` for the second issue.

This was uncovered during reconnect tests for gRPC-based operators which hook into worker `isActive`. 

## Type of change

## Checklist

Changes have been:

---

<details id="ai-disclosure">

- **Details**: Claude w/ Fable 5.1 

</details>

### DIFF
--- a/cmd/hatchet-migrate/migrate/migrations/20260908090000_v1_0_152.sql
+++ b/cmd/hatchet-migrate/migrate/migrations/20260908090000_v1_0_152.sql
@@ -1,0 +1,11 @@
+-- +goose Up
+-- +goose StatementBegin
+-- Identifies the listener session that last activated the worker, so only that session can
+-- deactivate it when listener sessions overlap.
+ALTER TABLE "Worker" ADD COLUMN "lastListenerSessionId" UUID;
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+ALTER TABLE "Worker" DROP COLUMN "lastListenerSessionId";
+-- +goose StatementEnd
--- a/internal/services/dispatcher/server.go
+++ b/internal/services/dispatcher/server.go
@@ -297,7 +297,7 @@ func (s *DispatcherImpl) ListenV2(request *contracts.WorkerListenRequest, stream
 	tenant := ctx.Value("tenant").(*sqlcv1.Tenant)
 	tenantId := tenant.ID
 	s.analytics.Count(stream.Context(), analytics.Worker, analytics.Listen)
-	sessionId := uuid.New().String()
+	sessionId := uuid.New()
 	workerId, err := uuid.Parse(request.WorkerId)
 
 	if err != nil {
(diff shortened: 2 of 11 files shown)


cmd/hatchet-migrate/migrate/migrations/20260908090000_v1_0_152.sql
```diff
@@ -0,0 +1,11 @@
+-- +goose Up
+-- +goose StatementBegin
+-- Identifies the listener session that last activated the worker, so only that session can
+-- deactivate it when listener sessions overlap.
+ALTER TABLE "Worker" ADD COLUMN IF NOT EXISTS "lastListenerSessionId" UUID;
```
mrkaye97: let's not use `IF NOT EXISTS` (I need to make a greptile rule for this or something) - same thing in the down

internal/services/dispatcher/server.go
```diff
@@ -332,28 +345,23 @@ func (s *DispatcherImpl) ListenV2(request *contracts.WorkerListenRequest, stream
...
-
-		s.l.Error().Ctx(ctx).Err(err).Msgf("could not update worker %s active status to true (session established %s, last session established %s)", request.WorkerId, sessionEstablished.String(), lastSessionEstablished)
+		s.l.Error().Ctx(ctx).Err(err).Msgf("could not activate worker %s for listener session %s", request.WorkerId, sessionId)
 		return err
 	}
 
 	fin := make(chan bool)
 
-	s.workers.Add(workerId, sessionId, newGRPCSubscribedWorker(stream, fin, workerId, s.defaultMaxWorkerLockAcquisitionTime, s.pubBuffer))
+	s.workers.Add(workerId, sessionId.String(), newGRPCSubscribedWorker(stream, fin, workerId, s.defaultMaxWorkerLockAcquisitionTime, s.pubBuffer))
```
mrkaye97: maybe we should update the `workers` struct to be:

```go
type workers struct {
	innerMap syncx.Map[uuid.UUID, *syncx.Map[uuid.UUID, *subscribedWorker]]
}
```

and pass the uuids everywhere instead of casting to string?

mrkaye97 (commented):
couple small comments

mrkaye97 (approved):

PR #4935, opened 2026-09-09
fix(python): duplicate actions error, dead listener

### LINKED ISSUE #4731: [BUG] Python SDK: Dead worker, but action listener lives on

### LINKED ISSUE #4922: Worker action registry silently overwrites on duplicate, and the action id is lowercased while the workflow name is not
## Summary

The worker's action registry maps `action_id -> callable` with a bare `dict[key] = value` and no duplicate check, so the last registration silently replaces the first.

Because the **action id is lowercased** while the **server-facing workflow name preserves case**, two functions declared with different Python names — `SendEmail` and `sendemail` — become two separate workflows on the engine that both route to a **single** function on the worker.

## Where

`sdks/python/hatchet_sdk/worker/worker.py`:

```python
for step in workflow.tasks:
    action_name = workflow._create_action_name(step)
    self._action_registry[action_name] = step   # no membership test
```

Case folding: `runnables/workflow.py` — `service_name = namespace + config.name.lower()` and `_parse_task_name` returns `(name or func.__name__).lower()`.
Divergent public name: `runnables/workflow.py` — `name = namespace + config.name`, case preserved.
Name inference: `hatchet.py` — `inferred_name = name or func.__name__` (no module, no qualname).

## Reproduction

Mechanism level: the engine needs a Postgres-backed gRPC server, so the only thing stubbed is the network call `AdminClient.put_workflow`. `Worker.register_workflow`, `_create_action_name`, `_parse_task_name` and `to_proto` are the real code paths.

**Case A — two functions, same explicit name:**

```
registry keys : ['process:process']
registry size : 1
resolved fn   : make_email.<locals>.process
billing fn    : make_billing.<locals>.process
warnings raised: NONE
exceptions     : NONE (both register_workflow calls returned)
worker would run for action 'process:process': {'who': 'email.process', 'charged': False}
EXPECTED for billing runs                    : {'who': 'billing.process', 'charged': True}
```

**Case B — no explicit name, case-only difference.** This is the sharper one: the developer wrote two functions with *different* Python names and passed no `name=` argument, so nothing in their source looks like a collision.

```
declared names: SendEmail / sendemail
registry keys : ['sendemail:sendemail']
resolved fn   : make_b.<locals>.sendemail
warnings: NONE
```

`Standalone.name` still reports `SendEmail`, so reading back the SDK's own view of the object does not reveal it either.

**Case C — server identity versus worker identity:**

```
workflow proto name A: SendEmail | step action A: ['sendemail:sendemail']
workflow proto name B: sendemail | step action B: ['sendemail:sendemail']
distinct workflows on server? True
same action id (same worker fn)? True
```

## Negative controls

```
# 1: genuinely distinct names do NOT collide
registry keys : ['charge:charge', 'notify:notify']
registry size : 2   (expected 2 -> no collision)

# 2: the SDK IS capable of raising on bad registration
ValueError: failed to register workflow: no_tasks. Workflows must have at least
one task registered before registering
```

## The silence

- `warnings.catch_warnings(record=True)` around both registrations in cases A and B: zero warnings
- no exception; both calls return normally
- `worker.py` contains no membership test, no `logger.warning`, no `warn(...)` at the assignment
- the runner logs an error only for a **missing** key; a key that resolves to the *wrong* function is indistinguishable from a correct one
- `Worker.start()` validates only that the registry is non-empty, never that it is complete relative to the workflows registered

## What the docs say

I grepped the Python SDK docstrings, `frontend/docs/content/docs/**` and the surrounding comments:

- **No documentation states that workflow or task names must be unique.** The only "unique name" statements concern webhook names and Go cron names.
- **No documentation states that names are lowercased.** The `name` parameter docstrings say only *"If not specified, defaults to the name of the function being wrapped"* — they promise the function's name, and the code returns `.lower()` of it.
- The public `Workflow.name` docstring says *"The (namespaced) name of the workflow"* and returns the case-preserved form, while the routing identity is the lowercased form.

## Note

The SDK enforces the analogous constraint elsewhere — `runnables/task.py` raises `ValueError("Conditions must have unique readable data keys.")` — and raises `IdempotencyCollisionError` with the colliding run id on the idempotency path. The vocabulary and the practice both exist; the registry assignment simply has neither.

## Not tested

- **The engine end to end.** No server was started and no network used; everything above is at the level of the SDK's naming, proto serialisation and registry. Whether the Go engine imposes an additional constraint — for example rejecting a `PutWorkflow` whose step action id already belongs to another workflow name — was **not verified by execution**. I read the admin server and repository code and found no such check, but I cannot call that exhaustive.
- Replay/retry routing: `Action.key` is `step_run_id/retry_count[/…]`, properly delimited, and I found no defect in it, but I did not exercise a replay.
- Child-workflow `child_index` deduplication across a retry. The Python side resets the spawn counter per attempt while the server dedups on parent plus index; whether a parent taking a different branch on retry can dedup a new child against a positionally-matching old one needs a live engine. The repo contains its own `examples/bug_tests/durable_child_key_duplicate_child/` suite in this area, so this may already be known. **I am not claiming a finding there.**
- Namespaces do prevent the collision across namespaces; within one (the default) they do not.
- Go and TypeScript SDKs were grepped, not audited.

I have not checked whether this was raised before; a pointer to an existing issue is welcome and I will close this in favour of it.

## Version

`hatchet-sdk` 1.40.0, source at `086a63f2245416296de288a944aad1b8357e63e9`, Python 3.12.

mrkaye97:
# Description

Fixes a few misc. issues in the Python SDK:

Fixes #4731
Fixes #4922

## Type of change

## Checklist

Changes have been:

---

<details id="ai-disclosure">

No AI here

</details>


mrkaye97:
@greptileai

grutt (approved):

PR #4944, opened 2026-09-10
fix(engine): durable dag getting stuck and not recovering

mrkaye97:
# Description

Fixes a couple more edge cases in durable dags. Basically, there was a case where writing tasks as skipped / cancelled would break the ordered delivery logic, which caused some bugs. Here we update the logic to rely on `satisfiedOrder` to make sure we're returning completions in the right order always.

And a couple other things:

1. Fixed CEL evaluation failures causing the operator to hang, so now we handle those
2. Expanded the idempotency key hash to include the target action if we have it, because every dag step was getting the same hash. I don't think this was actually a bug since we were protecting against nondeterminism internally, but probably a good idea regardless to be safe and make sure we're not spawning out of order

## Type of change

## Checklist

Changes have been:

---

<details id="ai-disclosure">

- **Details**: Fable to help debug the issues we were seeing in production and implement a fix
</details>


mrkaye97:
@greptileai

mrkaye97:
@greptileai

mrkaye97:
@greptileai

juliusgeo (approved):

PR #4987, opened 2026-09-19
feat: migrate engine protocol to ConnectRPC

abelanger5:
# Description

Migrates the existing gRPC server to use ConnectRPC instead, which has full compatibility with the existing gRPC setup. This is needed for broader API compatibility with serverless environments, and even frontends (with compatibility with `grpc-web`) as we're working on more streaming features. This is a pre-requisite for the serverless-operator work to be completed. 

The migration is pretty straightforward here because ConnectRPC is generated from protos as well, and the handler signatures for unary RPCs are the same. The primary difference is the streaming RPCs and the middleware, which has been extensively tested. 

Some explicit risks are:
1. The remove of the encoding step before locking the mutex for stream.Send. This means that we acquire a lock on the mutex before encoding the payload, which can add some latency / pessimistic locking behavior. This is a low risk, because the encoding time is miniscule compared to the network send time when we're in a flow control state (this change was theorized to solve a flow control issue, but it didn't end up being the root cause). 
2. Error code modification. The SDKs case on various gRPC error codes, and I (with Claude's assistance) audited the various places we're relying on gRPC errors across the SDKs to ensure compatibility, so this should be low / no risk. 
3. Changes to keepalive settings / gRPC streaming behavior. This is a bit of an unknown until we're deployed, but since gRPC relies on `http/2` for keepalive functionality the migration to the http2 server should preserve the existing behavior.  

## Type of change

## Checklist

Changes have been:

---

<details id="ai-disclosure">

</details>


internal/services/admin/v1/server.go
```diff
@@ -563,17 +559,20 @@ func (a *AdminServiceImpl) TriggerWorkflowRun(ctx context.Context, req *contract
...
 			if err != nil {
-				return nil, status.Errorf(codes.Internal, "failed to build idempotency collision error: %v", err)
+				return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to build idempotency collision error: %v", err))
 			}
 
-			return nil, st.Err()
+			collisionErr := connect.NewError(connect.CodeAlreadyExists, errors.New("idempotency key collision"))
+			collisionErr.AddDetail(detail)
+
+			return nil, collisionErr
```
abelanger5, line 575: need a close eye on this; double check that it's covered by an e2e test on each path

internal/services/admin/v1/server.go
```diff
@@ -36,13 +35,13 @@ func (a *AdminServiceImpl) CancelTasks(ctx context.Context, req *contracts.Cance
 	for _, idStr := range req.ExternalIds {
 		id, err := uuid.Parse(idStr)
 		if err != nil {
-			return nil, status.Error(codes.InvalidArgument, "invalid external id")
+			return nil, connect.NewError(connect.CodeInvalidArgument, errors.New("invalid external id"))
```
abelanger5, line 38: should double-check that returning a simple `err` (without wrapping it in `connect.NewError` does not leak any internal errors to the client

internal/services/admin/server_v1.go
```diff
@@ -88,17 +83,20 @@ func (a *AdminServiceImpl) triggerWorkflowV1(ctx context.Context, req *v1contrac
...
-				return nil, status.Errorf(codes.Internal, "failed to build idempotency collision error: %v", stErr)
+			if detailErr != nil {
+				return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to build idempotency collision error: %v", detailErr))
 			}
 
-			return nil, st.Err()
+			collisionErr := connect.NewError(connect.CodeAlreadyExists, errors.New("idempotency key collision"))
+			collisionErr.AddDetail(detail)
+
+			return nil, collisionErr
```
abelanger5, line 99: same comment here

internal/services/admin/server_v1.go
```diff
@@ -214,18 +212,21 @@ func (a *AdminServiceImpl) bulkTriggerWorkflowV1(ctx context.Context, req *contr
...
-			return nil, status.Errorf(codes.Internal, "failed to build idempotency collision error: %v", stErr)
+		if detailErr != nil {
+			return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("failed to build idempotency collision error: %v", detailErr))
 		}
 
-		return nil, st.Err()
+		collisionErr := connect.NewError(connect.CodeAlreadyExists, errors.New("idempotency key collision"))
+		collisionErr.AddDetail(detail)
+
+		return nil, collisionErr
```
abelanger5, line 229: same comment here

internal/services/dispatcher/subscribed_worker_v1.go
```diff
@@ -122,18 +121,6 @@ func (worker *subscribedWorker) sendToWorkerWithStream(
...
-
-	msg := &grpc.PreparedMsg{}
-	err := msg.Encode(worker.stream, action)
-	if err != nil {
-		encodeSpan.RecordError(err)
-		encodeSpan.End()
-		return fmt.Errorf("could not encode action: %w", err)
-	}
-
-	encodeSpan.End()
```
abelanger5, line 135: this was done to reduce the chance of flow control for large messages, but ended up not being the root cause of the issue -- the encoding time is very small relative to the network send time, which was what caused flow control

internal/services/grpc/middleware/auth.go
```diff
@@ -48,16 +48,9 @@ func (a *GRPCAuthN) Middleware(ctx context.Context) (context.Context, error) {
 	ctx = context.WithValue(ctx, analytics.APITokenIDKey, tokenUUID)
 	ctx = context.WithValue(ctx, analytics.TenantIDKey, tenantId)
 
-	span := trace.SpanFromContext(ctx)
-	telemetry.WithAttributes(span,
-		telemetry.AttributeKV{Key: "tenant.id", Value: tenantId},
-	)
```
abelanger5, line 54: we're removing traces / context here?

internal/services/grpc/middleware/auth.go
```diff
@@ -28,9 +26,11 @@ func NewAuthN(config *server.ServerConfig) *GRPCAuthN {
 	}
 }
 
-func (a *GRPCAuthN) Middleware(ctx context.Context) (context.Context, error) {
-	forbidden := status.Errorf(codes.Unauthenticated, "invalid auth token")
-	token, err := auth.AuthFromMD(ctx, "bearer")
```
abelanger5, line 33: existing clients are compatible, each SDK constructs `authorization: bearer <token>` and we only accept the first matched header

internal/services/grpc/server.go
```diff
@@ -236,170 +251,207 @@ func NewServer(fs ...ServerOpt) (*Server, error) {
...
-	burst := limit
-	limiter := middleware.NewHatchetRateLimiter(rate.Limit(limit), int(burst), s.l)
 
-	errorInterceptor := middleware.NewErrorInterceptor(s.a, s.l)
+	mux := http.NewServeMux()
+	routes := grpcRoutes{}
 
-	opts := []logging.Option{
-		logging.WithLogOnEvents(logging.StartCall, logging.FinishCall),
+	if s.ingestor != nil {
```
abelanger5: nit: each service should be required, and we should do the nil checks at the top - I don't think there are any cases where we partially register

internal/services/grpc/middleware/auth.go
abelanger5: nvm, found it below

mrkaye97 (approved):
I feel pretty good about this, should be easy enough to test on staging. there are no changes if we try to regenerate the SDKs right?

grutt (approved):
Approved pending staging upgrade testing

PR #5000, opened 2026-09-21
fix(engine): probe v1_task by primary key when resolving an external id

grutt:
# Description

Looking up a task by external id joined `v1_lookup_table` directly to `v1_task`, which Postgres planned as a merge across every daily partition. These queries now resolve the task key first and read `v1_task` by primary key so the scan prunes to one partition.

## Type of change

## What's Changed

- [x] `GetTaskByExternalId` looks up the task key, then reads `v1_task` by primary key
- [x] `RefreshTimeoutBy` and `ManualSlotRelease` use the same lookup
- [x] `ListMatchingTaskEvents` resolves lookup keys first, then joins `v1_task` by primary key

## Checklist

Changes have been:

- [x] Linted and formatted

## Testing

`EXPLAIN (ANALYZE, BUFFERS)` on a database with daily `v1_task` partitions: the previous join read every partition (several seconds, millions of rows). The rewritten single-id lookup prunes to one partition and returns in under a millisecond.

---

<details id="ai-disclosure">

- **Details**: Cursor agent (Grok) rewrote the task-lookup queries and regenerated sqlc.

</details>


pkg/repository/sqlcv1/tasks.sql
```diff
@@ -524,22 +528,35 @@ WITH input AS (
...
 FROM
-    v1_lookup_table l
-JOIN
-    v1_task t ON t.id = l.task_id AND t.inserted_at = l.inserted_at
+    looked_up l
+JOIN LATERAL (
+    SELECT *
+    FROM v1_task t
+    WHERE t.id = l.task_id AND t.inserted_at = l.inserted_at
+) t ON true
```
mrkaye97: join lateral is kind of odd here I think? what's the reason for needing this?

mrkaye97 (commented):
left one question about `JOIN LATERAL`

pkg/repository/sqlcv1/tasks.sql
grutt: the ai thinks it will force the prune, but in practice its the same plan.. i think can remove it

mrkaye97: yeah, I'd remove it

mrkaye97 (approved):

